PR #6, opened 2019-12-04
Ohw mike 20191204

MikeOpenHWGroup:
This change adds the capability to run the RISC-V compliance test suite independent of all other tests.  Currently only supports dsim.


MikeOpenHWGroup:
OK

PR #9, opened 2019-12-17
Metrics integration

aimeepsutton:
Makefile modifications to support simulation on Metrics platform: enable waveform dumping for DSim, enable separate directories for compile and running simulation. 
Added the file .metrics.json in the repository root dir. This is the integration point with the Metrics platform.


MikeOpenHWGroup:
Thanks Aimee - looks good.

Issue #52, opened 2020-02-11
Updates to simulation Makefile


MikeOpenHWGroup:
### Task Outcome
1. Add support for Mentor's _Questa_ SystemVerilog simulator.
2. Add the ability to run individual tests.

#### Questa.
At this point there are no rules to support Questa.  There are rules for Verilator, Cadence Xcelium (xrun) and Metrics dsim, so it should be straightforward to create new rules for Questa.

#### Running Individual Tests.
Section 3.2.2 of the [Verification Strategy](https://github.com/openhwgroup/core-v-docs/tree/master/verif/Common/OpenHWGroup_CORE-V_Verif_Strategy.pdf) (V0.4) discusses the testbench and testcases in sufficient detail for this task.  The Makefile rules and the directory with the testcase assembler code are organized in such a way that the user needs to run all tests in a directory.  This is OK for regression, but it is not ideal for development and debug of new tests.  Its desirable to be able to specify a single test.   This is expected to be a big task as it goes straight to the heart of how testcases and the Makefile are organized. 

### Location Information
The location of the Makefile in GitHub is https://github.com/openhwgroup/core-v-verif/blob/master/cv32/tests/core/Makefile.

### Completion Criteria
1. Questa: when `make questa-hello_world` (and similar) works the same as `make dsim-hello_world` (and similar).
2. Ability to specify single test on the command line.  E.G `make questa TEST=<test_name.S>`

### Additional context
In the near future, the above Makefile will be replaced by Makefiles in the sim directory:
https://github.com/openhwgroup/core-v-verif/tree/master/cv32/sim/core
https://github.com/openhwgroup/core-v-verif/tree/master/cv32/sim/uvmt_cv32

MikeOpenHWGroup:
Looks good Sebastien.

PR #54, opened 2020-02-12
Patch for issue #52

sjthales:
I just modified two files cv32/tests/core/Makefile and cv32/tests/core/firmware/start.S
if you want to run all test with questa, you can write : make questa-all
if you want to run a specific test, for instance add test, you can write : make questa-unit-test add
It is the same thing for compliance riscv, to run I_LBU_01 : make questa-unit-test  I_LBU_01
It works on my side.

### DIFF
--- a/cv32/tests/core/Makefile
+++ b/cv32/tests/core/Makefile
@@ -26,17 +26,17 @@ WAVES     = 0
 # vsim configuration
 VVERSION  = "10.7b"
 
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup:
Accepted.  I cannot test the vsim and questa rules, but I see what you are trying to achieve and it looks solid to me.  Also, the existing rules for Verilator and dsim run as before.

PR #76, opened 2020-03-12
Ohw mike 20200311

MikeOpenHWGroup:


MikeOpenHWGroup:
This pull-request is now "stale" as the issues it addresses are either resolved (see #81) or no longer required.

PR #79, opened 2020-03-13
Ohw mike 20200313

MikeOpenHWGroup:
Hi David.   This is an easy pull request.  The changes are:
- compile in all the stuff we commented out on Wednesday due to the whole UVM-1.2 mess (the Makefile now tells Cadence to use 1.2).
- use your report_server, not the overly verbose default one.
- minor tweaks to messaging (makes the logfile look cleaner).


MikeOpenHWGroup:
Thanks for the review David.

PR #93, opened 2020-04-14
Fix to assign enum to string

dawidzim:
Hey!
This pull request is for assign enum to string. Correct way to do this is to use method name():
`The name() method returns the string representation of the given enumeration value. If the given value is not a member of the enumeration, the name() method returns the empty string`


MikeOpenHWGroup:
Hi Dawid, thanks for the contribution.  I agree with your recommendation and have approved it.  The code you are updating is template generated, so I am passing this PR to David Poulin (@datum-dpoulin).  He is the author of the templates and so I'd be interested in his thoughts as well.

MikeOpenHWGroup:
Thanks for the contribution @dawidzim!

PR #96, opened 2020-04-27
MCY example for verilator testbench

nakengelhardt:
An example project that demonstrates how to use [mcy](https://github.com/YosysHQ/mcy).

This was mostly written to stay self-contained in the `cv32/sim/mcy` directory and not touch existing files, which leads to a lot of duplication. Needs discussion how to integrate.

The main issue is the makefile/build instructions. The makefile from `core` uses relative paths for includes so can't be launched in any other folders, but if the build instructions are duplicated they will diverge. The manifest for building the mutated design could perhaps be derived from the original with a sed script, so that it would automatically work with new code for small changes.

The reduced firmware that only runs the tests for the division instructions also contains a lot of duplicated files. It could be integrated as ifdefs, or moved inside the mcy folder so it doesn't clutter up the regular test files.


cv32/sim/mcy/cv32e40p_manifest.flist
```diff
@@ -0,0 +1,60 @@
...
+${DESIGN_RTL_DIR}/riscv_prefetch_buffer.sv
+${DESIGN_RTL_DIR}/riscv_prefetch_L0_buffer.sv
+${DESIGN_RTL_DIR}/riscv_L0_buffer.sv
+${DESIGN_RTL_DIR}/riscv_hwloop_regs.sv
+${DESIGN_RTL_DIR}/riscv_hwloop_controller.sv
+${DESIGN_RTL_DIR}/riscv_mult.sv
+${DESIGN_RTL_DIR}/register_file_test_wrap.sv
+${DESIGN_RTL_DIR}/riscv_int_controller.sv
+${DESIGN_RTL_DIR}/riscv_ex_stage.sv
+//${DESIGN_RTL_DIR}/riscv_alu_div.sv
```
MikeOpenHWGroup, line 53: For a demo this is find, but for production this is a dangerous practice.  Suggest we use the "real" manifest from the cv32e40p repo and use a script to swap out the RTL with the mutation file.

cv32/sim/mcy/Makefile
```diff
@@ -0,0 +1,543 @@
+###############################################################################
```
MikeOpenHWGroup, line 1: As far as I can tell, this Makefile makes no changes to the Makefile in ../core other than to add variables and targets for DIV_ONLY_FIRMWARE plus targets to run the mutated coded.   So it should be possible to reduce this makefile to `include $(PROJ_ROOT/cv32/sim/core/Makefile` and then add the aforementioned rules.

cv32/sim/core/Makefile
```diff
@@ -133,7 +133,7 @@ export DESIGN_RTL_DIR = $(CV32E40P_PKG)/rtl
 # Common Makefile:
 #    - Core Firmware and the RISCV GCC Toolchain (SDK)
 #    - Variables for RTL dependencies
-include ../Common.mk
+include $(PROJ_ROOT_DIR)/cv32/sim/Common.mk
```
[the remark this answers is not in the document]
MikeOpenHWGroup: Nothing is pretty when it comes to Makefiles.   I have no particular concern about short pathnames and the ability to run from anywhere is a bonus.

PR #100, opened 2020-04-30
Add debug req

silabs-PaulZ:
Add the ability to control the assertion of debug_req from mm_ram to the core.

### DIFF
--- a/cv32/tb/core/mm_ram.sv
+++ b/cv32/tb/core/mm_ram.sv
@@ -19,14 +19,16 @@ module mm_ram
  #(
      parameter RAM_ADDR_WIDTH    =  16,
                INSTR_RDATA_WIDTH = 128, // width of read_data on instruction bus
-               DATA_RDATA_WIDTH  =  32  // width of read_data on data bus
+               DATA_RDATA_WIDTH  =  32,  // width of read_data on data bus
+               DBG_START_ADDR    =  32'h1A110800, // Debug start address
+               DBG_ADDR_WIDTH    =  11
   )
   (
      input logic                          clk_i,
      input logic                          rst_ni,
 
      input logic                          instr_req_i,
-     input logic [RAM_ADDR_WIDTH-1:0]     instr_addr_i,
+     input logic [31:0]                   instr_addr_i,
      output logic [INSTR_RDATA_WIDTH-1:0] instr_rdata_o,
      output logic                         instr_rvalid_o,
      output logic                         instr_gnt_o,
@@ -77,7 +79,7 @@ module mm_ram
 
     // signals to ram
     logic                          ram_data_req;
-    logic [RAM_ADDR_WIDTH-1:0]     ram_data_addr;
+    logic [31:0]                   ram_data_addr;
     logic [31:0]                   ram_data_wdata;
     logic [31:0]                   ram_data_rdata;
     logic                          ram_data_we;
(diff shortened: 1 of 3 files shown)


cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -43,7 +43,7 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
                             parameter PULP_CLUSTER        =   0, //changed
                                       FPU                 =   0,
                                       PULP_ZFINX          =   0,
-                                      DM_HALTADDRESS      =  32'h1A110800,
+                                      DM_HALTADDRESS      =  32'h000F_0000,
```
MikeOpenHWGroup, line 46: We should not be changing this parameter without fully understanding the impact to the toolchain and the existing testcases.

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -111,8 +113,8 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
 
     // instantiate the core
     riscv_core #(
-                 .PULP_CLUSTER    (PULP_CLUSTER),
-                 .FPU             (FPU),
+                 .PULP_CLUSTER           (PULP_CLUSTER),
```
MikeOpenHWGroup, line 116: Extra whitespace (pls remove)

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -205,6 +207,8 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
          .irq_id_o       ( irq_id_in                      ),
          .irq_o          ( irq                            ),
 
+         .debug_req_o    ( debug_req                            ),
```
MikeOpenHWGroup, line 210: Extra whitespace

silabs-PaulZ:
Hi @MikeOpenHWGroup , I agree that the single memory is a better solution and this was part of my original pull request, but this required a change in the dm_halt_addr_i location that was raised as a concern in https://github.com/openhwgroup/cv32e40p/issues/321, hence the dual memory solution was provided. Another solution (not provided) is to remap the debug memory access region to a dedicated location in the memory within mm_ram.sv. This can be done as a later step to allow for a modifiable halt address. I'm looking forward to the TPE update. Let me know if you will need me to drop this PR or modify it.

Note that in addition to the debug entry region starting at dm_halt_addr_i, I would also like to have a separate dm_exception_addr_i region to handle exceptions during debug, which can affect the unified TPE.

cv32/tb/core/mm_ram.sv
```diff
@@ -75,26 +79,38 @@ module mm_ram
 
     // signals to ram
     logic                          ram_data_req;
-    logic [RAM_ADDR_WIDTH-1:0]     ram_data_addr;
+    logic [31:0]                   ram_data_addr;
```
MikeOpenHWGroup, line 82: You don't like parameters and I don't like magic numbers.  :-)   Not gonna go to battle over this, but puttin' it out there that I have a strong dislike for this coding style.

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -109,6 +111,37 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
       end
     end
 
+    // --------------------------------------------
```
MikeOpenHWGroup, line 114: This may have to change in the future.  The Google generator produces debug code in a single memory image.  However, we several weeks away from that, so it a good idea to bring this in now.

PR #101, opened 2020-04-30
Add possibility to use different gcc versions for test generation

ASintzoff:
These commits add a way to use another prefix like riscv-none-elf and also additional option like -mabi=ilp32.


MikeOpenHWGroup:
Hi @ASintzoff, this PR is old and the situation with respect to the tool-chain has not changed.  I will close this PR without merging it in.

PR #106, opened 2020-05-12
Iss integration

GTumbush:
Hi Mike,
  This pull request contains a few simple changes: 
1) Moved .tcl files to xrun directory. 
2) Deleted read from mvendorid in hello-world test. CSR mvendorid is not defined in cv32e40p. This causes an illegal instruction error with CV32E40P starting with hash 4e15e0a committed May 3. This is why the hello-world test was failing.

Thanks,
Greg


Silabs-ArjanB:
I don't fully agree with the fix done. The treatment of mvendorid in CV32E40P is actually a bug as reported in https://github.com/openhwgroup/cv32e40p/issues/335. It is fine if 'hello world' keeps failing until that bug has been fixed in the RTL.

GTumbush:
Hello Arjan,
    Since mvendorid is not implemented in the cv32e40p access of this register should produce an illegal instruction error, correct? If so, why do you not agree with deleting the read of this register in the hello world test?
Greg

Silabs-ArjanB:
If mvendorid is not supported it should be implemented as a CSR that returns 0 (and not an illegal instruction error, that is not an okay implementation).

GTumbush:
Hi Arjan,
  OK, I was not aware that @silabs-PaulZ introduced a bug with pull request https://github.com/openhwgroup/cv32e40p/pull/326/files  Thanks for pointing this out.

silabs-PaulZ:
The issue was recorded here: https://github.com/openhwgroup/cv32e40p/issues/335
The fix was submitted as a pull request: https://github.com/openhwgroup/cv32e40p/pull/337

PR #108, opened 2020-05-14
fix tests to use the new hardware performance monitors

silabs-PaulZ:
The CV32E40P core has an RTL update that will replace the performance counters with the HPM as specified in the privilege specification. The hello_world and sanity test pass, but it will break the firmware tests.

This PR will fix these firmware tests to work with the core update from:
https://github.com/openhwgroup/cv32e40p/pull/33


davideschiavone:
> 
> 
> Looks good to me @silabs-PaulZ. I will merge this in as soon as [openhwgroup/cv32e40p#339](https://github.com/openhwgroup/cv32e40p/pull/339) is merged by @davideschiavone.

done

PR #114, opened 2020-05-28
Fixes for supporting custom .c test and specific riscv_ebreak_test_0.S

GTumbush:
Fixes for supporting custom .c test
Addition of riscv_ebreak_test_0.S and changes to .mk to support this.
Fixed RISCV_EXE_PREFIX ?= issue in Common.mk

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -122,7 +122,7 @@ OVP_CTRL_FILE   = $(DV_OVPM_DESIGN)/riscv_CV32E40P.ic
 #
 # riscv toolchain install path
 RISCV            ?= /opt/riscv
-RISCV_EXE_PREFIX  = $(RISCV)/bin/riscv32-unknown-elf-
+RISCV_EXE_PREFIX ?= $(RISCV)/bin/riscv32-unknown-elf-
 
 # CORE FIRMWARE vars. All of the C and assembler programs under CORE_TEST_DIR
 # are collectively known as "Core Firmware".  Yes, this is confusing because
@@ -203,7 +203,7 @@ sanity: hello-world
 # We link with our custom crt0.s and syscalls.c against newlib so that we can
 # use the c standard library
 $(CUSTOM_DIR)/$(CUSTOM_PROG).elf: $(CUSTOM_DIR)/$(CUSTOM_PROG).c
-	$(RISCV_EXE_PREFIX)gcc -march=rv32imc -o $@ -w -Os -g -nostdlib \
+	$(RISCV_EXE_PREFIX)gcc -mabi=ilp32 -march=rv32imc -o $@ -w -Os -g -nostdlib \
 		-T $(CUSTOM_DIR)/link.ld  \
 		-static \
 		$(CUSTOM_DIR)/crt0.S \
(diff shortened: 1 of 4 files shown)


cv32/sim/Common.mk
```diff
@@ -203,7 +203,7 @@ sanity: hello-world
 # We link with our custom crt0.s and syscalls.c against newlib so that we can
 # use the c standard library
 $(CUSTOM_DIR)/$(CUSTOM_PROG).elf: $(CUSTOM_DIR)/$(CUSTOM_PROG).c
-	$(RISCV_EXE_PREFIX)gcc -march=rv32imc -o $@ -w -Os -g -nostdlib \
+	$(RISCV_EXE_PREFIX)gcc -mabi=ilp32 -march=rv32imc -o $@ -w -Os -g -nostdlib \
```
MikeOpenHWGroup, line 206: What does `-mabi=ilp32` do and why do we need it?   Does it affect other tests?

cv32/sim/Common.mk
```diff
@@ -262,6 +262,18 @@ $(CUSTOM)/dhrystone.elf: $(CUSTOM)/dhrystone.c
 		-I $(RISCV)/riscv32-unknown-elf/include \
 		-L $(RISCV)/riscv32-unknown-elf/lib \
 		-lc -lm -lgcc
+
```
MikeOpenHWGroup, line 265: Do we need this?   I would have expected the existing `make custom CUSTOM_PROG=<my_fancy_new_test_proj>` target to handle this.

cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -113,6 +113,13 @@ dhrystone: comp $(CUSTOM)/dhrystone.hex
 		+UVM_TESTNAME=uvmt_cv32_firmware_test_c \
 		+firmware=$(CUSTOM)/dhrystone.hex
 
+riscv_ebreak_test_0: comp $(CUSTOM)/riscv_ebreak_test_0.hex
```
MikeOpenHWGroup, line 116: Same as previous comment.

GTumbush: New code is the ebreak . S test.

When I try to make the riscv_ebreak_test_0 using make custom this is what I get:
make custom CUSTOM_PROG=riscv_ebreak_test_0 SIMULATOR=xrun XRUN_USE_ISS=YES CV32E40P_PKG=/wrk/gtumbush/mcu/OpenHW/cv32e40p/master

make: *** No rule to make target `/wrk/gtumbush/mcu/OpenHW/regression_test/cv32/sim/uvmt_cv32/../../../cv32/tests/core/custom/riscv_ebreak_test_0.c', needed by `/wrk/gtumbush/mcu/OpenHW/regression_test/cv32/sim/uvmt_cv32/../../../cv32/tests/core/custom/riscv_ebreak_test_0.elf'.  Stop.

.S tests don’t seem to be supported as a custom test.

cv32/sim/Common.mk
GTumbush: This is existing code, only the line feed is new.  The test specific code in this Makefile for hello_world, misalign, illegal, fibonacci, and dhrystone could be removed but then all users need to be aware of the new calling method (make custom CUSTOM_PROG=<my_fancy_new_test_proj>) and the ci script needs updating.

cv32/sim/Common.mk
GTumbush: mabi specifies integer and floating-point calling convention.

Without this option gcc will error with the following:
c1: error: requested ABI requires -march to subsume the 'D' extension

cv32/sim/uvmt_cv32/xrun.mk
MikeOpenHWGroup: Ah!   Good catch.  I'll take an action to figure out how to support the assembler (*.S) custom test-programs.   We'll want that for riscv-dv support for sure.   In the meantime, let's bring in your `riscv_ebreak_test_0` target as a temporary fix.

MikeOpenHWGroup:
Thanks for the quick replies Greg.  I'll merge this in.

PR #116, opened 2020-06-01
Pull Request for regression_test branch

MikeOpenHWGroup:
Several of the changes here are simply adding new tests and/or updates to the Makefiles.  There is little need to review these as changes will be coming to both in the coming few weeks.

The significant changes in this branch are to `.gitlab-ci.yml` and `.metrics.json`.  I have moderate confidence in the changes, as they are supporting regressions successfully.  However, both would benefit from a review before merging to the master branch.


.metrics.json
```diff
@@ -31,6 +31,15 @@
             "isPass":      "SIMULATION PASSED",
             "seed":        1
 	        },
+          {
+            "name":        "Previously generated riscv-dv test",
+            "build":       "uvmt_cv32",
+            "cmd":         "cd cv32/sim/uvmt_cv32; make asm ASM_PROG=riscv_ebreak_test_0 SIMULATOR=dsim USE_ISS=YES DSIM_WORK=/mux-flow/build/repo/dsim_work DSIM_RESULTS=/mux-flow/build/results",
+            "wavesCmd":    "cd cv32/sim/uvmt_cv32; make asm ASM_PROG=riscv_ebreak_test_0 SIMULATOR=dsim USE_ISS=YES DSIM_WORK=/mux-flow/build/repo/dsim_work DSIM_RESULTS=/mux-flow/build/results WAVES=1",
+            "metricsFile": "hello-world/metrics.db",
```
aimeepsutton, line 39: metricsFile property is set incorrectly.  Looking at the artifacts associate with this test, the correct value should be:
`'metricsFile" : "riscv_ebreak_test_0/metrics.dv"`

cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
```diff
@@ -114,6 +114,34 @@ interface uvmt_cv32_core_cntrl_if (
...
+    }
+    debug_module_halt_address: coverpoint dm_halt_addr {
+      bins low  = {32'h0000_0000, 32'h0000_FFFF};
+      bins med  = {32'h0001_0000, 32'hEFFF_FFFF};
+      bins high = {32'hF000_0000, 32'hFFFF_FFFF};
+    }
+    hart_id: coverpoint hart_id {
+      bins low  = {32'h0000_0000, 32'h0000_FFFF};
+      bins med  = {32'h0001_0000, 32'hEFFF_FFFF};
+      bins high = {32'hF000_0000, 32'hFFFF_FFFF};
```
aimeepsutton, line 139: Is your intention to specify a range with each of these bins? Right now you just have 2 explicit values in each.
You might want this, for example:
`bins low = {[32'h0000_0000 : 32'h0000_FFFF]};`

MikeOpenHWGroup: Good catch, thanks.   You are right, that is exactly what I wanted.

.metrics.json
MikeOpenHWGroup: Thanks.  Fixes in.  Passes the Metrics' JSON lint check.

PR #125, opened 2020-06-12
Update for the latest RTL (101f455de)

MikeOpenHWGroup:

Hi @GTumbush.  This is an update to point to the current head of the RTL (hash 101f455de).  As expected, a couple of changes to top-level interrupt pins was required.   There are also changes to the CSRs and I am worried that this will impact the ability of the OVPsim ISS to serve as a reference model when we add verification of interrupts.

I do not yet know if we should merge in this update, but I wanted you to be aware of the implications of the latest changes to the RTL.

### DIFF
--- a/cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
+++ b/cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
@@ -77,7 +77,7 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
     logic [31:0]                  data_rdata;
     logic [31:0]                  data_wdata;
 
-    logic [ 4:0]                  irq_id_out;
+    logic [ 5:0]                  irq_id_out;
     logic [ 4:0]                  irq_id_in;
 
     // Load the Instruction Memory 
@@ -125,7 +125,7 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
     end
 
     // instantiate the core
-    riscv_core #(
+    cv32e40p_core #(
                  .PULP_HWLP        (PULP_HWLP),
                  .PULP_CLUSTER     (PULP_CLUSTER),
                  .FPU              (FPU),
@@ -182,8 +182,8 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
          .irq_timer_i            ( core_interrupts_if.irq_timer      ),
          .irq_external_i         ( core_interrupts_if.irq_external   ),
          .irq_fast_i             ( core_interrupts_if.irq_fast       ),
-         .irq_nmi_i              ( core_interrupts_if.irq_nmi        ),
-         .irq_fastx_i            ( core_interrupts_if.irq_fastx      ),
+         //.irq_nmi_i              ( core_interrupts_if.irq_nmi        ),
+         //.irq_fastx_i            ( core_interrupts_if.irq_fastx      ),
 
          .debug_req_i            ( core_cntrl_if.debug_req           ),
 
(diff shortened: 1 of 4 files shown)


MikeOpenHWGroup:
Hi Greg.  Will you have a chance to look at this one?

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -77,7 +77,7 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
     logic [31:0]                  data_rdata;
     logic [31:0]                  data_wdata;
 
-    logic [ 4:0]                  irq_id_out;
+    logic [ 5:0]                  irq_id_out;
     logic [ 4:0]                  irq_id_in;
```
GTumbush, line 81: Shouldn't irq_id_in/out bit width match?

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
```diff
@@ -77,7 +77,7 @@ module uvmt_cv32_dut_wrap #(// DUT (riscv_core) parameters.
     logic [31:0]                  data_rdata;
     logic [31:0]                  data_wdata;
 
-    logic [ 4:0]                  irq_id_out;
+    logic [ 5:0]                  irq_id_out;
```
GTumbush, line 80: I'm not sure of the point of this if it is just cut down to [4:0] in the instantiation.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -147,24 +147,26 @@ module uvmt_cv32_step_compare
...
+             // MT: 2020-06-11
+             //"miex"    : csr_val = $root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.miex_q;
              "mtvec"   : csr_val = {$root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mtvec_q, 6'h0, 2'b01};
              "mtvecx"  : csr_val = {$root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mtvec_q, 6'h0, 2'b01};
              "mscratch": csr_val = $root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mscratch_q;
              "mepc"    : csr_val = $root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mepc_q;
              "mcause"  : csr_val = {$root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mcause_q[6], 
                                     25'b0, 
                                     $root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mcause_q[5:0]};
-             "mip"     : csr_val = {$root.uvmt_cv32_tb.dut_wrap.riscv_core_i.cs_registers_i.mip.irq_nmi,
```
GTumbush, line 158: MIP is still in https://core-v-docs-verif-strat.readthedocs.io/projects/cv32e40p_um/en/latest/control_status_registers.html Has the RTL been updated without updating the spec?

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
MikeOpenHWGroup: Yes it should (see my next comment).

cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv
MikeOpenHWGroup: You caught me.  :sheep: There has been so much churn on the interrupt interface of late (with maybe more to come), that I took the cheap-and-dirty hack-it-until-it-compiles approach.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
MikeOpenHWGroup: It still there, but its fields have changed.  Once the dust settles on the interrupts, we will need to re-sync the ISS and S&C module to match the RTL.

MikeOpenHWGroup:
Thanks for the review Greg.  I agree with your suggestions to properly fix the width of irq_id_out.  For now, I'd like to leave the mip check in step-and-compare commented out (see my comments).   I'll create an issue to re-review this with the designers once the interrupt spec stops churing.

PR #126, opened 2020-06-17
VCS support: additional preliminary modifications

ASintzoff:
One modification to fix a VCS compilation error.
The other ones are not directly linked to VCS but are useful to have cleaner repository.

### DIFF
--- a/cv32/bsp/Makefile
+++ b/cv32/bsp/Makefile
@@ -4,9 +4,10 @@ RISCV_GCC = $(RISCV_EXE_PREFIX)gcc
 RISCV_AR = $(RISCV_EXE_PREFIX)ar
 SRC = crt0.S handlers.S syscalls.c vectors.S
 OBJ = libcv-verif.a crt0.o handlers.o syscalls.o vectors.o
+CFLAGS ?= -Os -g -static -mabi=ilp32 -march=rv32imc -Wall -pedantic
 
 all:
-	$(RISCV_GCC) -Os -g -c -static $(SRC)
+	$(RISCV_GCC) $(CFLAGS) -c $(SRC)
 	$(RISCV_AR) rcs $(OBJ)
 
 clean:
--- a/cv32/bsp/Makefile
+++ b/cv32/bsp/Makefile
@@ -1,5 +1,5 @@
 RISCV            ?= /opt/riscv
-RISCV_EXE_PREFIX  = $(RISCV)/bin/riscv32-unknown-elf-
+RISCV_EXE_PREFIX ?= $(RISCV)/bin/riscv32-unknown-elf-
 RISCV_GCC = $(RISCV_EXE_PREFIX)gcc
 RISCV_AR = $(RISCV_EXE_PREFIX)ar
 SRC = crt0.S handlers.S syscalls.c vectors.S
--- a/cv32/tb/uvmt_cv32/imperas_iss.flist
+++ b/cv32/tb/uvmt_cv32/imperas_iss.flist
@@ -16,7 +16,7 @@
 // 
 
 // CV32 test bench files
-$(TBSRC_HOME)/uvmt_cv32/uvmt_cv32_step_compare.sv
+${TBSRC_HOME}/uvmt_cv32/uvmt_cv32_step_compare.sv
 ${TBSRC_HOME}/uvmt_cv32/uvmt_cv32_iss_wrap.sv
 
 // Vendor Libraries
--- a/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
+++ b/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
@@ -26,7 +26,7 @@ module CPU
     BUS SysBus
 );
 
-    import "DPI-C" context task ovpEntry(input int, input string, input string);
+    import "DPI-C" context task ovpEntry(input int i, input string s1, input string s2);
 
     export "DPI-C" task     busLoad;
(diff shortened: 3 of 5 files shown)


cv32/bsp/Makefile
```diff
@@ -1,12 +1,12 @@
 RISCV            ?= /opt/riscv
-RISCV_EXE_PREFIX  = $(RISCV)/bin/riscv32-unknown-elf-
+RISCV_EXE_PREFIX ?= $(RISCV)/bin/riscv32-unknown-elf-
```
MikeOpenHWGroup, line 2: This seems like a good idea.

cv32/bsp/Makefile
```diff
@@ -1,12 +1,12 @@
...
 RISCV_GCC = $(RISCV_EXE_PREFIX)gcc
 RISCV_AR = $(RISCV_EXE_PREFIX)ar
 SRC = crt0.S handlers.S syscalls.c vectors.S
 OBJ = libcv-verif.a crt0.o handlers.o syscalls.o vectors.o
 
 all:
-	$(RISCV_GCC) -Os -g -c -static $(SRC)
+	$(RISCV_GCC) -Os -g -c -static -mabi=ilp32 -march=rv32imc $(SRC)
```
MikeOpenHWGroup, line 9: I am very reluctant to accept this change for two reasons:
1. The author of the BSP is @craigblackmore and he has much more extensive knowledge of toolchain and BSP issues than I do.
2. It seems to me that this change restricts the machine to IMC instructions.   That is OK for now, but in a few weeks we will need to support Z and Xpulp instructions.

cv32/tb/uvmt_cv32/imperas_iss.flist
```diff
@@ -16,7 +16,7 @@
 // 
 
 // CV32 test bench files
-$(TBSRC_HOME)/uvmt_cv32/uvmt_cv32_step_compare.sv
+${TBSRC_HOME}/uvmt_cv32/uvmt_cv32_step_compare.sv
```
MikeOpenHWGroup, line 19: OK

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -26,7 +26,7 @@ module CPU
     BUS SysBus
 );
 
-    import "DPI-C" context task ovpEntry(input int, input string, input string);
+    import "DPI-C" context task ovpEntry(input int i, input string s1, input string s2);
```
MikeOpenHWGroup, line 29: We have to be very careful here.  We have noticed that each simulator supports the DPI in subtly different ways.  I will test this with DSIM and XRUN, but it might be safer to wrap this with `ifdef VCS` macros.  (Yes, I dislike this as well.)

cv32/bsp/Makefile
ASintzoff: Is this code relying on Z and Xpulp instructions?
The goal of this modification is to make explicit the ABI and the architecture as done in Common.mk

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
ASintzoff: FWIW, examples about DPI-C in System Verilog standard contain argument names.

cv32/bsp/Makefile
craigblackmore: I would suggest defining common options in a CFLAGS variable in Common.mk e.g.
```
CFLAGS = -Os -g -static -mabi=ilp32 -march=rv32imc -Wall -pedantic
```
then the make targets can use `$(CFLAGS)` rather than hard-coding the flags in multiple places.

In bsp/Makefile we can do:
```
CFLAGS ?= -Os -g -static -mabi=ilp32 -march=rv32imc -Wall -pedantic
```
and again use `$(CFLAGS)` instead of hard-coding the flags.

ASintzoff: OK

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
MikeOpenHWGroup: Oh I know.  I wish every vendor actually _read_ the LRM.  :wink:

cv32/bsp/Makefile
ASintzoff: A new version taking into account @craigblackmore suggestion is available.

ASintzoff:
Commit added regarding timeunit/timeprecision/timescale as discussed.

ASintzoff:
I plan to have a separate PR for `vcs.mk` itself once all the _preliminary_ modifications are merged.

MikeOpenHWGroup:
> I plan to have a separate PR for `vcs.mk` itself once all the _preliminary_ modifications are merged.

Perfect.   Please use the TIMESCALE variable in the Makefile.  This will ensure all simulators are using the same timescale.   If the format of that variable does not work as a VCS compile-time and/or run-time argument, feel free to add a new one and use it.

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
MikeOpenHWGroup: I tried it with DSIM and it worked as expected.

PR #130, opened 2020-06-19
vcs.mk: update to support VCS for uvmt_cv32 testbench

ASintzoff:
VCS support for `uvmt_cv32` testbench

`riscv-tests` and `riscv-compliance-tests` passed except:
- `I-NOP-01.S`
- `I-SB-01.S`
- `I-SH-01.S`


MikeOpenHWGroup:
Thanks @ASintzoff, it is great to have VCS support in our Makefiles and to have the environment pushed through compile and simulation with VCS.

PR #131, opened 2020-06-23
vcs support for core + fixes for uvm

masgia:
Signed-off-by: Massimiliano Giacometti <massimiliano.giacometti@hensoldt-cyber.com>

- VCS support for core tests
- adaptations in VCS flow for UVM tests

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -199,7 +199,7 @@ FIRMWARE_UNIT_TEST_OBJS   =  	$(addsuffix .o, \
 ###############################################################################
 # The sanity rule runs whatever is currently deemed to be the minimal test that
 # must be able to run (and pass!) prior to generating a pull-request.
-sanity: hello_world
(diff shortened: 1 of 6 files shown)


cv32/sim/Common.mk
```diff
@@ -199,7 +199,7 @@ FIRMWARE_UNIT_TEST_OBJS   =  	$(addsuffix .o, \
 ###############################################################################
 # The sanity rule runs whatever is currently deemed to be the minimal test that
 # must be able to run (and pass!) prior to generating a pull-request.
-sanity: hello-world
+sanity: hello_world
```
MikeOpenHWGroup, line 202: Hi @masgia, its probably not a good idea to change this: the CI config files invoke these targets and this could break them.   In addition, several Contributors have their own scripts that invoke these targets.

MikeOpenHWGroup:
Thanks @masgia - great to have VCS supported.

PR #144, opened 2020-07-07
Update debug_test for new trigger access change. 

silabs-PaulZ:
Also add ability to run xcelium in single step mode using XRUN_SINGLE_STEP variable

Address issue: https://github.com/openhwgroup/cv32e40p/issues/388

Signed-off-by: Paul Zavalney <paul.zavalney@silabs.com>

### DIFF
--- a/cv32/sim/uvmt_cv32/xrun.mk
+++ b/cv32/sim/uvmt_cv32/xrun.mk
@@ -76,14 +76,13 @@ comp: mk_xrun_dir $(CV32E40P_PKG) $(OVP_MODEL_DPI)
 		-elaborate
 #		$(XRUN_VELABCOVERAGE)
 
-XRUN_COMP_RUN = $(XRUN_COMP) $(XRUN_RUN_BASE_FLAGS)
+
+XRUN_SIM_PREREQ = comp
+XRUN_COMP_RUN = $(XRUN_RUN_FLAGS)
 
 ifeq ($(XRUN_SINGLE_STEP), YES)
 	XRUN_SIM_PREREQ = mk_xrun_dir $(CV32E40P_PKG) $(OVP_MODEL_DPI)
 	XRUN_COMP_RUN = $(XRUN_COMP) $(XRUN_RUN_BASE_FLAGS)
-else
-	XRUN_SIM_PREREQ = comp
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup:
Hi @GTumbush, can you have a look at this one?  I briefly looked at it, and it passed my regression using both dsim and xrun.  Having said that, the bulk of the changes impact `xrun.mk` and I do not reply on that one as much as you do.   I do not see anything here that would, but always better to be safer than sorry.   Thanks.

cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -76,12 +76,20 @@ comp: mk_xrun_dir $(CV32E40P_PKG) $(OVP_MODEL_DPI)
 		-elaborate
 #		$(XRUN_VELABCOVERAGE)
 
-XRUN_COMP_RUN = $(XRUN_COMP_ARGS) $(XRUN_RUN_BASE_FLAGS)
```
GTumbush, line 79: I believe this line is redundant since XRUN_COMP_RUN is assigned to either way in the next if/else block.

cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -76,12 +76,20 @@ comp: mk_xrun_dir $(CV32E40P_PKG) $(OVP_MODEL_DPI)
...
+XRUN_COMP_RUN = $(XRUN_COMP) $(XRUN_RUN_BASE_FLAGS)
```
GTumbush, line 79: I believe this line is redundant since XRUN_COMP_RUN is assigned to either way in the next if/else block.

silabs-PaulZ: I updated a change. Removing the redundant assignment doesn't work (I don't have an explanation) so I simply have it default to multi step and then overwrite if single step.

GTumbush: The mysteries of makefiles, ok no problem.  This can be merged.

MikeOpenHWGroup:
@Thanks @silabs-PaulZ and @GTumbush.  I'll merge this onto the master branch directly.

PR #145, opened 2020-07-08
Simplified always block calling compare

GTumbush:
Simplified always block calling compare, removed fork/join_any and #60ns delay


GTumbush:
Passes ci_check

PR #113, opened 2020-05-26
Add dm_exception_addr_i port to riscv_core. 

silabs-PaulZ:
This PR will add the dm_exception_addr_i to the top-level riscv_core.
This PR must be done in conjunction with the design update that adds this pin:  https://github.com/openhwgroup/cv32e40p/pull/353

This PR will fixe the following issues:
https://github.com/openhwgroup/cv32e40p/issues/185
https://github.com/openhwgroup/core-v-docs/issues/62

Signed-off-by: Paul Zavalney <paul.zavalney@silabs.com>

PR #148, opened 2020-07-09
add mtvec_addr_i port

silabs-PaulZ:
MTVEC will reset to mtvec_addr_i during first cycle of boot up.

Fixes issue https://github.com/openhwgroup/cv32e40p/issues/401

Requires RTL changes from: https://github.com/openhwgroup/cv32e40p/pull/403

Signed-off-by: Paul Zavalney <paul.zavalney@silabs.com>


MikeOpenHWGroup:
Looks good @silabs-PaulZ.  This PR is gated by https://github.com/openhwgroup/cv32e40p/issues/411.  Ready to integrate as soon as that is complete.

silabs-PaulZ:
Hi  @MikeOpenHWGroup , @Silabs-ArjanB , @davideschiavone ,

In order to resolve conflicts, I merged the design to the laster version on master which moves out the tracer instance from the core. I updated this PR with a bind of the tracer to the core in the dut_wap.sv. hello_world passes if I disable the ISS.

MikeOpenHWGroup:
Thanks @silabs-PaulZ, that was a good idea.  Alas, it is not going to do it.  The majority of our testing relies on the RM (ISS) to determine pass/fail, and even with self-checking tests the comparisons to the RM are important, so the tracer is not a nice-to-have.  Its a must-have.  I cannot move forward on this PR or https://github.com/openhwgroup/core-v-verif/pull/113 without the tracer.

Silabs-ArjanB:
@silabs-PaulZ Can you please make the following update in core-v-verif/cv32/sim/Common.mk

CV32E40P_HASH   ?= 3deb55860001f8ac39cefa80b90566170b12391f

After that I think this is a good pull request that also moves to a newer CV32E40P repos.

@MikeOpenHWGroup This pull request does not relate in any way to the tracer. That issue is solved in https://github.com/openhwgroup/core-v-verif/pull/113 that should be merged after this one.

davideschiavone:
I agree with @MikeOpenHWGroup , the tracer is a **_must have_** as it is used to work with the reference model.

davideschiavone:
@Silabs-ArjanB , @silabs-PaulZ , @MikeOpenHWGroup , why don't we create a new file that can acts as a top level (in the bhv folder, not synthesizable) that can be used so that the integration of the core+tracer is easy and transparent to the system hosting the core+tracer as before? If so we can create a task in the core repository

Something like:

module cv32e40p_wrap
(
SAMEPARAMETERs of CV
)
(
same interface of CV
)

cv32e40p core_i
(
...
);

cv32e40p_tracer tracer_i
(
...
);

endmodule

Silabs-ArjanB:
Hi @davideschiavone What you propose is exactly what Paul already did in https://github.com/openhwgroup/core-v-verif/pull/113.

davideschiavone:
Sorry @Silabs-ArjanB I don't see the wrap file. I mean, having a wrap file in the cv32 repository so that it is maitained by the core task group and can be used by core users without downloading other repositories

Silabs-ArjanB:
Hi @davideschiavone The tracer has been integrate here: https://github.com/silabs-PaulZ/core-v-verif/blob/74122489e7f0020f75571e5edcc23279ba4539f5/cv32/tb/uvmt_cv32/uvmt_cv32_dut_wrap.sv#L200 .

This pull request should not even mention the tracer at all as it is not intended to work with the latest RTL.

davideschiavone:
Hi @Silabs-ArjanB , thanks for the pointer but I did not mean that.
I meant having in the cv32e40p repository a wrapper file in the bhv folder that instantiate both the core and the tracer as it was before.
Such wrapper can then be used in all the systems that want to use the core has before (i.e. with the tracer) in an easier way. Like this we have both the comfort that there was before with the user that does not need to integrate by himself the tracer, and we decoupled synthesizable and not synthesizable files.

Silabs-ArjanB:
Hi @davideschiavone That we can do indeed, no problem, as long as it is in the bhv directory. Let's however first get #148 and #113 merged with the couple of already identified updates that are mentioned in these PRs. We can do the update you describe in a following (new) PR.

davideschiavone:
I suggest that the core-verif also instantiates the wrapper I am talking about as the tracer depends on internal signal of the core, and the user or verification enviroment should not care about it, and should not receive updates regarding these changes.

Idially, it should just instantiate the DUV

silabs-PaulZ:
I made the update to CV32E40P_HASH, hello world is running with ISS. 
Note that the tracer is instantiated in the cv32e40p_core.sv

PR #151, opened 2020-07-13
ci_check shebang fix for python path flexibility, allows for non-standard python3 executable locations in corporate environments

strichmo:


strichmo:
All /usr/bin/env does is locate python3 for you using PATH.  If "which python3" works (i.e. python3 is in your PATH), then this ought to work.

PR #171, opened 2020-07-28
add debug module. Update CSRs to match spreadsheet

duncangraham-Imperas:


vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -487,5 +492,9 @@ module CPU
         ovpEntry(ID, ovpcfg, elf_file);
         $finish;
     end
+    
+    final begin
+        ovpExit(ID);
```
MikeOpenHWGroup, line 497: We are going to have to find a way to handle this differently.   In a UVM environment there should be one-and-only-one `final` block in the entire environment.   Let me think on this.

PR #173, opened 2020-07-28
Xcelium simulator control common make flags

strichmo:
Here's the initial commits to support the discussed common simulation control flags with Xcelium
Documentation for the changes are provided in the README.md in cv32/sim/uvmt_cv32.  
I believe all changes are backwards-compatible with the exception of using xrun_results as a common results directory

I still need to document the common environment makefile variables but all discussed were added in this PR in a backwards-compatible way.

I will be following up with Questa sometime later this week


cv32/sim/xrun/indago.tcl
```diff
@@ -0,0 +1,6 @@
+# Indago waveform probe script
```
MikeOpenHWGroup, line 1: This is exactly the kind of file I was expecting to see in this directory.   I have thought for some time that these should be in `cv32/sim/tools/xrun` (not `cv32/sim/xrun`) since this would keep the _sim_ directory a little cleaner.   What do you think?

cv32/sim/xrun/probe.tcl
```diff
@@ -0,0 +1,5 @@
+database -open waves -default -incsize 2G
```
MikeOpenHWGroup, line 1: Same comment: maybe these should be in `cv32/sim/tools/xrun` (not `cv32/sim/xrun`).

cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -21,29 +21,107 @@
...
+COV_ARGS = -load cov_work/uvmt_cv32_tb/$(TEST)
+endif
+
+ifeq ($(call IS_YES,$(GUI)),YES)
+COV_ARGS += -gui
+else
+COV_ARGS += -execcmd "$(IMC_REPORT_ARGS)"
+endif
+
+################################################################################
```
MikeOpenHWGroup, line 108: Everything below this point is all good.  Thanks for the cleanup.

cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -21,29 +21,107 @@
...
+XRUN_DIR         ?= $(XRUN_RESULTS)/xcelium.d
 XRUN_SINGLE_STEP ?=
-#XRUN_VELABCOVERAGE = -covdut uvmt_cv32_tb -coverage function
-#XRUN_VRUNCOVERAGE  = -covoverwrite -covtest uvmt_cv32_tb -covscope uvmt_cv32_tb
-# VCOVERAGE     = imc -load cov_work/$(TESTBENCH)/$(TESTBENCH) -execcmd "report -summary"
+XRUN_ELAB_COV     = -covdut uvmt_cv32_tb -coverage b:e:f:t:u
+XRUN_RUN_COV      = -covoverwrite -covscope uvmt_cv32_tb
+NUM_TEST         ?= 2
+
+################################################################################
```
MikeOpenHWGroup, line 42: I get what you are trying to do here, and have no objections to the concepts.  A couple of questions:
1. What makes `ifeq ($(call IS_YES, ($MYVAR)),YES)` better? (Both are terrible, but that's "make" for you.)
2. Do we know this will work across multiple Cadence installations?

cv32/sim/uvmt_cv32/README.md
```diff
@@ -130,6 +130,92 @@ Set the shell variable SIMULATOR to `vcs` to simply run **`make <target>`**.
 with _vcs_ in several weeks.  If you need to update the Makefiles, please do
 so and issue a Pull Request.
 
+Common Makefile Flags
```
MikeOpenHWGroup, line 133: I'm supportive of this.  To make this work we'll need to recruit "tool primes" to help us keep this all working.  I am not confident that it will be possible to keep this stable, even for one tool across multiple member installations.

cv32/sim/uvmt_cv32/xrun.mk
strichmo: 1. I'm certainly not proud of that, but at least it kept the polarity of the statement consisten (i.e. that are no negations in the clause)
Taking the statement above, the other alternative I came up with was something like:
ifneq ($filter $(YES_VALS),$(MYVAR),)
# This is the positive clause!

Mentally I didn't like seeing a ifneq on a positive YES test. 

2. Fair question regarding Cadence.  I did update a couple of internal -execcmds to IMC because of deprecated commands.  In general what has worked reasonably well for me is to just put a stake in the ground (using a latest release or only one major release behind) and not be so aggressive about updating with latest and greatest.  Again this is only trying to support the absolute basics and I would expect certain power users to replace these in their own wrappers (should probably make these ?= now that I am thinking about this).

cv32/sim/uvmt_cv32/README.md
strichmo: Agreed on the tool primes.  I have Max ready to help with VCS.  Silicon Labs can do Xcelium and Questa, although I would like to find another helper for Questa.

This is all about minimum enablement to me.  If we can keep the feature creep out of here I hope we can maintain a basic level of functionality.  We will need to be careful about over-evolving this.

cv32/sim/xrun/indago.tcl
strichmo: Agree with this.  I will amend pull request to move those directories.  Other folks have a few scripts there so we will need to communicate the change.

cv32/sim/xrun/probe.tcl
strichmo: Yep...will amend.

cv32/sim/uvmt_cv32/xrun.mk
MikeOpenHWGroup: 1. Yes, we all _love_ Make.  :-)

2. I like the "stake in the ground" approach, so let's proceed.  In an abundance of caution I will ask another Contributor who uses Xcelium (@alfredoh1234) to review before merging.

MikeOpenHWGroup:
Hi @alfredoh1234.  I know that you are an active Contributor and are also using Cadence Xcelium, including the coverage collecting, reporting and viewing tools.   Can you have a look at this PR to ensure that it doesn't break anything in with your IT setup?   It _should_ be fine, but its always good to get a second opinion.

alfredoh1234:
@strichmo and @MikeOpenHWGroup , ere are my comments:
1. Why hardcode the path to the tools?
> CV_SW_TOOLCHAIN  ?= /opt/riscv
> RISCV            ?= $(CV_SW_TOOLCHAIN)

Why not keying-off tool environment variables instead? 

2. For consistency, can command-line arguments be set with YES instead of alternatively using 1/YES ?

3. Please remove hiding the coverage arguments inside intermediate variables
> ifeq ($(call IS_YES,$(COV)),YES)
> XRUN_USER_COMPILE_ARGS += $(XRUN_ELAB_COV)
> XRUN_RUN_COV_FLAGS += $(XRUN_RUN_COV)
> endif

Instead, add them to the XRUN_COMP and XRUN_PLUSARGS arguments list and let them resolve to "nul" if the makefile commandline does nto call for coverage.  

4. Please do not make it the default to change directory upon running a test. Why not make sure the tests can be ran from any directory and allow the user to be in whatever direcory they want to be? 
You may want to add a text message with the resolved command:
> "to check test results: cd $(XRUN_RESULTS)/$(CUSTOM_PROG)"

MikeOpenHWGroup:
Hi @alfredoh1234, thanks for the review.  I can comment on some of your feedback:

> 1. Why hardcode the path to the tools?
>> CV_SW_TOOLCHAIN ?= /opt/riscv
>> RISCV ?= $(CV_SW_TOOLCHAIN)

> Why not keying-off tool environment variables instead?

Perhaps you misunderstand the Make syntax here.  The conditional assignment (?=) allows the user to over-ride the variable either from the command-line or from environment variables.  So I think this syntax does what you want.

> 2. For consistency, can command-line arguments be set with YES instead of alternatively using 1/YES ?

I _believe_ this syntax allows you to set the variable using either YES or 1.  Its a way of trying to please everyone.  :wink:

> 3. Please remove hiding the coverage arguments inside intermediate variables
>> ifeq ($(call IS_YES,$(COV)),YES)
>> XRUN_USER_COMPILE_ARGS += $(XRUN_ELAB_COV)
>> XRUN_RUN_COV_FLAGS += $(XRUN_RUN_COV)
>> endif

I admit, I do not see the purpose of this either.  @strichmo, can you comment?

> 4. Please do not make it the default to change directory upon running a test. Why not make sure the tests can be ran from any directory and allow the user to be in whatever directory they want to be?
You may want to add a text message with the resolved command:

The requirement to launch tests from a specific directory has been in place from day one (actually, even before that at PULP).  In fact, I _want_ to keep that as it forces all users to look in the same place for useful information (Makefiles and READMEs).  What has been added is the ability to place compile/simulation output anywhere.  In any case, this command does not change the CWD of the user.

strichmo:
> @strichmo and @MikeOpenHWGroup , ere are my comments:
> 
> 1. Why hardcode the path to the tools?
> 
> > CV_SW_TOOLCHAIN  ?= /opt/riscv
> > RISCV            ?= $(CV_SW_TOOLCHAIN)
> 
> Why not keying-off tool environment variables instead?
> 

The updated documentation will highly recommend using CoreV environment variables (i.e. CV_ environment variables) 

To ease migration I wanted to make the current makefiles backwards compatible.  There is not a great way to inform all users of a backwards-compatible-breaking change.

> 1. For consistency, can command-line arguments be set with YES instead of alternatively using 1/YES ?

I have had many many opinions on this (1 vs Y vs YES) and came up with this scheme to support them all.  YES is still supported and we will just have to keep discipline on the makefiles to ensure that we apply the standards evenly through the IS_YES call.

> 2. Please remove hiding the coverage arguments inside intermediate variables
> 
> > ifeq ($(call IS_YES,$(COV)),YES)
> > XRUN_USER_COMPILE_ARGS += $(XRUN_ELAB_COV)
> > XRUN_RUN_COV_FLAGS += $(XRUN_RUN_COV)
> > endif
> 
> Instead, add them to the XRUN_COMP and XRUN_PLUSARGS arguments list and let them resolve to "nul" if the makefile commandline does nto call for coverage.
> 

The intermediate variables are for documentation purposes.  Near the top of the makefile a user can check the variables to determine exactly how the coverage elaboration and simulation are being performed.

> 1. Please do not make it the default to change directory upon running a test. Why not make sure the tests can be ran from any directory and allow the user to be in whatever direcory they want to be?
>    You may want to add a text message with the resolved command:
> 
> > "to check test results: cd $(XRUN_RESULTS)/$(CUSTOM_PROG)"

Agree...I will also adding post-step banners in a to highlight important steps and logfiles in a subsequent release as the current makefiles end up generating a wall of text too often.

Regarding directory usage that is something I agree with but that type of change would require changing to a script-implemented versus a make-implemented flow.  Granted I would rather be in a script driven flow but we don't quite have the time nor resources to make that happen now.  I can add this as a requested feature for later when we have some time to properly derive requirements and implement.

MikeOpenHWGroup:
Good discussion.  We each have a usage model for Makefiles that we are used to, and this is a reasonable approach towards a common, stable and documented flow.   I will merge this in.

alfredoh1234:
> Hi @alfredoh1234, thanks for the review. I can comment on some of your feedback:
> 
> > 1. Why hardcode the path to the tools?
> > 
> > > CV_SW_TOOLCHAIN ?= /opt/riscv
> > > RISCV ?= $(CV_SW_TOOLCHAIN)
> 
> > Why not keying-off tool environment variables instead?
> 
> Perhaps you misunderstand the Make syntax here. The conditional assignment (?=) allows the user to over-ride the variable either from the command-line or from environment variables. So I think this syntax does what you want.

The this should be an error, if CV_SW_TOOLCHAIN not set. If you blindly assign it to what may be a non-existent directory you are setting up a potential error that is very hard to debug.

PR #178, opened 2020-08-03
fix latching of exit value from virtual peripheral

strichmo:
rewrite test pass fail logic for both riscv and directed tests


strichmo:
@MikeOpenHWGroup I fixed the local issue of latching the exit value (non-blocking issue).  So now any directed test that does return EXIT_FAILURE will actually signal failure.

Secondly, I rewrote some of the logic in the base test to detect errors.  It seems we have 2 mechanisms here.  The exit code mechanism used by directed tests, and the test pass/test fail mechanism used by the riscv, random tests.  As long as the UVM sees one of the two mechanisms reporting a positive result I will pass the test.  If either report a fail then we will signal UVM_ERROR.  In absence of any signaling a UVM_ERROR is also reported.

PR #181, opened 2020-08-05
add explicit weak handlers for all external interrupts in cv32e40p

strichmo:

Proposed additional irq handlers for the mtvec table in the BSP
Please review and give any comments or suggested changes

### DIFF
--- a/cv32/bsp/README.md
+++ b/cv32/bsp/README.md
@@ -45,7 +45,7 @@ exception code `0`, they jump to the same location as exceptions, therefore the
 user software interrupt handler must also handle exceptions.
 
 The vector table is defined in `vectors.S` and may jump to one of the
-following three interrupt request handlers in `handlers.S`:
+following interrupt request handlers in `handlers.S`:
   * `u_sw_irq_handler` - handles user software interrupts and all exceptions.
   Saves all caller saved registers then checks `mcause` and jumps to the
   appropriate handler as follows:
--- a/cv32/bsp/README.md
+++ b/cv32/bsp/README.md
@@ -46,18 +46,54 @@ user software interrupt handler must also handle exceptions.
 
 The vector table is defined in `vectors.S` and may jump to one of the
 following three interrupt request handlers in `handlers.S`:
-  * `sw_irq_handler` - handles user software interrupts and all exceptions.
+  * `u_sw_irq_handler` - handles user software interrupts and all exceptions.
   Saves all caller saved registers then checks `mcause` and jumps to the
   appropriate handler as follows:
     - Breakpoint: jump to `handle_ebreak`.
     - Illegal instruction: jump to `handle_illegal`.
     - Environment call from M-mode: jump to `handle_ecall`.
     - Any other exception or user software interrupt: jump to `handle_unknown`.
-  * `verification_irq_handler` - calls `mret`. Usage of this interrupt is
-  still under discussion.
+  * `m_software_irq_handler` - handles machine-mode software interrupts
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_timer_irq_handler` - handles machine-mode timer interrupts
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_external_irq_handler` - handles machine-mode external interrupts
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_fast0_irq_handler` - handles machine-mode fast external interrupts (platform extension for CV32)  
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_fast1_irq_handler` - handles machine-mode fast external interrupts (platform extension for CV32)
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_fast2_irq_handler` - handles machine-mode fast external interrupts (platform extension for CV32)
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
+  * `m_fast3_irq_handler` - handles machine-mode fast external interrupts (platform extension for CV32)
+    - Currently jumps to `__no_irq_handler`.  Behavior to be defined in future commit.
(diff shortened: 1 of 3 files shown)


cv32/bsp/handlers.S
```diff
@@ -23,13 +23,110 @@
 .global __no_irq_handler
 .global sw_irq_handler
 .global verification_irq_handler
+.global software_irq_handler
```
jeremybennett, line 26: What is the difference between `software_irq_handler` and the existing `sw_irq_handler`?

cv32/bsp/handlers.S
```diff
@@ -23,13 +23,110 @@
...
+.weak fast7_irq_handler
+.weak fast8_irq_handler
+.weak fast9_irq_handler
+.weak fast10_irq_handler
+.weak fast11_irq_handler
+.weak fast12_irq_handler
+.weak fast13_irq_handler
+.weak fast14_irq_handler
+.weak fast15_irq_handler
+
```
jeremybennett, line 65: Do you really want these to be weak references? Are you expecting that the user will link in alternative implementations that override these?

Weak references are easy to get wrong, they have unexpected semantics, notably with libraries in ELF. They also have different semantics in PE (the file format on Windows), where they are weak aliases. I try to avoid them, unless there is an explicit compelling use case that requires them.

In this case, for verification, I can't see such a use case. Surely verifying hardware should expect fixed software behavior?  If there is an intention to provide explicit handlers later, it should be by updates to this BSP to add them, which would not need a weak reference anyway.

If you are going down this route, you should also make `__no_irq_handler`, `sw_irq_handler` and `verification_irq_handler` weak as well.

cv32/bsp/vectors.S
```diff
@@ -22,32 +22,33 @@ vector_table:
...
+	j fast7_irq_handler
+	j fast8_irq_handler
+	j fast9_irq_handler
+	j fast10_irq_handler
+	j fast11_irq_handler
+	j fast12_irq_handler
+	j fast13_irq_handler
+	j fast14_irq_handler
+	j fast15_irq_handler
+
```
jeremybennett, line 54: My thought is that if you are going to make this change, you should do it for all handlers, not just fast IRQ handlers.

Overall, adding explicit handlers is fine. But I don't think they should be weak references (certainly not for the verification use case). And if you want to have explicit handlers, make them all explicit.

Within this patch, the question of what appears to be duplication of the software IRQ handler has to be addressed. (see earlier comment).

cv32/bsp/handlers.S
strichmo: In interrupt testing I would like to assign a specific handler for each interrupt.  In the interrupt verification test I would prefer not to trust any internal CSRs for interrupt verification.  By installing a specific handler with a specific ID parameter I can be assured when I enter said handler that mcause and mip are correct.

If there is a better methodology for installing specific handlers in vectored mode I am open to it.  I am simply used to using weak handlers in embedded verification tests.  We commonly use this internally throughout our embedded verification codebase for our ARM-M products.

Agreed that all handlers should be weak for consistency if that's the path we choose eventually.

cv32/bsp/handlers.S
strichmo: software_irq_handler specifically maps to MSIP (mip[3] in vectored interrupt mode) versus the sw_irq_handler at mip[0].  The sw_irq_handler at mip[0] is still needed for handling basic exception in vectored mode.   

I would be open to a better name for the mip[3] handler.

strichmo: Excellent idea Craig, I will incorporate this into the change.  I will also document in the markdown.

cv32/bsp/vectors.S
strichmo: I would still really like for the handlers to be weak.  Referencing above in certain test scenarios (i.e. the interrupt test) I would add items to a handler that would simply be overhead in most other contexts.  

Please let me know if you know of an alternative to weak.  Is adding an optional callback preferable in the BSP?

MikeOpenHWGroup: I'll chime in on the handlers.  In verification we need the ability to deliberately break things and model (or react to) the expected response of the DUT when we do.  Typically this is done with callbacks or "hook" methods.  This is my first experience trying to do this with assembly, so any implementation advice is welcome, as long as we have ability to modify the "correct" behavior of the BSP so that we can abuse the DUT.

cv32/bsp/README.md
```diff
@@ -46,18 +46,54 @@ user software interrupt handler must also handle exceptions.
 
 The vector table is defined in `vectors.S` and may jump to one of the
 following three interrupt request handlers in `handlers.S`:
```
craigblackmore, line 48: Thanks for updating the README. Very minor note - need to remove "three" now we have more than three interrupt request handlers.

cv32/bsp/vectors.S
craigblackmore: Given there's a use case, I'm happy for them to be weak unless @jeremybennett would like to recommend an alternative.

cv32/bsp/README.md
strichmo: Thanks...will fix in another commit on this PR.

cv32/bsp/vectors.S
jeremybennett: I'm really uncomfortable with weak references for interrupt handlers.  I prefer Mike's approach. which is less tied to the specifics of ELF linking on Linux.  Create a set of default handlers using an indirection table. Something like (in pseudo-C)
```c
static void (*ext_handlers[NUM_INTERRUPTS]) (int a1, int a2, int a3) =
{ NULL, NULL, ... };


void generic_handler (unsigned int  num, int a1, int a2, int a3)
{
  if (ext_handlers[num])
    *(ext_handlers[num]) ()
  else
    <default handler>
}

void handler_callback (void (*handler) (unsigned int num, int a1, int a2, int a3)
{
  ext_handlers[num] = handler;
}
```
So your handler will do some default action, unless (presumably at start up), you call `handler_callback ()` to add a custom handler.  This code is somewhat simplified - you'd need error checking etc, but it shows the way.

PR #183, opened 2020-08-05
updates for CSRs and local interrupts

duncangraham-Imperas:
1)	Added Trigger Module registers
There is currently no trigger behavior but the CSRs should read and write correctly
2)	Added mtval, mcontext and scontext registers
write ignored/read zero
3)	Modified requested_csr_por.c test to remove User CSRs, Custom User CSRs and prvlvl/uhartid. I added the debug registers access that were commented out.
Do we need to discuss User CSRs and Custom User CSRs further? Should these be present (I do not think adding behaviour of the Custom User CSRs is something we can do at this stage). Please update this test to provide the definitive CSR access requirements if different from my modifications.
4)	Modified cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv to add tdata1 and tinfo but only to set to ignore to remove error messages below reported in tests.
Please update accordingly.
UVM_ERROR @ 124.000 ns : uvmt_cv32_step_compare.sv(196) reporter [STEP_COMPARE] index=tdata1 does not match a CSR name
UVM_ERROR @ 124.000 ns : uvmt_cv32_step_compare.sv(196) reporter [STEP_COMPARE] index=tinfo does not match a CSR name
5)	Modified vendor_lib/imperas/design/interface.sv for correct interrupt signals and debug signals. These are not connected in the testbench.
a.	remove User/Supervisor interrupt signals
b.	add localinterrupt signals 0-15 (interrupts 16-31)
c.	add interrupt acknowledge and interrupt id signals
d.	add debug signals to request entry to debug mode and to provide status when in debug mode


vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -30,8 +31,10 @@ module CPU
     import uvm_pkg::*;
 `endif
 
-    import "DPI-C" context task ovpEntry(input int, input string, input string);
+    import "DPI-C" context task ovpEntry(input int i, input string s1, input string s2);
+    `ifndef UVM
```
MikeOpenHWGroup, line 35: Actually, we need this import independent of UVM.  In the UVM environment this functional will be called from elsewhere in the environment.   I will fix that in a future pull-request.

MikeOpenHWGroup:
Thanks @duncangraham-Imperas.  I'll merge this onto the `rm_update` branch and kick the tires.

PR #185, opened 2020-08-06
Trigger Module Behaviour

duncangraham-Imperas:
Trigger module behaviour has been added to the RISCV reference model
removed old ISS from repo in vendor_lib/imperas_iss


cv32/tests/core/custom/requested_csr_por.c
```diff
@@ -260,8 +262,8 @@ int main(int argc, char *argv[])
     ++err_cnt;
   }
 
-  if (tdata1_rval != 0x28001000) {
-    printf("ERROR: CSR TDATA1 not 0x28001000!\n\n");
+  if (tdata1_rval != 0x28001040) {
```
MikeOpenHWGroup, line 265: I'll accept this in the interest of getting a clean pass for the purposes of integration.   After that, I will revert this until such time as [core-v-docs issue 143](https://github.com/openhwgroup/core-v-docs/issues/143) is resolved.

duncangraham-Imperas: When _readthedocs_, Trigger Data Register 1 specification, it describes bit 6 = 1 Match in M-mode and also bit 3 = 1 Match in U-Mode. I thought the reported error was to bit 3 being set and bit 6 was correctly set. If the correct value has neither bit set I can make this change. Please confirm the value required in the reference model and sorry for the confusion here. Thanks

PR #189, opened 2020-08-12
initial cv_regress release

strichmo:
@MikeOpenHWGroup  We will want to have a call over this particular pull request.  I'm sure all of this is non-controversial. :-)

- Introduces CWD portability (i.e. add wrappers and make sure scripts can be run from any directory)
-- Made slight changes to ci_check for directory robustness, but no functional change
-- regression testlist YAML NOT applied to ci_check yet
- First cut of the cv_regress script which generates various output regression products (metrics json, shell, other regression tools...) from common YAML regression lists
- Initial yaml regression lists for cv32
- Introduce bin/ directory for common utilities in core-v-verif
- README markdowns provided in all new bin and template directories for utility and YAML documentation

All changes should still be backwards compatible

### DIFF
--- a/ci/ci_check
+++ b/ci/ci_check
@@ -59,7 +59,8 @@ if (sys.version_info < (3,0,0)):
 name_of_ci_check_regression = 'cv32_ci_check_regression' # must match the name of one regression list in .metrics.json
 # This script is run from the "ci" directory, but the paths used by simulator
 # commands assume we are at the root of the repo.
-topdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+topdir = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
+print('ci_check: topdir  : {}'.format(topdir))
 
 ################################################################################
 # Methods....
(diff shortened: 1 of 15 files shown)


ci/ci_check
```diff
@@ -191,27 +194,26 @@ if not (prcmd):
...
-topdir = os.getcwd()
+os.chdir(topdir)
 
 #
 # Step 1: Run the User Regression for the UVM verification environment.
 #
 # .metrics.json is the CI regression config used by Metrics CI tools.
 if (uvm):
-    with open('.metrics.json') as f:
+    with open(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '.metrics.json')) as f:
```
MikeOpenHWGroup, line 216: This does not work for me. I get the following nasty-gram:
```
Traceback (most recent call last):
  File "./ci_check", line 216, in <module>
    with open(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '.metrics.json')) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/data/mike/GitHubRepos/openhwgroup/core-v-verif/intr_and_debug/../.metrics.json'
```

strichmo: I can replicate the failure.  Did not check the previous case where you actually run in the ci_check directory!

PR #192, opened 2020-08-14
Strichmo/riscv dv irq config

strichmo:
riscv-dv reconfiguration to support interrupts to prepare for random interrupt testing
all changes are in corev-dv and use factory to avoid riscv-dv code changes
added a new corev-dv target specifically for the cv32e40p

### DIFF
--- a/vendor_lib/google/corev-dv/corev_asm_program_gen.sv
+++ b/vendor_lib/google/corev-dv/corev_asm_program_gen.sv
@@ -35,7 +35,7 @@ class corev_asm_program_gen extends riscv_asm_program_gen;
     instr_stream.push_back(".include \"user_define.h\"");
     instr_stream.push_back(".section .text.start");
     instr_stream.push_back(".globl _start");
-    instr_stream.push_back(".section .text");
+    instr_stream.push_back(".section .init");
     if (cfg.disable_compressed_instr) begin
       instr_stream.push_back(".option norvc;");
     end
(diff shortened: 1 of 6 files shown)


vendor_lib/google/corev-dv/corev_asm_program_gen.sv
```diff
@@ -35,7 +35,7 @@ class corev_asm_program_gen extends riscv_asm_program_gen;
     instr_stream.push_back(".include \"user_define.h\"");
     instr_stream.push_back(".section .text.start");
     instr_stream.push_back(".globl _start");
-    instr_stream.push_back(".section .text");
+    instr_stream.push_back(".section .init");
```
MikeOpenHWGroup, line 38: What does this do?   (I'm always nervous poking on anything that interacts with the BSP...)

strichmo: @MikeOpenHWGroup  Good question.  Here's the history:
I had to make a change to the random constraints in riscv-dv to align our mtvec vector tables to 256 bytes (tvec_alignment = = 8).  
This in turn adds an .align 8 directive in the output assembly before the mtvec table, which you would expect.  However when actually linking the linker will apply that alignment to the **entire text section**.  in the previous code generator we put out initial start instruction in .text.  This moved our boot address from 0x80 to 0x100 (to satisfy alignment).  
I don't understand why the linker applies this directive to the entire section-I am very far from a linker script expert!

Since our TB cannot deal (yet) with a moving boot address one fix I found was to move the _start into the .init section.  This breaks the start away from the alignment and the linker script forces .init to 0x80 (our intended fixed boot address).

If you look at the other non-riscv-dv tests, the boot sections is actually in the .init section so I believe this is the actual intention of that section.  The above change seems to me to align these intentions with riscv-dv-generated assembler.

MikeOpenHWGroup: :+1:

PR #196, opened 2020-08-19
update for Imperas Reference Model

eroom1966:
Mike this contains the fixes for the reference model, and the target in xrun.mk in order to run the compliance framework in xcelium


vendor_lib/imperas/design/monitor.sv
```diff
@@ -202,9 +200,6 @@ module MONITOR
         MSWInt = SysBus.MSWInterrupt;
         MTInt  = SysBus.MTimerInterrupt;
         MEInt  = SysBus.MExternalInterrupt;
-        
-        DebugHlt      = SysBus.haltreq;
-        DebugResetHlt = SysBus.resethaltreq;
```
MikeOpenHWGroup, line 207: Hi @eroom1966, just to be sure I understand what is happening here...  the changes to this file remove Debug* inputs to the *monitor*, but not the *reference model*.  Is that correct?

eroom1966:
Hi Mike
Actually this code was non-functional anyway, simply wiring the nets together but nothing was sensitive to any changes
I think this is fine to take this as it exists today

PR #206, opened 2020-08-26
instruction interrupt functional coverage

strichmo:
fix interrupt assert checks for arbitration
random interrupt noise updates

note that ISA functional coverage still relies on ISS being present in simulation


cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
```diff
@@ -1546,12 +1550,12 @@ class uvme_rv32isa_covg extends uvm_component;
...
         while (1) begin
-           @(cntxt.isa_covg_vif.ins_valid);
-           sample(cntxt.isa_covg_vif.ins);
+            @(cntxt.isa_covg_vif.ins_valid);
+            sample(cntxt.isa_covg_vif.ins);
         end
     endtask
 
-    function void check_compressed(input ins_t ins);
+    function void check_compressed(inout ins_t ins);
```
MikeOpenHWGroup, line 1558: Nice catch!

cv32/env/uvme_cv32/uvme_cv32_cfg.sv
```diff
@@ -68,11 +68,16 @@ class uvme_cv32_cfg_c extends uvm_object;
    constraint defaults_cons {
       soft enabled                == 0;
       soft is_active              == UVM_PASSIVE;
-      soft scoreboarding_enabled  == 1;
+      soft scoreboarding_enabled  == 1;      
       soft cov_model_enabled      == 1;
       soft trn_log_enabled        == 1;
       soft sys_clk_period       == uvme_cv32_sys_default_clk_period; // see uvme_cv32_constants.sv
       //soft debug_clk_period       == uvme_cv32_debug_default_clk_period;
+
```
MikeOpenHWGroup, line 76: Hmmm.  I see what you're thinking here, but I am not sure we want to do this.  It is desirable to be able to merge functional coverage between tests that use the ISS and those that do not.  There is a limitation in dsim that requires the fcov code to be compiled and constructed across all sims in order to merge functional coverage.  So we want the ISA fcov to be compile and constructed, even if the ISS is not.  Of course, without the ISS, the ISA fcov will be zero.

I'll conduct an experiment to see whether we need a change here.

strichmo: Thanks for that info.  I can make the adjustment to avoid sampling in the interrupt_covg_if if the ISS isn't there.

strichmo: I agree on getting fcov separate from ISS.  I just have some random interrupt issues that I really want to get in front of the designers.  I can revisit this after I get those releases out.

strichmo: I probably should've just reused the core interrupts interrupts that was there.  No real reason why I diverged there.  I can resurrect if you'd like to keep things a bit more uniform.

strichmo: Regression flow: Intend to generate .metrics.json from cv_regress.  Support is already there but it's completely untested.  I don't suppose there's any way for me to actually get dsim access so we'll have to work out how to fix.  These jinja template flows are great once you get an initial template working and in place, but they always require a cycle of fixes to get the syntax right.

I'm pushing forward with cv_regress because we really need something for the non-dsim users to regress.  Bulk of work going forward will be random regression based.

MikeOpenHWGroup: OK, as long as we have a common source to define testcases and regressions that is fine.  If cv_regress produces .metrics.json as an output then we are golden.

strichmo:
I addressed the consistent fcov issue by removing that ifdef and solving my issue in a different manner.

MikeOpenHWGroup:
> I addressed the consistent fcov issue by removing that ifdef and solving my issue in a different manner.

Thanks @strichmo

PR #208, opened 2020-08-28
add source dependencies into the bsp makefile

strichmo:
do not automatically clean bsp when building a test

### DIFF
--- a/cv32/bsp/Makefile
+++ b/cv32/bsp/Makefile
@@ -4,12 +4,20 @@ RISCV_EXE_PREFIX ?= $(RISCV)/bin/riscv32-unknown-elf-
 RISCV_GCC = $(RISCV_EXE_PREFIX)gcc
 RISCV_AR = $(RISCV_EXE_PREFIX)ar
 SRC = crt0.S handlers.S syscalls.c vectors.S
-OBJ = libcv-verif.a crt0.o handlers.o syscalls.o vectors.o
+OBJ = crt0.o handlers.o syscalls.o vectors.o
+LIBCV-VERIF = libcv-verif.a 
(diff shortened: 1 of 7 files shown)


strichmo:
This solves an issue that came up when running many tests in parallel on a server farm.  Each test builds its own test code into machine code using the BSP to load into the simulation memory.  However the current bsp makefile will build every source fresh without dependencies.  Inevitably we hit race conditions where a "make bsp" from two tests collide and an object file is not found.  

With the proposed fix the bsp can be pre-built in a regression flow and each test will not rebuild the bsp when running "make bsp"

craigblackmore (changes requested):
Sorry just had chance to look at this as I was away Friday and Monday.

With these changes, the bsp won't get cleaned unless done explicitly, so I think we need to add `clean-bsp` as a dependency to the `clean_all` targets in `cv32/sim/core/Makefile` and `cv32/sim/uvmt_cv32/*.mk`.

Other than that, looks good.

strichmo:
Thanks for the feedback @craigblackmore !  I have pushed makefile updates to chain clean-bsp to clean_all targets.

craigblackmore (changes requested):
Thanks @strichmo. If you can add `clean-bsp` to the `clean` target in `cv32/sim/core/Makefile` too, then I think this is good to merge.

strichmo:
Added to sim/core/Makefile

craigblackmore (approved):

MikeOpenHWGroup:
Thanks for your input on this @craigblackmore.

Hi @strichmo.  I just approved this PR as it need an "approving reviewer".  In theory, you should be able to merge this now.  This will be a good test of your Committer privileges, so please give it a try.

PR #213, opened 2020-09-01
initial release of yaml test spec

strichmo:
tested in xcelium and questa
interrupt test, hello_world and corev tests can use new methodology
backwards compatible make targets remain in Makefiles until completely proven throughout dsim, xrun and vsim


cv32/sim/uvmt_cv32/vcs.mk
```diff
@@ -187,12 +187,12 @@ custom: $(VCS_SIM_PREREQ) $(CUSTOM_DIR)/$(CUSTOM_PROG).hex
 
 ################################################################################
 # Explicit target tests
-hello-world:  $(VCS_SIM_PREREQ) $(CUSTOM)/hello_world.hex
+hello-world:  $(VCS_SIM_PREREQ) $(CUSTOM)/hello-world.hex
```
MikeOpenHWGroup, line 190: I'm conflicted here.  I see what you are doing, but there are a lot of users who are "used to" the underscore.  I've gotten comments about it in the past.   Nevertheless, let's take the hit and clean this up.   I'll approve this.

MikeOpenHWGroup (commented):
Hi @strichmo .  This is all really great stuff.  I'd like to ensure I understand the direction a bit better before this gets merged in.  Can we have a concall to discuss?

cv32/sim/uvmt_cv32/vcs.mk
strichmo: Personally, I would rather have the underscore.  The issue here is that we have to align our agreement on the "name" of the test, or any test, as the yaml file now is a single source that drives the name used throughout the flows and we can't just write a redirection target to rename on-the-fly anymore.  

I thought most folks were used to the hyphen (it's all over our README markdown) so I went that direction.  Would be fine to change it back (and the relevant docs).  Your call.

MikeOpenHWGroup: You're right - underscores are all over the READMEs.  Sigh.  `hello-world` it is then.   I'll get a few comments bit when people see the benefits of single-source it will blow over quickly.

cv32/tests/core/corev-dv/corev_arithmetic_base_test/test.yaml
```diff
@@ -0,0 +1,7 @@
+# Test definition YAML for generated corev arithmetic base test
+
+name: corev_arithmetic_base_test_<RUN_INDEX>
+uvm_test: uvmt_cv32_firmware_test_c
+description: >
+    Math test generated by corev-dv
```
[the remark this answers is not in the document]
strichmo: I don't think it matters, but I released a newline fix anyway.

MikeOpenHWGroup (approved):
Great job on this @strichmo.  A large and significant contribution.   Thanks!

PR #214, opened 2020-09-01
rework clknrst_if clock and reset init behavior to match async reset strategy

strichmo:
avoid clock == x overlap with reset assertion
do not ring clocks until a stable reset cycle (1->0->1) is completed


MikeOpenHWGroup (changes requested):
Hi @strichmo, thanks for doing this, I know it was on my plate...

This PR seems to have a number of updates related to #213 embeded within it.  Can you revert back those files so that this PR only pulls in the clk-and-reset updates?

strichmo:
Aargh.  The rebase command did not work as I thought.  I should've checked that.  Let me just rebase this up to the current master.

strichmo:
Force pushed the rebased branch onto current master.

MikeOpenHWGroup (approved):

PR #216, opened 2020-09-02
Added parametrized width of instr_addr_i

dawidzim:
Hey!
I added `RAM_ADDR_WIDTH` for input port `instr_addr_i`.

### DIFF
--- a/cv32/tb/core/cv32e40p_tb_wrapper.sv
+++ b/cv32/tb/core/cv32e40p_tb_wrapper.sv
@@ -134,38 +134,38 @@ module cv32e40p_tb_wrapper
         #(.RAM_ADDR_WIDTH (RAM_ADDR_WIDTH),
           .INSTR_RDATA_WIDTH (INSTR_RDATA_WIDTH))
     ram_i
-        (.clk_i          ( clk_i                          ),
-         .rst_ni         ( rst_ni                         ),
-         .dm_halt_addr_i ( DM_HALTADDRESS                 ),
+        (.clk_i          ( clk_i                                     ),
+         .rst_ni         ( rst_ni                                    ),
+         .dm_halt_addr_i ( DM_HALTADDRESS                            ),
 
-         .instr_req_i    ( instr_req                      ),
-         .instr_addr_i   ( instr_addr[RAM_ADDR_WIDTH-1:0] ),
-         .instr_rdata_o  ( instr_rdata                    ),
-         .instr_rvalid_o ( instr_rvalid                   ),
-         .instr_gnt_o    ( instr_gnt                      ),
+         .instr_req_i    ( instr_req                                 ),
+         .instr_addr_i   ( {12'h000, instr_addr[RAM_ADDR_WIDTH-1:0]} ),
+         .instr_rdata_o  ( instr_rdata                               ),
+         .instr_rvalid_o ( instr_rvalid                              ),
+         .instr_gnt_o    ( instr_gnt                                 ),
 
-         .data_req_i     ( data_req                       ),
-         .data_addr_i    ( data_addr                      ),
-         .data_we_i      ( data_we                        ),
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Hi @dawidzim, thanks for the PR.   In general, I am a big fan of using parameters to define things like vector widths and so I really like this idea.  Having said that, not all members of the OpenHW community feel the same as I do and the Contributor who made specific updates to this area of the code is one of them.  So it would be best for him to comment.

Hi @silabs-PaulZ, I've fingered you as the reviewer of this PR.  I am inclined to accept it, but will give you the final say.

silabs-PaulZ:
Hi @dawidzim, @MikeOpenHWGroup ,

I personally like to parameterize as much as possible as well. For this case though, you cannot use the RAM_ADDR_WIDTH with instr_addr_i (nor dm_halt_addr_i). I would not be surprised if this change causes the debug_test to fail.

The RAM_ADDR_WIDTH is used to set the size of the RAM. The RAM_ADDR_WIDTH is set to 16 in mm_ram.sv, which is overwritten with a value of 20 in the uvmt_cv32_dut_wrap.sv, which is overwritten with a value of 22 in uvmt_cv32_tb.sv (a bit confusing for sure).

Now, the real problem is that when you execute debug instruction code, the code is located at instruction address 0x1A11_0800 (and it can be targeted anywhere in a 32 bit address region). This debug code is aliased to the end of the memory inside of mm_ram (2^RAM_ADDR_WIDTH - 2^DBG_ADDR_WIDTH = 2^22 - 2^14). In other words, instr_addr_i=0x1A11_0800 access the mm_ram addr=0x3FC000. Limiting the input instruction address will prevent accessing the debug code.

Bottom line, the input address width must match the full instruction bus address width [32 for the 32bit core] and not limited to the size of the RAM [22].

I have always insisted my designers use parameters as much as possible (and avoid use `defines for that matter) and I would encourage you to do the same...just make sure it is the correct parameter. On that note, I think we need a new parameter perhaps a generic ADDR_BUS_WIDTH or even better would be to use RAM_ADDR_WIDTH for the bus widths and make new parameters RAM_SIZE_WIDTH and corresponding DBG_SIZE_WIDTH for the actual memory size allocations.

Regards,
-PaulZ

MikeOpenHWGroup:
Hi @dawidzim, let me know if you are willing to drill into this further as per @silabs-PaulZ comments.  I will leave this PR open for one week.

dawidzim:
Ok. Can you look at `cv32e40p_tb_wrapper.sv` and in line `142` is: `         .instr_addr_i   ( instr_addr[RAM_ADDR_WIDTH-1:0] ),` that suggest me to use `RAM_ADDR_WIDTH` for parametrize width of `instr_addr_i`. Maybe we should change something in line `142` because now is situation:
`RAM_ADDR_WIDTH    = 20` so in instantion `ram_i` port `instr_addr_i` has value :`xxxxinstr_addr`. That cause problem in simulation

MikeOpenHWGroup:
I can envision two fixes.   The first, which may well be the better, is to update all modules to use a common set of parameters.  So a single RAM_ADDR_WIDTH parameter is used to define the instruction bus address with for all modules.

The second "fix" is a bit of a hack:
```
    .instr_addr_i   ( {12'h000, instr_addr[RAM_ADDR_WIDTH-1:0]} ),
```
I would accept the hack as an interim solution, but would rather see a proper fix.

dawidzim:
Hey!
As you suggest I added concatenation in RAM instance:
```
.instr_addr_i   ( {12'h000, instr_addr[RAM_ADDR_WIDTH-1:0]} ),
```

MikeOpenHWGroup (approved):
Thanks @dawidzim.   We all agree this is just an interim fix (we need to resolve all the issues identified by @silabs-PaulZ), but this is step in the right direction.

PR #220, opened 2020-09-07
WorkAround for Riviera-PRO, incompatible array assign

dawidzim:
Hey!
I added WA for Riviera-PRO, this WA is copy from ifdef for `DSIM`

### DIFF
--- a/cv32/tb/uvmt_cv32/uvmt_cv32_iss_wrap.sv
+++ b/cv32/tb/uvmt_cv32/uvmt_cv32_iss_wrap.sv
@@ -55,9 +55,7 @@ module uvmt_cv32_iss_wrap
 
    assign b1.Clk = clknrst_if.clk;
    assign step_compare_if.ovp_cpu_PCr = cpu.PCr;
-`ifndef DSIM
-   assign step_compare_if.ovp_cpu_GPR = cpu.GPR;
- `else
+`ifdef DSIM
    assign step_compare_if.ovp_cpu_GPR[31] = cpu.GPR[31];
    assign step_compare_if.ovp_cpu_GPR[30] = cpu.GPR[30];
    assign step_compare_if.ovp_cpu_GPR[29] = cpu.GPR[29];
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Sorry @dawidzim, I do not follow the purpose of this pull-request.   Maybe it is because I do not know what `WA` means.  :stuck_out_tongue_closed_eyes: 

In any case, the changes you made appear to be logically equivalent:
```
// original code
`ifndef DSIM
    // non-dsim-specific code
`else
   // dsim-specific code
`endif

// pull-request code
`ifdef DSIM
    // dsim-specific code
`else
   // non-dsim-specific code
`endif
```
Am I missing something?   We generally prefer the former style (using `ifndef`) because in some cases (most typically Verilator) we use that construct to exclude functionality altogether (that is, there is no else clause).

dawidzim:
Sorry, `WA` is work around, that was my mental shortcut
Riviera-PRO automatically define _VCP so in pull request I used:
```
`ifdef DSIM
    dsim code
`elsif _VCP
   riviera code
`else
non dsim/riviera code
`endif
```
I can change this to `ifndef`

danielmlynek:
In 1st place - this tool dependent ifdefs are needed because you are assigning array which are not assignment compatible according LRM. Maybe instead of workarounds for more strict tools you could fix the root of the issue and make those array item type to be of equivalent types. Now they are logic vs bit.

As per proposed change  I think that you are missing that elsif is used there. I think that this that it will work as you want if nor DSIM or _VCP will be defined then tool agnostic code will be used.:

`ifdef DSIM
    //dsim specific code
`elsif _VCP
   //riviera specific code
`else
    // tool-agnostic code
`endif

Maybe it can be written in more elegant way by using `define to get rid of code duplication for DSIM and _VCP:
`define DSIM_VCP_CODE <.....all those assignemts.... >
and the using `DSIM_VCP_CODE under DSIM _VCP ifdefs.

MikeOpenHWGroup (approved):

PR #222, opened 2020-09-08
fix compressed instruction decoding from ISS

strichmo:
Add UVM warnings for unmapped instructions
Add pseudo-op conversion to ISA instructions
Fill all known holes in missing instructions from the ISS
Includes update of cv32e40p repo hash to include compressed hint signal from tracer, (required)

### DIFF
--- a/cv32/sim/uvmt_cv32/Makefile
+++ b/cv32/sim/uvmt_cv32/Makefile
@@ -53,7 +53,7 @@ MKDIR_P = mkdir -p
 
 # Compile compile flags for all simulators (careful!)
 WAVES        ?= 0
-SV_CMP_FLAGS ?= "+define+CV32E40P_ASSERT_ON"
+#SV_CMP_FLAGS ?= "+define+CV32E40P_ASSERT_ON"
 TIMESCALE    ?= -timescale 1ns/1ps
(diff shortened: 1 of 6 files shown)


cv32/sim/uvmt_cv32/Makefile
```diff
@@ -53,7 +53,7 @@ MKDIR_P = mkdir -p
 
 # Compile compile flags for all simulators (careful!)
 WAVES        ?= 0
-SV_CMP_FLAGS ?= "+define+CV32E40P_ASSERT_ON"
+#SV_CMP_FLAGS ?= "+define+CV32E40P_ASSERT_ON"
```
MikeOpenHWGroup, line 56: Hmmm.  Why is this commented-out?

MikeOpenHWGroup (changes requested):

cv32/sim/uvmt_cv32/Makefile
strichmo: Trying to run without a bunch of errors.

strichmo: I'll get rid of it in this PR.

MikeOpenHWGroup (approved):

PR #223, opened 2020-09-08
ensure uvm_macros.svh is always sourced first in Xcelium

strichmo:


MikeOpenHWGroup (approved):
Looks good, thanks @strichmo

PR #226, opened 2020-09-10
remove unnecessary /*sparse*/

dawidzim:
Hey!
In this pull request I remove unnecessary `/*sparse*/`

### DIFF
--- a/vendor_lib/imperas/design/ram.sv
+++ b/vendor_lib/imperas/design/ram.sv
@@ -27,7 +27,7 @@ module RAM
 );
 
     // Sparse memory supported by all RTL simulators
-    reg /* sparse */ [31:0] mem [bit[29:0]];
+    reg [31:0] mem [bit[29:0]];
 
     Uns32 daddr4, iaddr4;
     Uns32 value;
--- a/vendor_lib/imperas/design/ram.sv
+++ b/vendor_lib/imperas/design/ram.sv
@@ -27,7 +27,7 @@ module RAM
 );
 
     // Sparse memory supported by all RTL simulators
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (commented):
Hi @dawidzim, thanks for the pull-request.   I have requested @eroom1966 to review it.

On the face of it, I see no urgency to remove the comment.  Is `/*sparse*/` used as a tool pragma?

dawidzim:
Hey @MikeOpenHWGroup, yes `/*sparse*/` is pragma used to sparse memory. But when is declared in wrong position like here, causes problems on compilation on our tool

eroom1966 (commented):
If I recall correctly one of the simulators requires the pragma before the declaration, and one of the simulators requires it after the declaration

@dawidzim  have you tested your change on all supported simulators to ensure we still get a sparse rather than dense memory ?

dawidzim:
@eroom1966 I think vsim and vcs and xcelium should work with this change. I don't know what about dsim. I can change this to
```
  reg /* sparse */ [31:0] mem [bit[29:0]]; `ifndef _VCP/* sparse */`endif
```
_VCP is automatically defined by simulator so this will work for us and for other simulators.
What you think about this solution

strichmo (changes requested):
If we are declaring the memory as an associative array, isn't the memory sparse already by definition.  The sparse pragma only has meaning if you are defining a full memory such as:

reg [31:0] mem [1023:0];

By defining a type bit[29:0] the memory has to be "sparse".

dawidzim:
@strichmo Thanks! I didn't notice that the array is associative. So I remove `sparse`

MikeOpenHWGroup (approved):

PR #231, opened 2020-09-12
add build configuration specifications

strichmo:
Build configurations enable a user to customize compile flags for their simulation build.
It also enables user to customize an Imperas ISS IC control file (one is built for each sim run now).  

Example:
% make test TEST=hello-world CFG=no_pulp

Notes:
- configs are in cv32/tests/cfg
- YAML is used for the configuration
- There is a "default.cfg"
- README.md added here:
https://github.com/strichmo/core-v-verif/blob/strichmo/pr/build_cfg/cv32/sim/uvmt_cv32/README.md#build-configurations
- Makes assumption that a single compilation configuration per workspace
- Always dumps ISS configuration via config file.  I like the self-documenting nature of that.

### DIFF
--- a/cv32/sim/uvmt_cv32/README.md
+++ b/cv32/sim/uvmt_cv32/README.md
@@ -93,10 +93,10 @@ Running the envrionment with Metrics [dsim](https://metrics.ca)
 ----------------------
 The command **make SIMULATOR=dsim sanity** will run the sanity testcase using _dsim_.
 <br><br>
-Setting a shell environment variable `SIMULATOR` to "dsim" will also define the
+Setting a shell environment variable `CV_SIMULATOR` to "dsim" will also define the
 Makefile variable SIMULATOR to `dsim` and you can save yourself a lot of typing. For
 example, in a bash shell:
-<br>**export SIMULATOR=dsim**
+<br>**export CV_SIMULATOR=dsim**
 <br>**make sanity**
 <br><br>
 The Makefile for dsim also supports variables to control wave dumping.  For example:
@@ -139,7 +139,7 @@ Current supported simulators: <i>Note that eventually all simulators will be sup
 
 | SIMULATOR | Supported |
 |-----------|-----------|
-|dsim       | No        |
+|dsim       | Yes       |
 |xrun       | Yes       |
 |questa     | Yes       |
 |vcs        | Yes       |
@@ -159,7 +159,7 @@ To run a simulation in interactive mode (to enable single-stepping, breakpoints,
 
 If applicable for a simulator, line debugging will be enabled in the compile to enable single-stepping.
 
-**make hello-world GUI=1**
+**make test TEST=hello-world GUI=1**
 
 ### Set the UVM quit count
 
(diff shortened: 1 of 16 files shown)


cv32/sim/uvmt_cv32/README.md
```diff
@@ -93,10 +93,10 @@ Running the envrionment with Metrics [dsim](https://metrics.ca)
 ----------------------
 The command **make SIMULATOR=dsim sanity** will run the sanity testcase using _dsim_.
 <br><br>
-Setting a shell environment variable `SIMULATOR` to "dsim" will also define the
+Setting a shell environment variable `CV_SIMULATOR` to "dsim" will also define the
 Makefile variable SIMULATOR to `dsim` and you can save yourself a lot of typing. For
 example, in a bash shell:
-<br>**export SIMULATOR=dsim**
+<br>**export CV_SIMULATOR=dsim**
```
MikeOpenHWGroup, line 99: Groan.  I know why you are doing this, but its gonna cause heartburn.   Nevertheless, let's merge it in now and get it over with.   We should prep a slide to explain these changes for the Tuesday status meeting.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -129,6 +129,7 @@ module uvmt_cv32_step_compare
            ignore = 0;
            csr_val = 0;
            case (index)
+             "marchid"       : csr_val = cv32e40p_pkg::MARCHID; // warning!  defined in cv32e40p_pkg
```
MikeOpenHWGroup, line 132: I expected a comment about this!   In my former lives I would have been vilified for this one two fronts:
1. There is a danger of a compile-dependency error.
2. Using RTL package data in a testbench means that the RTL and testbench always agree.

I have no answer for the first one.  I think we are safe for the second since the RM we use has a completely independent configuration for the value of `machid`.

cv32/tb/uvmt_cv32/uvmt_cv32_tb.sv
```diff
@@ -57,6 +58,11 @@ module uvmt_cv32_tb;
    * a few mods to bring unused ports from the CORE to this level using SV interfaces.
    */
    uvmt_cv32_dut_wrap  #(
+`ifdef NO_PULP
```
MikeOpenHWGroup, line 61: Arg.  More compile-time macros.   Don't see a way around it this time...

cv32/tb/uvmt_cv32/uvmt_cv32_tb.sv
```diff
@@ -171,7 +177,9 @@ module uvmt_cv32_tb;
         else if (core_sleep_o_d && irq_enabled)
           irq_deferint <= irq_enabled;
       end
-      assign iss_wrap.b1.irq_i = !iss_wrap.b1.deferint ? irq_deferint : irq_mip;
+
```
MikeOpenHWGroup, line 180: Why is this better?

cv32/tests/cfg/no_pulp.yaml
```diff
@@ -0,0 +1,9 @@
+name: no_pulp
+description: Sets all PULP-related parameters to 0
+compile_flags: >
+    +define+NO_PULP
+    +define+HAHAHA
```
MikeOpenHWGroup, line 5: You either left this here to see if I was watching, or there is an in-house SiLabs joke behind it.

MikeOpenHWGroup (commented):
Looks good.  I left a few comments, but none of them should gate the merge.   I'll push this through MetricsCI just to be safe.

MikeOpenHWGroup (approved):
Looks good - merging...

PR #234, opened 2020-09-14
fix bugs when user does not set makeargs

strichmo:
Simple point fix of cv_regress.


MikeOpenHWGroup (approved):
Subtle, yet important.

PR #236, opened 2020-09-14
XRUN now runs all ci_regression tests

MikeOpenHWGroup:
Easy, one-liner for you @strichmo.  This will add the corev-dv tests to ci_check with xrun.


strichmo (approved):
Verified locally.  

We can also extend this to Questa (vsim) as well.  I will go ahead and approve and issue a second PR for Questa to ci_check.

PR #238, opened 2020-09-15
Strichmo/pr/rand interrupt wfi

strichmo:
Assertion modeling fixes
WFI corev-vdv fixes to avoid test lockups by disabling all MIE interrupts then entering WFI


MikeOpenHWGroup (approved):
Looks good @strichmo, thanks.  This passes a local `ci_check` on my machine, so I'll merge it in now.

PR #244, opened 2020-09-17
fix csr coverage definitions

strichmo:
some pseduo-ops were still included as covergroups
those are now converted to ISA instructions


MikeOpenHWGroup (approved):
Good fix.  MetricsCI shows a slight bump in fcov for csr instructions.

MikeOpenHWGroup (approved):

Issue #239, opened 2020-09-15
Very basic performance counter test


Silabs-ArjanB:
Here is a very limited directed test that was used for debugging https://github.com/openhwgroup/cv32e40p/issues/429. Please use it as you see fit for verification of the performance counters.

[hpmcounter_test.txt](https://github.com/openhwgroup/core-v-verif/files/5224901/hpmcounter_test.txt)

eroom1966:
@Silabs-ArjanB @MikeOpenHWGroup 
I am working with @silabs-oysteink in order to incorporate the hpmcounters into the step/compare
we have initially seen an issue with intra instruction changes to mcycle - this may also affect other counters, so we need to devise a way to handle this efficiently

Let me try to explain the issue.
The RM always lags the RTL by one instruction retirement, the flow looks like this (pseudo-code)
while (1) {
  step_rtl_to_retire()
  step_ovp_to_retire()
  compare_state()
}

in reality, the step_rtl_to_retire() takes N cycles, depending upon the instruction type and whether any exceptions (sync/async) are raised.
Because it takes N cycles the value of mcycle will be modified N times, and we need to know the value of the mcycle which was visible at the point of instruction execution, for example, if we had the instruction
csrr t1, mcycle
the value written to t1 will not be the value of mcycle when the instruction retires, mcycle will have continued incrementing.

We need to inform the RM of the values of these realtime counters as they were 'architecturally visible' at the time of instruction execution - this will allow us to remain in sync

In conclusion, we cannot simply read the values of hpmcounter* at the point of instruction retire, it seems we need the values at the point of instruction execute.

all help gratefully received.

Silabs-ArjanB:
@eroom1966 FYI (not sure if this info will help you): The minstret counter in the RTL increments on 'issue' of an instruction instead of 'retirement'; this is easier to implement as only one instruction will be issued at a time whereas multiple instructions can retire at the same time in the RTL. The difference should however not be noticeable by a program executing on the RTL (as every issued instruction will eventually retire). Of course you are poking directly into some RTL registers, so for you this effect could be visible. Also please note that some fixes have been made but not yet merged: https://github.com/openhwgroup/cv32e40p/pull/504 (but these do not really impact what you are stating above)

Silabs-ArjanB:
[...]

PR #245, opened 2020-09-17
ISS debug support and infrastructure for randomized debug tests

silabs-oysteink:
PR includes:
*Latest updates from Imperas to make the debug_test pass with reference model enabled.
*A new basic uvm agent for generating random debug request
*Modifications to random code generator to enable generation of linkable debug code
*Test/target for randomized debug tests  - not yet passing
*Moved debug_test to programs/custom and included .yaml


cv32/sim/uvmt_cv32/xrun.mk
```diff
@@ -252,13 +252,6 @@ custom: $(XRUN_SIM_PREREQ) $(CUSTOM_DIR)/$(CUSTOM_PROG).hex
 
 ################################################################################
 # Explicit target tests
-debug_test: $(XRUN_SIM_PREREQ) $(CORE_TEST_DIR)/debug_test/debug_test.hex
```
MikeOpenHWGroup, line 255: I do not understand why you would remove this target.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -83,6 +86,16 @@ module uvmt_cv32_step_compare
       end
    endfunction // check_32bit
    
+   bit [31:0] [63:0] mhpmcounter;
+   always @(posedge `CV32E40P_CORE.ex_stage_i.clk) begin
+     if (`CV32E40P_CORE.ex_stage_i.csr_access_i) begin
+       //mhpmcounter[0] = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[0];
```
MikeOpenHWGroup, line 92: If these are not needed, please remove them.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -83,6 +86,16 @@ module uvmt_cv32_step_compare
...
+       //mhpmcounter[2] = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[2];
+       //$display("@csr_access_i=1 mcycle=%08x", mhpmcounter[0]);
```
MikeOpenHWGroup, line 94: I like lots of debug messaging in my code, but hate `$display()`.  If these have value, replace them with `uvm_info()` at an appropriate verbosity level.  Otherwise, delete them.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -130,10 +143,13 @@ module uvmt_cv32_step_compare
            csr_val = 0;
            case (index)
              "marchid"       : csr_val = cv32e40p_pkg::MARCHID; // warning!  defined in cv32e40p_pkg
-             "minstret"      : csr_val = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[`CV32E40P_CORE.cs_registers_i.csr_addr_i[4:0]][31:0];
-             "minstreth"     : csr_val = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[`CV32E40P_CORE.cs_registers_i.csr_addr_i[4:0]][63:32];
-             "mcycle"        : csr_val = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[`CV32E40P_CORE.cs_registers_i.csr_addr_i[4:0]][31:0];
-             "mcycleh"       : csr_val = `CV32E40P_CORE.cs_registers_i.mhpmcounter_q[`CV32E40P_CORE.cs_registers_i.csr_addr_i[4:0]][63:32];
+
+            
+             "mcycle"        : ignore = 1;
```
MikeOpenHWGroup, line 148: I assume this is temporary, pending a resolution to #239.  Is that correct?

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
```diff
@@ -186,9 +211,18 @@ module uvmt_cv32_step_compare
              check_32bit(.compared(index), .expected(iss_wrap.cpu.CSR[index]), .actual(csr_val));
 
         end // foreach (ovp.cpu.CSR[index])
+        
       `endif      
     endfunction // compare
-
+    
+    // RTL->RM CSR : mcycle, minstret, mcycleh, minstreth
```
MikeOpenHWGroup, line 218: Ah, I think I spoke to soon.   Is this the proposed fix from @eroom1966?  To wit:
```
...inject the performance counter CSR values into the RM, immediately prior to using them as part of an assignment to a GPR, this insures that the RTL and RM remain in sync with respect to data propagation.
```

cv32/tb/uvmt_cv32/uvmt_cv32_tb.sv
```diff
@@ -205,6 +207,20 @@ module uvmt_cv32_tb;
         end
       end
 
+      always @(posedge clknrst_if_iss.clk or negedge clknrst_if_iss.reset_n) begin
+        if (!clknrst_if_iss.reset_n) begin
+            iss_wrap.b1.haltreq <= 1'b0;
+        end else begin
```
MikeOpenHWGroup, line 213: Oh my!   This _is_ looking under the covers isn't it?   This will be very brittle if the RTL is updated.

cv32/tests/cfg/default.yaml
```diff
@@ -2,5 +2,5 @@ name: default
 description: Default configuration for CV32E40P simulations
 compile_flags: 
 ovpsim: >
-    --override root/cpu0/marchid=4
-    --showoverrides
+    --override root/cpu/marchid=4
```
MikeOpenHWGroup, line 5: @strichmo may not like this, but we can argue about it later.  :-)

lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
```diff
@@ -113,54 +100,28 @@ endfunction : build_phase
...
-            UVMA_RESET_STATE_PRE_RESET : drv_pre_reset (phase);
-            UVMA_RESET_STATE_IN_RESET  : drv_in_reset  (phase);
-            UVMA_RESET_STATE_POST_RESET: drv_post_reset(phase);
-         endcase
-      end
-   end
-   
+  
+   cntxt.vif.is_active =1;
+
```
MikeOpenHWGroup, line 105: I'm glad you simplified that.  The templates are a bit too complex.

vendor_lib/google/corev-dv/corev_debug_rom_gen.sv
```diff
@@ -0,0 +1,46 @@
+//Copyright 202[x] Silicon Labs, Inc.
```
MikeOpenHWGroup, line 1: 2020?

vendor_lib/google/corev-dv/corev_debug_rom_gen.sv
```diff
@@ -0,0 +1,46 @@
...
+//  
+//Unless required by applicable law or agreed to in writing, software
+//and hardware implementations thereof
+//distributed under the License is distributed on an "AS IS" BASIS,
+//WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, EITHER EXPRESSED OR IMPLIED.
+//See the License for the specific language governing permissions and
+//limitations under the License.
+//
+//
+class corev_debug_rom_gen extends riscv_debug_rom_gen;
```
MikeOpenHWGroup, line 19: Nice!

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -76,6 +79,10 @@ module CPU
         busStep;
     endtask
     
+    task atZero;
```
MikeOpenHWGroup, line 82: Really?  Why do we need this?   And if we need it, why create a task for it?

vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -198,6 +205,15 @@ module CPU
         end
     endfunction
     
+    function automatic void getCSR (input string index, output longint value);
+        if (CSR_rtl.exists(index)) begin
+            value = CSR_rtl[index];
+        end else begin
+            $display("CSR_rtl.exists(%s) = False", index);
```
MikeOpenHWGroup, line 212: Can we have a `uvm_info()` here as well?

MikeOpenHWGroup (commented):
Hi @silabs-oysteink and @eroom1966.  (I am assuming this is a joint effort.)  First of all _great job_ people.  This is not an insignificant amount of work.  @silabs-oysteink, thanks for following the UVM template structure (and I am impressed you followed it without needing support).

Before approving this I am going to merge it all onto a private fork/branch and regress it.   I wil lalso ask @strichmo to review as well.

In the meantime, I have made a few comments and it would be good if you replied.   Thanks!

cv32/sim/uvmt_cv32/xrun.mk
silabs-oysteink: Since I changed the test to use the new yaml setup I figured it would be less to maintain if I removed the explicit target. I think the interrupt test and some of the random tests also disappeared when converted to yaml?

vendor_lib/google/corev-dv/corev_debug_rom_gen.sv
silabs-oysteink: Yes of course, a bit to quick to copy from the template there :D

cv32/tb/uvmt_cv32/uvmt_cv32_tb.sv
silabs-oysteink: I agree, it isn't actually future proof, but for now this was the shortest path to get it up and running. Peeking at the state was chosen to be in line with how the interrupt code drives deferint signal. Using the internal debug_req_pending was to avoid the case where the RTL may miss a request if the debug_req_i is a single cycle. A simple FSM using only the external debug_req and internal state would perhaps be better.

cv32/sim/uvmt_cv32/xrun.mk
MikeOpenHWGroup: OK, I hadn't realized you'd converted to the yaml flow.

cv32/tb/uvmt_cv32/uvmt_cv32_tb.sv
MikeOpenHWGroup: For now we are OK.  In future revisions of this core we shall need to consider some "design for verification" techniques to ease integration of the Reference Model and ensure predictability of core behavior.

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
eroom1966: I have a tidy up in-flight, I was initially trying to ensure that @silabs-oysteink was reproducing my findings

cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv
eroom1966: I think this change will be permanent, and extended to all hpm counters

the strategy regarding performance counters (I believe) is as follows
The Architectural RM, cannot model the intricacies of the hardware performance counters - and in fact should not, as they are not architecturally defined. The registers are defined, but not their meanings/content.

What is important is that the program execution between the RTL and RM does not diverge due to these register values, therefore the approach is that the values used in the hpm registers is provided (injected) from the RTL to the RM, immediately prior to being used as a primary input to any instruction.

I added this behavior into the cv32/tb/uvmt_cv32/uvmt_cv32_step_compare.sv code

Taking the simple example of the mcycle register, when this is updated in the RTL, I monitor the control signal indicating it is being used, and pass the hpm register values into the RM. on the next instruction execution in the RM if we have an instruction, for example

csrr t1, mcycle

then we can be confident that mcycle will have the correctly calculated number of counted clocks written into the t1 register, even though the RM model has no notion of the micro-architectural dependencies stalls, branches, exceptions etc, that produced that value.

I am in the process of extending this to the remaining hpmcounters(3-31) and I have taken @Silabs-ArjanB initial testcase - which actually fails to pass (this is of course because the test itself is expecting some specific behavior which is not occurring), but correctly matches the behavior of the RM,  - more discussion on this topic here https://github.com/openhwgroup/core-v-verif/issues/239

MikeOpenHWGroup:
Good news: a MetricsCI regression of this PR ran cleanly.
Bad news: the debug_test errored out and failed due to the uvm_errors in the ISA functional coverage model.   I was able to "resolve" this by adding the following kludge to `uvme_rv32isa_covg.sv` (in function _get_csr_name_ starting at line 251):
```
            "dcsr", "dpc"    : begin
                `uvm_info("RV32ISA Coverage", $sformatf("get_csr_name(): CSR [%0s] not yet in functional coverage model.", s), UVM_DEBUG)
            end
```
I am aware that the functional coverage for Debug is still a work-in-progress, so if you like, you can push the above (or something equivalent) onto this PR and we are good-to-go pending @strichmo, review.

silabs-oysteink (commented):
> Good news: a MetricsCI regression of this PR ran cleanly.
> Bad news: the debug_test errored out and failed due to the uvm_errors in the ISA functional coverage model. I was able to "resolve" this by adding the following kludge to `uvme_rv32isa_covg.sv` (in function _get_csr_name_ starting at line 251):
> 
> ```
>             "dcsr", "dpc"    : begin
>                 `uvm_info("RV32ISA Coverage", $sformatf("get_csr_name(): CSR [%0s] not yet in functional coverage model.", s), UVM_DEBUG)
>             end
> ```
> 
> I am aware that the functional coverage for Debug is still a work-in-progress, so if you like, you can push the above (or something equivalent) onto this PR and we are good-to-go pending @strichmo, review.

Changes checked in :)

PR #249, opened 2020-09-22
Debug coverage and assertions

silabs-oysteink:


MikeOpenHWGroup (commented):
This all looks good @silabs-oysteink.  There are a few items I am going to ask @strichmo to review before approving.

strichmo (changes requested):
@silabs-oysteink @MikeOpenHWGroup A general request for this PR.  Even though SystemVerilog allows covergroups in modules we'd prefer to have the covergroups in the class-based world.  This enables common coverage control, instance naming and some other standardization.  

This does require some overhead.

The debug covergroups could be placed in a class similar to how it is done in:
cv32/env/uvme_cv32/cov/uvme_interrupt_covg.sv

The other complication is that you will need an interface to get the "signals" from the design into the class/covergroups.  That would require an interface definition connected to the signals you need (in assert module).  Then you could simply add that to the uvma_debug_cntxt_c as a virtual interface.  You can then configure the new debug covergroup class with the cntxt handle of the cv32 as the interrupt_covg does.

Please let me know if I can help plumb this into your PR if something is confusing.

strichmo (approved):
Thanks @silabs-oysteink !  Looks good to me.  @MikeOpenHWGroup should be ready for test and merge.

cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
```diff
@@ -279,5 +279,72 @@ interface uvmt_cv32_step_compare_if;
...
+    input logic [31:0] mcountinhibit_q,
+    input logic [63:0] mcycle,
+    input logic [63:0] minstret,
+    input logic inst_ret,
+    // WFI Interface
+    input logic core_sleep_o,
+
+    input logic csr_access,
+    input logic [1:0] csr_op,
+    input logic [11:0] csr_addr
```
MikeOpenHWGroup, line 339: There is a missing comma at the end of this line.  I'll fix it.

MikeOpenHWGroup (approved):

PR #250, opened 2020-09-22
Add option to select Toolchain from command-line

MikeOpenHWGroup:
Hi @strichmo, as discussed here is an attempt to allow users to select their toolchain **_and_** maintain backward compatibility for Contributors that do not want/need to change:

From the comment block in `core-v-verif/cv32/sim/Common.mk`:
```
# "Toolchain" to compile 'test-programs' (either C or RISC-V Assember) for the
# CV32E40P.   This toolchain is used by both the core testbench and UVM
# environment.  The assumption here is that you have installed at least one of
# the following toolchains:
#     1. GNU:   https://github.com/riscv/riscv-gnu-toolchain
#               Assumed to be installed at /opt/gnu.
#
#     2. COREV: https://www.embecosm.com/resources/tool-chain-downloads/#corev 
#               Assumed to be installed at /opt/corev.
#
#     3. PULP:  https://github.com/pulp-platform/pulp-riscv-gnu-toolchain 
#               Assumed to be installed at /opt/pulp.
#
# If you do not select one of the above options, compilation will be attempted
# using whatever is found at /opt/riscv using arch=unknown.
#
```

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -178,9 +178,10 @@ COREV_MARCH         ?= corev
 PULP_SW_TOOLCHAIN   ?= /opt/pulp
 PULP_MARCH          ?= unknown
 
-RISCV            = /opt/riscv
-RISCV_PREFIX     = riscv32-unknown-elf-
-RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
+CV_SW_TOOLCHAIN  ?= /opt/riscv
+RISCV            ?= $(CV_SW_TOOLCHAIN)
+RISCV_PREFIX     ?= riscv32-unknown-elf-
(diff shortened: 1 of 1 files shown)


strichmo (changes requested):
Could we leave CV_SW_TOOLCHAIN as the overall default.  I have the SiLabs folks using that, as it aligns with other Makefile defaults (such as CV_SIMULATOR).

I am requesting to restore these:
CV_SW_TOOLCHAIN  ?= /opt/riscv
RISCV            ?= $(CV_SW_TOOLCHAIN)

MikeOpenHWGroup:
Good point.  Pushed in an update restoring CV_SW_TOOLCHAIN as the user-override-able default.

strichmo (approved):
Cheers!

PR #252, opened 2020-09-23
update a few copyright notices

nakengelhardt:
I'm not sure if there's even anything left in these makefiles of the code that this copyright notice originally covered, but while it's still there, let's keep it up to date...

### DIFF
--- a/cv32/bsp/syscalls.c
+++ b/cv32/bsp/syscalls.c
@@ -1,7 +1,7 @@
 /* An extremely minimalist syscalls.c for newlib
  * Based on riscv newlib libgloss/riscv/sys_*.c
  *
- * Copyright 2019 Clifford Wolf
+ * Copyright 2019 Claire Wolf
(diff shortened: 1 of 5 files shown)


MikeOpenHWGroup (approved):
Thanks @nakengelhardt.  This raises an interest legal question.   It is my understanding that open-source licenses typically require down-stream modifications to preserve the original license header.  So what happens when the person or organization holding the original copyright changes their name?  Should the headers be updated, or not?

Fortunately, in the engineering circles we travel in (both open and closed source), everyone knows Claire Wolf so I'll merge this in.

nakengelhardt:
Oh right. I did check that the copyright notice itself doesn't have any particular legal value, but I didn't think of the "you must keep this header intact" clause... but I would guess that at worst, it can always be considered a new license grant of the same content by the original author. Since these licenses aren't exclusive, they have the right to license it again under different conditions.

PR #255, opened 2020-09-24
Updated various files because the register file bist wrapper has been…

Silabs-ArjanB:
… removed, pointed to latest RTL


MikeOpenHWGroup (approved):
Thanks @Silabs-ArjanB, I appreciate you fixing all those XMRs.

PR #257, opened 2020-09-27
add the interrupt_bootstrap directed test to verify mtvec_addr_i pins

strichmo:
Very simple directed test of mtvec_addr_i.  

Also adds following:

- Adds plusarg hook to set mtvec_addr_i as a short-term fix to set pins using existing interface initial block
- Add common makefile target to allow tests with both assembly and C source.  The bootstrap test uses this to implement a different start routine to avoid BSP's setting of mtvec.

### DIFF
--- a/cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
+++ b/cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
@@ -174,6 +174,20 @@ interface uvmt_cv32_core_cntrl_if (
     dm_exception_addr = 32'h1A11_1000;
     hart_id           = 32'h0000_0000;
 
+    // If a override is provided via plusarg then set bootstrap pins and adjust ISS model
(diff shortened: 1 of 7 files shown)


cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
```diff
@@ -174,6 +174,20 @@ interface uvmt_cv32_core_cntrl_if (
     dm_exception_addr = 32'h1A11_1000;
     hart_id           = 32'h0000_0000;
 
+    // If a override is provided via plusarg then set bootstrap pins and adjust ISS model
+    if ($value$plusargs("mtvec_addr=0x%x", mtvec_addr));
```
MikeOpenHWGroup, line 178: Clever!

MikeOpenHWGroup (approved):
Good contribution @strichmo.  I especially like the plusargs that throw overrides for OVPsim.

PR #260, opened 2020-09-29
debug covergroup fixes

strichmo:
adding a clocking block and rework the virtual interface on the debug… coverage to avoid Questa failure

### DIFF
--- a/cv32/env/uvme_cv32/cov/uvme_debug_covg.sv
+++ b/cv32/env/uvme_cv32/cov/uvme_debug_covg.sv
@@ -305,8 +305,6 @@ endclass : uvme_debug_covg
 function uvme_debug_covg::new(string name = "debug_covg", uvm_component parent = null);
     super.new(name, parent);
 
-    $display("Broke Questa");
-
     cg_debug_mode_ext = new();
     cg_ebreak_execute_with_ebreakm = new();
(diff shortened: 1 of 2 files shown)


cv32/env/uvme_cv32/cov/uvme_debug_covg.sv
```diff
@@ -298,6 +305,8 @@ endclass : uvme_debug_covg
 function uvme_debug_covg::new(string name = "debug_covg", uvm_component parent = null);
     super.new(name, parent);
 
+    $display("Broke Questa");
```
silabs-oysteink, line 308: Is this needed? If so, maybe `uvm_info would be better?

silabs-oysteink (approved):

cv32/env/uvme_cv32/cov/uvme_debug_covg.sv
strichmo: Ha!  That's a breadcrumb I was using to debug the actual problem.  I will remove it.

strichmo: Good catch.

MikeOpenHWGroup (approved):

MikeOpenHWGroup (approved):
Thanks for the quick fix @strichmo.  Merging...

PR #261, opened 2020-09-29
Update ISS/RM to support the WARL register fields defined for tdata1/dcsr

eroom1966:
this changes the behavior of some register fields in tdata1 and dcsr from R/W to WARL
This brings the RM into line with the latest RTL changes


MikeOpenHWGroup (approved):

MikeOpenHWGroup (approved):
Thanks @eroom1966, I will merge this in.

PR #271, opened 2020-10-03
enable full no_pulp regression with refined coverage in Xcelium

strichmo:


MikeOpenHWGroup (approved):
Nice cleanup of the RV32ISA coverage, thanks!    Obviously the `vRefine` files are tool-specific, but that is OK.  I'll conform that they do not impact dsim.

PR #272, opened 2020-10-05
CVA6 testbench 20201005

ASintzoff:
This is a testbench for CVA6 CPU.
It supports 64- and 32-bit versions of CVA6.
The cva6 testbench is located in a dedicated directory (cva6) and does not interfere with existing code (e.g. cv32).


MikeOpenHWGroup (changes requested):
Thanks for this PR @ASintzoff.  This is a big day for OpenHW to start integrating the verification environment for CVA6 into core-v-verif.

Maybe it would be a good idea to move the files in `tests` to `cva6/tests`.  I like the idea of having a common set of tests that would run on _any_ core and this is a good reason to put the compliance tests in a common directory.  However at this time it seems the CV32E4 and CVA6 integrate these tests differently, so for now I suggest moving the CVA6 version of these tests under the cva6 directory.

ASintzoff:
> Maybe it would be a good idea to move the files in `tests` to `cva6/tests`. I like the idea of having a common set of tests that would run on _any_ core and this is a good reason to put the compliance tests in a common directory. However at this time it seems the CV32E4 and CVA6 integrate these tests differently, so for now I suggest moving the CVA6 version of these tests under the cva6 directory.

@MikeOpenHWGroup : Update done according your suggestion.

ASintzoff:
> > Maybe it would be a good idea to move the files in `tests` to `cva6/tests`. I like the idea of having a common set of tests that would run on _any_ core and this is a good reason to put the compliance tests in a common directory. However at this time it seems the CV32E4 and CVA6 integrate these tests differently, so for now I suggest moving the CVA6 version of these tests under the cva6 directory.
> 
> @MikeOpenHWGroup : Update done according your suggestion.

@MikeOpenHWGroup Sorry, I miss some updates in the paths after the move of tests directory.

MikeOpenHWGroup (approved):
Thanks @ASintzoff!

PR #280, opened 2020-10-08
Bigger CSR access test

MikeOpenHWGroup:
Bigger version of the CSR access test.  Doesn't test the access modes of the CSRs any better, but it uses all available GPRs to do it, so it should produce a small bump in ISA Fcov.


strichmo (approved):
Thanks for incorporating my ifdef into your flow.  Pulled it and confirmed it works with no_pulp configs.

PR #282, opened 2020-10-12
Mcause

eroom1966:
@MikeOpenHWGroup 
this PR implements the update to the MCAUSE/EXCEPTION_CODE WLRL field
this is now produced as a masked value to only include bit[31] bit[4:0]


MikeOpenHWGroup (approved):
Thanks @eroom1966.  This isn't exactly the outcome we were hoping for, but it does make the RTL of the core and the ISS reference model bug-for-bug compatible.

PR #285, opened 2020-10-14
Support for multiple GCC toolchains

MikeOpenHWGroup:
Hi @strichmo, this is a high-urgency PR, please review as soon as you get a chance.

To mimic what we'll have on the MetricsCI system, I have installed the PULP toolchain at **/opt/riscv** and the [COREV toolchain](https://www.embecosm.com/resources/tool-chain-downloads/#corev) (from Embecosm) at **/opt/corev** on my local machine.  With these updates I can use the COREV variable to select the COREV toolchain for any testcase.   For example, the following does what you'd expect:
```
$ make test TEST=interrupt_test COREV=YES
```
Note that using the PULP toolchain, the `interrupt_test` test-program does not compile.

So, this all works for me locally.   I want to ensure that it doesn't break anything you have.   I'd also like to know why its necessary to pass variables explicitly to the `bsp` Makefile.

### DIFF
--- a/cv32/bsp/Makefile
+++ b/cv32/bsp/Makefile
@@ -21,3 +21,12 @@ $(LIBCV-VERIF): $(OBJ)
 
 clean:
 	rm -f $(OBJ) $(LIBCV-VERIF)
+
+
+vars:
+	@echo "make bsp variables:"
+	@echo "   CV_SW_TOOLCHAIN  = $(CV_SW_TOOLCHAIN)"
+	@echo "   RISCV            = $(RISCV)"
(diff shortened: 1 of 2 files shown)


strichmo (approved):
Changes look fine to me.

strichmo:
@MikeOpenHWGroup  To answer your question it has to do with exporting (in sh-ish environment, of which make is sh-based by default).  Variables in the envrionment can be exported (i.e. sub-shells inherit) or non-exported (not-inherited).  The rules are a bit funky...
For example on the make command line, any variable specified there are exported.
make A=3
# A is exported

But a simple variable assignment in a Makefile
A?=3
all:
  make -C bsp

will not export that A automatically.  You would need a "export" keyword there.

PR #289, opened 2020-10-16
enable random illegal instruction generation in core-dv

strichmo:


cv32/tests/programs/corev-dv/corev_rand_instr_test/corev-dv.yaml
```diff
@@ -5,7 +5,7 @@ uvm_test: corev_instr_base_test
 description: >
     RISCV-DV generated random instruction test
 plusargs: >
-    +instr_cnt=10000
+    +instr_cnt=50000
```
MikeOpenHWGroup, line 8: At this point we should probably crank this up for all corev/riscv-dv generated tests.   I'll do that on my next PR.

MikeOpenHWGroup (approved):

PR #290, opened 2020-10-17
Integrate RTL hash a26b194

MikeOpenHWGroup:
Hi @silabs-oysteink and @strichmo.  Here is my attempt to update the environment to point to the latest RTL (hash a26b194).  In the past week there have been some surprisingly large updates to the RTL and this had a big impact on the `debug_cov_assert_if` in the tb.  The `inst_ret` signal in the cs_registers module is gone, but fortunately the `id_valid` and `is_decoding` signals generated in the core module are still there so I created a logically equivalent signal that you'll spot right away.

This compiles and runs a ci_check regression without issue.  However, I am not in a position to understand if that `inst_ret` signal is still doing what the debug assertions assume it is doing.  @silabs-oysteink, you'll need to drill into that.

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -68,13 +68,17 @@ BANNER=*************************************************************************
 
 CV32E40P_REPO   ?= https://github.com/openhwgroup/cv32e40p
 CV32E40P_BRANCH ?= master
+#2020-10-14
+CV32E40P_HASH   ?= a26b194
 #2020-10-08
-CV32E40P_HASH   ?= f6196bf
+#CV32E40P_HASH   ?= f6196bf
 
 FPNEW_REPO      ?= https://github.com/pulp-platform/fpnew
 FPNEW_BRANCH    ?= master
+#2020-10-05
+FPNEW_HASH      ?= 5b2a9d4
 #2020-09-23
-FPNEW_HASH      ?= a0c021c360abcc94e434d41974a52bdcbf14d156
+#FPNEW_HASH      ?= a0c021c360abcc94e434d41974a52bdcbf14d156
 
 RISCVDV_REPO    ?= https://github.com/google/riscv-dv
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup:
Update: as suspected, the simply hack I put in to get the testbench to compile with the latest RTL does not work under all conditions.    When running either **debug_test**, **counters**, **perf_counters_instructions** or **cv32e40p_csr_access_test**, you will see the following:

```
UVM_ERROR @ 2374.800 ns : uvmt_cv32e40p_debug_assert.sv(283) reporter [CV32E40P_DEBUG_ASSERT] Minstret not counting when mcountinhibit[2] is cleared!
```
The time of the error is specific to each test - this is the first error from debug_test.  I have not investigated (might have time this evening).

silabs-oysteink:
I believe this line from cs_registers may help you:

>assign hpm_events[1]  = mhpmevent_minstret_i;                          // instruction counter

I guess using either hpm_events[1] or mhpmevent_minstret_i should do the trick.

MikeOpenHWGroup:
Thanks @silabs-oysteink, that works.  I've pushed it to this PR, please have a look.  Even though it works for our current regression, I am not certain it will work in the general case.

silabs-oysteink (approved):

Silabs-ArjanB:
Hi @MikeOpenHWGroup , @silabs-oysteink The mhpmevent_minstret_i replaces the former inst_ret signal from cv32e40p_cs_registers.sv. It is not logically equivalent because the retired instruction count was wrong and needed to be fixed. The new signal is different in the sense that illegal instructions, ecall, ebreak are not counted as retired instructions (see https://github.com/openhwgroup/cv32e40p/issues/533).

silabs-oysteink:
I guess this will probably break a couple of assertions, but they should be easy to fix once we get everything merged together.

strichmo (approved):

PR #295, opened 2020-10-21
added uvm macros, correct Makefile

dawidzim:
Hey!
In this pull request i added missing files in Makefile, also I added uvm to compilation for Riviera-PRO and `uvm_macros.svh` in `mm_ram.sv`


MikeOpenHWGroup (approved):
Thanks @dawidzim.  I am not sure how/why the inclusion of the uvm macros in mm_ram was not noticed before - maybe its simulator specific.  In any case, this is a good update.

PR #300, opened 2020-10-22
add in the single_step and debug ebreak tests

strichmo:


MikeOpenHWGroup (changes requested):
Hi @strichmo.  This is an obvious one to approve.  Could you rename `cv32_full_covg.no_pulp.yaml` to (say) `cv32_full_covg_no_pulp.yaml`?  The double extension confuses the Metrics flow.  My next PR would have this in there, but we may as well do it now.

strichmo:
Good call...pushed a change.

MikeOpenHWGroup (approved):

PR #307, opened 2020-10-26
fix the clean_all targets to remove core tests and add clean_compliance

strichmo:


MikeOpenHWGroup (approved):
Very proactive - thanks.

PR #311, opened 2020-10-28
Update Core TB and UVM env to match updated APU port map

MikeOpenHWGroup:
Update to match the new APU port map.


strichmo (approved):
All looks good.

PR #312, opened 2020-10-29
Improved randomization control

MikeOpenHWGroup:
Changes up the randomization of the memory stalls a little bit and updates the dsim makefile to run with random seeds by default.


strichmo (commented):
Agree with adding more flavors of randomization (i.e. knobs), although wouldn't it be nice to have this in a proper class someday!

However, I did over-bias no stalls in my original code for a reason.  Given that we turn off some checking when in "stall" mode I wanted to at least maintain 50-50 random testing distributions.

MikeOpenHWGroup:
Good catch @strichmo, I admit that I did not have CSR checking in mind when I turned up the bias toward stalls.   Restored.

I did keep the "max data, zero instruction" configuration as the fence.i test doesn't actually verify anything without it.   That should be safe as it is an explicit knob that will likely only be used for a single testcase.

> wouldn't it be nice to have this in a proper class someday!

I have an action from Rick to draw up a roadmap for core-v-verif - that will be on the list for sure.

strichmo (approved):

PR #317, opened 2020-10-30
Debug coverage and assertions update

silabs-oysteink:
Some bugfixes in cg/assertions and addition of some missing ones.
Added directed cases to debug_test to close some coverage holes.


strichmo (approved):
Changes look OK.

PR #323, opened 2020-11-02
random debug test workarounds

strichmo:
Rework corev-dv to avoid illegal ebreaks and drets in main program
ensure ebreakm is set when ebreak is encountered in main program
add an interrupt+debug request test

PR #324, opened 2020-11-02
random debug and interrupt test fixes

strichmo:
addressing comment from https://github.com/openhwgroup/core-v-verif/pull/323 regarding interrupt corev-dv test, I fixed inadvertent knob resetting of illegal instruction ratio
fix bug in step_compare where if interrupt and debug mode occur on same instruction retire, defer the interrupt until after debug mode finishes
add plusarg to disable step-compare check of dpc when running random interrupts and debug requests, will create issue with issue


MikeOpenHWGroup (approved):
Looks fine and runs through ci_check.

PR #335, opened 2020-11-05
Workaround for issue 294

masgia:


MikeOpenHWGroup (approved):
Thanks @masgia.  The updates to `uvmt_cv32_tb_ifs.sv` are ugly (/\*logic\*/), but nevertheless work for me on Cadence Xcelium (xrun) and Metrics dsim.

I am going to ask @strichmo to have a look before we merge this in.    Assuming we accept this PR, it might make sense to remove the logic token altogether.

silabs-oysteink (approved):
Looks good. It could be an idea to add a comment in the affected interface describing why this change has been made - for future reference.

strichmo (approved):

strichmo:
Thanks Max for this PR.  I'm going to merge, but will issue another PR later today to remove all "logic" on inputs.  There should be no issue for Xcelium, Questa nor Dsim to make these wires and will look much cleaner.

PR #341, opened 2020-11-05
Silabs oysteink debug cov update

silabs-oysteink:
Added directed tests for missing debug coverage.
Split debug_test in two - putting most trigger related tests into debug_test_trigger.
Not that this new test requires gcc 9.3.0 to get the correct timing needed.
The test is also known to fail in step-and-compare at the moment, so it should be run with USE_ISS=NO

When the current failures with ISS are fixed, we need to run with RTL from PR #569 or later, as this introduced changes to the tracer to allow for ebreak instructions at a trigger address not to retire.

ci_check has been run and passes for me.


MikeOpenHWGroup:
Thanks for this @silabs-oysteink.  I'm going to punt this one over to @strichmo for review.   He is more familar with debug verif on this project _and_ I am currently without an ISS on any of my machines that run dsim because of an IT issue (should be back on my feet tomorrow).

strichmo (approved):

PR #345, opened 2020-11-09
Added check that expected # of UVM tests actually ran

GTumbush:
modified hello-world.c so it would not compile.  ci_check catches that this test did not run and reports failure.

### DIFF
--- a/ci/ci_check
+++ b/ci/ci_check
@@ -40,7 +40,6 @@
 #         "cmd" key.
 #      3. Handle Verilator a little more elegantly.
 #      4. Terminate if compile fails.
-#      5. Check that all specified tests ran.
 ################################################################################
 
 import json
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (changes requested):
Thanks for taking this on Greg.  This is a big step in the right direction.  A couple of additional things I'd ask for:

1. Item #5 in the TODO list in the comment header should be removed.
2. When I ran `./ci_check -s dsim` on a machine that does not have dsim installed the script hangs because of the brain-dead method I use to check for failure.

GTumbush:
Deleted Note #5 and added check that specified simulator exists

MikeOpenHWGroup (approved):
Awesome @GTumbush, thanks much.

PR #348, opened 2020-11-10
many interrupt functional coverage fixes

strichmo:


MikeOpenHWGroup (approved):
Wow that `corev_ecall_instr_stream` is a bit of a sledge hammer.

PR #351, opened 2020-11-10
Filling coverage hole for illegal_insn + debug_req

silabs-oysteink:
Added directed test and assertion for debug_req_i while executing an illegal instruction. depc is set to entry point of exception handler before debug entry. Test is part of debug_test_trigger. Had to alter timings for the trigger tests after this change.


MikeOpenHWGroup (changes requested):
Hi @silabs-oysteink, this submission looks good but the debug_test_trigger test still fails.  Is that expected?

silabs-oysteink:
It may fail if it is not compiled with gcc 9.3.0, as different compiler versions may alter the timing of the debug_req and irq signals from the virtual peripheral.

Given the correct gcc, it will still fail with ISS due to issue #353 and cv32e40p issue(s) #548 and #549. #353 has a fix in the last PR from Lee, but I don't know about the other two.

silabs-oysteink:
Do no merge yet, I will move the last subtest into a separate test.

MikeOpenHWGroup:
OK, thanks for the heads up.  I added a "DO NOT MERGE" label to minimize the chances that I'll merge this by mistake.  Please remove that label when you are ready.

MikeOpenHWGroup (approved):

PR #358, opened 2020-11-11
fixes corner cases in TB with irq wakeup and debug_req wakeup from wfi

strichmo:
also turn down some test times on the interrupt random tests


MikeOpenHWGroup (approved):
Looks good.   As discussed, it would be useful to add some comments to describe the deferred interrupt and debug logic.  In the interests of time, I'll merge these changes in now (pending a ci_check regression running now).

PR #359, opened 2020-11-11
Use state machine in step & compare instead of asynchronous always blocks

GTumbush:
Updated uvmt_cv32_step_compare.sv to use a state machine instead of asynchronous always blocks. Passes ci_check in lee_moore branch that runs normal regression tests and in addition interrupt_test, corev_rand_interrupt_0, and corev_rand_interrupt_1. All pass.

Note, this branch is from https://github.com/GTumbush/core-v-verif/tree/lee_moore hash 12b6874 with log "Move stop/start_clk to state machine". It does not contain Lee's latest modifications to this file. It does not solve issues https://github.com/openhwgroup/cv32e40p/issues/548 and https://github.com/openhwgroup/cv32e40p/issues/549 . Since these issues do not gate RTL freeze I want to get this upgrade in master ASAP.

Please ignore changes to corev_asm_program_gen.sv I do not know why this is flagged as changed.


MikeOpenHWGroup (approved):
Ah, that's better.  Looks good and `ci_check` likes it, so I'm merging it in.  Thanks @GTumbush

MikeOpenHWGroup (approved):

PR #365, opened 2020-11-12
Move to latest RTL

Silabs-ArjanB:
Moving to the latest RTL

By the way: @MikeOpenHWGroup, maybe as a side effect of the previous RTL commit  (that added the names to generated blocks) the issues mentioned in https://github.com/openhwgroup/cv32e40p/issues/572 have been solved (as I had to rewrite the blocks mentioned by the dsim warnings). I have no way to check with dsim myself though.

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -72,7 +72,7 @@ export SHELL = /bin/bash
 
 CV32E40P_REPO   ?= https://github.com/openhwgroup/cv32e40p
 CV32E40P_BRANCH ?= master
-CV32E40P_HASH   ?= edfe222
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for this @Silabs-ArjanB.

Just for future reference, it is not wrong to use the entire hash when pointing to the cv32e40p, but in general the first 7 digits of the hash are sufficient.

Also, thanks for the "free" clean-up of the lint (non) issue.  I can confirm that dsim is much happier now.  BTW, I filed that at an issue with Metrics and they agreed that they have a bug.   Obviously it does not gate us in any way.

PR #367, opened 2020-11-13
Update to fill coverage holes

silabs-oysteink:
Added covergroups for ebreak + dcsr.ebreakm==0
Fixed bad assertions for c/ebreak
Excluded PULP-only states from cross between FSM state and debug_req_i.
Rewrote the code to assert debug_req at reset.
Added a new directed test to hit debug_req_i at state BOOT_SET. This uses the same code as debug_test_reset, but a different yaml to set an extra plusarg.


silabs-oysteink:
I added another virtual sequencer and randomized debug_test_boot_set. Please check carefully if this is correct, seems ok to me.

MikeOpenHWGroup (approved):
Looks good @silabs-oysteink.  Due to the changes to the way debug reset is driven I was a bit more cautious with this one and ran a mini regression (more than just ci_check) before merging it in.

PR #370, opened 2020-11-13
fix jalr_cg, was missing the zero bin

strichmo:
was also computing imm as pc-relative but this was not correct

### DIFF
--- a/cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
+++ b/cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
@@ -589,7 +589,7 @@ class uvme_rv32isa_covg extends uvm_component;
         cp_rs1   : coverpoint get_gpr_name(ins.ops[2].val, ins.ops[2].key, "jalr") {
             bins gprval[] = {[zero:t6]};
         }        
-        cp_offset: coverpoint get_pc_imm(ins.ops[1].val, ins.pc, "jalr") {
+        cp_offset: coverpoint get_imm(ins.ops[1].val, "jalr") {
             bins neg  = {[$:-1]};
             bins zero = {0};
             bins pos  = {[1:$]};
@@ -1412,7 +1412,7 @@ class uvme_rv32isa_covg extends uvm_component;
             bins gprval[] = {[s0:a5]};
         }
         cp_shamt5   : coverpoint get_imm(ins.ops[2].val, "c.slli" ) {
-            bins zero = {0};
+            // Note that zero is reserved (used for HINTs)
             bins pos  = {[1:$]};
         }
     endgroup
@@ -1427,7 +1427,7 @@ class uvme_rv32isa_covg extends uvm_component;
             bins gprval[] = {[s0:a5]};
         }
         cp_shamt5   : coverpoint get_imm(ins.ops[2].val, "c.srli" ) {
-            bins zero = {0};
+            // Note that zero is reserved (used for HINTs)
             bins pos  = {[1:$]};
         }
     endgroup
@@ -1444,7 +1444,7 @@ class uvme_rv32isa_covg extends uvm_component;
             bins gprval[] = {[s0:a5]};
         }
         cp_shamt5   : coverpoint get_imm(ins.ops[2].val, "c.srai" ) {
-            bins zero = {0};
+            // Note that zero is reserved (used for HINTs)
             bins pos  = {[1:$]};
         }
     endgroup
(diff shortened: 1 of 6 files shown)


cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
```diff
@@ -589,7 +589,7 @@ class uvme_rv32isa_covg extends uvm_component;
         cp_rs1   : coverpoint get_gpr_name(ins.ops[2].val, ins.ops[2].key, "jalr") {
             bins gprval[] = {[zero:t6]};
         }        
-        cp_offset: coverpoint get_pc_imm(ins.ops[1].val, ins.pc, "jalr") {
+        cp_offset: coverpoint get_imm(ins.ops[1].val, "jalr") {
```
MikeOpenHWGroup, line 592: Good catch!   Pretty sure this was my bad.

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
```diff
@@ -1412,7 +1412,7 @@ class uvme_rv32isa_covg extends uvm_component;
             bins gprval[] = {[s0:a5]};
         }
         cp_shamt5   : coverpoint get_imm(ins.ops[2].val, "c.slli" ) {
-            bins zero = {0};
+            // Note that zero is reserved (used for HINTs)
             bins pos  = {[1:$]};
```
MikeOpenHWGroup, line 1416: Yes, but don't we want to cover HINTs?

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
```diff
@@ -1844,15 +1843,17 @@ class uvme_rv32isa_covg extends uvm_component;
...
-                    `uvm_info("RV32ISA Functional Coverage", $sformatf("jalr_cg: ins.ops[0].val = %0s, ins.ops[1].val = %0s, ins.ops[2].val = %0s",
-                                                                       ins.ops[0].val, ins.ops[1].val, ins.ops[2].val), UVM_DEBUG)
-                    // If operand1 is a consant (C) then assume operand1 is r0
-                    // and move the constant to operand2 to maintain consistent cg interface
-                    //if (ins.ops[1].key[0]) begin
-                    //    `uvm_info("RV32ISA Functional Coverage", $sformatf("jalr_cg: ins.ops[1].key[0] = %0s", ins.ops[1].key[0]), UVM_DEBUG)
-                    //    ins.ops[2] = ins.ops[1];
-                    //    ins.ops[1].key = "R:"; ins.ops[0].val = "zero";
-                    //end
+                    // Usually the Decoder from ISS presents all three operands such as R1:s6 C:785 R2:s6
```
MikeOpenHWGroup, line 1846: How do you know this?

MikeOpenHWGroup (changes requested):
Couple of questions about this one - see my review comments.

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
strichmo: Examined lots of logs.  Ideally we would put asserts on the output of 3rd party ISS we don't have direct control of stimulus, but I can't say that is worth the effort to implement.  Yet another reason we should be pulling our functional coverage from our own monitor/tracer so we can append appropriate assertion code.

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
strichmo: For all of the standard (RV32I) hints we get coverage through bins in the standard ISA instructions.  (Hints in RV32I are a set of computational instructions with rd=x0).

For RV32C I just added the zero register to appropriate C instructions that may encode hints so we will get coverage.

I also added hint ratio into the rand_instr and rand_instr illegal tests to generate these hints.

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
MikeOpenHWGroup: > Yet another reason we should be pulling our functional coverage from our own monitor/tracer so we can append appropriate assertion code.

Absolutely agreed!

strichmo:
I force-pushed an updated branch, rebased to master to alleviate merging issues.

cv32/env/uvme_cv32/cov/uvme_rv32isa_covg.sv
MikeOpenHWGroup: I understand and agree with your first two points.   However, I do not see why the immediate value in `c.slli rd, uimm` cannot be zero for non-hint instructions.  Of course,  `c.slli x8 0` won't actually _do_ anything, but isn't that the point?

In the interests of moving forward quickly, I'll approve/merge this PR and open an issue.

MikeOpenHWGroup (approved):

PR #376, opened 2020-11-17
CSR inst fcov

MikeOpenHWGroup:
This will bring CSR instruction fcov to 100%
image


strichmo (approved):
Thanks for getting these closed....looks good!

PR #379, opened 2020-11-17
coverage closure fixes

strichmo:
address riscv-dv randomization issue with negative numbers and c.lui
add vrefines for unused rounding add and mult values
add corev-dv interrupt tests to hit c.lw and c.sw more often


MikeOpenHWGroup (approved):
Looks good Steve.   ci_check passes with both xrun and dsim.

PR #387, opened 2020-11-19
update path for core tests

dawidzim:
Hey!
In this pull request i correct path to CORE_TEST and added core tests for Riviera-PRO:
core_tests                  = ['misalign', 'illegal', 'dhrystone', 'fibonacci', 'riscv_ebreak_test_0']
Signed-off-by: Dawid Zimonczyk <dawidz@aldec.com.pl>

### DIFF
--- a/cv32/sim/Common.mk
+++ b/cv32/sim/Common.mk
@@ -203,7 +203,7 @@ ASM_DIR   ?= $(ASM)
 # Note that the DSIM targets allow for writing the log-files to arbitrary
 # locations, so all of these paths are absolute, except those used by Verilator.
 # TODO: clean this mess up!
-CORE_TEST_DIR                        = $(PROJ_ROOT_DIR)/cv32/tests/core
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup (changes requested):
Hi @dawidzim.  I am happy to see the support of the "core" testbench with Riviera-PRO.  Rather than adding additional tool-specific targets for each test-program, it would be better to have a generic target that works for any test-program.   For example with Verilator these targets work:
```
make custom CUSTOM_PROG=misalign
make custom CUSTOM_PROG=illegal
```
I think it would be better to have similar targets for Riviera-PRO.

Better yet, its on the TODO list (but not gating RTL Freeze for CV32E40P) to have the core testbench work with the same targets as the UVM environment.

dawidzim:
Hey @MikeOpenHWGroup 
I added option `riviera-custom`, so if we want run test on Riviera-PRO we should do:
```
make riviera-custom CUSTOM_PROG=illegal
```

MikeOpenHWGroup (approved):

PR #389, opened 2020-11-19
Remove rand weights of dist

dawidzim:
Hey!
I remove rand weights of dist
Signed-off-by: Dawid Zimonczyk <dawidz@aldec.com.pl>

### DIFF
--- a/vendor_lib/google/corev-dv/corev_interrupt_csr_instr_lib.sv
+++ b/vendor_lib/google/corev-dv/corev_interrupt_csr_instr_lib.sv
@@ -32,25 +32,31 @@ class corev_interrupt_csr_instr_stream extends riscv_load_store_rand_instr_strea
   } interrupt_csr_action_enum;
 
   rand interrupt_csr_action_enum action;
+`ifdef _VCP 
+  int unsigned wgt_random_mie = 1;
+  int unsigned wgt_random_mstatus_mie = 3;
+`else
   rand int unsigned wgt_random_mie;
   rand int unsigned wgt_random_mstatus_mie;
+`endif
   rand reg[31:0] rand_mie_setting;
 
+`ifndef _VCP
   constraint ordering_wgt_c {
     solve wgt_random_mie before action;
     solve wgt_random_mstatus_mie before action;
   }
-
+`endif
   constraint dist_action_c {
     action dist  { RANDOM_MIE :/ wgt_random_mie,
                    RANDOM_MSTATUS_MIE :/ wgt_random_mstatus_mie };
   }
-
+`ifndef _VCP
   constraint default_wgt_c {
     soft wgt_random_mie == 1;
     soft wgt_random_mstatus_mie == 3; 
   }
-
+`endif
   riscv_instr_name_t allowed_mie_instr[$] = {CSRRW, CSRRC, CSRRS};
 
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup (changes requested):
I am not sure this is a god idea @dawidzim.  We have spent considerable time tuning corev-dv (an extension of Google's riscv-dv) to reliably achieve high functional and code coverage in regression.

I will ask @strichmo, our resident expert on riscv-dv, to review this pull-request.

strichmo (changes requested):
I'm not sure why the rand qualifier on the weights are removed.  Is it an issue with the simulator or an issue with code-style/functionality?

danielmlynek:
It is both coding style and simulator issue.
It is not good coding style in my opinion. I rather doubt if you want to have random weights - it would be hard to analyze what would be the final distibution. 
Weights are rather parameters which should be known before randomization.

MikeOpenHWGroup:
> It is not good coding style in my opinion. I rather doubt if you want to have random weights

I cannot comment about this specific instance, but I have often used random weights in past projects.  It can be a powerful method of creating unforeseen stimulus patterns or device configurations.

danielmlynek:
My comment is based on my opinion from simulator point of view. My point is that having random weights cause that in fact the distribution is not definded. Each randomization is done with different weights. If I would like to have random weights I would randomize them (separately) once per some simulation period as a configuration.
I've also noticed that your weights are constrainted to constant values by some soft constraints. I thought that you rather want to have them const. You need to define on them conflicting constraint to make them random.
If you are saying that having random weight is important for verification process - then thank you for your feedback - we will take it into account.

MikeOpenHWGroup:
Thanks for the additional information @dawidzim.  When it comes to such matters, there is the question about what the language simulator should support and how the coder should use the language.

Both randomization of weights and soft constraints are legal as so far as the SV LRM is concerned, so I would argue that all SV simulators must support both.

As for coding styel, in this instance it makes sense to randomize distributions as it means we will get a different distribution of MIE and MSTATUS_MIE actions for each simulation run.   Without that the base class (riscv_load_store_rand_instr_stream) will generate a fixed ratio, regardless of the number of sims and this has made closing coverage difficult.

Whether or not soft constraints should be used is almost a matter of religion and I avoid discussing such topics.  :wink:

So I agree with @strichmo, that we should leave the code as-is.   However, if the code as written is not supported by Riviera-PRO, please do update this pull request with a macro to avoid such constructs at compile-time.   I believe the macro \_VCP has been used in the past for this.

danielmlynek:
>I believe the macro _VCP has been used in the past for this.

We will uptade PR with ifdefs

>As for coding styel, in this instance it .....

That is exactly what I was writing about - different dist weights for each simulation run or at least for some number of randomizations is great idea. But rerandomizing dist weights in each randomization is not so good idea.

MikeOpenHWGroup (approved):
Thanks for this @dawidzim.   This runs clean under DSIM.

MikeOpenHWGroup:
Hi @strichmo, I am happy with this PR as it stands now.  Since you've requested changes in your previous review, GitHub needs you to re-review before it can be merged in.   Thanks.

strichmo (approved):

PR #393, opened 2020-11-19
add wfis directly to compressed lw sw stream to enhance coverage

strichmo:

### DIFF
--- a/vendor_lib/google/corev-dv/corev_load_store_instr_lib.sv
+++ b/vendor_lib/google/corev-dv/corev_load_store_instr_lib.sv
@@ -52,5 +52,43 @@
     end
   endfunction : randomize_offset
 
- endclass : corev_compressed_load_store_stress_instr_stream
+  function void post_randomize();
+    super.post_randomize();
+
+    instr_list[0].comment = "corev_compressed_load_store_stress";
+  endfunction : post_randomize
+
+endclass : corev_compressed_load_store_stress_instr_stream
+
+class corev_compressed_load_store_wfi_stress_instr_stream extends corev_compressed_load_store_stress_instr_stream;
+  
+  rand bit first_instr_wfi;
+  rand bit last_instr_wfi;  
+
+  `uvm_object_utils(corev_compressed_load_store_wfi_stress_instr_stream)
+  `uvm_object_new
+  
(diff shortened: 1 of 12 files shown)


vendor_lib/google/corev-dv/corev_load_store_instr_lib.sv
```diff
@@ -52,5 +52,43 @@
     end
   endfunction : randomize_offset
 
- endclass : corev_compressed_load_store_stress_instr_stream
+  function void post_randomize();
```
MikeOpenHWGroup, line 55: Generally I recommend having a one-to-one mapping of classes to files.

formal/tests/cv32e40p_formal_tb.sv
```diff
@@ -0,0 +1,222 @@
+// Copyright 2020 Silicon Labs, Inc.
```
MikeOpenHWGroup, line 1: Wow!   This one caught me off guard.   Not against it (far from it), but it doesn't seem to fix with this particular pull-request.   Also, we ought to make this part of an overall formal strategy.

MikeOpenHWGroup (changes requested):
Couple of comments/questions.

formal/tests/cv32e40p_formal_tb.sv
strichmo: Oh yikes!  That never should've gone in.  That's our local stuff...sorry!!!!  I will pull it out of the PR.

vendor_lib/google/corev-dv/corev_load_store_instr_lib.sv
strichmo: I go back and forth on whether sequence "libraries" should be multiple classes per file or one-per.  I think what generally works best is to create a new directory for sequences and go one class per file.  How do you feel about making that change with the current corev-dv sequences?

There's only 6 class definitions to refactor.

strichmo: I went ahead and made the change to break out corev-dv directed streams into their own class files (and their own directory).

MikeOpenHWGroup (approved):

PR #395, opened 2020-11-21
Fetch enable

MikeOpenHWGroup:
Minor update to cover this from the [User Manual](https://core-v-docs-verif-strat.readthedocs.io/projects/cv32e40p_um/en/latest/sleep.html#startup-behavior):
```
After initial assertion of fetch_enable_i, the fetch_enable_i signal
is ignored until after a next reset assertion.
```
Also:
- removed an unused signal in the control interface
- renamed clock_en to pulp_clock_en (to match RTL primary input)
- got rid of fcov for static primary inputs to the DUT
- clone the latest RTL


cv32/tb/uvmt_cv32/uvmt_cv32_tb_ifs.sv
```diff
@@ -199,27 +200,40 @@ interface uvmt_cv32_core_cntrl_if (
...
 
   function void stop_fetch();
     drv_cb.fetch_en <= 1'b0;
+	lfsr_reset      <= 1'b1;
     `uvm_info("CORE_CNTRL_IF", "uvmt_cv32_core_cntrl_if.stop_fetch() called", UVM_DEBUG)
   endfunction : stop_fetch
 
+  // LFSR used to "randomly" toggle fetch_en to show that
+  // the core ignores fetch_en after its initial assertion.
+  assign fb = !(lfsr[15] ^ lfsr[13] ^ lfsr[12] ^ lfsr[10]);
```
strichmo, line 225: I will approve this as a quick way to get coverage, but this should really be done as a special vseq started in the UVM test.

strichmo (approved):
Would greatly prefer that the lfsr be implemented as a UVM vseq that can be selected with plusarg or randomly selected.

PR #402, opened 2020-11-24
Crosses

MikeOpenHWGroup:
Two very ugly updates to increase ISA cross coverage of <inst>, c.jal and to add some p.clipr illegal instructions.


strichmo (approved):
Thanks...will merge and go!

Issue #253, opened 2020-09-23
Compliance test cases have mismatches in signature checks


Silabs-ArjanB:
In core-v-verif (hash 891805fe096181421896b640450d72a771744d98) the ci_check fails on xrun-illegal and xrun-hello-world. There are also various 'compliance test failures' visible but they do not end up in below summary:

TOOL:	xrun(64)	20.06-a001: Exiting on Sep 23, 2020 at 05:26:29 CDT  (total: 00:00:08)
../cv32/sim/uvmt_cv32/xrun_results/illegal/xrun-illegal.log:                   **SIMULATION FAILED - ABORTED**
../cv32/sim/uvmt_cv32/xrun_results/hello-world/xrun-hello-world.log:                   **SIMULATION FAILED - ABORTED**
../cv32/sim/uvmt_cv32/xrun_results/corev_jump_stress_test_1/xrun-corev_jump_stress_test_1.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/corev_arithmetic_base_test_0/xrun-corev_arithmetic_base_test_0.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/corev_jump_stress_test_0/xrun-corev_jump_stress_test_0.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/corev_arithmetic_base_test_1/xrun-corev_arithmetic_base_test_1.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/corev_rand_instr_test_0/xrun-corev_rand_instr_test_0.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/riscv_ebreak_test_0/xrun-riscv_ebreak_test_0.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/csr_instructions/xrun-csr_instructions.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/corev_rand_instr_test_1/xrun-corev_rand_instr_test_1.log:                        SIMULATION PASSED
../cv32/sim/uvmt_cv32/xrun_results/riscv_arithmetic_basic_test_0/xrun-riscv_arithmetic_basic_test_0.log:                        SIMULATION PASSED

CI Check FAILED with unknown failures.

MikeOpenHWGroup:
Investigating.  Thanks for the notification.

[...]

Issue #50, opened 2020-02-10
Complete RISC-V ISA Verification


MikeOpenHWGroup:
### Task Outcome
1. 100% functional coverage of all RISC-V ISA instructions supported by CV32E40P.
2. 100% pass rate of a regression that achieves the above.

### Background information
Completing this task will require a combination of RISC-V compliance testing, random instruction stream testing and semi-directed testing.

### Location Information
This goal can only be achieved using the `uvmt_cv32` environment.

### Completion Criteria
A Metrics **_dsim_** regression on the master branch that achieves the "Task" Outcome", above.

MikeOpenHWGroup:
Update:

> 100% functional coverage of all RISC-V ISA instructions supported by CV32E40P.

Functional coverage of the RISC-V ISA current stands at 95.9%

> 100% pass rate of a regression that achieves the above.

Done.

I will keep this issue open until the ISA Fcov achieves 100% **_or_** a waiver is generated.

MikeOpenHWGroup:
ISA Fcov is currently > 99%.  All holes are in cross-coverage and are of the form:
```
<instr>, c.jal
<instr>, c.jalr
<instr>, mret
<instr>, dret
```
Work is continuing to close these holes, but it is recommended that this coverage _not_ gate RTL Freeze for CV32E40P.

MikeOpenHWGroup:
Whoops, this issue should have been closed when the CV32E40P v1.0.0 was signed-off.

PR #406, opened 2020-11-26
Compliance Signatures

MikeOpenHWGroup:
This PR enables signature testing of individual test-programs from the riscv-compliance test-suite.  Updated instructions for the user are in the README at cv32/sim/uvmt_cv32.

TODOs before the end of the week:

- [ ] Create a script that runs all compliance tests, including signature file checks
- [ ] Update simulator-specific makefiles (may do that as a set of task-issues to other Contributors.


strichmo:
@MikeOpenHWGroup regarding a script to run the compliance. I would like to make sure this uses the cv_regress infrastructure. I already have a shell script template output and it should be easy to add any hooks. That way anyone with bash and a simulator can run our compliance.

MikeOpenHWGroup:
My bad @strichmo, I did not even think about `cv_regress`, as I always use MetricsCI for regressions.  Later today I will be pushing in an update to this branch that uses a Make target to run a full compliance regression, including a signature check of all tests.   This will, I believe, satisfy issues #50 and #253, so we can pass RTL Freeze.

Would it be OK to get that in first and then get compliance running under cv_regress?

strichmo:
Looks good to me.  Thanks for taking on such a large task. I will have updates for cv_regress and Questa support in a subsequent PR.

strichmo (approved):

PR #414, opened 2020-12-01
Clone RTL from cv32e40p_v1.0.0

MikeOpenHWGroup:
The significant change here is to `Common.mk` to clone the cv32e40p from the RTL Freeze tag.   Other changes are to support a latent couple of bugs in Verilator support.


cv32/sim/Common.mk
```diff
@@ -83,18 +86,27 @@ COMPLIANCE_BRANCH ?= master
...
-  TMP = git clone -b $(CV32E40P_BRANCH) --single-branch $(CV32E40P_REPO) --recurse $(CV32E40P_PKG)
+  TMP = git clone -b $(CV32E40P_BRANCH) --single-branch $(CV32E40P_REPO) $(CV32E40P_PKG)
 endif
 
 ifeq ($(CV32E40P_HASH), head)
   CLONE_CV32E40P_CMD = $(TMP)
 else
   CLONE_CV32E40P_CMD = $(TMP); cd $(CV32E40P_PKG); git checkout $(CV32E40P_HASH)
 endif
+
```
strichmo, line 103: Would like for this code to be in one ifeq endif section.  I believe the code as written could cause a race condition as Make tries to solve CLONE_CV32E40P_CMD as make tries to determine the proper resolution order of variables.

strichmo (changes requested):

cv32/sim/Common.mk
MikeOpenHWGroup: Good point.  Pushed in a change that does it with nested ifeq/endif sections.

strichmo (approved):
Been burned by this before...thanks for the fix!

Silabs-ArjanB:
Hi @strichmo @MikeOpenHWGroup can you please undo this commit? The problem with it is the fact that a tag might actually change before we actually freeze the RTL. core-v-verif should only use the hash and at RTL freeze we will check that the hashed version has the right tag. This is already an issue now as Davide found a synthesis reason why we need to change the RTL and therefore we will end up re-applying the same tag (cv32e40p_v1.0.0) to a new hash (which has not been made yet). We want to have an easy check that core-v-verif was run on the correct hash and the tag usage does not provide such a check.

MikeOpenHWGroup:
I understand and agree with your point @Silabs-ArjanB that the verification needs to re-done on the actual, _final_, tag used to point to the RTL Freeze version of the cv32e40p repository.   I wanted to test the required updates to the Makefiles before the actual freeze.   Its a given that we may need to update it before the actual freeze.

Question: does GitHub allow for re-use of tag names?   That is, can a tag point to (say) hash 1234567 and then later be updated to hash7654321?   I had not expected this to be the case.  So, if the tag needs to move, we should use a new name (e.g. cv32e40p_v1.0.0_yyyymmdd).

MikeOpenHWGroup:
Its been clear that both the tag and hash must be (and shall be) recorded.  I was not aware that tags could move.  A flaw in my opinion.  It means that the tag is a mere convenience and that the hash is the only accurate reference to a "freeze".

PR #421, opened 2020-12-04
add more pulp instructions for code coverage

strichmo:
adds more more clipr pulp case and pulp.abs to illegal instruction test


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
We generated PRs for the same thing.  :-)

PR #426, opened 2020-12-08
Move GitHub Pages to openhwgroup/core-v-verif

MikeOpenHWGroup:
Hi @rickoco, this is a big, but simple, PR.  The objective is to move the coverage reports from my fork to the openhwgroup repo.   This PR will do two things:
1. Delete the old `doc` directory.
2. Create a new `docs` directory (as needed by GitHub Pages) which contains the content from the old `doc` directory, plus the CV32E40P coverage reports and waivers.


DBees (approved):
To me this seems ok. Having said that, I am not that familiar with the core-v-verif repository and contents.

MikeOpenHWGroup:
Thanks Duncan.

PR #427, opened 2020-12-08
fix atohex to atoi

strichmo:


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Drat!   I missed those - was looking in `c_check_imm()`.  Thanks for the save.

PR #450, opened 2021-01-28
Add argument to printf of misa check. Enhance misa check for PULP

GTumbush:
Fixes 2 issues with the MISA check in hello-world.c
1) printf had no argument for format specifier %x
2) check would fail if PULP was defined.

### DIFF
--- a/cv32/tests/programs/custom/hello-world/hello-world.c
+++ b/cv32/tests/programs/custom/hello-world/hello-world.c
@@ -27,6 +27,13 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#ifdef NO_PULP
+#define EXP_MISA 0x40001104
+#else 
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (approved):
Good one @GTumbush.  By coincidence, I just noticed the format specifier and have it fixed on a branch that I haven't yet PR'ed.  You win.  The other fix is also good.   As an aside, I've been noodling on how best to bring in parameters for compiling test-programs.   You fix fits well into that and it gets you going, let's use it for now.

MikeOpenHWGroup:
Hi @GTumbush.  Can you update your commit to use the `-s` or `--signoff` argument on the command line.  I need to be getting better about enforcing that going forward.   The command line is:
```
$ git commit -m 'Clever, concise commit message' -s
$ git push <remote> <branch>
```

PR #451, opened 2021-01-31
Prototype of "core" testbench for CVA6

MikeOpenHWGroup:
This pull-request demonstrates the proposed directory and file structure to support the CVA6 in core-v-verif and provides minimal `Make` targets to see how it will work.  Note that the only the "core" testbench is supported and all it does is compile and simulate long enough to reset the core.  There are targets for both Verilator and Metrics dsim.  (The CVA6 RTL currently does not compile with Cadence Xcelium.)  The README at `cva6/sim/core` _should_ be sufficient to get you started.   Note that you do not need a proper toolchain at time time.

@strichmo, I am asking for you to review as you are interested in the the additions to core-v-verif to support multiple cores.  This is intended to me a mini-demo of that.  I _think_ it is in-line with what we discussed last week.

@JeanRochCoulon, I am asking for you to review because you are a key Contributor to the cva6.

@ASintzoff, I am asking for you to review as you made the initial contibutions to the `cva6` directory in core-v-verif.  I would like to re-arrange much of this to be compatible with the dual "core" and "uvm" structure that will be used by all CORE-V cores verified in core-v-verif.


strichmo (approved):
I don't see any issues with these changes.

MikeOpenHWGroup:
This PR has aged-out and now cannot be merged in.  The CVA6 _is_ integrated, so this PR is no longer needed anyway.

PR #454, opened 2021-02-04
Move ci_check to bin

MikeOpenHWGroup:
@strichmo, this is the ci_check move we discussed today.   I am piggy-backing this onto a development branch I have had going for a while that should also be merged in.  These piggy-back changes only affect the `core` testbench and are there to help some of the Verilator-user questions I have been getting of late.


strichmo (approved):
Looks good.  Should not be a big deal to rebase my moves for the scalable testbench

PR #456, opened 2021-02-09
Multi-core multi-testbench extensions for core-v-verif

strichmo:
Here is the change to essentially create core-specific testbenches for core-v-verif development.

Highlights:
- core-v-verif/sim directory now contains common makefile infrastructure (Common.mk and uvmt makefiles)
- CV_CORE environment variable can now be used to delineate the core in which you wish to run tests.  This may be overridden on any make command line
- Introduced PATHed _makeuvmt_ command which will redirect your make call to any sim/uvmt directory based on CV_CORE (even on command line)
- Deprecated _makecv32_
- All _cv32 directories renamed to generic equivalents
- cv32/ is renamed cv32e40p
- All UVM classes/infrastructure renamed for core (e.g. uvmt_cv32 -> uvmt_cv32e40p or respective core)

Tested:
- Metrics
- Questa
- Xcelium

Not tested:
- Made best attempt to port but not test VCS


MikeOpenHWGroup (approved):
Whew!   Well, that was a lot.  It all looks good to me and I agree that we should merge it to the a `core-v-verif_multi`branch to allow others to readily see what is being proposed.

I have a number of specific comments which you'll find in the "Conversation" section of this PR.

MikeOpenHWGroup:
Hi @strichmo, there is a _lot_ here.  Below are my specific comments.  Anything marked with a "???" is something I'd like more information about.  It probably makes sense for us to discuss over zoom.

1. bin/Makefile.uvmt and bin/makeuvmt:
   - ??? Re-direction scripts to point makefile to a specific core.
         Requires CV_CORE to be set.

2. cv32/env/uvme_cv32/riscv_dv_extension/*
   - Deleted since is was never used.

3. cv32/tests/programs/custom/riscv_ebreak_test_0/test.yaml
   - ??? Not sure why this is deleted.

4. cv32/tests/uvmt_cv32/base-tests/*.sv
   - ??? I understand why these could be deleted, but I still see a future for
         UVM testcases in this ENV.  Let's discuss.

5. cv32/* --> cv32e40p/*
   - renaming for actual core names.
   - ??? Where does CV_CORE_LC get defined?

6. cv32e40p/env/uvme/corev-dv/*
   - new home of corev-dv
   - ??? do we one one per core or can a layered approach work?

7. cv32e40p/env/uvme/corev-dv/instr_lib/.keep
   - Unless this is intensional, should be deleted.
   - If its intentional, recommend renaming to .gitkeep (a Git convention).

8. Move cov model to uvme_cv32e40p_cov_model
   - Agreed.

9. cv32/env/uvme_cv32/ral
   - Should we delete these? Unless core-v-verif codes up its own SV RM,
     its difficult to imagine core-v-verif will ever use the RAL.

10. cv32e40p/sim/mcy
    - *** Recommend moving this to cv32e40p/sim/tools/mcy. (I've already done
          that on a separate, not yet merged, branch).

11. cv32e40p/tests/uvmt/vseq/uvmt_cv32e40p_vseq_lib.sv
    - ??? Is this flagged as a new file by GitHub because the uvmt_cv32
          directory is renamed to uvmt?

12. sim/uvmt/*
    - I just now noticed that there is a top-level "sim" directory.

13. vendor_lib/google/corev-dv/*.sv
    - ??? So, these are now corev-dv base classes and the core-specific
          corev-dv classes are in $CV_CORE/env/uvme/corev-dv, is that correct?

GTumbush:
I'd like to attend this zoom meeting. My only conflict is 3:15pm-4:30pm MST.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Looks good @strichmo.   I'm bringing this in.

PR #471, opened 2021-02-18
Fix for metrics-regress on GitHub Action

MikeOpenHWGroup:
The "fix" is to point `.github/worksflows/metrics-regress.yml` to the correct regression in .metrics.json:
```
 METRICS_REGRESSION_NAME: cv32e40p_ci_check_regression
```
The rest of it is comments and license headers.


bin/metrics-regress
```diff
@@ -1,4 +1,33 @@
 #!/usr/bin/python3
```
strichmo, line 1: Minor note.  Usually like to do /usr/bin/env python3 just in case an install is using a custom Python (or a virtual env).
I know we aren't supposed to manually but provides a little resilience to Python environment.

strichmo (approved):
Looks good!

PR #499, opened 2021-03-03
append mergetest

silabs-robin:
Signed-off-by: Robin Pedersen <Robin.Pedersen@silabs.com>


strichmo (approved):
Looks like your PR was validated properly.  Thanks!

Issue #484, opened 2021-02-26
apply RUN_INDEX to all tests in core-v-verif


strichmo:
### Task Outcome
The common makefile infrastructure should be updated such that all individual test results are appended with a RUN_INDEX make variable.  Unless otherwise specified this RUN_INDEX is always 0.

### Location Information
Common makefile infrastructure
Note that corev-dv tests should have RUN_INDEX removed from their respective YAML test configurations.

### Completion Criteria
- Ensure that all ci_check tests can be regressed.
- Create a new regression or edit an older regression to run multiple iterations of a custom program and ensure that the custom program can be regressed for multiple random iterations with no change to its current test configuration YAML.

strichmo:
Was merged and reviewed in E status meeting.

PR #504, opened 2021-03-04
Issue #484 Added default run index to all test outputs

silabs-hfegran:


MikeOpenHWGroup (approved):
Hi @silabs-hfegran.  Whew!  A lot of small changes to lot of files.  Thanks for handling this.

I considered asking for an update to the README, but we do not discuss RUN_INDEX anywhere, even for the places we use it already, so let's leave that for another time.

PR #509, opened 2021-03-08
Updated Riscv-dv to latest hash

silabs-hfegran:
Summary of changes:

        - Removed jump to init_machine_mode (function location inlined
          by Google, no longer needed).

        - Reverted previous mret->ret change to init_machine_mode
          generator function to avoid an infinite loop.

        - Changed setup_mmode_reg to not enable MIE bit.
          This is instead done by mret at the end of code section
          by copying MPIE->MIE. (cf. riscv priv. spec.)
          This prevents an infinite loop if an interrupt request is
          present prior to init_machine_mode completion.

        - Updated common interrupt handler jump for increased robustness
          with large programs. (Ports over change from RISCV-DV
          upstream).

        - Added empty custom csr definitions for compatibility with
          latest Riscv-dv.

### DIFF
--- a/cv32e40p/sim/Common.mk
+++ b/cv32e40p/sim/Common.mk
@@ -21,7 +21,7 @@ CV_CORE_TAG    ?= cv32e40p_v1.0.0
 
 RISCVDV_REPO    ?= https://github.com/google/riscv-dv
 RISCVDV_BRANCH  ?= master
-RISCVDV_HASH    ?= 10fd4fa8b7d0808732ecf656c213866cae37045a
+RISCVDV_HASH		?= 0b625258549e733082c12e5dc749f05aefb07d5a
 
 COMPLIANCE_REPO   ?= https://github.com/riscv/riscv-compliance
 COMPLIANCE_BRANCH ?= master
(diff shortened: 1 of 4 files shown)


cv32e40p/sim/Common.mk
```diff
@@ -21,7 +21,7 @@ CV_CORE_TAG    ?= cv32e40p_v1.0.0
 
 RISCVDV_REPO    ?= https://github.com/google/riscv-dv
 RISCVDV_BRANCH  ?= master
-RISCVDV_HASH    ?= 10fd4fa8b7d0808732ecf656c213866cae37045a
+RISCVDV_HASH		?= 0b625258549e733082c12e5dc749f05aefb07d5a
```
MikeOpenHWGroup, line 24: @silabs-halfdan, just FYI, you can use the most significant seven hex-digits when referencing a git hash.  So this could be `RISCVDV_HASH ?= 0b62525`.  (I learned this several weeks after generating the previous version of this Makefile.)  The full hash is only required in the event of a collision.

MikeOpenHWGroup (approved):

Issue #495, opened 2021-03-02
add the EMBench benchmark as a regression and individual tests to the verification testbench


strichmo:
### Task Outcome
It would be a great benefit to both the 40X and 40S projects to enable a simulation suite of appropriate benchmarks to measure both current processor performance as well as deltas in performance as pipeline changes are introduced during development.

One possible source of benchmarks suitable for embedded processors would be the EMBench suite of benchmarks freely available at:

https://github.com/embench/embench-iot

This is a collection of benchmarks developed specifically for smaller RISCV processors. 

A process for completing this task should include the following:

clone the package and attempt to use its build scripts to generate ELFs
See if there is a user hook to insert a test_done write similar to .c files in the cv32e40p BSP
Ensure the address map of the linker scripts are correct with respect to the CV32E* map
Decrease simulation time to 
Use user hooks to measure number of cycles for the benchmark
Determine how to incorporate build into cv32e40p in core-v-verif.  Ideally the external package can be cloned as a "Vendor_lib" with a customization layer, and/or Makefile updates.
In conjunction with Jenkins cycle performance scraping, create a regression of benchmarks

### Completion Criteria
- A set of makefile targets can be used to run the EMBench suite
- A regression target enables EMBench to be executed

MikeOpenHWGroup:
This is a good idea, assuming that minimum performance levels for a given configuration are well defined in the specification and the project has performance targets it must achieve.  The test should:
- configure the DUT, TB and stimulus such that the performance is predictable.
- measure the performance and provide an unambiguous pass/fail.

Issue #494, opened 2021-03-02
add EEMBC Coremark test


strichmo:
### Task Outcome
The coremark suite is a general purpose processor benchmark and the below link is EEMBC's MCU version of it.  Similar to EMBench we should have a version of this running our the 40X and 40S with performance tracking.

https://github.com/eembc/coremark

### Completion Criteria
The benchmark executes on the 40X

strichmo:
Robin Pedersen from Silicon Labs is working on this and will add the PR.

MikeOpenHWGroup:
See my comments for issue #495.  They apply here as well.

PR #510, opened 2021-03-08
Removed magic numbers for mm_ram addresses and added a cycle counter

silabs-mateilga:

The magic numbers used to signify the mm_ram addresses have been made in to parameters for readability.

A cycle counter has been added to mm_ram. The motivation for this is to enable benchmarks running on the core to access and record the cycle count for a given testcase, which will be necessary for implementing EMBench and Coremark in core-v-verif.

This PR is relevant to issues #494 and #495.

### DIFF
--- a/cv32e40p/tb/core/mm_ram.sv
+++ b/cv32e40p/tb/core/mm_ram.sv
@@ -88,6 +88,10 @@ module mm_ram
     localparam int                        MMADDR_RNDNUM     = 32'h1500_1000;
     localparam int                        MMADDR_TICKS      = 32'h1500_1004;
 
+    // UVM info tags
+    localparam string                     MM_RAM_TAG = "MM_RAM";
+    localparam string                     RNDSTALL_TAG = "RNDSTALL";
+
     // mux for read and writes
     enum logic [2:0]{RAM, MM, RND_STALL, ERR, RND_NUM, TICKS} select_rdata_d, select_rdata_q;
 
@@ -193,7 +197,7 @@ module mm_ram
 `ifndef VERILATOR
         if (!$test$plusargs("rand_stall_obi_disable")) begin
             if ($test$plusargs("max_data_zero_instr_stall")) begin
-                `uvm_info("RNDSTALL", "Max data stall, zero instruction stall configuration", UVM_LOW)
+                `uvm_info(RNDSTALL_TAG, "Max data stall, zero instruction stall configuration", UVM_LOW)
                 // This "knob" creates maximum stalls on data loads/stores, and
                 // no stalls on instruction fetches.  Used for fence.i testing. 
                 rnd_stall_regs[RND_STALL_DATA_EN]     = 1;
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (changes requested):
Love it!   Thanks @silabs-mateilga for the search-and-destroy of those magic numbers.   You have done a great service to the core-v-verif community.

I wonder if the mm_ram is the best place for the cycle counter.  This might not be the most obvious location and if the mm_ram is replaced, the cycle counter could disappear and the benchmarks will be impacted.  It seems to be that this could be  a feature of the testbench itself, or even a component of the testbench.

strichmo:
For now, the Virtual Peripheral in mm_ram is the best place we have to enable tests (in C or assembly) to control the testbench.  Even if we had a proper UVM-agent peripheral, I would argue that it makes sense, especially in a benchmark context, to support this functionality.  This would include the other VP features in the mm_ram.

Once we take on the creation of a UVM agent proper, I would argue for the porting of VP features one-for-one in the new peripheral.

cv32e40p/tb/core/mm_ram.sv
```diff
@@ -451,6 +475,12 @@ module mm_ram
         
         end else if (select_rdata_q == RND_NUM) begin
             data_rdata_mux = rnd_num;    
+        end else if (select_rdata_q == TICKS) begin
+            data_rdata_mux = cycle_count_q;
+
+            if (cycle_count_overflow_q) begin
+                $fatal(1, "cycle counter read after overflow");
```
MikeOpenHWGroup, line 482: Please replace the call to $fatal() with an invocation of the \`uvm_fatal() macro.  (Yes, I _know_ some historical code still uses $fatal(), but we should not add new ones.)

MikeOpenHWGroup (approved):
Ah, I missed the part where the cycle count is actually readable and reset-able.  In that case, I agree that mm_ram is (currently) the proper place for this to be.  @silabs-mateilga, I'll update the virtual peripheral documentation for that and ask you to review it.

Speaking of things I missed the first time, I've made a comment about the use of $fatal().  @silabs-mateilga, please address that and I'll approve this PR.   Thanks!

cv32e40p/tb/core/mm_ram.sv
strichmo: @MikeOpenHWGroup This one might be my fault.  I steered them to $fatal as I thought this code has to run under Verilator??

MikeOpenHWGroup: Yeah, we need a definitive statement on how we support Verilator.   That's on me.  For now, please use the uvm messaging service (\`uvm_fatal(), \`uvm_error(), \`uvm_info(), etc.) and I'll make the appropriate fixes/hacks after the fact to keep Verilator going.

silabs-mateilga: Thanks for the input! I'll change the call to the uvm macro, expect an update tomorrow.

MikeOpenHWGroup: :+1:

silabs-mateilga: Fix is in. I also took it one step further and updated other $fatal calls in the file, and parameterized the info tag for the uvm messaging calls.

MikeOpenHWGroup (approved):

PR #528, opened 2021-03-23
Hotfix: xrun debug

silabs-hfegran:
Fixes issues with tcl script that prevents xrun debugging with WAVES=1


strichmo (approved):
thanks for fix!  I'll make a separate one for the 40x, which I'm sure requires as well.

PR #532, opened 2021-03-25
Fixes to reenable debug_test in ci_check

silabs-mateilga:
Quick fix to get debug_test to pass until we have verification of bus errors in place.


strichmo (approved):
Thanks for the fix!

Issue #490, opened 2021-03-01
next-gen ISA coverage model: enhance intra-instruction coverage with better bins and crosses


strichmo:
### Task Outcome
The covergroups and bins in the next-generation RISC-V ISA coverage model need to include more bins and crosses within each individual instruction

### Background information
The RV32 ISA coverage model in the CV32E40P defined instruction coverage with basic bins.  For example for an instruction with 2 register operands and 1 immediate signed operands the typical coverage bins would include:

- All possible register indices for the first register operand.
- All possible register indices for the second register operand
- For the immediate, 3 bins for zero, positive and negative

In general no crosses were defined within the instruction coverage bin.

For improvements in coverage:
- Crosses of register indices and operands should be included
- More bins for immediates should be included.  The bins should be defined relative to the width of the immediate and instruction

### Location Information
Covergroup definitions in core-v-verif/lib/uvm_agents/uvma_isa_covg

### Completion Criteria
Successful coverage report generated with the enhanced ISA coverage bins

Issue #489, opened 2021-03-01
next-gen ISA coverage model: use RVFI as monitor


strichmo:
## Task Title
Port next-generation coverage model to RVFI

### Task Outcome
The UVM ISA coverage model agent monitors executed instructions via the RVFI agent versus using OBI or other tracer source.
The UVM ISA coverage model extracts operand information (i.e. contents of registers) from RVFI to supplement operands of GPRs in the coverage model.

### Background information
This task should be started after https://github.com/openhwgroup/core-v-verif/issues/488 is completed and the RVFI monitor is working for a core.  Note that it may be useful to enable multiple input paths 

### Location Information
core-v-verif/lib/uvm_agents/uvma_isa_covg

### Completion Criteria
- Released demonstrated functional coverage agent working with RVFI monitor and RVFI agent.
- Demonstrated at least one instruction functional covergroup and bins updated with register operand values driven by RVFI monitor

strichmo:
The RVFI now drives the ISACOV model within the wb_controller_e40x branch.  This change will be merged to all other branches in the next merge cycle.

Issue #488, opened 2021-03-01
next-gen ISA coverage model: create uvma_covg and integrate with OBI monitor


strichmo:
### Task Outcome
A UVM agent is developed to track functional coverage.  This should be an extensible, reusable UVM agent that be reused against all cores.

- Input of coverage model comes from an analysis port.  This will initially be OBI for the purpose of this task, but will be extended to RVFI in future releases.
- Have no connections to the ISS.
- Logs all instructions using standard UVM template monitor logging capabilities.
- Must only track ISA-specified instructions.  No pseudo-instructions.
- Enables dynamic configuration via configuration class of supported extensions
- Initially supports CV32E40P (RV32IMCZifencei) with ability to extend beyond.
- Creates classes based on standard UVM object conventions to model instructions

### Location Information
core-v-verif/uvm_agents/uvma_covg

### Completion Criteria
The UVM agent is released and demonstrated on the CV32E40P meeting all above requirements.

### Additional context
Conceptual block diagram: credit @MikeOpenHWGroup 
[OBI_Coverage.docx](https://github.com/openhwgroup/core-v-verif/files/6062590/OBI_Coverage.docx)

strichmo:
Notes on implementation.  @silabs-robin has started to implement the Agent and has some basic infrastructure together.  https://github.com/silabs-robin/core-v-verif/tree/isa_agent_newcg_repeat

Based on that work I would like to put together some more detailed requirements

## Assumptions
- All of the following is contained in a uvm_agent _uvma_isa_covg_ 
- HINTs are not included

## Enumerated types
- Create an enumerated type for ISA instruction opcodes.  Currently include the IMC types.  Each type name should match ISA and be organized by RISCV extension.  No pseduo-ops.
- Create an enumerated type for ISA instruction _group_.  The types represent the general operation that the processor performs and groups instructions together.  (ALU, LOAD, STORE, BRANCH, JUMP, CSR, MMU, EXCP, PRIV, WFI).  
Each instruction opcode belongs to one and only one group.
Note that the term should be _instruction group_ for this enum.  Instruction type refers to the ISA instruction type (i.e. R-type, I-type, S-type, etc.)
Create a package level function or static method in instr class to fetch group from instruction)
- Enumerated type for GPR registers.  Name each register via ISA name (x<n>) and not the ABI name.   The ABI name should never show up in the coverage model.
- Enumerated type for CSR names
- Enumerated type for modes

## Instruction class
[...]

Issue #487, opened 2021-03-01
next-gen ISA coverage model: create or integrate an open-source disassembler


strichmo:
### Task Outcome
For the next-generation OpenHW testbenches, we need a disassembler that can take a binary op-code and emit relevant fields related to the instruction.  The disassembler would be invoked on a per-instruction basis.

For this task the disassembler should support 32bit RV32IMC for integration into the next-generation coverage model on the CV32E40P.  However it should be easily extensible to the B and P extensions for future integration in the 40X.  Additionally the Zce standard will become a requirement (when/if ratified) for the 40S/X.

Performance considerations must be taken into account.  Ideally the disassembler will be called on-the-fly per each detected op-code.  Therefore the disassembler call could be via the DPI or into natively compiled SystemVerilog.

### Location Information
Must be easily integratable into core-v-verif.  
Can be C source that is checked in, sourced from an external repo, or internally-developed SystemVerilog that is compiled into simulations.

### Completion Criteria
- Support for the CV32E40P extensions (RV32IMCZifencei)
- Integration of the disassembler into the CV32E40P next-generation coverage model

PR #533, opened 2021-03-26
Create a UVM agent for ISA coverage

silabs-robin:
This fulfils the initial criteria of https://github.com/openhwgroup/core-v-verif/issues/488

It introduces an agent which monitors activity in the 40x's tracer module (rvfi should be used later https://github.com/openhwgroup/core-v-verif/issues/489).
The monitor decodes/disassembles the received instruction and parses it into a uvm_object representing the instruction (later, https://github.com/openhwgroup/core-v-verif/issues/487 should enable full support for decoding).
The uvm object is passed to the coverage model which samples into covergroups defined for RV32IMCZifencei_Zicsr (more tweaks and refinements later https://github.com/openhwgroup/core-v-verif/issues/490 and bottom of https://github.com/openhwgroup/core-v-verif/issues/488).

The name "uvma_isa" was chosen (instead of the prescribed "uvma_covg" or "uvma_isa_covg") because it matches the existing naming-patterns for both 1) other agents, and 2) covergroups.

More polish is needed (as noted above), but the system stands by itself and demonstrates a complete architecture and coverage is disabled by default.


silabs-robin:
Usage:
1. Within `cv32e40x/env/uvme/uvme_cv32e40x_cfg.sv`, set `isa_cfg.enabled` to 1 (is there a better way to toggle this?).
2. Run any test in uvmt with `COV=1`.
3. Use your preferred simulator and coverage database viewer.

cv32e40x/env/uvme/uvme_cv32e40x_cfg.sv
```diff
@@ -35,6 +35,7 @@ class uvme_cv32e40x_cfg_c extends uvm_object;
    //rand int unsigned             debug_clk_period;
    
    // Agent cfg handles
+   rand uvma_isa_cfg_c        isa_cfg;
```
MikeOpenHWGroup, line 38: Its not clear from the class or object name that this is a configuration object for a functional coverage model.  Perhaps `uvma_isa_fcov_cfg_c`?

lib/uvm_agents/uvma_isa/uvma_isa_mon.sv
```diff
@@ -0,0 +1,128 @@
...
+  fork
+    begin
+      forever begin
+        uvma_isa_mon_trn_c mon_trn;
+
+        @(cntxt.vif.retire);
+
+        mon_trn = new();
+        mon_trn.instr = new();
+        mon_trn.instr.name =
```
MikeOpenHWGroup, line 66: This is hard to follow.   Either:
- use a case statement (nested, if required), or
- improve the formatting for readability.

lib/uvm_agents/uvma_isa/uvma_isa_if.sv
```diff
@@ -0,0 +1,23 @@
...
+//     https://solderpad.org/licenses/
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+interface uvma_isa_if;
```
MikeOpenHWGroup, line 17: Where is this instantiated in the environment?   Is another component expected to source 32-bit instructions on this interface?

MikeOpenHWGroup (changes requested):
Thanks for this @silabs-robin.  I am _very_ pleased to see this work starting - this coverage model is hugely important for all CORE-V cores and it will be heavily re-used.  I have added a few comments, most of them minor and there are two that are more important than the others:
1. Please give some thought to a more descriptive name for this agent.  "isa" is very generic and does not convey that this is a coverage model.
2. I found the code in `uvma_isa_mon_c::run_phase()` very difficult to follow (even though I am familiar with the ISA).  Se my comments in that file.

cv32e40x/env/uvme/uvme_cv32e40x_cfg.sv
```diff
@@ -90,8 +92,15 @@ class uvme_cv32e40x_cfg_c extends uvm_object;
       obi_instr_cfg.read_enabled  == 1;
       obi_data_cfg.write_enabled  == 1;
       obi_data_cfg.read_enabled   == 1;
+      isa_cfg.enabled              == 0;
```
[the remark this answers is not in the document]
strichmo: @MikeOpenHWGroup This was my idea.  This is a very early version of the new func cov model and really we shouldn't be using it until it's more functional (at least with IMC coverage).  Therefore to avoid polluting fcov stats in any regressions I proposed to leave this off and contributors who work to update the model will enable it via a local edit.  Should add a FIXME here however at the least.

cv32e40x/env/uvme/uvme_cv32e40x_pkg.sv
```diff
@@ -38,6 +38,7 @@ package uvme_cv32e40x_pkg;
    import uvml_hrtbt_pkg  ::*;
    import uvml_sb_pkg     ::*;
    import uvml_trn_pkg    ::*;  
+   import uvma_isa_pkg::*;
```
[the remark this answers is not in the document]
strichmo: Agreed, this agent's main purpose is to take ISA-based functional coverage so cov has to end up in the agent name.  ISA agent is just too general.

lib/uvm_agents/uvma_isa/instr.sv
```diff
@@ -0,0 +1,74 @@
+// Copyright 2021 OpenHW Group
```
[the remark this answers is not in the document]
strichmo: This should be named in accordance with the agent and should include a trn suffix.  Since this will always be a passive agent, it's not a sequence item, but certainly is a "transaction" within the framework/terminology of our OpenHW agents.

lib/uvm_agents/uvma_isa/uvma_isa_if.sv
strichmo: Perhaps add a little bit more self-documentation here, that this is a temporary interface.  I would imagine that we would replace this with an exposed analysis port for OBI (or RVFI directly, if ready).

lib/uvm_agents/uvma_isa/uvma_isa_mon.sv
strichmo: Note that this will be replaced with another decoder solution.

MikeOpenHWGroup: OK

silabs-robin:
Thanks you @MikeOpenHWGroup  and @strichmo  for reviewing this PR.
I agree with all of the feedback and have tried to address it in the latest commits.
(I'll reply to some of the comments above with some elaboration too).

lib/uvm_agents/uvma_isa/instr.sv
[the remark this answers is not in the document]
silabs-robin: I don't see a solution that satisfies all constraints.

If one leave it as is, it doesn't follow the repo's naming scheme.
If the name reflects "rv32imcz", it conflicts with the task's "ability to extend beyond".
If suffixing "_trn", it should extend uvml_trn_mon_trn_c and conflicts with mon_trn.
If mon_trn holds the data instead, then the name reflects "instruction" even less.
If "instr" is a struct typedef, then it conflicts with "standard UVM object conventions to model instructions".

For now, I've changed it to "uvma_isacov_instr".

cv32e40x/env/uvme/uvme_cv32e40x_pkg.sv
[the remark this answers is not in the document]
silabs-robin: I changed the name to "uvma_isacov", but if we would like something else I can set it to that.

Rationale:
If possible, I would prefer to keep the convention of "uvma_<name>" and avoid "uvma_<name1>_<name2>", because the former is consistent and simple while the latter adds a cognitive overhead in deciphering each term of our multi-word names (e.g. "uvma_abc_xyz_mon_trn_logger").
The reason I avoided "uvma_isa**f**cov" is 1) too many codes and abbreviations are difficult, and 2a) "isa coverage" names a type of coverage and 2b) arguably suggests a sub type of "functional coverage" (isa declares functionality, is not rtl, is not e.g. performance reqs).
The only precedence of "fcov" seems to be in "isa_fcov_holes", and all examples of "covg" seem to be for classes containing covergroups.

lib/uvm_agents/uvma_isa/uvma_isa_if.sv
silabs-robin: I might have misunderstood "Input of coverage model comes from an analysis port"?
This way of using interface is the same as all the other "Agent interfaces" in the tb.
But maybe the task's intention was to use TLM ports for input to the agent itself, and not just to the "cov_model" as it is now.

I added a TODO comment about updating the interface, both regarding RVFI and potentially TLM port.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @silabs-robin, the updates look good.  Will merge this in.

PR #543, opened 2021-03-29
add CV_CORE_PATH for optionally having rtl repo dir symlinked locally

silabs-robin:
This PR adds the possibility to point a core directory in `core-v-cores` to a local directory via symlink. This is instead of cloning.
The benefit is that one doesn't have to be afraid of some `make clean` or other scripts accidentally wiping the core directory.
Hence one can safely work on a core's RTL and run ci_check, clean-build core-v-verif, etc.

(This is something we had locally, intermediately before the "multi-design" core-v-verif repo was ready. I am merely porting it over to the github repo now.)

### DIFF
--- a/mk/Common.mk
+++ b/mk/Common.mk
@@ -76,22 +76,26 @@ $(error Must define a COMPLIANCE_REPO to use the common makefile)
 endif
 
 ###############################################################################
-# Generate command to clone the core RTL
-ifeq ($(CV_CORE_BRANCH), master)
-  TMP = git clone $(CV_CORE_REPO) $(CV_CORE_PKG)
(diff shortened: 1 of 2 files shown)


mk/Common.mk
```diff
@@ -76,22 +76,26 @@ $(error Must define a COMPLIANCE_REPO to use the common makefile)
 endif
```
silabs-robin, line 76: This diff of `Common.mk` is a bit difficult to read because `diff` gets confused by the similarity of the lines.
It is easier read by opening the entire file.

MikeOpenHWGroup (approved):

PR #550, opened 2021-04-06
disable scheduled regression on the e40p

strichmo:
using workflow_dispatch in schedule to convert this action to manual-start only


GTumbush (commented):
I thought we were going to leave regressions running since we still have 4 ISS licenses.

strichmo:
I would like to see the results of Mike's manual run on Friday with regards to length of time of regression before re-enabling this.  We can decide on a schedule after that.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
@GTumbush, are you good with this change?  If so, please hit the approve button.   Thanks.

GTumbush (approved):
Approved

PR #554, opened 2021-04-08
Integrate disassembler/decoder into ISA coverage model

silabs-robin:
This PR integrates a disassembler/decoder into the isa coverage agent in accordance with https://github.com/openhwgroup/core-v-verif/issues/487.
It makes use of a few source files from riscv-isa-sim in order to extract fields and names of instruction word binaries.
The ad-hoc decoder in uvma_isacov_mon is replaced by this integrated disassembler.

Usage:
1. In "uvme_cv32e40x_cfg.sv", `isacov_cfg.enabled` should be `== 1`.
2. In "cv32e40x/sim/uvmt", run `make dpi_dasm`
3. Run any `make test ...` and append these variables to the command:
    * `USER_COMPILE_FLAGS='+define+DPI_DASM'`
    * `USER_RUN_FLAGS='-sv_lib $(DPI_DASM_PKG)/libdpi_dasm.so'`
4. (One can read more in [lib/dpi_dasm/README.md](https://github.com/silabs-robin/core-v-verif/blob/8c14a4c4ea99d2f9c31a34be8a808009242519b9/lib/dpi_dasm/README.md) and [lib/uvm_agents/uvma_isacov/README.md](https://github.com/silabs-robin/core-v-verif/blob/8c14a4c4ea99d2f9c31a34be8a808009242519b9/lib/uvm_agents/uvma_isacov/README.md))

Testing:
I have tested this, but with the current unavailability of the ISS the testing is not 100%.
In any case, the default behavior should be that the disassembler and uvma_isacov are not enabled and the tests themselves should not be affected (passive agent).
Without the "USER_" flags above, normal behavior is untouched as far as I could tell while testing different combinations of make args.

NB:
RVC (compressed instructions) is not yet implemented because RVFI (or equivalent) first needs to replace the tracer so that non de-compressed instruction binaries can be fed to the disassembler and coverage model.
But the disassembler does support "C" and much more, and we have tested this functionality outside of the integration.


lib/dpi_dasm/dpi_dasm_imports.svh
```diff
@@ -0,0 +1,58 @@
+import "DPI" set_config = function void dasm_set_config(int bit_size, string isa,
```
MikeOpenHWGroup, line 1: Please add a license comment header to this file.

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -17,7 +17,7 @@
 
 
 typedef enum {
-  UNKNOWN,
+  UNKNOWN,  // TODO this should not be needed?
```
MikeOpenHWGroup, line 20: I think you are right.  Probably safe to leave it in now.

Just FYI, and not related to this PR, it would be nice at some point to have deliberate functional coverage for specific classes of illegal instructions.

MikeOpenHWGroup (changes requested):
Happy to see this PR @silabs-robin, thanks.  I have left a couple of comments in specific files all minor stuff).

Question about `lib/dpi_dasm/config.h`. It is an empty file so GitHub won't allow me to comment on the file directly.  Do we need it?  If its there for possible future use perhaps a comment header to that effect would be helpful.

silabs-robin:
> Please add a license comment header to this file.

Indeed you are right. Somehow I always forget to think about the licenses.
I added a license header to "dpi_dasm.cxx" as well as "dpi_dasm_imports.svh".

I also wrote some comment in "config.h" explaining that it is there as a disassembler dependency and is intentionally empty.

MikeOpenHWGroup:
> Indeed you are right. Somehow I always forget to think about the licenses.

I also do too!

> I added a license header to "dpi_dasm.cxx" as well as "dpi_dasm_imports.svh".

Thanks.

> I also wrote some comment in "config.h" explaining that it is there as a disassembler dep
Looks good.

MikeOpenHWGroup (approved):
Due to the use of non -standard (for OpenHW) licenses in this pull-request, I requested a review from the Eclipse Foundatioin.  They have completed a high level triage and indicated we may now check the content into our project’s source code repository. Eclipse will perform a comprehensive analysis at a later date based on IP work queue.

The review is documented in https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23234

PR #564, opened 2021-04-19
Updating references/bindings in toplevel tb

silabs-oysteink:
Updated references in uvmt_cv32e40x_tb to reflect RTL changes. (Controller and interrupt controller was moved out of ID stage and into toplevel).

Signed-off-by: Oystein Knauserud <Oystein.Knauserud@silabs.com>

### DIFF
--- a/cv32e40x/sim/Common.mk
+++ b/cv32e40x/sim/Common.mk
@@ -15,7 +15,7 @@ export SHELL = /bin/bash
 
 CV_CORE_REPO   ?= https://github.com/openhwgroup/cv32e40x
 CV_CORE_BRANCH ?= master
-CV_CORE_HASH   ?= b62250f81acc8505556b43e341c19668b4f8a5d8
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup (changes requested):
Thanks for this @silabs-oysteink.   It looks to me that the reference to nodes internal to the E40X core have been updated and this PR is to get the TB to match.   You may not be aware that the core-v-verif clones a specific hash of the RTL, so this PR should also include an update to `cv32e40x/sim/Common.mk` to get the specific hash of the RTL that implements these updated nodes.

silabs-oysteink:
Updated the Common.mk to point to the latest RTL.

MikeOpenHWGroup (approved):
Thanks @silabs-oysteink, that does the trick.   For future reference, you can simplify your life somewhat by only using the first seven digits from the git hash.  (Of course, using the full hash is not wrong.)

PR #568, opened 2021-04-21
Bugfix uvma_isacov GUI=1

silabs-robin:
The [PR about integrating a disassembler](https://github.com/openhwgroup/core-v-verif/pull/554) broke runs with GUI=1.
Sorry I was not aware of this until I tried GUI=1 just now.

Explanation:
The use of DPI includes a shared object file (.so).
By default this .so is not built and included in simulations.
It is ok to run without the .so and still have references to it in the code, as long as one don't attempt to call it.
Hence, I used ifdef to make such usage optional.
But running in gui (xrun) treats this differently and complained about the missing symbol linkage even when it would not get called.

Solution:
Ifdef the entire function that refers to the dpi function.
(As opposed to ifdef on only _the call to the function_ that refers to the dpi function).


MikeOpenHWGroup (approved):
This PR makes use of a compiler macro `DPI_DASM` that was introduced to the code on 2021-04-06.  In general it is bad practise to use compiler macros and I would rather not introduce more.   Having said that, the macro already exists and this is a bug fix so I will approve/merge and open an issue to investigate the removal of this macro.

PR #573, opened 2021-04-22
update RM licensing

eroom1966:
This was an issue highlighted by @strichmo regarding the license checkout procedure
This PR resolves that issue

I should add I have only modified CV32E40P, I think we need to discuss the best way to handle the additional cores in terms of the structure.


MikeOpenHWGroup:
Hi @strichmo, I've reviewed and this all looks good, but as @eroom1966 indicated that this PR is in response to an issue you flagged, its better if you review before we merge.

strichmo (approved):
OK with changes.  Tested locally and seems to work fine.

PR #574, opened 2021-04-22
enable dpi_dasm for coverage model across all simulators

strichmo:
PR enables the new ISA cov model (and the DPI_DASM built on Spike) in all simulators by default.
define for DPI_DASM is removed
Checked in Linux64 verison of dpi_dasm shared library that should work across simulators (tested with xrun, vsim and dsim)
Made attempts to update VCS and Riviera but not tested
make dpi_dasm should work by default on Xcelium and Mentor (target only needed to edit, rebuild dpi_dasm shared lib)

Note that the coverage model is still disabled (run-time) by default until covergroups are implemented and tested.


lib/dpi_dasm/dpi_dasm.cxx
```diff
@@ -109,10 +110,10 @@ extern "C" const char* get_insn_name(uint64_t insn) {
     initialize_disassembler();
   }
 
-  insn            = endian_swap(insn, gv_is_big_endian);
-  string insn_str = gp_disassembler->lookup(insn & 0xFFFFFFFF)->get_name();
-
-  return insn_str.data();
+  insn       = endian_swap(insn, gv_is_big_endian);
+  gs_ret_str = gp_disassembler->lookup(insn & 0xFFFFFFFF)->get_name();
```
silabs-hfegran, line 114: Any reason why we have to use a global variable here?

mk/Common.mk
```diff
@@ -610,14 +610,15 @@ vcs-unit-test:  vcs-run
...
 
-DPI_DASM_SRC = $(DPI_DASM_PKG)/dpi_dasm.cxx $(DPI_DASM_PKG)/spike/disasm.cc $(SPIKE_PKG)/disasm/regnames.cc
-DPI_DASM_OUT = $(DPI_DASM_PKG)/libdpi_dasm.so
+DPI_DASM_SRC    = $(DPI_DASM_PKG)/dpi_dasm.cxx $(DPI_DASM_PKG)/spike/disasm.cc $(SPIKE_PKG)/disasm/regnames.cc
+DPI_DASM_ARCH   = $(shell uname)$(shell getconf LONG_BIT)
+DPI_DASM_LIB    = $(DPI_DASM_PKG)/lib/$(DPI_DASM_ARCH)/libdpi_dasm.so
 DPI_DASM_CFLAGS = -shared -fPIC -std=c++11
-DPI_DASM_SVDPI_PATH ?= . -please_set_DPI_DASM_SVDPI_PATH_to_where_svdpi.h_is
-DPI_DASM_INC = -I $(DPI_DASM_PKG) -I $(DPI_DASM_SVDPI_PATH) -I $(SPIKE_PKG)/riscv -I $(SPIKE_PKG)/softfloat
+DPI_DASM_INC    = -I$(DPI_DASM_PKG) -I$(DPI_INCLUDE) -I$(SPIKE_PKG)/riscv -I$(SPIKE_PKG)/softfloat
```
silabs-hfegran, line 617: I would propose that we do not call the DASM-related variables that reference modified spike-files SPIKE_* but rather something akin to "DASM_SPIKE_*" or "SPIKE_DASM_*" in case other spike files are included in the future. There are differences between files coming directly from the spike repository and the ones we included. (Specifically small changes to remove decoding instructions into pseudo instructions which spike does by default). 

This comment applies to all included simulator-specific make files as well (e.g. clean_spike make targets)

silabs-hfegran (changes requested):

MikeOpenHWGroup (commented):
Thanks @strichmo.  Looks good to me. I'll hold off approving until you've has a chance to respond to @silabs-hfegran's  comments.

mk/Common.mk
```diff
@@ -610,14 +610,15 @@ vcs-unit-test:  vcs-run
...
+DPI_DASM_CXX    = g++
```
silabs-robin, line 618: There is a reason I used `c++` and not `g++`, and an example is shown in the attached screenshot.
However, I think the right choice is `g++` and _if_ there are any troubles (which our future setup should be able to catch) then fix it from there.
I just confirmed that isacov itself appears to work fine with the .so you provided and even after rebuilding the .so.

image

silabs-robin:
I read through the diff.
Apart from my comment about g++ and my [confusion about uvm field macros](http://verificationhorizons.verificationacademy.com/volume-7_issue-2/articles/stream/are-ovm-and-uvm-macros-evil-a-cost-benefit-analysis_vh-v7-i2.pdf) I found nothing to object to.
Looks good to me.

lib/dpi_dasm/dpi_dasm.cxx
strichmo: Yes, when running in Questa and passing back a const char* from a stack-based string object the char* buffer was regularly corrupted.  The global ensures the c_str() buffer is persistent though the SystemVerilog side copying out the string buffer.

mk/Common.mk
strichmo: Good call.  Will uniquify this particular spike.

mk/Common.mk
strichmo: My guess is perhaps this is side-effect of the global string issue above.  At this point we generally universally use GNU tools to build both applications and firmware code so for sake of consistency I generally point C builds to gcc and C++ to g++ by default and the user can retarget the make variables as needed.

strichmo:
> I read through the diff.
> Apart from my comment about g++ and my [confusion about uvm field macros](http://verificationhorizons.verificationacademy.com/volume-7_issue-2/articles/stream/are-ovm-and-uvm-macros-evil-a-cost-benefit-analysis_vh-v7-i2.pdf) I found nothing to object to.
> Looks good to me.

Good question:
My philosophy on skipping uvm_field_macros: DON'T!

I've had a lot of discussion with Mentor folks regarding this and I don't think it's good practice for a verification team.  Sure there's overhead, but there's more overhead in time lost when someone downstream uses your class, calls .clone() and can't figure out why it won't copy the fields correctly.   And this debug is usual indirect through another UVM agent/component consuming your object.  

It's just too easy to go ahead and implement the macros for "data-ish" object (sequence items, sequences, monitor transactions, etc.)

The only caveat I will add is that it's generally a good practice to improve debuggability of your objects by customizing do_print() (and do_record() for recording) to provide context-specific printing (i.e. you can remove don't care fields, clean up printing, etc.)  To do this you would or-in UVM_NOPRINT | UVM_NORECORD for each field that you want to customize the printing of.

strichmo:
@silabs-robin @silabs-hfegran @MikeOpenHWGroup All changes should be addressed in latest commit.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Great question/comment @silabs-robin, and thanks for passing that article along.   I'll second everything @strichmo said.

Simulator memory footprint and run-time performance are rarely a problem in verification, because the real bottleneck is engineering efficiency.  Today's SV simulators can fit a full-up UVM verificaiton environment full of macros, coverage, constraints, RAL plus a 2M-gate-equivalent RTL model and fit it into 8GiB of processor memory.   It may run 2x slower than the equivalent "verilog-only" testbench, but in return it will find more bugs and close coverage faster.

The boss will always buy you a bigger/faster computer, but they will never give you more time on the schedule.  :wink:

Based on @strichmo's latest comment, I will merge this PR.   Thanks @all.

PR #580, opened 2021-04-30
add in support for IO memory updating the RM memeory view

eroom1966:
This pull Request includes the feedback of IO updates for the virtual peripherals so that the RM sees the same values read from IO as the RTL

### DIFF
--- a/cv32e40p/tb/core/mm_ram.sv
+++ b/cv32e40p/tb/core/mm_ram.sv
@@ -471,13 +471,17 @@ module mm_ram
             data_rdata_mux = core_data_rdata;
         end else if(select_rdata_q == RND_STALL) begin
             data_rdata_mux = rnd_stall_rdata;
+            uvmt_cv32e40p_tb.iss_wrap.ram.RND_STALL = data_rdata_mux;
 `ifndef VERILATOR
             `uvm_fatal(MM_RAM_TAG, $sformatf("out of bounds read from %08x\nRandom stall generator is not supported with Verilator", data_addr_i));
 `endif
         end else if (select_rdata_q == RND_NUM) begin
             data_rdata_mux = rnd_num;
+            uvmt_cv32e40p_tb.iss_wrap.ram.RND_NUM = data_rdata_mux;
+
         end else if (select_rdata_q == TICKS) begin
             data_rdata_mux = cycle_count_q;
+            uvmt_cv32e40p_tb.iss_wrap.ram.TICKS = data_rdata_mux;
 `ifndef VERILATOR
             if (cycle_count_overflow_q) begin
                 `uvm_fatal(MM_RAM_TAG, "cycle counter read after overflow");
--- a/cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv
+++ b/cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv
@@ -234,7 +234,7 @@ module uvmt_cv32e40p_step_compare
     endfunction // compare
     
     int cycles = 0;
-    always @(posedge uvmt_cv32e40p_tb.dut_wrap.cv32e40p_wrapper_i.core_i.clk) begin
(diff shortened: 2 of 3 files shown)


cv32e40p/tb/core/mm_ram.sv
```diff
@@ -471,13 +471,17 @@ module mm_ram
             data_rdata_mux = core_data_rdata;
         end else if(select_rdata_q == RND_STALL) begin
             data_rdata_mux = rnd_stall_rdata;
+            uvmt_cv32e40p_tb.iss_wrap.ram.RND_STALL = data_rdata_mux;
```
MikeOpenHWGroup, line 474: Why does the `iss_wrap.ram` need to know about stalls?

cv32e40p/tb/core/mm_ram.sv
```diff
@@ -471,13 +471,17 @@ module mm_ram
...
 `ifndef VERILATOR
             `uvm_fatal(MM_RAM_TAG, $sformatf("out of bounds read from %08x\nRandom stall generator is not supported with Verilator", data_addr_i));
 `endif
         end else if (select_rdata_q == RND_NUM) begin
             data_rdata_mux = rnd_num;
+            uvmt_cv32e40p_tb.iss_wrap.ram.RND_NUM = data_rdata_mux;
```
MikeOpenHWGroup, line 480: Dito.

cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv
```diff
@@ -233,6 +233,11 @@ module uvmt_cv32e40p_step_compare
       `endif      
     endfunction // compare
     
+    int cycles = 0;
+    always @(posedge uvmt_cv32e40p_tb.dut_wrap.cv32e40p_wrapper_i.core_i.clk) begin
```
MikeOpenHWGroup, line 237: Can we make that `CV32E40P_CORE.clk`?  (The macro is defined at the top of the file.)

MikeOpenHWGroup (changes requested):
Hi @eroom1966, I've added a couple of comments/questions to `mm_ram` and requested a small change to `step_compare`.

Looking at the changes to `imperas/design/ram.sv`, I will reiterate my comments from the meeting.  It looks to me like we are replicating the virtual peripherals from the core's memory in the RM's memory.   This creates a significant maintenance headache, especially with all the "magic numbers" that have to be kept aligned.

I'll comment on this further in our email thread.

cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv
eroom1966: Aah, sorry missed that - changing it to the MACRO is fine.

cv32e40p/tb/core/mm_ram.sv
eroom1966: I was putting this in as forward planning, I recall something @strichmo mentioned that when stalls were in play the RTL/RM did not stay in sync.
I think anything with backdoor access, has to notify the RM memory.

cv32e40p/tb/core/mm_ram.sv
eroom1966: again as above

cv32e40p/tb/core/mm_ram.sv
MikeOpenHWGroup: Good thinking.

cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv
MikeOpenHWGroup: No worries, that's what PR reviews are for.  :-)  Seriously, I am rapidly becoming quite the fan of this workflow.

Can you push in a change to use the macro?

MikeOpenHWGroup (approved):

PR #585, opened 2021-05-07
Update Vplan location and template

MikeOpenHWGroup:
This _looks_ like a huge PR, but isn't.  The most visible change in this PR is to restructure the docs directory into `Reports` and `VerifPlans`.  None of the content of Reports has been changed, just moved.

This PR does include a change to the simulation Vplan spreadsheet template and VerificatonPlanning101 to clarify some confusion about the distinction between and ISS and RM.  Thanks for @silabs-hfegran for pointing that out.


MikeOpenHWGroup:
Thanks @strichmo.  I just merged #586, which launched a couple CI regressions.  Merging this will also launch CI since we haven't configured it to be smart enough not to regress updates to spreadsheets.   So I'll hold off merging this in for now.

PR #592, opened 2021-05-12
Expand ISACOV with sequences, groups, and hazards

silabs-robin:
Hi. Sorry for making a PR from someone else's fork, but the reason is that I want to continue adding features that build upon the work in their branch. And I would also like to have each feature in their own PRs, such that both reviews and changes are more focused, flexible, and decoupled. This doesn't follow the typical flow so I wasn't sure about the best way to proceed.

Description:
Adds coverage for sequences of instructions and data dependency hazards.
Also defines csr covergroups and enum.
And it organizes into a new ...tdefs.sv file.

@strichmo, do you know any reason it should not be merged?
Is it for example not complete enough so it should wait?
An alternative might be having a temporary branch in openhwgroup/core-v-verif?

I have two PR's that (independently) build upon this PR: https://github.com/openhwgroup/core-v-verif/pull/593 https://github.com/openhwgroup/core-v-verif/pull/594.

### DIFF
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
@@ -78,40 +78,51 @@ task uvma_isacov_mon_c::sample_instr();
   @(cntxt.vif.retire);
 
   mon_trn = uvma_isacov_mon_trn_c::type_id::create("mon_trn");
-  mon_trn.instr = new();
+  mon_trn.instr = uvma_isacov_instr_c::type_id::create("mon_instr");
 
   instr_name = dasm_name(cntxt.vif.insn);
   if (instr_name_lookup.exists(instr_name)) begin
-    mon_trn.instr.name = instr_name_lookup[instr_name];
+    mon_trn.instr.name = instr_name_lookup[instr_name];    
   end else begin
     mon_trn.instr.name = UNKNOWN;
     `uvm_warning("ISACOVMON", $sformatf("TODO: error couldn't look up '%s'", instr_name));
   end
-  mon_trn.instr.name =
-    cntxt.vif.is_compressed ?
-      (mon_trn.instr.name == JAL) ?
-        (cntxt.vif.insn[11:7] == 5'b00000) ?
-          C_J :
-        (cntxt.vif.insn[11:7] == 5'b00001) ?
-          C_JAL :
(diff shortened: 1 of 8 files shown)


lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -78,40 +78,51 @@ task uvma_isacov_mon_c::sample_instr();
...
-        UNKNOWN :
-      (mon_trn.instr.name == LW) ?
-        C_LW :
-      UNKNOWN :
-    mon_trn.instr.name;  // TODO get non de-compressed binary input instead of this
-
-  mon_trn.instr.rs1 = dasm_rs1(cntxt.vif.insn);
-  mon_trn.instr.rs2 = dasm_rs2(cntxt.vif.insn);
-  mon_trn.instr.rd = dasm_rd(cntxt.vif.insn);
+  // mon_trn.instr.name =
```
strichmo, line 90: @silabs-robin I commented out this bit of code because I didn't think it was needed anymore due to the use of the Spike DISASM.  Is this true?  Was I correct to comment out?

strichmo (changes requested):
Just a single comment regarding some code I commented out in ISACOV monitor.  Please confirm that my removal was valid.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-robin: That code was temporarily there because RVFI is not yet the source of incoming instructions.

Currently the tracer is used to obtain retired instructions. The problem is that it never gives compressed instructions (it gives de-compressed plus a flag).
Hence I created a very limited translator so I could test sampling of covergroups for compressed instructions.

The commenting out is OK because it will be incomplete either way.
It will be incomplete until we have actual compressed instruction word binaries coming in.

_(ps, this is part of why I want to check the correctness of the coverage model by diffing a recreated trace log; because it is wrong to do compressed disassembly of a non-compressed instruction and merely switching to rvfi won't be evidence enough that it has become correct)_

strichmo: Understood and agree with the reasoning.  Thanks, I'll push this review through.

strichmo (approved):

PR #594, opened 2021-05-12
Add more register crosses to individual ISA coverage

silabs-robin:
Description:
Adds crosses of register coverpoints.
Example `rd` crossed with `rs1`.
By default these covers are disabled, and can be enabled via the config object.

Continues from https://github.com/openhwgroup/core-v-verif/pull/592

ps:
I don't like how flags must be passed to cg instantiation like this:
`addi_cg   = new("addi_cg", .reg_crosses_enabled(cfg.reg_crosses_enabled));`.
But SystemVerilog doesn't allow using the class handle of `cfg` in the constant expression of the `with` clause, and I found no other option. (i.e. the simulator claims so, but I wasn't 100% convinced after reading the SV formal syntax).


MikeOpenHWGroup (approved):

MikeOpenHWGroup (approved):
Merging this in as the PR looks good, but I've noticed a failure of the debug_test in the CI regression.

PR #602, opened 2021-05-18
Strichmo/pr/cov warn suppression

strichmo:
Add another warning suppression related to coverage to Xcelium
The warning is informational only.


MikeOpenHWGroup (approved):
Looks good.  Running a local regression and will merge if/when they pass.

MikeOpenHWGroup:
Local regression of e40x ran clean.  Had a hickup on e40p that I cannot reproduce, so I will merge this and investigate off-line.

PR #616, opened 2021-05-20
Temporarily remove "WFI" interrupt tests

MikeOpenHWGroup:
As discussed in today's VTG meeting.  These failures are known false negatives that give grief to integrations onto the release branches.  We can reinstate once the false negative is resolved.

Note that these will need to be pushed to the e40x/release and e40s/release branches.


strichmo (approved):
Changes OK.with me

PR #625, opened 2021-05-24
add support for minstret(h) from REF -> DUT Compare

eroom1966:
This PR Adds full support for minstret prediction, propagated to a GPR bia a CSR read instruction


cv32e40p/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40X.sv
```diff
@@ -0,0 +1,603 @@
+/*
```
MikeOpenHWGroup, line 1: Hi @eroom1966, I wonder if this file shouldn't be located someplace else such as `cv32e40x/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40X.sv`?

MikeOpenHWGroup (commented):
Looks good @eroom1966.   Before approving I'd like to resolve the review I started on  `cv32e40p/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40X.sv `.   @strichmo, do you have an opinion?

cv32e40p/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40P.sv
```diff
@@ -575,20 +575,18 @@ module CPU
     endfunction
     
     initial begin
-        if ($test$plusargs("USE_ISS")) begin
```
strichmo, line 578: The if ($test$plusargs("USE_ISS")) code needs to be there to facilitate run-time disablement of the ISS for certain tests.

Is there another reason this was removed (i.e. was it breaking something)?

strichmo (changes requested):

cv32e40p/vendor_lib/imperas/riscv_CV32E40P_OVPsim/sv/riscv_CV32E40X.sv
strichmo: Yes, it should be on the CV32E40X branch?

I'm a bit confused with this update as it appears that riscv_CV32E40X.sv and riscv_CV32E40P.sv have different sets of interfaces (The X version has updated RVVIs to incorporate interrupt, debug, etc.).  Is the intention to ensure that E40P and E40X ISS models are compatible with the newer interface definitions?

eroom1966:
@MikeOpenHWGroup 
I think this PR is no longer valid - am I correct, if so I will close ?

strichmo:
I believe this PR should be closed without merging as well.

MikeOpenHWGroup:
Agree.  The PR is now very stale and has multiple conflicts.

eroom1966:
closing as no longer relevant

PR #634, opened 2021-05-25
fix running xcelium and Indago in GUI mode with waves

strichmo:
will not automatically start running simulation when starting


MikeOpenHWGroup (approved):
All good with me.  I'll wait for @silabs-hfegran to comment before merging.

silabs-hfegran:
Looks good too me, 

I presume this was some misconfiguration on one of my checked out repos - The plain openhw core v verif seems to work fine. Thanks for looking into this, Steve.

MikeOpenHWGroup (approved):

PR #636, opened 2021-05-31
Diffing isacov log vs tracer log

silabs-robin:
This change is for checking that ISA coverage samples correctly (try to detect false positives/negatives), and includes some corresponding fixes to the sampling/decoding itself.
It changes formatting in isacov's logging to be compatible with 40x's tracer log [1].
It also adds to get_instr_type() and convert2string() to handle those cases which were missing.
Essentially, since incoming binary words are decoded into uvm objects (which coverage samples) and these objects are printed to a log, they can be compared against a reference log (if the logs match, then the uvm objects should have correct data in them).

This diffing of logs actually did uncover some instructions being sampled wrong.
For example, branch instructions had the offset's sign bit wrong (due to Spike internals) and both B-type and U-type needed shifting of the result obtained from dpi_dasm/Spike disassembler, "M" extension weren't logged properly, etc.
Coverage for those items were then false before, until now being found and fixed.
NB! RVC (compressed) has not yet been tested thoroughly because we wait for RVFI to deliver actual compressed instructions to isacov (diffing the logs will fail on compressed instrs unless filtered [2] away).

By default, the log diffing is not enabled.
Since it needs to be done as post-processing, I added the line `$(POST_TEST)` to uvmt/xrun.mk `test:` target.
It allows e.g. `make test TEST=hello-world POST_TEST='make isacov_logdiff'` or it can be used with ci_check, and can potentially be used for any other kind of desired post-processing.

[1] The tracer log format is not standardized. I think it should have tab-separated columns, comma-separated within-column data, and allow for leading/trailing spaces per cell. And it should maybe have defined what gets printed as hex/decimal/string etc. I also don't like how non-general the filtering of the logs and file names must be because of this.

[2] One can use 'sed -i "s/c\.//" $(ISACOV_TRACELOG) && make isacov_logdiff' to filter compressed instructions before diffing the logs.

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
@@ -522,7 +522,7 @@ bind cv32e40x_wrapper
    assign core_cntrl_if.clk = clknrst_if.clk;
 
    always @(dut_wrap.cv32e40x_wrapper_i.tracer_i.retire) -> isacov_if.retire;
-   assign isacov_if.insn = dut_wrap.cv32e40x_wrapper_i.tracer_i.insn_val;
+   assign isacov_if.instr = dut_wrap.cv32e40x_wrapper_i.tracer_i.insn_val;
    assign isacov_if.is_compressed = dut_wrap.cv32e40x_wrapper_i.tracer_i.insn_compressed;
 
    // Capture the test status and exit pulse flags
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_if.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_if.sv
@@ -23,7 +23,7 @@ interface uvma_isacov_if;
   //   1) use RVFI, and/or 2) use a TLM port instead.
 
   event        retire;
-  logic [31:0] insn;
+  logic [31:0] instr;
   logic        is_compressed;
 
 endinterface
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
@@ -108,6 +108,9 @@ function string uvma_isacov_instr_c::convert2string();
   if (name inside {SLLI, SRLI, SRAI}) begin
     return $sformatf("%s x%0d, x%0d, 0x%0x", name.name(), rd, rs1, rs2);
   end
+  if (name == FENCE_I) begin
+    return $sformatf("fence.i");
+  end
 
   // Printing based on instruction format type
   if (itype == R_TYPE) begin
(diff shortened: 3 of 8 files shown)


MikeOpenHWGroup (commented):
It is good to see that the coverage model is getting the attention it deserves and these fixes should all be recognized as "good catches".  It would be good to get these cov model fixes in as soon as possible.

Before we commit the scriptware to implements the tracer log diff'ing it would be useful to have a strategy discussion.  As a general rule statements like "changes formatting in isacov's logging to be compatible with 40x's tracer log" and that sed/awk one-liner in the `isacov_logdiff` target of `uvmt.mk` gives me pause.  This diff'ing capability will be very brittle and difficult to maintain.

Also, it would be a good idea to make the dumping/diff'ing controllable from the UVM testcase, rather than a Makefile target.

@silabs-robin, I have a suggestion.  Let's close this PR and issue two news ones.  The first should implement the coverage model fixes.  This is high-value and should get into our regressions as soon possible.   The second PR can implement the logging/diff'ing and we can use that PR to discuss strategy.

silabs-robin:
> [...] sed/awk one-liner [...] This diff'ing capability will be very brittle and difficult to maintain.

I think so too. That is what I was thinking about regarding "how non-general the filtering of the logs and file names must be". I don't immediately have any clue for how to clean it. The specificity of the tracer format might be unavoidable, but extracting the data would be a lot cleaner if the format was different. I don't think sed/awk is a bad fit, rather the log format is currently clunky.

> Also, it would be a good idea to make the dumping/diff'ing controllable from the UVM testcase, rather than a Makefile target.

Maybe yes. It could potentially fit more purely into the conceptual framework of UVM etc. Though we do have some alright makefile customs and I think the target I added was somewhat straight-forward. I don't have a strong opinion on it and am willing to follow whatever the consensus is.

> I have a suggestion. Let's close this PR and issue two news ones. The first should implement the coverage model fixes. This is high-value and should get into our regressions as soon possible. The second PR can implement the logging/diff'ing and we can use that PR to discuss strategy.

Sorry if I appear a bit lazy here, but could we do them both in this PR? Do we currently have regressions with isacov enabled and corresponding coverage databases that we want to have a lot of history for, or anything similar to that?
Though, I could split it in two PRs. Then everything should be kept except for anything makefile-related. The simplest solution would be to make a new commit just deleting those lines and keeping this PR. Or I could learn how to use more advanced features of git to extract certain commits, or create a patch and just apply it to two fresh branches as two single commits.

MikeOpenHWGroup:
> Sorry if I appear a bit lazy here, but could we do them both in this PR? 

In the interests of making progress, I will agree and merge this in as it passes a local regression.  However, I am still uncomfortable maintaining support for anything this brittle in the env, so I have opened issue #639 to discuss the topic.

MikeOpenHWGroup (approved):

PR #640, opened 2021-06-01
Add TINFO, MCONTEXT and SCONTEXT to instr_csr_t

MikeOpenHWGroup:
TINFO, MCONTEXT and SCONTEXT are missing from instr_group_t and this is causing UVM warnings, so I took the liberty of adding them.  Also reformatted the members of the enum to line up the "=" (so it looks like a lot was changed).

Signed-off-by: MikeOpenHWGroup <mike@openhwgroup.org>


strichmo (approved):
Looks OK to me.
Thanks for adding.

PR #643, opened 2021-06-02
cleanup of uvma_debug_drv_c::drv_req()

MikeOpenHWGroup:
This PR resolves (solves?) the issue we have been seeing with the `debug_test_boot_set` test-program when running under DSIM.  I have demonstrated to myself that the test now passes (and behaves as expected) with both DSIM and XRUN.

### DIFF
--- a/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
+++ b/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
@@ -113,21 +113,15 @@ task uvma_debug_drv_c::run_phase(uvm_phase phase);
 endtask : run_phase
 
 
-
+// WARNING If no time is consumed by this task, a zero-delay oscillation loop will occur and stall simulation
 task uvma_debug_drv_c::drv_req(uvma_debug_seq_item_c req);
-   `uvm_info("DEBUGDRV", $sformatf("Driving debug: %s",req.sprint()), UVM_HIGH); 
-   cntxt.vif.drv_cb.debug_drv <= 1'b1;
- //  while(1) begin
-       repeat (req.active_cycles) @(cntxt.vif.mon_cb);
-       cntxt.vif.drv_cb.debug_drv <= 1'b0;           
- //  end
 
-   // TODO Implement uvma_debug_drv_c::drv_req()
-   //      Ex: cntxt.vif.drv_cb.abc <= req.abc;
-   //          cntxt.vif.drv_cb.xyz <= req.xyz;
-   
-   // WARNING If no time is consumed by this task, a zero-delay oscillation loop will occur and stall simulation
-   
+   `uvm_info("DEBUGDRV", $sformatf("Driving debug:\n%s",req.sprint()), UVM_HIGH)
+   @(cntxt.vif.drv_cb) cntxt.vif.drv_cb.debug_drv <= 1'b1;
+   repeat (req.active_cycles) @(cntxt.vif.mon_cb);
+   cntxt.vif.drv_cb.debug_drv <= 1'b0;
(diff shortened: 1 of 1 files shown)


strichmo (approved):
I believe this should be OK.  The clocking block wait should wait on a clock edge (I've also seen ##0 used if you have a default clocking block).

lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
```diff
@@ -113,21 +113,15 @@ task uvma_debug_drv_c::run_phase(uvm_phase phase);
 endtask : run_phase
 
 
-
 task uvma_debug_drv_c::drv_req(uvma_debug_seq_item_c req);
-   `uvm_info("DEBUGDRV", $sformatf("Driving debug: %s",req.sprint()), UVM_HIGH); 
+
+   @(cntxt.vif.drv_cb); // WARNING If no time is consumed by this task, a zero-delay oscillation loop will occur and stall simulation
```
AnnaAdmin, line 118: I'm not familiar with the code so this may be obvious to others: why does the delay at the beginning of the task use drv_cb while the rest of the code is synced by mon_cb?

aimeepsutton, line 118: Why does this delay use `drv_cb` while the rest of the task is synced by `mon_cb`?

strichmo: Unless there is some clock gating with the interface I am not aware of (there shouldn't be), the code should really be using the same clocking block for all driving activities.  Since the clock is the same it's probably OK functionally but not the best style.

lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
```diff
@@ -113,21 +113,15 @@ task uvma_debug_drv_c::run_phase(uvm_phase phase);
...
+   `uvm_info("DEBUGDRV", $sformatf("Driving debug:\n%s",req.sprint()), UVM_HIGH)
    cntxt.vif.drv_cb.debug_drv <= 1'b1;
- //  while(1) begin
-       repeat (req.active_cycles) @(cntxt.vif.mon_cb);
-       cntxt.vif.drv_cb.debug_drv <= 1'b0;           
- //  end
+   repeat (req.active_cycles) @(cntxt.vif.mon_cb);
```
aimeepsutton, line 121: would the following have the same effect as the proposed change?
`repeat (req.active_cycles + 1) @(cntxt.vif.mon_cb)`

lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
MikeOpenHWGroup: That could be a gaff on my part.  Agree that it _should_ work either way and that it is not the best style.

lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
MikeOpenHWGroup: For DSIM, yes it would.  For XRUN, the debug request is asserted for two clock cycles.  I believe (don't know) that there is a delta-cycle issue here that is being evaluated differently by DSIM and XRUN.

MikeOpenHWGroup:
Thanks for approving this @strichmo.  I will hold off merging it in until I've had a chance to discuss in person with @aimeepsutton.  While this update resolves the problem, there is a difference in behaviour between XRUN and DSIM that I do not understand.

MikeOpenHWGroup:
Update.  @aimeepsutton and I had a chance to review and discuss this PR and there is no reason not to merge it in, so I will do so.  Having said that, there is still a difference in behaviour between DSIM and XRUN when running this code which needs to be understood.  I will create an issue to track that discussion.

PR #648, opened 2021-06-03
Fix verilator for CV32E40P

MikeOpenHWGroup:
Please hold you nose and review.  I just noticed that the fixes put in for the cycle-counter virtual peripheral do not work for Verilator.

Also points to the latest version of CV32E40P RTL.

### DIFF
--- a/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
+++ b/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
@@ -113,21 +113,15 @@ task uvma_debug_drv_c::run_phase(uvm_phase phase);
 endtask : run_phase
 
 
-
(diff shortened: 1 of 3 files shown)


lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
```diff
@@ -113,21 +113,15 @@ task uvma_debug_drv_c::run_phase(uvm_phase phase);
 endtask : run_phase
 
 
-
```
strichmo, line 116: I seem to recall these changes were part of another PR.  Did you mean to merge these together.  (I'm OK with change but wanted to make sure we weren't double-changing and perhaps creating an inadvertent conflict.)

strichmo (changes requested):
Please take a look at the comment about the debug_drv change.

strichmo (approved):
Approving as the change should be identical.

PR #659, opened 2021-06-04
Fix a couple of typos in ISACOV

MikeOpenHWGroup:
Hi @silabs-robin.  This PR is to fix a typo I introduced in `uvma_isacov_tdefs.sv`.  While I was looking into this I noticed a second typo in `uvma_isacov_cov_model.sv`.

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -136,7 +136,7 @@ covergroup cg_csritype(
   cp_csr: coverpoint instr.csr {
     bins CSR[] = {[USTATUS:VLENB]} with (cfg_illegal_csr[item] == 0);
   }
-  cp_immu: coverpoint instr.immu[4:0];
+  cp_immu: coverpoint instr.immu[31:12];
 endgroup : cg_csritype
 
 covergroup cg_cr(
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
@@ -311,7 +311,7 @@ typedef enum bit[CSR_ADDR_WL-1:0] {
   TDATA3         = 'h7A3,
   TINFO          = 'h7A4,
   MCONTEXT       = 'h7A8,
-  SMCONTEXT      = 'h7AA,
+  SCONTEXT       = 'h7AA,
   DCSR           = 'h7B0,
   DPC            = 'h7B1,
   DSCRATCH0      = 'h7B2,


silabs-robin:
Hi @MikeOpenHWGroup sorry for a late reply.

First, I want to reply about `immu`.

Good catch btw; it was a subtle error.
The spec for `CSRR*I` actually talks about `uimm` meaning "unsigned immediate" and is hence probably unrelated to U-type immediate. Another evidence for this is in the instruction set listing, where the top of the page lists the relevant instruction type formats and U-type is not among them. That is one reason I don't think `immu` should be used here.
Another reason I would prefer not using the `immu` field is that it requires using indices like `[a:b]` and as we just saw this can be a bit fragile.
I think maybe both `immu[4:0]` and `immu[31:12]` are not right, because the `CSRR*I` immediate should essentially be `rs1` (either exactly or zero-extended, preferably exactly).

In my opinion `cp_immu: coverpoint instr.immu[...];` should _maybe_ be `cp_uimm: coverpoints instr.rs1;`?

silabs-robin:
Regarding SCONTEXT, I don't really know what it is. But I guess that is the part you wanted to fix in the first place so that one is probably alright then.
I couldn't find it in the priv spec, but it is in the 40P user manual, so maybe it is either a 40P thing or it is in the debug spec or something else.
Then I only have my comments above regarding immu in csr itype.

MikeOpenHWGroup:
It _is_ confusing @silabs-robin, and the RISC-V spec has changed the nomenclature it uses for CSR fields over the years.  As of riscv-spec-20191213 the CSRR*I instructions all use `uimm[4:0]`, so I think we should also use that in our coverage model.  

> The spec for CSRR*I actually talks about uimm meaning "unsigned immediate" and is hence probably unrelated to U-type immediate. Another evidence for this is in the instruction set listing, where the top of the page lists the relevant instruction type formats and U-type is not among them. That is one reason I don't think immu should be used here.

Keep in mind that CSR instructions are not considered to be one of the R-, I-, S-, B-, U-, or J-type instructions.

> In my opinion `cp_immu: coverpoint instr.immu[...];` should maybe be `cp_uimm: coverpoints instr.rs1;`

Not sure I agree.  For CSRR*I instructions, bit field 19:15 contains a 5-bit unsigned immediate value, not an index a GPR.

silabs-robin:
> Keep in mind that CSR instructions are not considered to be one of the R-, I-, S-, B-, U-, or J-type instructions.

This surprised me. I-type is the only one with `funct3` and a continuous immediate at bits `[31:20]`.
To compare with something similar, the shift instructions (SLLI etc) "are encoded as a specialization of the I-type format" even if the `shamt` is a field of its own, and also the EBREAK instruction is I-type even if it has special constraints on `rs1`.
Though the Zicsr chapter doesn't seem to mention "type"...

Actually I just found a bit of information on this.
_"SYSTEM instructions are used to access system functionality that might require privileged access and are encoded using the I-type instruction format."_.
And the CSR instructions are SYSTEM instructions.

> the CSRR*I instructions all use uimm[4:0]

This is true. One could add `uimm` to `uvma_isacov_instr_c`? One might also add `shamt` and some other special fields to that class (slli currently repurposes the R-type cg to get shamt from rs2, `cg_rtype slli_cg;  // TODO own cg?`). I think that might be more complex to manage (relating it to e.g. dpi_dasm) than to just repurpose names like slli did with rs2, but I am not sure which alternative would be better.
Note also that `uimm[4:0]` is replaced with `uimm` in the instruction set listing, and that `uvma_isacov_instr_c` only holds the field as seen in the instruction format without any zero-extending to 32-bit.
Maybe the naming here is a bit prone to being confused.

`uimm` must be bits [19:15] (same as rs1).
But `immu[31:12]` (U-type immediate) are bits [31:12], so it was called a typo.
And `immu[4:0]` (also U-type immediate) are bits [16:12], so I don't think this is the CSRs' `uimm` either, and I am not sure those bits even exist as U-type immediate is [31:12] in the specs format and in the instr class.
I think one could use either `immu[19:15]` (I don't like using U-type fields in a I-type instruction but I am ok with it), or one could create an alias for rs1 and call it `uimm`, or one could just use rs1 directly, or maybe something else.

This reply became a bit longer than I anticipated.

MikeOpenHWGroup:
>This reply became a bit longer than I anticipated.

Its a good discussion, but maybe we could move it to an issue or an in-person discussion.   In the meantime, we need to either reject or accept the update to `lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv`.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -136,7 +136,7 @@ covergroup cg_csritype(
   cp_csr: coverpoint instr.csr {
     bins CSR[] = {[USTATUS:VLENB]} with (cfg_illegal_csr[item] == 0);
   }
-  cp_immu: coverpoint instr.immu[4:0];
```
silabs-robin, line 139: My understanding is that `immu[4:0]` is not correct. The CSR format's `uimm[4:0]` refers to a different set of bits, so this would sample something else than what is intended.

silabs-robin (changes requested):
I am requesting a change because it appears to me that the proposed sampling does not match the bits corresponding to the field I think it should match. Although I do also think that the code that was there from before was not right either.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
MikeOpenHWGroup: OK, can I leave that with you?   I am not intimately familiar with how the new ISACOV uses spike to unpack instructions so I'm not the best person to resolve this.

silabs-robin: > OK, can I leave that with you?

Jepp. I can fix it. Then it is probably fine go ahead and merge this PR.

I created https://github.com/openhwgroup/core-v-verif/issues/664 as a reminder, but I don't have permissions to assign myself.

strichmo (approved):

PR #663, opened 2021-06-10
Rvfi rvvi initial e40x

eroom1966:
First stage of a commit to add the new CV32E40X feature requirements

I have also added a test which you can choose to keep/discard - this test exhaustively accesses all addressable CSR registers in order to ensure that registers which should not be present, cause an illegal instruction exception


strichmo (approved):
Thanks Lee for the addition.  I will merge, and then merge in another PR to fix a few items with e40x to get our tests back to the previous state.

PR #669, opened 2021-06-15
NMI handler added to corev-dv

silabs-mateilga:
NMI handler generation addet to corev-dv for 40x, and related address hardcoding added to the linker. Needs to be revisited when the address is determined(and if it's ever revised) and the handler itself need to be verified when the NMI test capability is added to core-v-verif.

### DIFF
--- a/cv32e40x/env/corev-dv/cv32e40x_asm_program_gen.sv
+++ b/cv32e40x/env/corev-dv/cv32e40x_asm_program_gen.sv
@@ -247,7 +247,43 @@ class cv32e40x_asm_program_gen extends corev_asm_program_gen;
     end
     gen_section(get_label($sformatf("%0smode_intr_handler", mode_prefix), hart),
                 interrupt_handler_instr);
+
+
+    //generate nmi section
+    gen_nmi_handler_section(hart);
+
   endfunction : gen_interrupt_handler_section
+
+  // generate NMI handler.
+  // will be placed at a fixed address in memory, set in linker file
+  //TODO: verify correct functionality when NMI test capability is ready
+  virtual function void gen_nmi_handler_section(int hart);
+    string nmi_handler_instr[$];
+
+    // Insert section info so linker can place
+    // debug code at the correct adress   
+    instr_stream.push_back(".section .nmi, \"ax\"");
+    
+    // read relevant csr's 
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mepc", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mcause", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mtval", cfg.gpr[0]));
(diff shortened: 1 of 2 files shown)


cv32e40x/env/corev-dv/cv32e40x_asm_program_gen.sv
```diff
@@ -247,8 +247,44 @@ class cv32e40x_asm_program_gen extends corev_asm_program_gen;
...
+    
+    // read relevant csr's 
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mepc", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mcause", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mtval", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mie", cfg.gpr[0]));
+
+    //set error and end
+    nmi_handler_instr.push_back($sformatf("li a0, test_ret_val"));
+    nmi_handler_instr.push_back($sformatf("li t0, test_fail"));
```
strichmo, line 275: I got an assembler error that test_fail was not defined.  I wonder if it wouldn't be easier just to do a straight-forward jump (jr) to test_done directly after doing the "system interrogation CSR reads" above?

strichmo (changes requested):
There is an assembler fail when I try to use this.  Have a suggestion for simplifying in the comment.

cv32e40x/env/corev-dv/cv32e40x_asm_program_gen.sv
```diff
@@ -247,8 +247,38 @@ class cv32e40x_asm_program_gen extends corev_asm_program_gen;
...
+    
+    // read relevant csr's 
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mepc", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mcause", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mtval", cfg.gpr[0]));
+    nmi_handler_instr.push_back($sformatf("csrr x%0d, mie", cfg.gpr[0]));
+
+    nmi_handler_instr.push_back($sformatf("la x%0d, test_done", cfg.scratch_reg));
+    nmi_handler_instr.push_back($sformatf("c.jr x%0d", cfg.scratch_reg));
+
```
strichmo, line 275: Sorry to split hairs on this, but general code in corev-dv "infrastructure" code must work on any RISC-V implementation regardless of whether the C extension is implemented.  In other words, corev-dv should always "work" on a core that just implements I.  Thus, any code that generates any "C" instructions directly must take care that the main configuration object in riscv-dv says that c extension is indeed supported.  You should just use "jr" pseudo-op here.

Note that the assembler will substitute compressed instructions if it can (based on instruction and register usage) and you specified the arch as supporting C (all of our tests do).

strichmo (changes requested):

strichmo (approved):

PR #670, opened 2021-06-15
Move the "Verification Strategy"

MikeOpenHWGroup:
This is the first pull request to move the "Verification Strategy" from core-v-docs repo to the docs directory in core-v-verif.  This PR simply installs a copy of the current docs here.  The second PR will be on core-v-docs to remove the docs from that repo.


strichmo (approved):
Looks good.  Verified new location is still build-able.

PR #682, opened 2021-06-17
Rvfi rvvi initial e40x shared

eroom1966:
This pull request will add the restructured RM using shared data through the DPI


strichmo:
Confirmed update does not cause any additional failures in either checking mode.  Merging.

PR #689, opened 2021-06-21
update for support of the PMA and bus faults on E40X

eroom1966:
@strichmo 
Steve this PR adds in the PMA for the E40X, you should now be able to access the parameters in a similar structure to that of the RTL definition, adding --showoverrides will provide the full commandline option

this also has the Busfaults part of the interface, I dont have a way to drive this, but managed to artificially create faults and observe that an exception is taken by the RM, of course the simulation then ends as it immediately deviates

If you want to have a shared desktop to go through this, we can arrange that for Mon/Tues (I am on vacation Wed-Fri)
We should try to include a PMA test, and a busfault test if possible


eroom1966:
A further comment here
I want to consolidate the two imperas installation folders - as they contain the same data.
Where do you think it should be located in the source tree, as it is common to both cores, and in the future will be common to additional cores.

strichmo:
> A further comment here
> I want to consolidate the two imperas installation folders - as they contain the same data.
> Where do you think it should be located in the source tree, as it is common to both cores, and in the future will be common to additional cores.

@eroom1966 @MikeOpenHWGroup Good question: I'll start an email on this with my opinion to keep it separate from this PR.

strichmo:
I will go ahead and merge this PR into rvfi_rvvi_initial_e40x.  **This should be the last update for that branch.**

Then I will merge into cv32e40x/dev using a separate PR.

PR #690, opened 2021-06-21
Pma directed test

silabs-robin:
This PR adds directed tests for PMA features.

The tests correspond to the PMA vPlan entries about directed tests.
DUT pma config params are set from the following chain of causality: "make" command line, yaml cfg files, tb, wrapper, dut.
Several parts of this depend on features that have not yet been implemented in RTL, so they have been commented away and the plan is to gradually enable them as the RTL becomes ready.
The tests cannot be fully run with iss enabled.

The ifdefs in `tb` I think are necessary, but I am open to suggestions for how to avoid them.
And the parameter configs in `tb` could for example be moved to a separate file to keep things clean. Though I think it could be wise to wait a bit as there is work being done on riscv-dv that might impact these decisions.


strichmo (changes requested):
I would like to see a more rigorous approach to the actual PMA configurations.  I has assumed we would have a set of PMA configurations explicitly specified in the vplan indexed 0 to <n>.  I would assume PMA config 0 is empty.  

To specify the PMA regions your code in uvmt_cv32e40x_tb.sv should go into the testbench pacakge.  The code itself is fine.  Based upon setting and ifdef of PMA_CFG<n> you would simply ifdef in a selection and set the PMA parameters accordingly to be picked up by the dut_wrap instantiation as you show.

silabs-robin:
> I would like to see a more rigorous approach to the actual PMA configurations.

(more comments on this below)

> I had assumed we would have a set of PMA configurations explicitly specified in the vplan indexed 0 to . I would assume PMA config 0 is empty.

If implementing those configs, then I also would expect it to be like that.

> To specify the PMA regions your code in uvmt_cv32e40x_tb.sv should go into the testbench pacakge.

I can fix this. That is a cleaner way of organizing it.

--------

Configs:

I would like to weigh the pros and cons.
The configs I wrote were chosen because we hadn't (and still haven't) decided on infrastructure and these configs can hit every test item marked as directed test while using the default unmodified linker script.
The reason we ended up with the list of configs is because Henrik and I were discussing how the original "Regions" test items could be practically tested. The list was made so we could have a feasibly testable limited set of possibilities while still hitting important scenarios.
I think it should be decided how comprehensive the pma directed tests should be and how to delineate responsibilities between directed and constrained random for the pma. And I would like to have a clearer view of why the complexity cost is worth it.

I see at least four alternatives.
1) One is to let the directed tests continue to use their own configs (separate from the vPlan configs) and let the constrained random tests use the vPlan configs. _(risks directed tests not hitting all edge cases)_.
2) Wait for pma riscv-dv to further explore what the needs are for such an infrastructure and set of configs, then update the directed tests according to findings. _(risks unproportionately distributing our workloads)_.
3) adding the configs now, split the directed tests so that each test uses one known config. _(risks becoming high overhead)_.
4) adding the configs now, adding infrastructure to let the tests be aware of which config is currently selected and use if-statements to enable certain checks for selected configs _(risks becomming complicated/unwieldy)_.

cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
```diff
@@ -19,7 +19,58 @@
 `define __UVMT_CV32E40X_CONSTANTS_SV__
 
 
-
+   `ifdef PMA_CUSTOM_CFG
```
MikeOpenHWGroup, line 22: Does this _need_ to be done with compile macros (tick defines)?

silabs-robin: I think they are needed, but I am not 100% sure. I tried to look for alternatives but I couldn't find any that works. For example, plusargs don't seem to be usable at elaboration-time. One could maybe use simulator-specific command line arguments to set parameter values. These pma values are fed to the parameters of the DUT instantiation, so that limits our options.

MikeOpenHWGroup: Parameters and generate statements?

silabs-robin: It might work, but then I think one would need simulator-specific command line arguments to set the parameters. At least I don't know any other way to let `tests/cfg/` yaml files decide about parameters.

strichmo: The "coverage" goals of the directed test should be very modest.  I would just like to see the configs used by the directed test (maybe it's just 2 configs at most) come from the same pool of PMA configs that the random tests use.  

The primary goals for directed tests are:
- Pipeclean design
- Pipeclean ISS model and integration of PMA configuration of ISS
- Facilitate any directed coverage goals that may come up after debug

strichmo:
> > I would like to see a more rigorous approach to the actual PMA configurations.
> 
> (more comments on this below)
> 
> > I had assumed we would have a set of PMA configurations explicitly specified in the vplan indexed 0 to . I would assume PMA config 0 is empty.
> 
> If imple

The "coverage" goals of the directed test should be very modest.  I would just like to see the configs used by the directed test (maybe it's just 2 configs at most) come from the same pool of PMA configs that the random tests use.  

The primary goals for directed tests are:
- Pipeclean design
- Pipeclean ISS model and integration of PMA configuration of ISS
- Facilitate any directed coverage goals that may come up after debug

cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
strichmo: Passing parameters to a simulator are not commonly used (at least in my experience) and generally awkward compared to the relatively simple way one specifies a macro (e.g. +define+PMA_CFG2)

silabs-robin:
> I would just like to see the configs used by the directed test (maybe it's just 2 configs at most) come from the same pool of PMA configs that the random tests use.

That makes sense. Although it isn't easily doable at the moment. Just as for the dv testing, we need a configurable memory model and custom linker scripts.
I expect that it should not be complicated to later just update the C defines I used and point them to appropriate addresses in the vplan-based configs.

cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
MikeOpenHWGroup: Sorry to be pedantic about this, but we are heading in the wrong direction.  A set of `ifdef PMA_CFG_n` is going to be lead to a rats nest of code.  I see the point about parameters - that is not going to work either (my bad).

Could we not have a config object that defines the PMA config?  For now, the PAM config will be defined in the test-program yaml to control corev-dv and the BSP.   Values from the yaml can then be constraints on the UVM environment.  (At some future date, when corev-dv is integrated into the UVM environment, we may be able to use proper constraints on the PMA config from the UVM test-case.)

silabs-robin: > Sorry to be pedantic about this

That is absolutely fine by me. I usually enjoy pursuing crystal-like results, even if I often can't reach it.

> Could we not have a config object that defines the PMA config?

Objects as in class instances, like the "cfg" we have in UVM?
I just gave it a try.
In a module, one can apparently make variables like `uvmt_cv32e40x_test_cfg_c cfg` and initialize them like `initial void'(uvm_config_db#(uvmt_cv32e40x_test_cfg_c)::get(null, "", "cfg", cfg));`.
But the values need to be available at elaboration-time and they need to be constant expressions, so one can't instantiate modules with `.MYPARAM(cfg.myconst)`.
Xcelium had this to say:
_"Reference to automatic variables and elements or members of dynamic variables is not allowed outside procedural blocks"_
and
_"Illegal operand for constant expression"_.
Here, the `const int unsigned myconst` I used might not even be a constant expression. I think I remember some thing about `const` not necessarily being a "constant expression"? And the `cfg` object itself is not contant.
The "constant expression" error is about the same I saw trying generate-statement with plusargs.

silabs-robin:
I just updated ISS' PMA parameters. ISS support for that is not yet on cv32e40x/dev branch, however.

To summarize:
This PR adds the vPlan's directed tests.
Several parts are commented out and meant to be included as the RTL implementation progresses.
Configs are set via cli and yaml, but ifdefs are used and all my attempts at finding other solutions failed.
ISS is configured, but the shared object files are not in the PR's target branch, and there is mismatch between RVFI/RVVI.
The configs are custom and does not match the vPlan, but to fix that we must both decide on the custom linker script infrastructure and also get a memory model that lets us configure regions freely.

silabs-robin:
Hi @strichmo is this PR ok to merge now?

strichmo (approved):

strichmo:
OK to merge now from my side.

PR #700, opened 2021-06-30
fix ebreak causing debug entry

strichmo:
use dcsr cause to determine external entry in RVFI


strichmo:
Pushing this through so that design folks can fix RVFI on ebreak entry.

Issue #664, opened 2021-06-10
isacov: The uimm field of the CSR immediate coverpoint is sampled incorrectly


silabs-robin:
More details here https://github.com/openhwgroup/core-v-verif/pull/659

PR #713, opened 2021-07-07
Fix cp_uimm in cg_csritype

silabs-robin:
As per https://github.com/openhwgroup/core-v-verif/issues/664 this PR updates `cp_uimm` in sampling of `cg_csritype`.

The problem was a slightly confusing term in riscv formats, `uimm` (_"unsigned immediate"_) which name is used only in csr instructions and some compressed instructions. It is not the same as how `imm` differs between the instruction formats (e.g. U-type, R-type, etc). Hence, the existing `immu[31:12]` did not sample the immediate of csr instructions.

ps. There is also a question of whether "special" fields like `uimm` and `shamt` should be made into new properties of the instruction class or if it is fine to reuse the existing canonical fields like `rs1` and `rs2`. (i vote for the latter)

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -136,7 +136,7 @@ covergroup cg_csritype(
   cp_csr: coverpoint instr.csr {
     bins CSR[] = {[USTATUS:VLENB]} with (cfg_illegal_csr[item] == 0);
   }
-  cp_immu: coverpoint instr.immu[31:12];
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @silabs-robin, approved.

FWIW, my thoughts on "special" fields:
- `shamt` should not be a property in our instruction class.  This is a concept from the ABI and we are not verifying the ABI.  We should restrict our ISA model to the hardware view.
- I am less clear on `uimm`.  I was operating under the notion that this was part of the hardware defintion of the ISA, but I may be wrong about that.

I too prefer to use the existing canonical fields like rs1 and rs2.

PR #721, opened 2021-07-13
Some updates to run core testbench with questa

pascalgouedo:
Hi,
These updates allowed to compile the core testbench on questa and to run some of the custom tests (dhrystone, coremark, fibonnaci, hello-world).
Regards,
Pascal.


cv32e40p/tb/core/mm_ram.sv
```diff
@@ -472,18 +472,24 @@ module mm_ram
         end else if(select_rdata_q == RND_STALL) begin
             data_rdata_mux = rnd_stall_rdata;
 `ifndef VERILATOR
+`ifdef __UVMT_CV32E40P_TB_SV__
```
MikeOpenHWGroup, line 475: A comment and a question about this:

- Comment: I am not a fan of introducing a new compile-time macro and associated "ifdefs"
- Question: Where is the macro `__UVMT_CV32E40P_TB_SV__` defined?

Perhaps we could globally replace all `ifndef VERILATOR with `ifndef __UVMT_CV32E40P_TB_SV__`?

MikeOpenHWGroup (changes requested):
Thanks for this @pascalgouedo!   Before merging this in, I'd like to resolve the question/comments I made about the `__UVMT_CV32E40P_SV_TB__` macro you introduced.

cv32e40p/tb/core/mm_ram.sv
pascalgouedo: It is defined at the beginning of the UVMT tb.
I used it to be able to compile core tb when using this mm_ram.sv file.
I added it around iss_wrap functions.

Globally replacing it would not work as ifndef VERILATOR is sometimes used alone.

MikeOpenHWGroup: Of course!  In the compile-guards.  Stupid of me to forget this.  Clever.

I am not convinced that the code in the core testbench cannot be arranged such that VERILATOR cannot be replaced with UVMT_CV32E40P_TB_SV.  However, you are correct that it will not be as simple as a global search and replace, so I will approve/merge this PR and open an issue to get that done later.

MikeOpenHWGroup (approved):

PR #725, opened 2021-07-13
remove more long tests from DSIM

strichmo:


MikeOpenHWGroup (approved):
Hmmm.   That is a lot of "long" tests to remove.

PR #729, opened 2021-07-15
fix issue of incrementing order, and location of [mem]

eroom1966:
@strichmo @MikeOpenHWGroup 
This should also allow me to further investigate the PMA test
Passes all local ci_check tests


strichmo (approved):
Passes ci_check.  Thanks for this @eroom1966

PR #733, opened 2021-07-16
Cv32e40x/merge dev to rel 36

strichmo:


strichmo:
I found an issue with non-Xcelium simulators with OVPSIM memory modeling and UVM hdl deposits.  This will need to be addressed in cv32e40x/dev before a merge can propagate through the CIs properly.

PR #734, opened 2021-07-20
create an RVVI_memory interface to hold memory variable in lieu of st…

strichmo:
…ring-based HDL path

some simulators could not use an associative array with uvm_hdl_deposit


MikeOpenHWGroup (approved):
Looks good to me @strichmo, but this needs to be reviewed and approved by @eroom1966 before merging (regardless of what GitHub says).

strichmo:
@MikeOpenHWGroup @eroom1966 The PR is updated to move memory() interface instantiation into the RAM model.  

rel_check passes on all 3 E cores with Xcelium with this change.

PR #741, opened 2021-07-21
Cv32e40p/merge master to rel 9

strichmo:
Merge back to cv32e40p/release from master


MikeOpenHWGroup (commented):
Is there anything to do here?   This PR has 604 file changes, yet it is a merge from master to cv32e40p/release which is already in sync with the master.    A merge of `strichmo/cv32e40p/merge_master_to_rel_9` to `master` yields an "Already up to date" message from git.

I wonder if this isn't a vacuous merge?

strichmo:
This should be a real merge.  This populates all of the new methodology updates into the cv32e40p.  This includes just about everything such as ISACOV, latest RVFI/RVVI agents, etc.  The only one of these that directly would affect cv32e40p is the Imperas single-source update (this should pass based on tests, but should be verified with a rel_check CI run on this merge, I have already run this locally before PRing).

MikeOpenHWGroup:
Ah, I get it now. My second cup of coffee this morning kicked in.  Stand-by.

PR #745, opened 2021-07-21
add initial support for CV32E40S

eroom1966:
This PR adds in support for CV32E40S

Note that the RM enables user mode and MISA[U]=1
however in the RTL, this is not yet supported, so will cause differences
Reference
https://github.com/openhwgroup/cv32e40s/issues/29


strichmo:
Merging this to get the model into the testbench.  Will follow up with my own PR to update RTL to get MISA updated and some other test updates.

PR #746, opened 2021-07-22
PMA test fix

strichmo:
Adding initialization to PMA aligned I/O load test to ensure ISS matches
Test still fails because RTL does not have data PMA implemented


silabs-robin:
Hi. I had (locally) done some tweaks to test out the problems that erooM was reporting, https://github.com/openhwgroup/core-v-verif/pull/747 (made PR for it now, for easy comparison).
It seems to target some of the same things as this PR. Not sure which solution is preferable.

PR #751, opened 2021-07-23
Strichmo/pr/e40x rel to master 210723

strichmo:
Merge from cv32e40x/release to get the updated model from Imperas
basic MISA cleanup for the S for user mode
Known failures in ci_check - RTL bugs need to be fixed, will file issues after merge


MikeOpenHWGroup (approved):
Thanks @strichmo, looks good.  It took me a second to figure out what "CORESB" in `uvme_cv32e40s_core_sb.sv` was.  My brain translated it into "CORES B", instead of "CORE SB".  :-)

PR #755, opened 2021-07-27
Changed cross module references

silabs-oysteink:
Interrupt assert module was using ID stage signals to check for WFI instructions.
The new controller is using WB stage signals, and thus changes the assert module to use the same.


MikeOpenHWGroup:
Hi @silabs-oysteink.  I would expect that a change to the XMRs in the testbench would also require a change in the RTL to match.  This is handled by updating the `CV_CORE_HASH` variable in the [Common Makefile](https://github.com/openhwgroup/core-v-verif/blob/master/cv32e40x/sim/Common.mk) at \<core\>/sim/Common.mk.

silabs-oysteink:
The RTL used already has this change (it was brought in with the new controller). This is an update to the testbench to account for the RTL changes.

MikeOpenHWGroup:
> The RTL used already has this change

Ah, I had to look a little deeper to understand.  In the RTL, both the `id_stage` and `wb_stage` have _instr_valid_ and _rdata_ nets, which is why the testbench compiles before and after this PR is merged in.

MikeOpenHWGroup (approved):

strichmo:
Thanks for approval Mike.  I was a little late to the review.  Oystein's change holds to the original intent of the property-testing that there is an upper bound to the latency between the core committing to a wfi and the actual core_sleep_o.  The new controller commits to an instruction in writeback instead of instruction decode.  

Thanks for the review!

PR #767, opened 2021-08-04
Reinstate debug assertions

silabs-robin:
This PR re-enables 40x' debug assertions, re-wiring some connections and re-writing some of the assertions.

It passes in both sim and formal (NB! Needs assumes to limit obi wait states).
Though it is not fully 100% complete, most of it has been restored.
This should be thoroughly reviewed, and I will try to assist in that. Don't hold back any criticism or questions.
I already suspect that some changes are necessary, but I don't know what the right call is.

@silabs-oysteink


cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -381,16 +380,28 @@ bind cv32e40x_wrapper
...
+      .illegal_insn_i(dut_wrap.cv32e40x_wrapper_i.core_i.ex_wb_pipe.illegal_insn),
+      .wb_illegal(dut_wrap.cv32e40x_wrapper_i.core_i.ex_wb_pipe.illegal_insn),
+      .wb_valid(dut_wrap.cv32e40x_wrapper_i.core_i.wb_stage_i.instr_valid),  // TODO:ropeders ex_wb_pipe.instr_valid),
+      .wb_halt(dut_wrap.cv32e40x_wrapper_i.core_i.wb_stage_i.ctrl_fsm_i.halt_wb),
+      //TODO:ropeders .illegal_insn_q(dut_wrap.cv32e40x_wrapper_i.core_i.controller_i.controller_fsm_i.illegal_insn_q),
+      .ecall_insn_i(dut_wrap.cv32e40x_wrapper_i.core_i.ex_wb_pipe.ecall_insn),
+      .debug_req_i(dut_wrap.cv32e40x_wrapper_i.core_i.controller_i.controller_fsm_i.pending_debug),  // TODO:ropeders "debug_req_i"?
       .debug_mode_q(dut_wrap.cv32e40x_wrapper_i.core_i.controller_i.controller_fsm_i.debug_mode_q),
       .dcsr_q(dut_wrap.cv32e40x_wrapper_i.core_i.cs_registers_i.dcsr_q),
       .depc_q(dut_wrap.cv32e40x_wrapper_i.core_i.cs_registers_i.dpc_q),
```
silabs-robin, line 407: About signal names in the interface (like "depc"), I think fixing that could wait?
The debug coverage must also be brought back, so I was thinking about doing a proper cleanup of all the signals then, to also purge any no-longer-needed signals.
Does that make sense or should it be done differently?

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -55,46 +56,63 @@ module uvmt_cv32e40x_debug_assert
...
     // Assertions
     // ---------------------------------------
 
     // check that we enter debug mode when expected. 
     // CSR checks are done in other assertions
     property p_enter_debug;
-        $changed(debug_cause_pri) && (debug_cause_pri != 3'b000) && !cov_assert_if.debug_mode_q
-        |-> decode_valid [->1:2] ##0 cov_assert_if.debug_mode_q;
+        $changed(debug_cause_pri) && (debug_cause_pri != 0) && !cov_assert_if.debug_mode_q
+        |-> decode_valid [->1:20] ##0 cov_assert_if.debug_mode_q;
```
silabs-robin, line 86: This was changed from `[->1:2]` to `[->1:20]` because obi could just stall indefinitely and deny debug entry.
This goto condition was confusing throughout the assertion set, so I would like to discuss what the best solution is.

From before, I think `1:2` represented the "currently executing ID" and "next arising execution in ID"? But now the decisions are being taken from WB so the conditions must necessarily change.
One proposal (that is implemented in some of the other assertions), is to use this:
`!decode_valid [->1] ##0 decode_valid [->1]`
to mean "wait til the current execution is done, then wait til we have continued again".

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
     a_cebreak_debug_mode: assert property(p_cebreak_debug_mode)
-        else
-            `uvm_error(info_tag,$sformatf("Debug mode with wrong cause after ebreak, case = %d",cov_assert_if.dcsr_q[8:6]));
+        else `uvm_error(info_tag,$sformatf("Debug mode with wrong cause after ebreak, case = %d",cov_assert_if.dcsr_q[8:6]));
+
 
     // c.ebreak without dcsr.ebreakm results in exception at mtvec
     // Exclude single stepping as the sequence gets very complicated
+
     property p_cebreak_exception;
```
silabs-robin, line 139: This property is particularly difficult to capture.
The intent is to check _"c.ebreak without dcsr.ebreakm results in exception at mtvec"_, but that statement is not universally true without qualifications? Formal showed that an intercepting haltreq or irq could break that assertion.

Hence I added "throughout" conditions for not seeing haltreq or irq until we have arrived at the next execution.

One challenge here is to not over-constrain the antecedent, so that we "semi-vacuously" passes certain cases. But we also don't want a "liberal" consequent with some "||" that could falsely excuse real problems.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
-                                                                   (cov_assert_if.id_stage_pc == cov_assert_if.mtvec);
+        disable iff(!cov_assert_if.rst_ni)
+        $rose(cov_assert_if.is_cebreak) && !cov_assert_if.debug_mode_q && !cov_assert_if.debug_req_i
+        && !cov_assert_if.dcsr_q[2] && !cov_assert_if.dcsr_q[15]
+        ##0 (!cov_assert_if.debug_req_i && !cov_assert_if.irq_ack_o throughout !decode_valid[->1])
+        ##0 (!cov_assert_if.debug_req_i && !cov_assert_if.irq_ack_o throughout decode_valid[->1])
+        |->
+        !cov_assert_if.debug_mode_q && (cov_assert_if.mcause_q[5:0] === cv32e40x_pkg::EXC_CAUSE_BREAKPOINT)
+        && (cov_assert_if.mepc_q == pc_at_ebreak) && (cov_assert_if.id_stage_pc == cov_assert_if.mtvec);
+        // TODO:ropeders need assertions for what happens if cebreak and req/irq?
```
silabs-robin, line 148: >         // TODO:ropeders need assertions for what happens if cebreak and req/irq?

Since haltreq and irq is excluded from the case above, I wonder if we should add new assertions that explicitly check such cases.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+
     a_cebreak_exception: assert property(p_cebreak_exception)
         else
             `uvm_error(info_tag,$sformatf("Exception not entered correctly after c.ebreak with dcsr.ebreak=0"));
 
+
     // ebreak without dcsr.ebreakm results in exception at mtvec
     // Exclude single stepping as the sequence gets very complicated
+
     property p_ebreak_exception;
```
silabs-robin, line 159: Same problem**s** as with cebreak exception above.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+        else `uvm_error(info_tag,$sformatf("Debug mode not restarted after c.ebreak"));
+
 
     // ebreak during debug mode results in relaunch
+
     property p_ebreak_during_debug_mode;
-        $rose(cov_assert_if.is_ebreak) ##0 cov_assert_if.debug_mode_q   |-> decode_valid [->2] ##0 cov_assert_if.debug_mode_q && 
-                                                     (cov_assert_if.id_stage_pc == halt_addr_at_entry); // TODO should check no change in dpc and dcsr
+        $rose(cov_assert_if.is_ebreak) && cov_assert_if.debug_mode_q
+        |-> !decode_valid [->1] ##0 decode_valid [->1]
```
silabs-robin, line 191: Note:
Using `decode_valid` in the "consequent" is much safer (imo) than having it in the antecedent.
The antecedent should have as few constraints as possible, so we don't accidentally over-constrain and ignore cases that we actually wanted to consider.
The consequent could (imo) have as many "&&" factors as we want, because we will(?) be corrected if we expect too much (but "||" terms seem really dangerous to me).

silabs-robin: But... Gotos in consequent might skip past some critical point, so the "&&"ing should be pretty rigorous and demanding.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -225,56 +267,74 @@ module uvmt_cv32e40x_debug_assert
         else
             `uvm_error(info_tag, $sformatf("WFI in debug mode cause core_sleep_o=1"));
 
-    // Debug request while sleeping makes core wake up and enter debug mode
-    // wit cause=haltreq
+
+    // Debug request while sleeping makes core wake up and enter debug mode with cause=haltreq
+
     property p_sleep_debug_req;
```
silabs-robin, line 273: Originally, there was 2 implications in this property. Was there some particular reason for that that I failed to heed?

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -381,16 +380,28 @@ bind cv32e40x_wrapper
       .id_stage_instr_valid_i(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.instr_valid),
       .id_stage_instr_rdata_i(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.instr.bus_resp.rdata),
       .id_stage_is_compressed(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.is_compressed),
-      .id_valid(dut_wrap.cv32e40x_wrapper_i.core_i.id_valid),
-      .is_decoding(dut_wrap.cv32e40x_wrapper_i.core_i.is_decoding),
-      .id_stage_pc(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.pc),
+      .id_stage_wfi_insn     (dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.id_ex_pipe_o.wfi_insn),  // TODO:ropeders "ex_stage"?
+      .id_valid(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.id_valid),
+      .is_decoding(1),  //dut_wrap.cv32e40x_wrapper_i.core_i.is_decoding),  // TODO:ropeders figure out correct source
```
silabs-oysteink, line 385: The signal is_decoding does not exist in RTL anymore, and cases using that signal within these assertions should probably change to look at the WB stage instead.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -55,46 +56,63 @@ module uvmt_cv32e40x_debug_assert
...
+    cov_assert_if.wb_stage_instr_valid_i
+    && (cov_assert_if.wb_stage_instr_rdata_i == 32'h0000_9002);
 
-  assign cov_assert_if.is_mulhsu = cov_assert_if.is_decoding &
-                                   cov_assert_if.id_stage_instr_valid_i & 
-                                   cov_assert_if.id_stage_instr_rdata_i[31:25] == 7'h1 &
-                                   cov_assert_if.id_stage_instr_rdata_i[14:12] == 3'b010 &
-                                   cov_assert_if.id_stage_instr_rdata_i[6:0]   == 7'h33;
+  assign cov_assert_if.is_mulhsu =
+    cov_assert_if.is_decoding
```
silabs-oysteink, line 68: Could remove this signal, is_decoding does not exist in RTL anymore.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
```diff
@@ -130,8 +130,20 @@ interface uvmt_cv32e40x_debug_cov_assert_if
...
+    input  [31:0] ex_stage_instr_rdata_i,
+    input  [31:0] ex_stage_pc,
+    input         wb_stage_instr_valid_i,
+    input  [31:0] wb_stage_instr_rdata_i,
+    input  [31:0] wb_stage_pc, // Program counter in writeback
+    input         wb_illegal,
+    input         wb_valid,
+    input         wb_halt,
+    input         wb_stage_wfi_insn,
     input         is_decoding,
```
silabs-oysteink, line 147: Remove

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -55,46 +56,63 @@ module uvmt_cv32e40x_debug_assert
...
+    && cov_assert_if.id_stage_instr_valid_i
+    && (cov_assert_if.id_stage_instr_rdata_i[31:25] == 7'h1)
+    && (cov_assert_if.id_stage_instr_rdata_i[14:12] == 3'b010)
+    && (cov_assert_if.id_stage_instr_rdata_i[6:0]   == 7'h33);
+
+
+  assign decode_valid = cov_assert_if.id_stage_instr_valid_i && (cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::FUNCTIONAL);
```
silabs-oysteink, line 75: Old controller made decisions based on state of the ID stage. New controller looks at WB (except for jumps in ID and branhces in EX), so I think that the assertions should probably only look at WB stage signals if possible

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-oysteink: I would prefer if WB signals were used. The intention of the original assertion was that if a change in predicted debug case was seen, the next decoded instruction should be in debug mode.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -55,46 +56,63 @@ module uvmt_cv32e40x_debug_assert
...
-        else
-            `uvm_error(info_tag, $sformatf("Debug mode not entered after exepected cause %d", debug_cause_pri));
+        else `uvm_error(info_tag, $sformatf("Debug mode not entered after exepected cause %d", debug_cause_pri));
+
+
+    // Check that depc gets the correct value when debug mode is entered.
 
-    // Checck that depc gets the correct value when debug mode is entered.
     property p_debug_mode_pc;
-        $rose(first_debug_ins) |-> cov_assert_if.debug_mode_q && (prev_id_pc == halt_addr_at_entry) && (cov_assert_if.depc_q == pc_at_dbg_req);
```
silabs-oysteink, line 92: Should prev_id_pc really be used here?

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -55,46 +56,63 @@ module uvmt_cv32e40x_debug_assert
...
+        $rose(first_debug_ins)
+        |-> cov_assert_if.debug_mode_q && (prev_id_pc == halt_addr_at_entry);
```
silabs-oysteink, line 97: Should prev_id_pc really be used here?

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-oysteink: Yes, this is a hard case as there are so many other events that take priority over an ebreak without dcsr.ebreakm set.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-oysteink: I agree, we should probably have separate asserts for that.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+        else `uvm_error(info_tag,$sformatf("Exception not entered correctly after ebreak with dcsr.ebreak=0"));
+
+
     // c.ebreak during debug mode results in relaunch of debug mode
 
     property p_cebreak_during_debug_mode;
-        $rose(cov_assert_if.is_cebreak) ##0 cov_assert_if.debug_mode_q  |-> decode_valid [->2] ##0 cov_assert_if.debug_mode_q  &&
-                                                       (cov_assert_if.id_stage_pc == halt_addr_at_entry); // TODO should check no change in dpc and dcsr
+        $rose(cov_assert_if.is_cebreak) && cov_assert_if.debug_mode_q
+        |-> !decode_valid [->1] ##0 decode_valid [->1]
```
silabs-oysteink, line 178: I would prefer WB signals here too

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+        ##0 cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == halt_addr_at_entry);
```
silabs-oysteink, line 179: .. and here, why id_stage_pc?

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+
 
     // ebreak during debug mode results in relaunch
+
     property p_ebreak_during_debug_mode;
-        $rose(cov_assert_if.is_ebreak) ##0 cov_assert_if.debug_mode_q   |-> decode_valid [->2] ##0 cov_assert_if.debug_mode_q && 
-                                                     (cov_assert_if.id_stage_pc == halt_addr_at_entry); // TODO should check no change in dpc and dcsr
+        $rose(cov_assert_if.is_ebreak) && cov_assert_if.debug_mode_q
+        |-> !decode_valid [->1] ##0 decode_valid [->1]
+        ##0 cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == halt_addr_at_entry);
```
silabs-oysteink, line 192: id_stage_pc again

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+        else `uvm_error(info_tag, "Trigger match with tdata[2]==0 resulted in debug mode");
+
 
     // Exception in debug mode results in pc->dm_exception_addr_i
+
     property p_debug_mode_exception;
-        $rose(cov_assert_if.illegal_insn_i) && cov_assert_if.debug_mode_q && cov_assert_if.is_decoding |-> (decode_valid & cov_assert_if.id_valid) [->2] ##0 cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == exception_addr_at_entry);
+        $rose(cov_assert_if.illegal_insn_i) && cov_assert_if.debug_mode_q
+        |=> (decode_valid && cov_assert_if.id_valid) [->1]
+        ##0 cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == exception_addr_at_entry);
```
silabs-oysteink, line 231: Much ID stage above

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -113,94 +130,119 @@ module uvmt_cv32e40x_debug_assert
...
+            $sformatf("Exception in debug mode not handled incorrectly. dm=%d, pc=%08x",
+                cov_assert_if.debug_mode_q, cov_assert_if.id_stage_pc));
+
 
     // ECALL in debug mode results in pc->dm_exception_addr_i
     property p_debug_mode_ecall;
-        $rose(cov_assert_if.ecall_insn_i) && cov_assert_if.debug_mode_q  && cov_assert_if.is_decoding && cov_assert_if.id_stage_instr_valid_i
-        |-> (decode_valid & cov_assert_if.id_valid) [->1:3] ##0 cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == exception_addr_at_entry);
+        $rose(cov_assert_if.ecall_insn_i) && cov_assert_if.debug_mode_q
+        && cov_assert_if.is_decoding && cov_assert_if.id_stage_instr_valid_i
```
silabs-oysteink, line 243: is_decoding not in RTL anymore

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -289,21 +349,44 @@ module uvmt_cv32e40x_debug_assert
...
+    sequence s_dmode_dret_pc_ante;  // Antecedent
+        cov_assert_if.debug_mode_q && cov_assert_if.is_dret
+        ##1 (!cov_assert_if.debug_req_i && !cov_assert_if.irq_ack_o throughout decode_valid[->1]);
+    endsequence
+
+    sequence s_dmode_dret_pc_conse;  // Consequent
+        decode_valid && !cov_assert_if.debug_mode_q && (cov_assert_if.id_stage_pc == cov_assert_if.depc_q);
+    endsequence
+
+    cov_dmode_dret_pc : cover property(s_dmode_dret_pc_ante |-> s_dmode_dret_pc_conse ##0 (cov_assert_if.depc_q != 0));
```
silabs-robin, line 371: The reason I split `property` into two `sequence`s is because I wanted to check this cover. It was not obvious whether the assertion was just trivially passing or not.
If this style is difficult to read I am ok with changing it.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -289,21 +349,44 @@ module uvmt_cv32e40x_debug_assert
...
-        else
-            `uvm_error(info_tag, "Executing dret in M-mode did not result in illegal instruction");
+        else `uvm_error(info_tag, "Executing dret in M-mode did not result in illegal instruction");
+
+
+    // dret in D-mode will restore pc (if no re-entry or interrupt intervenes)
+
+    sequence s_dmode_dret_pc_ante;  // Antecedent
+        cov_assert_if.debug_mode_q && cov_assert_if.is_dret
+        ##1 (!cov_assert_if.debug_req_i && !cov_assert_if.irq_ack_o throughout decode_valid[->1]);
```
silabs-oysteink, line 364: Interrupts are ignored during debug mode, so irq_ack_o should always be 0. No harm done in leaving it in though

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -289,21 +349,44 @@ module uvmt_cv32e40x_debug_assert
...
+        // TODO:ropeders also assert, stays in mmode until decode_valid if no debug_request
     endproperty
 
-    a_dmode_dret : assert property(p_dmode_dret)
-        else
-            `uvm_error(info_tag, "Dret did not cause correct return from debug mode");
+    a_dmode_dret_exit : assert property(p_dmode_dret_exit)
+        else `uvm_error(info_tag, "Dret did not exit debug mode");
+
+    // TODO:ropeders what is missing from these dret assertions?
```
silabs-robin, line 388: >     // TODO:ropeders what is missing from these dret assertions?

Formal showed that the original assertion could be broken with intervening re-entry or interrupts. Therefore I tried to split the checking into multiple assertions, but I have a feeling that maybe not all cases are checked and there should be added some assertions.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -447,40 +568,29 @@ module uvmt_cv32e40x_debug_assert
...
-                    else if(cov_assert_if.dcsr_q[2])
-                        debug_cause_pri <= 3'b100;
-                    else
-                        debug_cause_pri <= 3'b000;
-
-                end
-
-            end else if(cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::DBG_TAKEN_IF) begin
-                if(cov_assert_if.debug_req_i) begin
+            if((cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::FUNCTIONAL) && !cov_assert_if.debug_mode_q) begin
```
silabs-oysteink, line 583: Her you should be able to check if ctrl_fsm_ns == DEBUG_TAKEN and not in debug mode

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -381,16 +380,28 @@ bind cv32e40x_wrapper
       .id_stage_instr_valid_i(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.instr_valid),
       .id_stage_instr_rdata_i(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.instr.bus_resp.rdata),
       .id_stage_is_compressed(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.is_compressed),
-      .id_valid(dut_wrap.cv32e40x_wrapper_i.core_i.id_valid),
-      .is_decoding(dut_wrap.cv32e40x_wrapper_i.core_i.is_decoding),
-      .id_stage_pc(dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.if_id_pipe_i.pc),
+      .id_stage_wfi_insn     (dut_wrap.cv32e40x_wrapper_i.core_i.id_stage_i.id_ex_pipe_o.wfi_insn),  // TODO:ropeders "ex_stage"?
```
silabs-oysteink, line 383: The id_ex_pipe_o.wfi_insn indicates WFI in EX as your comment suggests

silabs-oysteink (changes requested):
In general I would like to remove most depencies on ID stage signals where possible.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
silabs-robin: Fixed.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-robin: Fixed.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
silabs-robin: Fixed.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
strichmo: That makes sense to me.

strichmo (changes requested):
When I pulled this PR in to test I had to move the *started_decoding_in_debug* logic delcaration slightly higher in the file to avoid a compile error.

```
     // Detect first instruction of debug code
+    logic started_decoding_in_debug;
 
     assign first_debug_ins =
         cov_assert_if.debug_mode_q && cov_assert_if.wb_valid
         && !first_debug_ins_flag && started_decoding_in_debug;
 
-    logic started_decoding_in_debug;
```

silabs-robin:
@strichmo 

> When I pulled this PR in to test I had to move the _started_decoding_in_debug_ logic delcaration slightly higher in the file to avoid a compile error.

Aha. That makes sense.
I think it was unfortunate timing that your checkout happened just after I made some commits trying to fix Øystein's feedback and I have only tested with Jasper.
I must make sure (once again, after adding new commits) that simulation runs ok before merge.

> Otherwise this PR looks OK to me.

That is good. I must still iron out some kinks, addressing all of Øystein's feedback.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
silabs-robin: Will likely leave this as is for now, and fix when doing the coverage.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-robin: Let me see if I understood that correctly.

First there was:
```
if(cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::DECODE & !cov_assert_if.debug_mode_q) begin
    ...
end else if(cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::DBG_TAKEN_IF) begin
    ...
```

Now it is:
```
if((cov_assert_if.ctrl_fsm_cs == cv32e40x_pkg::FUNCTIONAL) && !cov_assert_if.debug_mode_q) begin
    ...
```

You propose(?):
```
if((cov_assert_if.ctrl_fsm_ns == DEBUG_TAKEN) && !cov_assert_if.debug_mode_q) begin
```

Would that be better?
Using FUNCTIONAL (as it is now) has the risk of accidentally updating the prediction in between DEBUG_TAKEN and entry, but we have not seen that it is possible?
Using DEBUG_TAKEN (as proposed) updates less often, so it risks missing cases?

silabs-robin:
This part needs careful review.

```
    sequence s_conse_next_retire;  // Should only be used in consequent (not antecedent)
        ($fell(cov_assert_if.wb_stage_instr_valid_i) [->1]  // Finish current preoccupation
            ##0 cov_assert_if.wb_valid [->1])  // Go to next WB done
        or
        (cov_assert_if.wb_valid [->1]  // Go directly to next WB done
            ##0 (cov_assert_if.dcsr_q[8:6] == 3))  // Need good reason to forgo $fell(instr_valid)
        ;
    endsequence
```

Explanation:
In the consequents (of `antecedent |-> consequent`), there was previously used `decode_valid[->1:2]` etc.
It is not straight forward to substitute it with `wb_valid[->1]` because of halt/kill signals etc.
I made this sequence, which is intended as a way to go to the cycle where our consequent should have its effect.
My worry is that it potentially can excuse bad behavior.
Example: I previously experimented with `debug_req[->1] ##0 wb_valid[->1]` but this seems bad, because an antecedent could fail our expectations but formal can potentially create a debug_req to make up for it.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-robin: @silabs-oysteink do you think that such new assertions should be made as part of this PR?

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -225,117 +281,182 @@ module uvmt_cv32e40x_debug_assert
...
 
     // Check that trigger regs cannot be written from M-mode
     // TSEL, and TDATA3 are tied to zero, hence no register to check 
+
     property p_mmode_tdata1_write;
-        !cov_assert_if.debug_mode_q && cov_assert_if.csr_access && cov_assert_if.csr_op == 'h1 && cov_assert_if.id_stage_instr_rdata_i[31:20] == 'h7A1 |-> ##2 $stable(cov_assert_if.tdata1);
+        !cov_assert_if.debug_mode_q && cov_assert_if.csr_access && cov_assert_if.csr_op == 'h1
+        && cov_assert_if.wb_stage_instr_rdata_i[31:20] == 'h7A1
+        |->
+        ##0 $stable(cov_assert_if.tdata1) [*4];
```
silabs-robin, line 426: @silabs-oysteink could you provide some more insight into this? I couldn't see this assertion in the vplan and I don't know the uarch well enough.
Do you think there is any reason to continue using id_stage?
Do you think the stability criterion should be different?

silabs-robin:
Hi @silabs-oysteink and @strichmo . I think the PR is ready to be merged now.
I updated it to work with the latest cv32e40x/dev, and both formal and simulation passes.
The only things outstanding are those comments above and that which I have started fixing as part of the debug coverage task.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
silabs-oysteink: The purpose of this assertion is to check that one cannot write to tdata1 outside of debug mode. As the 40X now does CSR writes in WB, WB signals should be used. I'm not sure about the stability criteria though, it should be enough to check stability before/after the actual (attempted) write.

silabs-oysteink (approved):

silabs-robin:
Sorry about the noise, but I remembered one potential problem!
The OBI "err" signal was reintroduced in core-v-verif, and since there is special handling of wb_valid in rtl/rvfi I disabled "err" in formal. This means that the assertions _can fail(!)_ if an obi "err" signal comes at the right time.
If we know that random tests used in regressions do not introduce "err" then I think there will be no problems from merging this, but if they do then a solution must be found first.

strichmo (approved):

PR #770, opened 2021-08-04
change options to objdump for better debug

strichmo:
use isa registers instead of ABI (x0-x31)
delineate compressed instructions in pnemonics with c. prefix


MikeOpenHWGroup (approved):
...I take your word for it... (and I hate you for forcing to me read the man page for objdump.)

PR #774, opened 2021-08-06
Strichmo/pr/uvma obi e40s

strichmo:
Port the E40S testbench to use the OBI UVM memory agent.
Remove the code for the deprecated agent
Remove the deprecated agent from the E40X and the E40S testbenches.


strichmo:
Metrics runs look OK.  It is understood that the E40S is not completely stabled with respect to ci_check, but expected passing tests did pass.

MikeOpenHWGroup (approved):

PR #808, opened 2021-08-12
remove absolute addresses within virtual peripheral sequences

strichmo:
Virtual peripherals should be installable at *any* address.  This PR provides a method and VP base variables to ensure that implementations can retrieve a register offset index independent of the the implementation's absolute address.


MikeOpenHWGroup (commented):
This is something I've thought about as well, so I am happy to see this.   However, maybe I need a second cup of coffee, but I cannot see where the decoding for the individual virtual peripherals is done in this PR.

strichmo:
The register_vp_seq method is where this magic happens.  Please bear with me as I have one more change I'd like to make (in an upcoming PR).

The memory slave sequence as below maintains a hash of addresses that point to vp_seq handles.  Thus we can do an O(1) lookup for each address to see if it maps to a vp_seq on the fly.

The register_vp_seq passes in a base address and (for now) number of words to claim.  But I'm going to change this such that the sequence itself will contain information as to how many words it uses via a method.  The environment should not be configuring this...it is a "property" of the vp seq itself.

https://github.com/openhwgroup/core-v-verif/blob/5174c2b575f3baf16360afe5d2bb6afe25e5257d/lib/uvm_agents/uvma_obi_memory/src/seq/uvma_obi_memory_slv_seq.sv#L174

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
> The register_vp_seq method is where this magic happens.

Ah, maintenant, je comprends.  Using the hash lookup is slick.

PR #812, opened 2021-08-14
change register_vp_seq to define number of register words from the cl…

strichmo:
…ass definition via a pure virtual method

add instruction bus fault support with directed test
add directed instruction bus error test


strichmo:
All 3 cores pass ci_check on Xcelium and Metrics.  Merging...

PR #814, opened 2021-08-16
Cleanup of trailing whitespace

silabs-hfegran:
Removed trailing whitespace from source files for preexisting files in the 40x environment.


silabs-hfegran:
I've left out top level (lib/40p/40s) for now - a cleanup would be good there as well, but should perhaps be synced throughout the 40p/40s/40x/other branches

strichmo:
Was this scripted in any way?  IS there something we could morph into a utility perhaps?

strichmo (approved):

silabs-hfegran:
> Was this scripted in any way? IS there something we could morph into a utility perhaps?

Yes, I did this with a shell command - easy to integrate into scripts if desired,
for .sv-files as follows:
`find . -type f -name '*.sv' -exec sed --in-place 's/[[:space:]]\+$//' {} \+`

PR #821, opened 2021-08-19
Pma directed, fix trap handler

silabs-robin:
This PR updates the pma directed tests so that the trap handler returns consistently correct.

Previously:
Several parts of the test has been commented away, waiting for the RTL to get the new features implemented. Now we want to start re-enabling these parts but an issue was found in the test's trap handler.
When inducing pma exceptions, the trap handler didn't return correctly for all cases.
(Hadn't been tested because there was nothing to test it on.)
(The typical PC+4/2 can't be used when jumping to arbitrary locations.)

Solution:
Use function calls ("provoke()") as a way to determine where the test continues execution.
Additionally, explicitly manage the stack pointer and save/restore GPRs.

The test passes when using a 40x from one of Øystein's branches and without iss, but it should not work with 40x current master.
@silabs-oysteink can you confirm whether this test works as expected when running with your latest rtl fixes? Both with and without iss?

### DIFF
--- a/cv32e40x/tests/programs/custom/pma/pma.c
+++ b/cv32e40x/tests/programs/custom/pma/pma.c
@@ -31,8 +31,31 @@
 static volatile uint32_t mcause = 0;
 static volatile uint32_t mepc = 0;
 static volatile uint32_t mtval = 0;
+static volatile uint32_t retpc = 0;
 
-static void (*cause_instr_access_fault)(void) = (void (*)(void))IO_ADDR;
+// Exception-causing instructions
+static void (*instr_access_fault)(void) = (void (*)(void))IO_ADDR;
+void misaligned_store(void) {
+  uint32_t tmp;
+  tmp = 0xBBBBBBBB;
+  __asm__ volatile("sw %0, 1(%1)" : "=r"(tmp) : "r"(IO_ADDR));
+}
+void load_misaligned_io(void) {__asm__ volatile("lw t0, 3(%0)" : : "r"(IO_ADDR));}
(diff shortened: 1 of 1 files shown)


cv32e40x/tests/programs/custom/pma/pma.c
```diff
@@ -31,8 +31,50 @@
...
+static void (*instr_access_fault)(void) = (void (*)(void))IO_ADDR;
+void misaligned_store(void) {
+  uint32_t tmp;
+  tmp = 0xBBBBBBBB;
+  __asm__ volatile("sw %0, 1(%1)" : "=r"(tmp) : "r"(IO_ADDR));
+}
+void load_misaligned_io(void) {__asm__ volatile("lw t0, 3(%0)" : : "r"(IO_ADDR));}
+void load_misaligned_iomem(void) {__asm__ volatile("lw t0, 0(%0)" : : "r"(MEM_ADDR_1 - 3));}
+void load_misaligned_memio(void) {__asm__ volatile("lw t0, 0(%0)" : : "r"(IO_ADDR - 1));}
+void store_first_access(void) {__asm__ volatile("sw %0, 2(%1)" : : "r"(0x11223344), "r"(IO_ADDR));}
```
silabs-robin, line 46: Admittedly, it is not very clean that these parts are defined so far away from the context where they are used. These magic numbers make sense within that context.
If anyone can conjure a better way to solve this problem then I would gladly follow that route instead.

strichmo (approved):
Looks like a reasonable way to arrange.  Code is pretty easy to follow.  It will be nice to get rid of the commenting on the asserts, but we'll get to that as the RTL comes up.

PR #817, opened 2021-08-18
Added temporary hack (and todos) to support updated rvfi interface.

silabs-halfdan:
Signed-off-by: Halfdan Bechmann <halfdan.bechmann@silabs.com>

PR #822, opened 2021-08-20
incorporate new RVFI debug signals and latest RTL

strichmo:
Manually integrate https://github.com/openhwgroup/core-v-verif/pull/817 with the latest cv32e40x RTL.

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
@@ -68,7 +68,7 @@
             atomic         : 0}
          };
    `else
-      parameter int unsigned               CORE_PARAM_PMA_NUM_REGIONS = 0;
+      parameter int                        CORE_PARAM_PMA_NUM_REGIONS = 0;
       parameter cv32e40x_pkg::pma_region_t CORE_PARAM_PMA_CFG[-1:0] =  '{default:cv32e40x_pkg::PMA_R_DEFAULT};
    `endif
 
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
@@ -69,7 +69,7 @@
          };
    `else
       parameter int unsigned               CORE_PARAM_PMA_NUM_REGIONS = 0;
-      parameter cv32e40x_pkg::pma_region_t CORE_PARAM_PMA_CFG[0:0] = '{'z};
+      parameter cv32e40x_pkg::pma_region_t CORE_PARAM_PMA_CFG[-1:0] =  '{default:cv32e40x_pkg::PMA_R_DEFAULT};
    `endif
 
 
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_dut_wrap.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_dut_wrap.sv
@@ -43,8 +43,8 @@ module uvmt_cv32e40x_dut_wrap
   import cv32e40x_pkg::*;
 
   #(// DUT (riscv_core) parameters.
-    parameter NUM_MHPMCOUNTERS    =  1,
-    parameter int unsigned PMA_NUM_REGIONS =  0,
+    parameter              NUM_MHPMCOUNTERS    =  1,
+    parameter int          PMA_NUM_REGIONS =  0,
     parameter pma_region_t PMA_CFG[(PMA_NUM_REGIONS-1):0] = '{default:PMA_R_DEFAULT},
     // Remaining parameters are used by TB components only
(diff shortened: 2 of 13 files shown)


strichmo:
This PR will pass rel_check fine with xcelium, however the negative-index issue on PMA_CFG needs to be fixed to avoid compile failure with DSIM.  The issue to fix is here:

Once fixed I will re-regress and merge-pointing to the RTL fix.

strichmo:
Issue with RTL was resolved by changing types to int instead of unsigned int.  This fixes the DSIM issue.

cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv
```diff
@@ -68,8 +68,8 @@
             atomic         : 0}
          };
    `else
-      parameter int unsigned               CORE_PARAM_PMA_NUM_REGIONS = 0;
-      parameter cv32e40x_pkg::pma_region_t CORE_PARAM_PMA_CFG[0:0] = '{'z};
+      parameter int                        CORE_PARAM_PMA_NUM_REGIONS = 0;
+      parameter cv32e40x_pkg::pma_region_t CORE_PARAM_PMA_CFG[-1:0] =  '{default:cv32e40x_pkg::PMA_R_DEFAULT};
```
MikeOpenHWGroup, line 72: Why is the range of `CORE_PARAM_PMA_CFG` set to [-1:0] in the default case of the ifdef?  Earlier versions of DSIM have not handled this correctly and I am not sure I see the utility of negative indicies in any case.

cv32e40x/tb/uvmt/uvmt_cv32e40x_dut_wrap.sv
```diff
@@ -43,9 +43,9 @@ module uvmt_cv32e40x_dut_wrap
   import cv32e40x_pkg::*;
 
   #(// DUT (riscv_core) parameters.
-    parameter NUM_MHPMCOUNTERS    =  1,
-    parameter int unsigned PMA_NUM_REGIONS =  0,
-    parameter pma_region_t PMA_CFG[(PMA_NUM_REGIONS ? (PMA_NUM_REGIONS-1) : 0):0] = '{default:PMA_R_DEFAULT},
+    parameter              NUM_MHPMCOUNTERS    =  1,
+    parameter int          PMA_NUM_REGIONS =  0,
+    parameter pma_region_t PMA_CFG[(PMA_NUM_REGIONS-1):0] = '{default:PMA_R_DEFAULT},
```
MikeOpenHWGroup, line 48: With `PMA_NUM_REGIONS` set to 0 (the default), this again creates a parameter with a negative index.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -43,7 +43,7 @@ module uvmt_cv32e40x_tb;
 `endif
 
    parameter int PMA_NUM_REGIONS = 0;
-   parameter cv32e40x_pkg::pma_region_t PMA_CFG[(PMA_NUM_REGIONS ? (PMA_NUM_REGIONS-1) : 0):0] = '{default:PMA_R_DEFAULT};
+   parameter cv32e40x_pkg::pma_region_t PMA_CFG[(PMA_NUM_REGIONS-1):0] = '{default:PMA_R_DEFAULT};
```
MikeOpenHWGroup, line 46: Dito.

MikeOpenHWGroup (commented):
Thanks @strichmo.  This compiles and runs locally for me, but I have a few questions about negative indices for parameters.

cv32e40x/tb/uvmt/uvmt_cv32e40x_dut_wrap.sv
strichmo: The Cores team ran some experiments with VCS which initially had the same issue.  They determined that if the typing of the parameter was an int instead of an _unsigned int_ then then simulator was OK with the index and did not pop any errors.  DSIM also seems to be OK with this after the last release (7/21).  The Cores team believes this is cleaner code and wants to resolve the issue this way.

https://github.com/openhwgroup/cv32e40x/issues/195

MikeOpenHWGroup: > if the typing of the parameter was an int instead of an unsigned int then then simulator was OK with the index and did not pop any errors

Arg.  Seems like we are experimenting with simulator implementations rather than following the LRM.   On the other hand, I see the logic: if `PMA_NUM_REGIONS` is an _int_ (a signed data type), then  `PMA_NUM_REGIONS-1` must also be an _int_ and the "-1" value is interpreted as one would expect.  It seems that we've already expended effort on this, so let's go with it.   It will be interesting to see what other simulators (e.g. Aldec) do with this.

MikeOpenHWGroup (approved):

PR #823, opened 2021-08-24
add initial b extension support

strichmo:
- ISACOV model updated based on reviews for A B and C extensions
- basic support for the B extension added including a directed test
(Note that this will require a toolchain with experimental B support.  The default is no B extension.)
- rework the derivation of MARCH in all firmware and bsp builds for uvmt testbench.  This may require communication as usage has changed (and a review).

This PR passed a DSIM regression with cv32e40x rel check and cv32e40p ci check


strichmo:
Previously the MARCH variables in the Makefile were assigned to the RISCV_PREFIX tuple (e.g. riscv32-unknown-elf).  However that isn't really the march.  The march is supplied with a --march switch to the compiler (gcc).  The MARCH variables were actually being assigned to the "vendor" field of the gcc target triplet: machine-vendor-operatingsystem

Now a user can specify RISCV_MARCH (or COREV_MARCH, or PULP_MARCH, etc.) on the command line to override the actual march that the BSP and firmware will be built.  Thus a B-extension test can specify rv32imc_zba_zbb or an atomic test can specify rv32imac, etc.

cv32e40x/tests/cfg/default.yaml
```diff
@@ -2,8 +2,7 @@ name: default
 description: Default configuration for CV32E40X simulations
 compile_flags:
 ovpsim: >
-    --override root/cpu/misa_Extensions=0x001104
```
MikeOpenHWGroup, line 5: Why are these OVPsim overrides no longer necessary?

MikeOpenHWGroup (commented):
Looks good.  One question about `cv32e40x/tests/cfg/default.yaml`.

cv32e40x/tests/cfg/default.yaml
MikeOpenHWGroup: Ah.  I had forgotten about this.   A far better approach IMHO (as you know I like to keep as much cfg info as possible under direct control of the UVM testcase and this is a step in that direction).

Issue #727, opened 2021-07-14
Heartbeat monitor creates long queue of objects 


strichmo:
### Type
* Memory leak - does not cause functional failure, but unnecessarily bloats memory usage in core-v-verif simulations

### Steps to Reproduce
Run any core-v-verif cv32e40x testcase:
% makeuvmt test TEST=hello-world

You will need a profiler to directly see the queue (or use a debugger).  For example this is the size of the heartbeat monitor's queue with the _interrupt_test_ in Xcelium:

         2972: struct {...} queue worklib.uvml_hrtbt_pkg::uvml_hrtbt_mon_c@2970_1.timestamps [ 13.95 Mbytes ] 325913 elements

strichmo:
With this PR this issue should be fixed:
https://github.com/openhwgroup/core-v-verif/pull/833

PR #833, opened 2021-08-28
Adding fixes to heartbeat monitor

datum-dpoulin:
Added performance improvements and documented the codebase.

### DIFF
--- a/lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
+++ b/lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
@@ -1,19 +1,13 @@
-// 
 // Copyright 2020 OpenHW Group
-// Copyright 2020 Datum Technology Corporation
+// Copyright 2021 Datum Technology Corporation
 // 
-// Licensed under the Solderpad Hardware License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-// 
-//     https://solderpad.org/licenses/
-// 
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-// 
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+// Licensed under the Solderpad Hardware License v 2.1 (the "License"); you may not use this file except in compliance
+// with the License, or, at your option, the Apache License version 2.0.  You may obtain a copy of the License at
+//                                        https://solderpad.org/licenses/SHL-2.1/
+// Unless required by applicable law or agreed to in writing, any work distributed under the License is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations under the License.
 
 
 `ifndef __UVML_HRTBT_MON_SV__
@@ -21,23 +15,22 @@
 
 
 /**
- * Component implementing a per-phase timeout that can be reset. Enables test
- * benches to terminate simulation only once the design has achieved a quiet
- * state, without the use of '#' delays.
+ * Component implementing a per-phase timeout that can be reset.  Enables test benches to terminate simulation only once
+ * the design has achieved a quiet state, without the use of '#' delays.
  */
 class uvml_hrtbt_mon_c extends uvm_component;
    
    // Configuration
    bit           enabled          = 1;
-   int unsigned  startup_timeout  = uvml_hrtbt_default_startup_timeout ;
-   int unsigned  heartbeat_period = uvml_hrtbt_default_heartbeat_period;
-   int unsigned  refresh_period   = uvml_hrtbt_default_refresh_period  ;
+   int unsigned  startup_timeout  = uvml_hrtbt_default_startup_timeout ; ///< Time period by which a heartbeat MUST be observed
+   int unsigned  heartbeat_period = uvml_hrtbt_default_heartbeat_period; ///< Timer duration for each component
+   int unsigned  refresh_period   = uvml_hrtbt_default_refresh_period  ; ///< Interval between calls to eval_heartbeat()
    
    // State
-   uvml_hrtbt_entry_struct  timestamps[$];
+   uvml_hrtbt_entry_struct  timestamps[int unsigned];
    bit                      observed_heartbeat = 0;
    
-   
+
    `uvm_component_utils_begin(uvml_hrtbt_mon_c)
       `uvm_field_int(enabled         , UVM_DEFAULT)
       `uvm_field_int(startup_timeout , UVM_DEFAULT)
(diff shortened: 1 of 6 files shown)


lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
```diff
@@ -237,72 +230,78 @@ endtask : phase_loop
...
+               `uvm_info("HRTBT_MON", $sformatf("Removing component from heartbeat list: id (%0d)", timestamps[ii].id), UVM_DEBUG)
             end
             else begin
-               `uvm_info("HRTBT", $sformatf("Removed component from heartbeat list: owner (%s), id (%0d)", timestamps[ii].owner.get_full_name(), timestamps[ii].id), UVM_DEBUG)
+               `uvm_info("HRTBT_MON", $sformatf("Removing component from heartbeat list: owner (%s), id (%0d)", timestamps[ii].owner.get_full_name(), timestamps[ii].id), UVM_DEBUG)
             end
-         end
-         else begin
-           new_timestamps.push_back(timestamps[ii]);
+            timestamps.delete(ii);
```
MikeOpenHWGroup, line 249: Soooo, lemme see if I got this right: `timestamps` has been redefined as an associative array (was a queue).  I would expect this to result in modest memory savings.  The big change is that the "matured" timestamps are deleted from the array and _not_ pushed onto another queue.  This is where the memory savings are.  Do I have that right?

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
```diff
@@ -237,72 +230,78 @@ endtask : phase_loop
...
    if (id == 0) begin
-      id = $random();
+   	pick_new_id = 1;
+   end
+   else if (timestamps.exists(id)) begin
+   	pick_new_id = 1;
+   	`uvm_warning("HRTBT_MON", $sformatf("Specified heartbeat ID (%0d) is already in use.  A new random ID will be picked.", id))
    end
+   do begin
+   	id = $urandom();
```
MikeOpenHWGroup, line 276: Does this need to be random?   Why not define `id` as a static member which is simply incremented each time a new one is needed?

MikeOpenHWGroup (commented):
Thanks for this @datum-dpoulin.  I have reviewed the changes and left a couple of comments/questions.  Please have a look at those.

I am running a local regression which looks fine so far and I do expect it to pass (will add a comment once it has completed).

@strichmo, you orginated the issue on this (#727), so you should be the one to approve this PR.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_pkg.flist.xsim
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Solderpad Hardware License v 2.1 (the "License"); you may not use this file except in compliance
+// with the License, or, at your option, the Apache License version 2.0. You may obtain a copy of the License at
+//                                        https://solderpad.org/licenses/SHL-2.1/
+// Unless required by applicable law or agreed to in writing, any work distributed under the License is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+
+// Directories
+-i ${DV_UVML_HRTBT_PATH}
```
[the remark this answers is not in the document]
datum-dpoulin: Sure thing! I couldn't remember if we still had the special xsim versions anymore. Good catch. Definitely one thing we gotta get them away from.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
datum-dpoulin: Yes I was copying all timestamps every time. Not sure why. This just deletes individual entries in an associative array.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
datum-dpoulin: It was already random, I didn't wanna change things.

MikeOpenHWGroup: OK, at some point this should be discussed.  Results from $urandom() calls are not constrainable.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_pkg.flist.xsim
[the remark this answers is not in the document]
MikeOpenHWGroup: Great.  Please push in an update that removes this file.   Thanks.

MikeOpenHWGroup (commented):
Note: my local regression passed (as expected).

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
```diff
@@ -237,72 +230,78 @@ endtask : phase_loop
...
-            timestamps.push_back(new_timestamps[ii]);
-         end
-      end
    end
    
 endtask : eval_heartbeat
 
 
-function void uvml_hrtbt_mon_c::heartbeat(uvm_component owner, int id=0);
+function void uvml_hrtbt_mon_c::heartbeat(uvm_component owner, int unsigned id=0);
```
strichmo, line 263: This method should bail out (without registering anything) if the heartbeat monitor is disabled.  This should alleviate the memory usage issues in any tests that may not want to use these.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
```diff
@@ -237,72 +230,78 @@ endtask : phase_loop
...
-           new_timestamps.push_back(timestamps[ii]);
+            timestamps.delete(ii);
          end
       end
       
-      if (new_timestamps.size() == 0) begin
-         `uvm_info("HRTBT", "Heartbeat count is 0. Dropping objection", UVM_NONE)
+      // 3. Exit loop if all entries are gone
+      if (timestamps.size() == 0) begin
+         `uvm_info("HRTBT_MON", "Heartbeat count is 0. Dropping objection", UVM_NONE)
```
strichmo, line 255: This can clutter logs.  Would prefer a higher verbosity level (UVM_HIGH)

strichmo (changes requested):

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
datum-dpoulin: Excellent point Steve, I will add this in.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_mon.sv
datum-dpoulin: I agree, with the exception of phases in which heartbeats were actually monitored. Then I'd put it at UVM_MEDIUM/LOW.

strichmo (approved):
Thanks for the fixes.

lib/uvm_libs/uvml_hrtbt/uvml_hrtbt_pkg.flist.xsim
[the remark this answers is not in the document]
datum-dpoulin: Crap, I forgot to get rid of it I think.

PR #834, opened 2021-08-31
Debug coverage update

silabs-robin:
Coverage class updated to new controller.

Debug coverage is re-enabled (upon COV=1), and covergroups have been slightly updated.
In addition, unused signals were removed from the interface and some signals got updated.
(There are also a few lines of whitespace fixes that I couldn't resist.)

Both formal and simulation tests are passing (ps, the cov model itself is not used in formal).
I synced with upstream cv32e40x/dev and resolved the conflicts.
On the "debug_test" it shows 90.61%, but I have not tried to merge coverage results from several other tests.

@silabs-oysteink  I think your eyes would be valuable on the assertions file and the tb connections.


cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -348,34 +348,24 @@ module uvmt_cv32e40x_tb;
     uvmt_cv32e40x_debug_cov_assert_if debug_cov_assert_if(
       .clk_i(clknrst_if.clk),
       .rst_ni(clknrst_if.reset_n),
```
strichmo, line 350: Could we push this instance a bit down into the hierarchy (perhaps at the wrapper level)?  It might make the signal paths a bit easier to read.

strichmo (changes requested):
Please review single comment.  I don't consider it a must-fix.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
silabs-robin: Fixed.
I changed it to use "bind" for both the interface instantiation and for the assertions.
Simulation still generates a sensible db and it passes debug_test, and formal shows that assertions are still checked.

strichmo: Thanks!

strichmo (approved):

strichmo:
I'm good for merge when @silabs-oysteink is OK as well.

silabs-oysteink (approved):

PR #837, opened 2021-09-01
Changed atoi and atohex to $sscanf as per LRM 6.16.9

silabs-hfegran:
atoi and atohex sign-extends and truncates without warning,
Incorrect behavior when address width >32 bits or MSB=1

(setting `TEST_PLUSARGS=+nmi_addr=0xbc000100` replicates the problem this commit fixes)

### DIFF
--- a/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
+++ b/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
@@ -1,4 +1,4 @@
-// 
+//
 // Copyright 2020 OpenHW Group
 // Copyright 2020 Datum Technology Corporation
 // Copyright 2020 Silicon Labs, Inc.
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Good catch @silabs-hfegran!  This issue could have been resolved by replacing `atoi()` with `atohex()`, but your implementation that replace both with `$sscanf()` is better because unexpected behaviour is caught.

I notice that a 0 return value of `read_cfg_plus_xlen()` is the same whether the user provides an incorrect plusarg or none at all:
```
   // Read plusargs for defaults
   if (read_cfg_plusarg_xlen("hart_id", hart_id)) begin
      hart_id_plusarg_valid = 1;
      hart_id.rand_mode(0);
   end
```
Is there a way to produce an error (or warning) if the user attempts to provide a value > 32-bits?

silabs-hfegran:
> Good catch @silabs-hfegran! This issue could have been resolved by replacing `atoi()` with `atohex()`, but your implementation that replace both with `$sscanf()` is better because unexpected behaviour is caught.

The original used both atoi and atohex, and it was actually the atohex that caused issues with 0xbc000100 - it got converted to a signed int then sign extended when assigned to the bit-array.

silabs-hfegran:
> I notice that a 0 return value of `read_cfg_plus_xlen()` is the same whether the user provides an incorrect plusarg or none at all:
> 
> ```
>    // Read plusargs for defaults
>    if (read_cfg_plusarg_xlen("hart_id", hart_id)) begin
>       hart_id_plusarg_valid = 1;
>       hart_id.rand_mode(0);
>    end
> ```
> 
> Is there a way to produce an error (or warning) if the user attempts to provide a value > 32-bits?

MAX_XLEN is 128-bits if I remember correctly, and the upper address bits ([127:32] for cv32) are constrained to 0 - that is how this error popped up in the first place. So a value >32 bits will set some of the upper bits and the constraint will fail

MikeOpenHWGroup (approved):

PR #845, opened 2021-09-02
sync design files with master for better merging of model updates

strichmo:


strichmo:
Forcing this through to enable @eroom1966 to proceed with his model updates.

PR #846, opened 2021-09-02
revert the RVVI/order increment code

eroom1966:
@strichmo Steve looks like this change has passed all of the CI tests, please take a look


strichmo:
Based on local regressions:
This PR is fine for CV32E40X.  rel_check passes fine.
There is an issue for the CV32E40P. in that the valid signaling changed for step and compare.  As we are planning to replace the step and compare module (for UVM) I have a patch to fix this in step and compare.  This passed a full regression last night.  

I will go ahead and approve this PR, then I will follow up with another PR with this patch.  No E40P GitHub CIs should be affected since this is PR is on the cv32e40x/dev branch.

strichmo (approved):

PR #853, opened 2021-09-09
random instruction bus fault test and directed data bus faults

strichmo:
Add a random test for instruction bus faults
Add a directed data_bus_error test for testing RTL turnon.  The ISS hooks are not complete yet, so must be run with USE_ISS=0
Add NMI handler to the CV32E40X BSP.


MikeOpenHWGroup (approved):
Looks good to me @strichmo (although I am not an expert in the `bsp`).  Will merge pending a successful local ci_check regression.

PR #840, opened 2021-09-01
Cleanup (or waive) lint issues

MikeOpenHWGroup:
This PR contains several updates to address Verissimo lint issues:

**cv32e40p/tb/core/mm_ram.sv:**
 - created setup_transaction() function for common code between reads and writes.

**cv32e40p/tb/uvmt/uvmt_cv32e40p_tb.sv:**
 - remove dead code (unused variables)

**lib/uvma_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv:**
 - add waiver pragmas

**lib/uvma_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv:**
 - add waiver pragmas

**lib/uvma_agents/uvma_obi_memory/src/comps/uvma_obi_memory_seq_item_log.sv:**
 - add waiver pragmas

@datum-dpoulin, please have a look at the updates to the `obi_memory` agent files.  Verissimo has (correctly) identified redundant code.  In at least one case (uvma_obi_memory_drv.sv) I think the redundancy is acceptable, so I have waived it.  The other redunancies are (I believe) all related to the boiler-plate implementation of the agent.  I have waived these as well, but we may want to update the templates.

PR #854, opened 2021-09-09
Address lint errors in CV32E40P environment

MikeOpenHWGroup:
This pull request address a few more of the SV/UVM lint errors reported by [Verissimo](https://dvteclipse.com/core5verif-verissimo/1/main/index.html).  @strichmo, some of this was merged onto the master branch already in PR #840.  I'm asking for a PR to cv32e40p/dev because I goofed up on at least one waiver, and I should have been PR'ing to dev all along (my bad).

- `mm_ram.sv`: The UVM environment no longer uses mm_ram, but the "redundant code" lint error was accurate, so I cleaned that up.
- `uvmt_cv32e40p_tb.sv` and `uvmt_cv32e40p_base_test.sv`: removed unused declarations and waived the "do not use `$display()`" rule (because it does not apply in these cases - see my comments to see if you agree).
- `uvma_obi_memory_cov_model`, `uvma_obi_memory_drv.sv` and `uvma_obi_memory_seq_item_logger.sv`: these are interesting.  The linter is flagging redundant code across separate files(!).  This is very cool.  I put waivers in place for these because much of the redundant code is template generated.  @datum-dpoulin, this is something you may want to look at in your templates, but we can live with it for now.

### DIFF
--- a/cv32e40p/tb/core/mm_ram.sv
+++ b/cv32e40p/tb/core/mm_ram.sv
@@ -190,6 +190,28 @@ module mm_ram
     int                            errno;
     string                         error_str;
 
+    // Common code used by both reads and writes
+    function void setup_transaction();
+
+       data_req_dec = data_req_i;
+       if ( (data_addr_i >= dm_halt_addr_i) &&
+            (data_addr_i < (dm_halt_addr_i + (2 ** DBG_ADDR_WIDTH)) )
+          ) begin
+          // remap debug code to end of memory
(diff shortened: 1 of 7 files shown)


datum-dpoulin (approved):
Waivers make sense. Good catch on the UVM utility macro mismatch

MikeOpenHWGroup:
> Waivers make sense. 

Thanks.

> Good catch on the UVM utility macro mismatch

Yeah, that is an example of a useful linter error.  Could eventually cause problems if someone wanted to start using the do_*() methods.

strichmo (approved):

PR #861, opened 2021-09-14
unify global symbols in readelf file

strichmo:
add readelf to common core configuration to set bootstrap pins automatically based on common "symbols" from the readelf.
note that plusargs in core_cntrl_cfg still take precedent if both defined.


strichmo:
The specific issue being addressed here is that any instruction faults taken early in a test (before loading the mtvec via a CSR instruction) would vector to a non-existent handler.

strichmo:
I'm going to pull this to reconsider the paradigm.

PR #862, opened 2021-09-14
Prevent regression scripts from parsing primary prompt variable

silabs-hfegran:
Excluded the primary prompt variable (PS1) from being parsed
into the regression enviornment.

Reasoning: Custom user prompts may cause regression scripts to
fail due to incompatible shell functionality settings.


strichmo (approved):
Thanks for the fix.

PR #863, opened 2021-09-14
Cv32e40p/merge dev to rel 27

strichmo:


MikeOpenHWGroup (approved):
Thanks @strichmo.  A good chunk of this is "just" linter fixes, but I will still wait for a local regression to pass before merging (half hour).

PR #864, opened 2021-09-14
Cv32e40x/merge dev to rel 46

strichmo:

PR #865, opened 2021-09-14
merge from cv32e40x/release and cv32e40p/release to master

strichmo:


MikeOpenHWGroup (approved):
Same review as #864.  Passed local regression of both E40X and E40P.

PR #880, opened 2021-09-20
Update the interrupts vplan

silabs-robin:
Updates the interrupts vplan that was previously copied over from 40P so that it now reflects features of the 40X.

Changelog:

General
* Moved items of the same feature adjacent to each other
* Updated "requirement location" to point to 40x user manual
* Added some minor comments on existing items
* (Hope I didn't ruin formatting, because excel defaults faught against the existing style used)
* Updated from "N/A" to "assertion coverage" for some assertion checks
* Cleanup up comments

Interrupt interface
* Removed "_o" suffix from irq_ack/irq_id (signals kept in rtl but not as top lvl pins)
* Added item about nmi and irq_ack
* Added item about rvfi_intr and irq_ack

Interrupt CSR
* Edited "MSTATUS MIE and MIE" to be clear, "MSTATUS[MIE] and MIE"

Interrupt instruction
* Removed Xpulp item
* Added all riscv extensions that are new in 40x contra 40p
* Updated wording of some "instructions interruped" items
* Added Zifencei item
* Added X interface instructions interrupted item
* Added illegal instructions interrupted item
* Removed "Counters" extension
* Added check that killed instructions have no side-effects

Eventually taken
* Added progress assertion for pending enabled interrupts

RVFI
* Added several items about rvfi interrupt signaling
* ...

NMI
* Added several items about nmi
* ...
* Added progress assertion for NMIs
* Added item for updating mstatus.mpie
* Updated mepc/mcause/mstatus items to check vs RM, instead of assertions
* Added item for nmip bit cover statement for debug/not


strichmo:
Setting DO NOT MERGE until review is completed.

silabs-robin:
Is it better to wait until other vplans are also done? Fence.i and possibly others too?

strichmo:
Yes I was going to propose that today.  Let's get started on the fence.i vplan and combine those reviews.  I will create an issue today and make a MM announcement that an "offline" review can start for the interrupts above.

silabs-robin:
I pushed an update.
One can merge with "squash" to keep some KBs of cruft away from the history.
If there is much more feedback then I will queue up changes so there aren't too many pings about updates.

MikeOpenHWGroup (approved):

PR #895, opened 2021-09-24
initial RVVI NMI interrupt signaling

strichmo:
The directed test data_bus_error should enable testing NMI handler entry once implemented in OVPSIM


strichmo:
Please note that a bug in the RTL where the wrong mcause code is used (store code for loads and load code for stores) is still in master.  This PR, once merged, should fix that:

https://github.com/openhwgroup/cv32e40x/pull/244

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Looks good.   Passes local regression.

PR #897, opened 2021-09-27
Strichmo/pr/rvfi cycle cnt

strichmo:
add a fetch_constant plusarg for benchmarks and performance simulations
add a cycle count to the rvfi monitor log messages

### DIFF
--- a/cv32e40x/env/uvme/vseq/uvme_cv32e40x_core_cntrl_fetch_toggle_seq.sv
+++ b/cv32e40x/env/uvme/vseq/uvme_cv32e40x_core_cntrl_fetch_toggle_seq.sv
@@ -64,6 +64,10 @@ function uvme_cv32e40x_fetch_toggle_seq_c::new(string name = "");
     fetch_toggle_mode = FETCH_RANDOM_TOGGLE;
     fetch_toggle_mode.rand_mode(0);
   end
+  else if ($test$plusargs("fetch_constant")) begin
+    fetch_toggle_mode = FETCH_CONSTANT;
+    fetch_toggle_mode.rand_mode(0);
+  end
 
 endfunction : new
 
--- a/cv32e40x/regress/cv32e40x_hello_world.yaml
+++ b/cv32e40x/regress/cv32e40x_hello_world.yaml
@@ -1,20 +1,20 @@
 # YAML file to specify the ci_check regression testlist.
-name: cv32e40p_ci_check
-description: Commit sanity for the cv32e40p
+name: cv32e40x_ci_check
+description: Commit sanity for the cv32e40x
 
 builds:
   corev-dv:
     cmd: make comp_corev-dv
-    dir: cv32e40p/sim/uvmt
+    dir: cv32e40x/sim/uvmt
 
-  uvmt_cv32e40p:
+  uvmt_cv32e40x:
     cmd: make comp
-    dir: cv32e40p/sim/uvmt
+    dir: cv32e40x/sim/uvmt
 
 tests:
   hello-world:
-    build: uvmt_cv32e40p
+    build: uvmt_cv32e40x
     description: UVM Hello World Test
-    dir: cv32e40p/sim/uvmt
+    dir: cv32e40x/sim/uvmt
     cmd: make test COREV=YES TEST=hello-world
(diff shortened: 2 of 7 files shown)


lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -64,28 +64,34 @@ interface uvma_rvfi_instr_if
...
   // Local variables
   // -------------------------------------------------------------------
+  bit [CYCLE_CNT_WL-1:0] cycle_cnt = 0;
 
   // -------------------------------------------------------------------
   // Begin module code
   // -------------------------------------------------------------------
-  
+
+  always @(posedge clk) begin
```
MikeOpenHWGroup, line 79: Should this not be gated by reset?

cv32e40x/env/uvme/vseq/uvme_cv32e40x_core_cntrl_fetch_toggle_seq.sv
```diff
@@ -64,6 +64,10 @@ function uvme_cv32e40x_fetch_toggle_seq_c::new(string name = "");
     fetch_toggle_mode = FETCH_RANDOM_TOGGLE;
     fetch_toggle_mode.rand_mode(0);
   end
+  else if ($test$plusargs("fetch_constant")) begin
```
MikeOpenHWGroup, line 67: We are sprinkling `$test$plusargs` checks throughout our code rather than centralizing things in cfg objects that can be constrained in a UVM testcase.   Just because Google does it...

MikeOpenHWGroup (changes requested):
A coupla questions for you.

cv32e40x/regress/cv32e40x_hello_world.yaml
```diff
@@ -1,20 +1,20 @@
...
 tests:
   hello-world:
-    build: uvmt_cv32e40p
+    build: uvmt_cv32e40x
     description: UVM Hello World Test
-    dir: cv32e40p/sim/uvmt
+    dir: cv32e40x/sim/uvmt
     cmd: make test COREV=YES TEST=hello-world
-    num: 500
+    num: 100
```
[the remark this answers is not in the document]
strichmo: It had some initial value when we getting stalls working properly in various forms (the old tracer, RVFI, the new E40X controller, etc.)  It also had some value on this PR (i.e. manually adding stall disable to check into why runtimes were different).

I would be fine with either:
- deleting or
- adding some description comments that this is a boilerplate, example regression for individual testing

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
strichmo: In my opinion, cycle counts should be immutable for a given simulation run.  In other words, you shouldn't see a cycle count twice, even if you are applying full resets to the DUT.

cv32e40x/env/uvme/vseq/uvme_cv32e40x_core_cntrl_fetch_toggle_seq.sv
strichmo: Yeah...I am painfully aware of this.  
After I get the PMA/address randomization PR done, I will address this as the next methodology issue to tackle.  Once that is implemented, we can forbid $test$plusargs and $value$plusargs outside of the developed infrastructure.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
MikeOpenHWGroup: Ah.  I misunderstood the purpose.  I assumed this would be applied to a check of the `cycle` CSR.

cv32e40x/regress/cv32e40x_hello_world.yaml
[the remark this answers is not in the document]
MikeOpenHWGroup: If you have gotten value out of this single-test-program regression, then that is a signal not to delete it.  Let's add some descriptive comments.

Issue #896, opened 2021-09-26
CV32E40P Test Program is not Compiling


drac98:
I'm new to RISC-V and I have been trying to run a simulation of CV32E40P with Verilator.

## CV32E40P Test Programs Compiling Error

I followed the instructions given [here](https://github.com/openhwgroup/core-v-verif/tree/2f8d8d3a4413b87e95fc4f399ff711475473e41d/cv32e40p/sim/core).
- Point your environment variable RISCV to your RISC-V toolchain. Call make to run the default test (hello_world).

Compiler : `corev-openhw-gcc-ubuntu2004-20200913`
Verilator : `4.212 2021-09-01 rev v4.212-20-g27c08a65`

### Compile error

Terminal output:
https://pastebin.com/b67RYu6K

- [x] Cloning CV32E40P RTL model
- [x] Compiling CORE TB and CV32E40P with Verilator
- [x] Compiling BSP
- [ ] Compiling test-program ../../tests/programs/custom/hello-world/hello-

### Steps to Reproduce

1. Install recommended toolchain
2. Clone the `core-v-verifa` repo
2. go to `CV32E40P/sim/core` directory
3. run `make`

If more information is needed, please let me know.

MikeOpenHWGroup:
Hi @drac98, thanks for bringing this issue to our attention.  I have been able to reproduce the issue and am working on a fix.

MikeOpenHWGroup:
Hi @drac98, with #899 merged in I believe things should be working for you now.  Please check and let me know.   FYI, this fix is on the master branch and will be propagated to the `cv32e40p/dev` and `cv32e40p/release` branches at the end of this week.

drac98:
Hello @MikeOpenHWGroup , the fix works. Thank you.

PR #899, opened 2021-09-27
Fix up core testbench for Verilator and DSIM

MikeOpenHWGroup:
Hi @strichmo, this addresses issue #896 for Verilator and DSIM.  I _think_ it also addresses Xcelium, but my connection to the servers I use that support Xcelium are not available this morning.

### DIFF
--- a/cv32e40p/sim/core/Makefile
+++ b/cv32e40p/sim/core/Makefile
@@ -56,6 +56,10 @@ CV_CORE_UC     = $(shell echo $(CV_CORE) | tr a-z A-Z)
 # Compile compile flags for all simulators
 SV_CMP_FLAGS =
 
+# Default "custom test-program"
+CUSTOM_PROG  ?= requested_csr_por
+TEST         ?= hello-world
+
 ###############################################################################
 # Common Makefiles:
 #  -Variables for RTL and other dependencies (e.g. RISCV-DV)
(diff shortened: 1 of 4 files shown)


cv32e40p/sim/core/Makefile
```diff
@@ -421,21 +412,33 @@ xrun-all: $(CV_CORE_PKG) $(RTLSRC_PKG) $(RTLSRC) $(TBSRC_PKG) $(TBSRC)
...
-xrun-hello-world: xrun-all $(VERI_CUSTOM)/hello-world/hello-world.hex
-	$(XRUN) \
-		$(XRUN_FLAGS) \
-		-f $(CV_CORE_MANIFEST) \
-		$(TBSRC_PKG) $(TBSRC) \
-		+firmware=$(VERI_CUSTOM)/hello-world/hello-world.hex
-
-.PHONY: xrun-custom
-xrun-custom: xrun-all $(VERI_CUSTOM)/$(CUSTOM_PROG)/$(CUSTOM_PROG).hex
+#.PHONY: xrun-hello-world
```
strichmo, line 415: Can we just get rid of these targets instead of commenting?

strichmo (changes requested):
Thanks for aligning the custom and test target names.  I will look to get the UVMT makefiles cleaned up accordingly.

There is a single request to consider removing a commented-out target.

cv32e40p/sim/core/Makefile
MikeOpenHWGroup: Yes we can.  Good catch.

strichmo (approved):

PR #905, opened 2021-09-29
Update the counters vplan (DO NOT MERGE)

silabs-robin:
Most importantly, the vplan had outdated definitions of the "event selectors" vs what is now in the 40x user manual.
I also fixed some other items that I found could be changed:

Count inhibit
* Added mcountinhibit check for indices [3:31] (inhibiting mhpmcounters)
* Added check that inhibiting mcycle/minstret doesn't inhibit mhpmcounters

Minstret
* Move "instruction count" item from mcycle to minstret

Overflow
* Move the two overflow assertions from mcycle to hold for all counters

Mhpmcounter
* Updated counter selectors (names/numbers) according to newest 40x spec
* Added assertion that "minstret-like" events count with correct timing
* Added check for different NUM_MHPMCOUNTERS


silabs-robin:
Superseded by https://github.com/openhwgroup/core-v-verif/pull/907

PR #908, opened 2021-09-30
refactor code

eroom1966:
@strichmo 
Hi Steve, just some local refactoring - this should have no effect outside of the module itself
all the CI tests pass for me on both E40P and E40X (on this branch)


strichmo (approved):
Confirmed passes regression-merging changes.

PR #910, opened 2021-10-04
Create a coverage model for 40x Counters

silabs-robin:
Made covergoups according to the vplan's functional coverage. Also added links to coverage in the vplan.
The covergroups are in `cv32e40x/env/uvme/cov/uvme_counters_covg.sv` and only relies on a port from the rvfi agent.


cv32e40x/env/uvme/cov/uvme_counters_covg.sv
```diff
@@ -0,0 +1,195 @@
...
+
+
+covergroup cg_counters (int num_mhpmcounters)
+  with function sample(uvma_rvfi_instr_seq_item_c#(ILEN, XLEN) rvfi);
+
+  `per_instance_fcov
+
+  cp_inhibit_mcycle : coverpoint rvfi.csrs["mcountinhibit"].get_csr_retirement_data()[0];
+  cp_inhibit_minstret : coverpoint rvfi.csrs["mcountinhibit"].get_csr_retirement_data()[2];
+  cp_is_csr_read : coverpoint ((rvfi.insn[6:0] == 7'b 1110011) && (rvfi.insn[13:12] != 2'b 00) && (rvfi.insn[11:7] != 0)) {
```
strichmo, line 29: Good coverpoints, however it feels that we are reimplementing some decoding that occurs already within ISACOV.  Do you think we could make this coverage model "driven" by ISACOV's sequence item?  We could attach the RVFI instruction sequence handle to the ISACOV object to efficiently pass through the CSR data.

silabs-robin: > Do you think we could make this coverage model "driven" by ISACOV's sequence item?

In total, I am not sure if we should.
Looking at the cases here: 1) mcountinhibit[idx] does not benefit (AFAICT), 2) cp_is_csr_read could avoid the decoding but extracting the info from an isacov object is not less complex, 3) rvfi.dbg_mode doesn't benefit, 4) "rvfi.insn[31:20] == (12'h B00 + idx)" could maybe not benefit so much(?) because it would then use an enum which is not as easy to compare against arithmetic.
So I think it currently is a tiny amount of decoding, using standard riscv knowledge, and without much overhead.
(But I am not opposed to it; I will change it if there is good reason or are more votes in favor of change.)

cv32e40x/env/uvme/cov/uvme_counters_covg.sv
```diff
@@ -0,0 +1,195 @@
...
+    bins mhpm_idx = {1};
+  }
+  cp_is_mhpm_idx_h : coverpoint (rvfi.insn[31:20] == (12'h B80 + idx)) {
+    bins mhpm_idx_h = {1};
+  }
+  cp_is_dbg_mode : coverpoint rvfi.dbg_mode {
+    bins dbg_mode = {1};
+  }
+
+  x_check_mhpm : cross cp_inhibit, cp_event, cp_is_csr_read, cp_is_mhpm_idx {
```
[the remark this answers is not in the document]
silabs-robin: Fixed

cv32e40x/env/uvme/cov/uvme_counters_covg.sv
MikeOpenHWGroup: > I will change it if there is good reason or are more votes in favor of change.

I vote in favour of the change.

It is true that doing a "little bit" of decoding in the coverage model is both quick and easy and it may be overkill to use ISACOV sequence items here.  Nevertheless, it is important that all items in our coverage model that _can_ be driven by ISACOV sequence items _are_ driven by ISACOV sequence items.  Otherwise we will soon find ourselves in a position where "most off" the coverage is driven by ISACOV and the only way to find out what isn't is to make a change a "see what breaks".  This will quickly develop into an ever-growing maintenance headache.

Please do make the necessary changes to remove all instruction decoding from this coverage code and use ISACOV sequence items for that.  A little bit of extra work now will say you a lot of debugging later.

silabs-robin: > We could attach the RVFI instruction sequence handle to the ISACOV object to efficiently pass through the CSR data.

This (picture below) is what was intended, right?
I can try to fix the rest of it tomorrow.

image

strichmo: I believe this is OK.  ISACOV is pretty much completely coupled to RVFI and I couldn't well justify copying the CSR information out for each instruction so attaching a handle is easy and cheap.

PR #915, opened 2021-10-05
Update TB after irq_ack and irq_id were made internal to the core.

silabs-oivind:
CI clean.


MikeOpenHWGroup (approved):
Thanks @silabs-oivind!

PR #919, opened 2021-10-05
instruction error directed test enhancement

strichmo:
Add test to directed test where an OBI error occurs in a prefetched instruction that is not executed.


MikeOpenHWGroup:
Ahhh.  I get it now.   Nice use of compiler macros to help the reader (me) understand.   Thanks.

PR #931, opened 2021-10-12
Connected x-interface in dut wrapper.

silabs-oysteink:
The latest versions of the E40X includes the eXtension interface on the toplevel and wrapper. Not connecting this make core-v-verif unable to compile. Connected this interface in the dut wrapper get compilation to work.


strichmo:
Passes a local ci_check

PR #935, opened 2021-10-13
add the Verilab SVLIB to cv32e40x

strichmo:
To provide useful common utilities (including regexp parsing, extended file operators, time functions, etc.) the Open Source Verilab SVLIB library will be cloned and compiled into all core-v-verif simulations.

The SVLIB itself is licensed under Apache 2.0, however no source code is included in the environment.  The SVLIB is hosted at a bitbucket site and is cloned as other 3rd party libraries (e.g. riscv-dv and Spike)

This has been tested with dsim, Xcelium, and Questa.  The Riviera and VCS makefiles are updated, but untested.

A common makefile target is added to enable rebuilding the SVLIB DPI shared library (make svlib).  A 64-bit Linux shared library is checked into this PR which should work for any Linux users of core-v-verif.


cv32e40x/vendor_lib/verilab/README.md
```diff
@@ -0,0 +1,33 @@
...
+- File pathname manipulation
+- Environment variable queries
+- Time functions
+
+## Documentation
+
+For more details on using the SVLIB, please refer to their documentation PDF.  This PDF can be found either on the Verilab website
+or by cloning the Bitbucket SVLIB repository into vendor_lib (`make clone_svlib`) and finding the documenation under `svlib/doc`
+
+## CORE-V_VERIF COmpilation
```
MikeOpenHWGroup, line 17: Typo: "COmpilation" s.b. "compilation".  (Wouldn't nag about something this simple, but I'm already asking for a change to `uvmt_cv32e40x_tb.sv`.)

mk/Common.mk
```diff
@@ -162,6 +162,21 @@ else
 endif
 # DPI_DASM Spike repo var end
 
+###############################################################################
+# Generate command to clone Spike for the Disassembler DPI (used in the isacov model)
```
MikeOpenHWGroup, line 166: This comment header is for Spike, not SVLIB.

MikeOpenHWGroup (changes requested):
This is a great addition to core-v-verif @strichmo, thanks for implementing it.   I have a couple of minor comments on a few files.

I am wondering why svlib in placed in `cv32e40x/vendor_lib` rather than `vendor_lib`?  Wouldn't we want to make this easily available to all environment?

cv32e40x/vendor_lib/verilab/README.md
strichmo: Fixed.  Push incoming.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb.sv
```diff
@@ -612,6 +612,20 @@ module uvmt_cv32e40x_tb;
 
    assign core_cntrl_if.clk = clknrst_if.clk;
 
+   // Informational print message on loading of OVPSIM ISS to benchmark some elf image loading times
+   // OVPSIM runs its initialization at the #1ns timestamp, and should dominate the initial startup time
+   longint start_ovpsim_init_time;
+   longint end_ovpsim_init_time;
+   initial begin
```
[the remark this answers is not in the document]
strichmo: Normally I would agree.  However the nature of this code is that it is trying to "guesstimate" when OVPSIM has started, and this it has some very hacky timestamps.  I generally don't combine threads with explicit time control (or event waits, etc.)  I can raise an issue with Imperas to benchmark on their side rather than use this hack.

PR #939, opened 2021-10-13
Cv32e40x/dev

eroom1966:
@strichmo 
This pull request adds an errata to the E40P to model the **incorrect** behavior for the 0.13.2 debug spec, this is an additional parameter to create this errata behavior

>         .defer_step_bug      = True,        // enable erratum step behavior

The E40X does not have this errata applied, and hence will fail until the RTL is updated from @silabs-oysteink 1.0.0 debug spec inclusion, unless the parameter is set using an override


strichmo:
@eroom1966 Sorry if I am overlooking but I don't see the override switch to set this parameter in the model.  Could you please provide guidance as to what switch I should set for the E40P?

PR #950, opened 2021-10-18
Performance counters update

silabs-hfegran:
- Updated performance counter tests for 40x events
- Added performance counter test for interrupt events to the interrupt test.

### DIFF
--- a/cv32e40x/tests/programs/custom/hpmcounter_basic_test/test.yaml
+++ b/cv32e40x/tests/programs/custom/hpmcounter_basic_test/test.yaml
@@ -2,5 +2,3 @@ name: hpmcounter_basic_test
 uvm_test: uvmt_$(CV_CORE_LC)_firmware_test_c
 description: >
     HPM Counter Basic Test (donated by Arjan Bink)
-plusargs: >
-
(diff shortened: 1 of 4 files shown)


cv32e40x/tests/programs/custom/hpmcounter_hazard_test/hpmcounter_hazard_test.c
```diff
@@ -50,10 +50,38 @@ static int chck(unsigned int is, unsigned int should)
   return err;
 }
 
+static int chck_le(unsigned int is, unsigned int should)
```
strichmo, line 53: Any reason this is static?

cv32e40x/tests/programs/custom/mhpmcounter29_csr_access_test_2/test.yaml
```diff
@@ -2,6 +2,4 @@ name: mhpmcounter29_csr_access_test_2
 uvm_test: uvmt_$(CV_CORE_LC)_firmware_test_c
 description: >
     CSR access test with NUM_MHPMCOUNTER = 29 (CANNOT run with ISS)
-ovpsim: >
```
strichmo, line 5: So we don't need to suppress OBI-I stalls for this test?

strichmo (changes requested):

PR #960, opened 2021-10-20
enable random instruction bus errors in corev-dv

strichmo:
- for corev-dv tests, use the initial address in ram section as a mtvec bootstrap such that exceptions occurring at time zero will be handled properly
- remove the OBI to RVVI instruction bus fault connection
- use RVFI instead to detect and signal RVVI ovpsim instruction bus faults
- add one-shot error mode to only fire one instruction bus error in corev-dv tests, because the generated program cannot handle multiple errors


lib/uvm_agents/uvma_rvvi/uvma_rvvi_sqr.sv
```diff
@@ -21,7 +21,6 @@
 `define __UVMA_RVVI_SQR_SV__
 
 `uvm_analysis_imp_decl(_rvfi_instr)
-`uvm_analysis_imp_decl(_obi_i)
```
MikeOpenHWGroup, line 24: I fell off the banner cart here...  OBI memory transactions are removed (entirely it seems).  Why were they ever part of the RVVI agent?

MikeOpenHWGroup (commented):
Hi @strichmo.  I can see no fault in this PR, but I am not understanding something pretty fundamental - see my comment in `lib/uvm_agents/uvma_rvvi/uvma_rvvi_sqr.sv`

lib/uvm_agents/uvma_rvvi/uvma_rvvi_sqr.sv
MikeOpenHWGroup: Thanks!

Issue #997, opened 2021-10-31
Wrong number of arguments specified for 'interrupt' attribute


MikeOpenHWGroup:
### Type
* Compile error

### Steps to Reproduce
1. Check out hash `7f9ff9881fd399af9f786a0aa5bf1ee0b5ab54a4` of core-v-verif (current head of `cv32e40x/dev`).
2. $ cd core-v-verif/cv32e40x/sim/uvmt
3. $ make test TEST=interrupt_test

### Error message from [Metrics log](https://openhwgroup.metrics.ca/logs?externalIds[]=2519536f-f7a9-4295-853e-ff5d721c0f4d):
```
/opt/riscv/bin/riscv32-unknown-elf-gcc \
	 \
	-Os -g -static -mabi=ilp32 -march=rv32imc -Wall -pedantic \
	 \
	-I ../../tests/asm \
	-I /mux-flow/build/repo/cv32e40x/bsp \
	-o /mux-flow/build/results/default/interrupt_test/0/test_program/interrupt_test.elf \
	-nostartfiles \
	/mux-flow/build/repo/cv32e40x/tests/programs/custom/interrupt_test/interrupt_test.c \
	-T /mux-flow/build/repo/cv32e40x/bsp/link.ld \
	 -L /mux-flow/build/results/default/interrupt_test/0/test_program/bsp \
	-lcv-verif

In file included from /mux-flow/build/repo/cv32e40x/tests/programs/custom/interrupt_test/interrupt_test.c:5:0:
/mux-flow/build/repo/cv32e40x/tests/programs/custom/interrupt_test/interrupt_test.h:73:1: error: wrong number of arguments specified for 'interrupt' attribute
 __attribute__((interrupt ("machine"))) void m_software_irq_handler(void);
 ^~~~~~~~~~~~~

/mux-flow/build/repo/cv32e40x/tests/programs/custom/interrupt_test/interrupt_test.h:74:1: error: wrong number of arguments specified for 'interrupt' attribute
 __attribute__((interrupt ("machine"))) void m_timer_irq_handler(void);
 ^~~~~~~~~~~~~
```

strichmo:
@MikeOpenHWGroup I seem to remember this being an issue with using an older toolchain.  Could you show me which version of gcc you are using?

MikeOpenHWGroup:
gcc 9.3.0 on my Ubuntu20.04 desktop and 4.8.5 on CMC's CentOS7 machines.

[...]

PR #1002, opened 2021-11-05
New image with updated toolchain and dsim

MikeOpenHWGroup:


MikeOpenHWGroup:
For the record:
- this is in response to #997
- the new toolchain is `riscv32-embecosm-gcc-ubuntu2004-20211027` which is based on gcc version `12.0.0 20211027`.

MikeOpenHWGroup:
I am merging this in now because it unlocks CI on all outstanding PRs.

PR #1006, opened 2021-11-08
fix label mismatch in fenci_instr_lib

strichmo:


strichmo:
Pushing obvious fix to continue merge cycle.

PR #1013, opened 2021-11-10
update the E40X hash to latest

strichmo:
Updates PMA policies per E40X User Manual updates.


MikeOpenHWGroup (approved):
> Updates PMA policies per E40X User Manual updates

I'll bet there is a story there.  :wink: 

Updates look good - will merge pending successful local regression.

PR #1023, opened 2021-11-16
corev_pma_rand_pma_test fixes for instruction corruption

strichmo:
This PR fixes two major issues with the _corev_pma_rand_pma_test_ in cv32e40x regressions.

- OVPSIM does not seem to handle fetches from the _end_ of memory (i.e 0xffff_fffc) well.
- super.post_randomize() usage in many PMA instructions streams is cleaned up to avoid inserting instructions from super classes that are not properly constrained in accordance with the intended class.  This caused address generation for these instructions to be not-as-intended which could overwrite instructions in the flat memory space.


silabs-hfegran (approved):
Looks good to me

PR #1025, opened 2021-11-16
connect nmip CSR RVFI to ISS NMI signal

strichmo:
Enables data bus errors in debug mode to be properly signaled for NMI pending.

### DIFF
--- a/cv32e40x/env/uvme/uvme_cv32e40x_core_sb.sv
+++ b/cv32e40x/env/uvme/uvme_cv32e40x_core_sb.sv
@@ -349,8 +349,9 @@ function void uvme_cv32e40x_core_sb_c::check_csr(uvma_rvfi_instr_seq_item_c#(ILE
                                                  uvma_rvvi_state_seq_item_c#(ILEN,XLEN) rvvi_state);
 
    foreach (rvfi_instr.csrs[i]) begin
-      string csr = rvfi_instr.csrs[i].csr;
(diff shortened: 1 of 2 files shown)


cv32e40x/env/uvme/uvme_cv32e40x_core_sb.sv
```diff
@@ -360,17 +361,24 @@ function void uvme_cv32e40x_core_sb_c::check_csr(uvma_rvfi_instr_seq_item_c#(ILE
          `uvm_fatal("CORESB", $sformatf("CSR %s from RVFI does not exist in RVVI state interface", csr));
       end
 
+      // Adjust CSR mask
```
MikeOpenHWGroup, line 364: I'm not going to gate merging on this question, but I am curious why this mask is necessary in this context.  Why would we skip dcsr.nmip if the core is not in debug mode?

MikeOpenHWGroup (approved):
Looks good and passes local regression.  I wish all PRs were this easy.  :-)

cv32e40x/env/uvme/uvme_cv32e40x_core_sb.sv
strichmo: This is an artifact of the RTL.  The DCSR also has no meaning in machine mode as it is not accessible.

MikeOpenHWGroup: Thanks!

PR #1026, opened 2021-11-17
set test timeout from plusarg 

strichmo:
Default timeout left unchanged.
Also fixes a fixme in the base test.


MikeOpenHWGroup (approved):
Passes local regression.

Had to look up why the RVFI monitor needed to know the NMI address, but I get it now...

PR #1028, opened 2021-11-18
Create bus errors (obi/rvfi) scoreboard for 40x

silabs-robin:
This PR adds a scoreboard that intends to check that OBI transactions with "err" bus faults becomes visible on RVFI.
The in-code "NaturalDocs" comment describes how it works.

I have tested with the directed tests `data_bus_error` and `instr_bus_error` and those work as expected.
`corev_rand_data_obi_err`, `corev_rand_data_obi_err_debug`, `corev_rand_instr_obi_err`, `corev_rand_instr_obi_err_debug` were also tried, and they did pass before pulling latest `cv32e40x/dev` and now I get ISS miscompares and infinite loops (should not be this scoreboard's fault). Note that `corev_rand_instr_obi_err_debug` did give a warning as it completed with one pending bus fault that hadn't seen its handler yet.
Passed ci_check when it ran during the VTG today.

If anyone thinks there could be a better way of checking this, please just let me know. This is what I managed to come up with, but I don't have a strong feeling about how good it is. It is not a big class, but I found that some of the logic is a bit tricky.
I would like to test this more but I don't think we have any additional bus error tests than those mentioned above?

@silabs-hfegran, since we discussed a little bit, if you have time to review that would be appreciated.


cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
```diff
@@ -0,0 +1,283 @@
...
+
+  super.check_phase(phase);
+
+  // Check OBI D-side
+  assert (cnt_obid_trn > 0)
+    else `uvm_warning(info_tag, "zero D-side OBI transactions received");
+  assert (cnt_obid_trn >= cnt_obid_err)
+    else `uvm_error(info_tag, "obid 'err' transactions counted wrong");
+  assert (cnt_obid_trn != cnt_obid_err)
+    else `uvm_warning(info_tag, "all the D-side OBI transactions were errs");
```
silabs-hfegran, line 188: L186's >= overlaps L186's !=, is it necessary to both warn and err in case of >= ? 

Same goes for the I-side further down.

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
```diff
@@ -0,0 +1,283 @@
...
+endfunction : check_phase
+
+
+function bit uvme_cv32e40x_buserr_sb_c::should_instr_err(uvma_rvfi_instr_seq_item_c#(ILEN,XLEN) rvfi_trn);
+
+  uvma_obi_memory_addr_l_t  err_addrs[$];
+  bit [31:0]                rvfi_addr = rvfi_trn.pc_rdata;
+
+  // Extract all addrs from queue of I-side OBI "err" transactions
+  foreach (obii_err_queue[i]) err_addrs[i] = obii_err_queue[i].address;
```
silabs-hfegran, line 245: easier to read if using begin-end construct instead of one-liner loops (and often also less error prone)

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
```diff
@@ -0,0 +1,283 @@
...
+
+  foreach (err_addrs[i]) begin
+    bit compressed =
+      (rvfi_trn.insn[1:0] != 2'b 11)
+      && !({rvfi_addr[31:2], 2'b 00} inside {err_addrs});
+    bit [31:0] hi_addr = err_addrs[i] + 4;
+    bit [31:0] lo_addr = err_addrs[i] - (compressed ? 2 : 4);
+
```
silabs-hfegran, line 253: This is a matter of preference, logic is not wrong - for readability I would suggest perhaps changing it to something like the following snippet. The non-inclusive range is not as intuitive to read and the "inside lo:hi" construction makes the intention obvious:
`bit [31:0] err_addr_range_hi = err_addrs[i] + 4 + 1;`
`bit [31:0] err_addr_range_lo = err_addrs[i] - (compressed ? 2 : 4) - 1;`
`if (rvfi_addr inside {[err_addr_range_lo:err_addr_range_hi]})`

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
silabs-robin: I don't think it could both warn and err?
The case of `<` would give an error.
The case of `==` would give a warning.
One could have made it err on `<=` and only have err without warning, but I didn't see a reason to forbid `==` even if it is very suspicious so I made it a warning.

silabs-hfegran: my bad - I misinterpreted things here; disregard this comment

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
silabs-robin: I can change this if one more person opines and creates a majority in favor of changing it. (I couldn't find that the style guidelines dictate anything on this?)

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
silabs-robin: I can change this if one more person opines and creates a majority in favor of changing it. (I couldn't find that the style guidelines dictate anything on this?)

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
strichmo: I would agree with @silabs-hfegran

MikeOpenHWGroup (approved):
I like this, thanks for the PR.

cv32e40x/env/uvme/uvme_cv32e40x_buserr_sb.sv
MikeOpenHWGroup: I generally favour `low:high` ranges for readability.  Having said that, I would not advocate for the exclusion of non-inclusive ranges in the style guide because they can be useful in generate loops.

strichmo (approved):

Issue #1030, opened 2021-11-20
Should we suppress ReadingOutputModport dsim warning?


MikeOpenHWGroup:
[CV32E40X issue #259](https://github.com/openhwgroup/cv32e40x/issues/259) describes a ReadingOutputModport compile-time warning emitted by Metrics dsim.   This warning is not seen with Cadence Xcelium and in a discussion with Metrics it was agreed that reading from an output modport does not violate the LRM.

I would like to propose that we suppress this dsim warning, but will only do so if @strichmo agrees it is safe to so.

strichmo:
Given that Xcelium does not issue a warning, I am OK with suppressing this warning.

I will reassign back to @MikeOpenHWGroup to make the change.

MikeOpenHWGroup:
This issue can be closed when pull-request #1034 is merged in.

PR #1034, opened 2021-11-23
Suppress dsim compile warning and support for core testbench

MikeOpenHWGroup:
The primary purpose of this PR is to suppress dsim ReadingOutputModport warnings (see #1030).

I am sneaking in several updates for the core testbench as well (actually most of the changes in this PR).  Note that the core TB cannot compile (yet), but I want to get these updates in to help the USouthhampton team.


MikeOpenHWGroup:
Sorry, I had pushed in everything _but_ the dsim warning suppression.  It's in now.

strichmo (approved):

PR #1036, opened 2021-11-24
Remove redundant use of always for a_dmhaltaddr_stable

MikeOpenHWGroup:
For reasons unknown, the addition of `a_dmhaltaddr_stable` cause Metrics DSIM simulations of some (but not all) testcases to drastically slow down. For example, "make test TEST=interrupt_test" goes from ~8.5 minutes to more than 87 minutes.  Oddly, the simulation runs correctly, it just takes a very long time.   I have raised a support issue with Metrics.

The code that causes this issue is changed in this PR.  A redundant whitespace (`2'b 00`) is also fixed, but that is of no consequence to either xrun or dsim.  Please have a look and satisfy yourself that I have not broken anything.

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
@@ -550,7 +550,7 @@ module uvmt_cv32e40x_debug_assert
     end
     a_dmhaltaddr_stable : assert property (
         fetch_enable_i_sticky
-        |=>
+        |->
         $stable(cov_assert_if.dm_halt_addr_i)
         ) else `uvm_error(info_tag, "dm_halt_addr_i changed after fetch_enable_i");
 
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
@@ -543,12 +543,12 @@ module uvmt_cv32e40x_debug_assert
     a_dmhaltaddr_stable : assert property (
         (cov_assert_if.fetch_enable_i && cov_assert_if.rst_ni)
         |=>
-        always $stable(cov_assert_if.dm_halt_addr_i)
+        $stable(cov_assert_if.dm_halt_addr_i)
         ) else `uvm_error(info_tag, "TODO");
 
     // Should be word-aligned
     a_dmhaltaddr_aligned : assert property (
-        cov_assert_if.dm_halt_addr_i[1:0] == 2'b 00
+        cov_assert_if.dm_halt_addr_i[1:0] == 2'b00
         ) else `uvm_error(info_tag, "TODO");
 
 
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
@@ -540,11 +540,19 @@ module uvmt_cv32e40x_debug_assert
     // Check that "dm_halt_addr_i" is correct
 
     // Should be stable after "fetch_enable_i"
+    logic fetch_enable_i_sticky;
+    always @(posedge cov_assert_if.clk_i or negedge cov_assert_if.rst_ni) begin
+        if (!cov_assert_if.rst_ni) begin
+            fetch_enable_i_sticky <= 0;
+        end else if (cov_assert_if.fetch_enable_i) begin
+            fetch_enable_i_sticky <= 1;
+        end
+    end
     a_dmhaltaddr_stable : assert property (
-        (cov_assert_if.fetch_enable_i && cov_assert_if.rst_ni)
+        fetch_enable_i_sticky
         |=>
         $stable(cov_assert_if.dm_halt_addr_i)
-        ) else `uvm_error(info_tag, "TODO");
+        ) else `uvm_error(info_tag, "dm_halt_addr_i changed after fetch_enable_i");
 
     // Should be word-aligned
     a_dmhaltaddr_aligned : assert property (


cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -543,12 +543,12 @@ module uvmt_cv32e40x_debug_assert
     a_dmhaltaddr_stable : assert property (
         (cov_assert_if.fetch_enable_i && cov_assert_if.rst_ni)
         |=>
-        always $stable(cov_assert_if.dm_halt_addr_i)
+        $stable(cov_assert_if.dm_halt_addr_i)
```
silabs-robin, line 546: I am not sure if this change is alright or not...

First, I think dsim spawns one thread for each cycle where the antecedent is fulfilled, and all of these threads run forever until the test exits. That is what I suspect.

Secondly, I am sceptical to whether the semantics of the assertion should be changed like that.
The user manual says _"Do not change after enabling core via fetch_enable_i"_. That is why I don't think it is sufficient with only `|=> $stable()`, and is why I added `always`.

Thirdly, I think some helper logic might solve this. We could maybe have the consequent without "always", as you propose (`|=> $stable(...)`), but change the antecedent to use a "sticky" register.
```
    logic fetch_enable_i_sticky;
    always @(posedge cov_assert_if.clk_i or negedge cov_assert_if.rst_ni) begin
        if (!cov_assert_if.rst_ni) begin
            fetch_enable_i_sticky <= 0;
        end else if (cov_assert_if.fetch_enable_i) begin
            fetch_enable_i_sticky <= 1;
        end
    end
    a_dmhaltaddr_stable : assert property (
        fetch_enable_i_sticky
        |=>
        $stable(cov_assert_if.dm_halt_addr_i)
        ) else `uvm_error(info_tag, "dm_halt_addr_i changed after fetch_enable_i");
```

silabs-robin: (I think dsim could maybe fix that by not starting a new thread each time, but rather check if another "long-running checker" is already checking the same thing?)

MikeOpenHWGroup: > First, I think dsim spawns one thread for each cycle where the antecedent is fulfilled, and all of these threads run forever until the test exits.

Possibly.  I would expect to see an ever increasing memory consumption by the dsim process and this has not been observed.   

> Secondly, I am skeptical to whether the semantics of the assertion should be changed like that.
The user manual says "Do not change after enabling core via fetch_enable_i". That is why I don't think it is sufficient with only |=> $stable(), and is why I added always.

If that is the case, then I am not sure property `a_dmhaltaddr_stable` does what we want.  Non-overlapping implications (`|=>`) define synchronous properties.

> Thirdly, I think some helper logic might solve this.

Good idea.  SVA code is powerful and very dense, and "helper logic" is a great way to clarify intent (at the expense of more code).

> I think dsim could maybe fix that...

Agreed.  An issue has been opened Metrics on this. Let's see what they respond with before we do anything.

MikeOpenHWGroup:
Hi @silabs-robin.  You are right on both fronts.  The dsim developers at Metrics agree with your interpretation of the assertion semmantics and recognize the excessive runtime as a bug in their tool.  They recommended a workaround that is virtually identical to the one you proposed.  I will push that in shortly.

Just so we have a record of this, below is the response from Metrics (their ticket  number 1149):
```
R&D has analyzed the problem and has confirmed that the poor performance is due to the way SVAs are implemented in DSim. We have ideas about how to improve performance in these circumstances but it's a major undertaking. The task has been added to our backlog.

Here is some advice from R&D about this particular situation:

The code is LRM compliant, and the "always" in the assertion is not redundant, but actually changes the meaning of the assertion.

With the always, it means "check that this variable is stable FOR THE REST OF THE SIMULATION". So the assertion evaluation will only ever succeed at the end of the simulation. Without the always, the variable is checked only on the next clock, then the assertion's evaluation ends.

To rewrite the assertion to mean the same thing but without the negative performance, you could introduce a variable that turns to true the first time the antecedent is true, then assert newVar |=> $stable(..).

Currently DSim should issue the warning:
=W:[ExcessiveAssertionsAtEnd]
at the end of the simulation, to direct the user's attention to this assertion and its behaviour.
```

MikeOpenHWGroup:
Hi @silabs-robin, please have another look at this one.  I've pushed-in an update which I believe implements a workaround and it passes my local regressions.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -540,15 +540,23 @@ module uvmt_cv32e40x_debug_assert
...
+        if (!cov_assert_if.rst_ni) begin
+            fetch_enable_i_sticky <= 0;
+        end else if (cov_assert_if.fetch_enable_i) begin
+            fetch_enable_i_sticky <= 1;
+        end
+    end
     a_dmhaltaddr_stable : assert property (
-        (cov_assert_if.fetch_enable_i && cov_assert_if.rst_ni)
+        fetch_enable_i_sticky
         |=>
```
silabs-robin, line 553: Come to think of it...
Shouldn't this `|=>` be turned into a `|->` now?
Previously there was `fetch_enable |=> stable` (check 1 cycle later), but the new `fetch_enable_i_sticky <= 1` is already a registered signal (assigned 1 cycle later). Using both the registered sticky signal and also `|=>` will do the check 2 cycles later (1 cycle too much)?

MikeOpenHWGroup: @silabs-robin, once again you are correct.  I was convinced by your argument, but needed to conduct an experiment to be sure, so I put this assertion into a standalone module and ran a couple of tests.

If **fetch_enable_i** is asserted one cycle before **dm_halt_addr**, then `fetch_enable |=> stable` passes, bit `fetch_enable |-> stable` fails (as required by the spec).   This is a great example of an "off-by-one-cycle bug".

I will push in an update to replace the nonoverlapping sequential implication `|=>` with the overlapping sequential implication `|->`.

Good catch!

MikeOpenHWGroup:
Thanks for your help with this @silabs-robin.   Much appreciated.

PR #1038, opened 2021-11-26
Implement the cva6 configurations defined in cva6 specification

JeanRochCoulon:
Implement the cva6 configurations defined in cva6 specification
target variable can take cv64a6_imacfd_sv39, cv32a6_imac_sv0,
                                       cv32a6_imac_sv32, cv32a6_imacf_sv32


cva6/sim/cva6.yaml
```diff
@@ -50,11 +50,11 @@
   tool_path: SPIKE_PATH
   tb_path: TB_PATH
   precmd: >
-    make -C <path_var> verilate variant=<variant> user-define="+define+WT_CACHE"
+    make -C <path_var> verilate target=<target> user-define="+define+WT_CACHE"
   cmd: >
     <path_var>/work-ver/Variane_testharness +time_out=40000000 +debug_disable=1 <elf>
   postcmd: >
-    <tool_path>/spike-dasm < ./trace_rvfi_hart_00.dasm > log
+    grep "ffffffe0005e5cd4" ./trace_rvfi_hart_00.dasm
```
MikeOpenHWGroup, line 57: I call these "magic numbers", but they work without explanation.  We should either not use them, or use variable names that have some meaning.  Also, it is likely this number must match another number in a separate file.   It would be useful if this file and the others that need this magic number to obtain its value from a common source.

MikeOpenHWGroup (commented):
Looks good @JeanRochCoulon.  I think that @ASintzoff should review and approve this PR before I merge it.  Also, I have one question about `cva6/sim/cva6.yaml`.

cva6/sim/cva6.py
```diff
@@ -861,7 +857,19 @@ def load_config(args, cwd):
...
+      args.isa  = "rv64gc"
+    elif args.target == "cv32a6_imac_sv0":
+      args.mabi = "ilp32"
+      args.isa  = "rv32imac"
+    elif args.target == "cv32a6_imac_sv32":
+      args.mabi = "ilp32"
+      args.isa  = "rv32imac"
+    elif args.target == "cv32a6_imafc_sv32":
+      args.mabi = "ilp32"
+      args.isa  = "rv32imacf"
```
ASintzoff, line 871: Please use canonical order for `args.isa` (`rv32imafc`) to be sure compilers will recognize correctly the extensions.

cva6/sim/cva6.yaml
JeanRochCoulon: @MikeOpenHWGroup , This modification has been removed from the current PR. New #1039 PR contains it, with the explanation of "magic numbers"

cva6/sim/cva6.py
JeanRochCoulon: Corrected. Thanks

ASintzoff (approved):

PR #1044, opened 2021-12-03
bus error test fixes and incorporate multi-bit traps

strichmo:
The new RVFI multi-bit trap (as described in the E40X User Manual) is now integrated into the testbench.
The 4 random bus error tests are now mostly passing.
The OBI scoreboard is broken for some tests, but those will be described in an issue to be fixed by @silabs-robin later.
Advance the RTL hash to the latest E40X master.


MikeOpenHWGroup (approved):
Looks good @strichmo, thanks.   (More trap bits, oh boy!)

Passes a local regression.

PR #1047, opened 2021-12-06
Cv32e40x/merge dev to rel 54

strichmo:
Merge passes Xcelium.


MikeOpenHWGroup (approved):
Thanks @strichmo, looks good.   Most of these I have seen before as pull-requests to `dev`.  Will merge once a local regression passes.

PR #1051, opened 2021-12-07
fix issue 1046

strichmo:
cleanup all signaling of the NMI faults to use RVFI CSR mcause signals in lieu of address matches


MikeOpenHWGroup (commented):
Looks good @strichmo.   I will not merge until @silabs-robin has had a chance to review and my local regression passes.

silabs-robin (approved):
Hi. Sorry for late reply, I got swamped.
This PR looks good to me.

It seems most of it is about replacing "nmi" with "nmi_load"/"nmi_store".
That one if-else that I raised an issue about is fixed and looks more like I'd expect it to.
I guess "bus_fault" is referring to instr-side bus fault which cause exceptions, while "nmi_load/store_fault" refers to data-side bus fault which cause non-maskable interrupts? That naming is slightly confusing to me, but it is ok.

Issue #1045, opened 2021-12-03
install latest GCC compiler on OpenHW Metrics cloud


strichmo:
### Task Outcome
The Metrics cloud is updated with an installation of a top-of-tree gcc that provides full basic support for the 1.0.0 Zba, Zbb, Zbc, and Zbs extensions.  The minimal level of support required is that the assembler and the binutils support all opcode pnemonics for the above extensions supported by the cv32e40x and cv32e40s.

When the tool is installed, the cv_regress defaults for the cv32e40x and cv32e40s should point to the GNU toolchain as its default.  This should be introduced via a PR to the dev branches for both cores.

The toolchain installation should be the latest top-of-tree GCC from the Embecosm Jenkins server:
https://www.embecosm.com/resources/tool-chain-downloads/#riscv

### Background information
The current _default_ toolchain for test development on CORE-V cores is the COREV toolchain provided by Embecosm.  This toolchain is a gcc-based toolchain that incorporates an up-to-date version of gcc with optimizations for COREV cores (e.g. PULP).  However maintaining the top-of-tree gcc to COREV is not a trivial process and thus the RISCV updates to gcc will lag in the COREV toolchain.  The Zb* extensions need to be supported now to test these instructions in the CV32E40X, as the instructions are already implemented.

The current top-of-tree gcc has the required level of support to facilitate random tests and some optimization in directed tests.
The earliest that the COREV toolchain may be updated with these changes is late Q1 of 2022.

### Location Information
The toolchain install directory of the Metrics cloud.

### Completion Criteria
Zb extension tests are added to ci_check/rel_check and these pass in the Metrics cloud.

MikeOpenHWGroup:
Hi @strichmo and @aimeepsutton, I have opened service ticket #1155 with Metrics for this task.

silabs-hfegran:
@MikeOpenHWGroup is this issue resolved?

MikeOpenHWGroup:
Yes, indeed.

PR #1050, opened 2021-12-06
alter the default toolchain in cv_regress per core, use gnu for e40x …

strichmo:
…and e40s due to Zb support

Signed-off-by: Steve Richmond <Steve.Richmond@silabs.com>

PR #1053, opened 2021-12-07
Update Metrics docker image

MikeOpenHWGroup:
Docker image with dsim version `20211122.2.0` and the `riscv32-embecosm-gcc-ubuntu1804-20211128` toolchain.

Should resolve issue #1045 which in turn unlocks pull-request #1050.

### DIFF
--- a/.metrics.json
+++ b/.metrics.json
@@ -7,109 +7,109 @@
     "list": [
       {
         "name":     "uvmt_cv32e40p",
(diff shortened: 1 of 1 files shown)


strichmo (approved):
Thanks for getting this updated.

silabs-robin (approved):
I don't have time at the moment to learn how this works. But since it seems to "just" be version numbers of some package config that are updated, then I don't object to merging.

PR #1054, opened 2021-12-08
Fixed rvvi memory backdoor for misaligned accesses to volatile memory

silabs-hfegran:
This commit addresses a regression error that is observed with the fence.i tests, where the ISS memory backdoor corrupts ISS memory due to erroneous handling of misaligned data addresses.

### DIFF
--- a/lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_drv.sv
+++ b/lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_drv.sv
@@ -143,6 +143,10 @@ endfunction : end_of_elaboration_phase
 
 task uvma_rvvi_ovpsim_drv_c::stepi(REQ req);
 
+   bit[31:0] mem_val_update = 0;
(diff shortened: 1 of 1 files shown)


silabs-robin:
This looks right to me.
Previously, the memory would get updated for address `mem_addr >> 2`, but that would update to a word-aligned address regardless of whether the access was actually word-aligned or not.

ps. I am still, in general, confused about our policy about the "volatile" memory regions, but regardless of that I think this PR correctly fixes a real problem.

MikeOpenHWGroup:
Hi @silabs-hfegran, this PR looks good but git seems to think there is a conflict.  Can you have a look?

silabs-hfegran:
> Hi @silabs-hfegran, this PR looks good but git seems to think there is a conflict. Can you have a look?

Hi @MikeOpenHWGroup , 
the merge conflict should be resolved now

MikeOpenHWGroup (approved):
Thanks for cleaning up the conflict @silabs-hfegran.  My local tests pass so this is good-to-go.

PR #1056, opened 2021-12-08
Use recent CVA6 version

ASintzoff:
To ensure people will use a recent CVA6 version when executing scripts (cva6/regress/) from core-v-verif, update the default value of CVA6_HASH variable.

### DIFF
--- a/cva6/README.md
+++ b/cva6/README.md
@@ -35,8 +35,8 @@ Run one of the shell scripts:
 - `source cva6/regress/dv-riscv-tests.sh`:
 [riscv-tests](https://github.com/riscv/riscv-tests) test suite.
 
-These tests are using [riscv-dv](https://github.com/ThalesGroup/riscv-dv)
-as environment (forked from https://github.com/google/riscv-dv).
+These tests are using [riscv-dv](https://github.com/google/riscv-dv)
+as environment.
 
 ## Environment variables
 Other environment variables can be set to overload default values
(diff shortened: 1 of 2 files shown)


cva6/README.md
```diff
@@ -51,7 +51,7 @@ The default values are:
 
 - `CVA6_REPO`: `https://github.com/openhwgroup/cva6.git`
 - `CVA6_BRANCH`: `cva6_reorg`
-- `CVA6_HASH`: `95b1070a76d2399ff4a95af70082514b0d0c5743`
+- `CVA6_HASH`: `bb9821d85f88d4e615bfa9bb35c9a78ed6f0b579`
```
MikeOpenHWGroup, line 54: Hi @ASintzoff, I suggest that we do not place these specific hash values in the README.  This creates a maintenance issue since both the README and the install script need to be kept consistent.  The README should indicate that a specific hash can be selected with an environment variable and that the install script sets the default.

MikeOpenHWGroup (changes requested):

cva6/README.md
ASintzoff: Good point. I just force-pushed a fixed version of the README.

MikeOpenHWGroup (approved):

cva6/README.md
MikeOpenHWGroup: Thanks!

PR #1059, opened 2021-12-09
Fixed PMA region configs wrong order

silabs-hfegran:
Coverage analysis revealed a problem with the parsing of the PMA configs (parsed in reverse order), leading the configs to not capture the original test intent. This introduced the need for bug fixes in the linker and corev-dv memory region generation as well. 

Additionally, some regions were changed to main memory to fix lacking coverage.

### DIFF
--- a/cv32e40x/env/corev-dv/ldgen/cv32e40x_ldgen.sv
+++ b/cv32e40x/env/corev-dv/ldgen/cv32e40x_ldgen.sv
@@ -348,7 +348,7 @@ function void cv32e40x_ldgen_c::create_pma_section_file(string filepath);
         $fdisplay(fhandle_pma, { indent(L1), ".region_", $sformatf("%0d %0s", i, section_location), ":" });
         $fdisplay(fhandle_pma, { indent(L1), "{" });
         $fdisplay(fhandle_pma, { indent(L2), "KEEP(*(.region_", $sformatf("%0d", i), "));" });
-        $fdisplay(fhandle_pma, { indent(L1), "} > region_", $sformatf("%0d", i) });
+        $fdisplay(fhandle_pma, { indent(L1), "}"});
       end
     end
     $fdisplay(fhandle_pma, "}");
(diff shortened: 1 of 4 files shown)


MikeOpenHWGroup:
Hi @silabs-hfegran, this looks fine to me but I think one of @strichmo or @silabs-robin should approve this before merging.  In the meantime I have a couple of comments:

1. I am not sure my understanding of how PMA region addressing works is correct. I have posed a question in #1060 about it.
2. There are a _lot_ of magic numbers in this code and this pull-request adds a lot more.  It would be handy to have these defined as localparams (or equivalent) with human-readable-names.

silabs-hfegran:
> Hi @silabs-hfegran, this looks fine to me but I think one of @strichmo or @silabs-robin should approve this before merging. In the meantime I have a couple of comments:
> 
> 1. I am not sure my understanding of how PMA region addressing works is correct. I have posed a question in [Question about PMA region configs #1060](https://github.com/openhwgroup/core-v-verif/issues/1060) about it.
> 2. There are a _lot_ of magic numbers in this code and this pull-request adds a lot more.  It would be handy to have these defined as localparams (or equivalent) with human-readable-names.

1. I will get back to you under question #1060

2. The only new magic numbers introduced in this commit is the addresses in the yaml-config, set to be within an executable region. Previously the order of the regions were reversed, and the address was in what was then an executable region. The only thing that changed with the pma-configs themselves is a reversing of the indexing order (both in the array declaration, and the order of the items themselves, as the order observed in simulation did not match what was expected). I verified that the order appears correct under simulation now and should better cover the cases where we have overlapping regions where priority has to be taken into account.

silabs-hfegran:
As for the addresses themselves, these are somewhat arbitrary addresses set up to attempt to get decent coverage of the different possible PMA configurations with a minimal amount of test configurations. e.g. nested overlapping regions, full memory space, regions non-overlapping with no directly adjacent regions, non-overlapping with adjacent regions and so on.

strichmo (approved):

silabs-robin (approved):
I have read through and discussed my questions with Henrik in DMs.
All the changes seem logical to me.

PR #1066, opened 2021-12-10
Add CSR access test for Xsecure CSRs

silabs-robin:
Just simple read/write/set/clear accesses to {cpuctrl, secureseed0, secureseed1, secureseed2}.
This test failed before ISS support was added, but passes after the fixes made in https://github.com/openhwgroup/core-v-verif/pull/1061 .


MikeOpenHWGroup (approved):
Thanks @silabs-robin, this looks good.  I'm approving it on inspection only as this new test is not part of the cv32e40s-ci-check-dev regression, so it cannot break the build, even f the test fails.

PR #1067, opened 2021-12-10
Fixed JAL instruction in corev-dv fence.i test

silabs-hfegran:
JAL takes symbol, not a number-immediate. In the test,
linker failures occurred in cases when the store
fence.i exec stream randomization inserted a JAL instruction.

The instruction in question is supposed to never execute,
so the jump symbol is set to be at the beginning of the stream,
thus intentionally creating an infinite loop that will never
be entered.

Additionally, compiler guards were added around the instructions
that are being intentionally overwritten, to avoid potential 
problems with alignment and compressed instructions. 

### DIFF
--- a/cv32e40x/env/corev-dv/cv32e40x_fencei_instr_lib.sv
+++ b/cv32e40x/env/corev-dv/cv32e40x_fencei_instr_lib.sv
@@ -115,6 +115,7 @@ class corev_store_fencei_exec_instr_stream extends riscv_load_store_rand_instr_s
   function void post_randomize();
     riscv_instr        instr;
     riscv_pseudo_instr pseudo;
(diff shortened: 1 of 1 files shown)


silabs-hfegran:
This needs to be updated with the same changes for branch-instructions as well

silabs-robin:
Thanks for figuring this out and fixing it.
The test was difficult to write in the first place because of the interplay between dv code and the compiler. But you seem to have sorted out these two remnant issues.

silabs-hfegran:
@MikeOpenHWGroup This commit fixes compile errors for fence.i-tests with pma_test_config_3 following the pma-config changes.

MikeOpenHWGroup (approved):

Issue #926, opened 2021-10-08
Use of $urandom()


MikeOpenHWGroup:
In core-v-verif there are 8 files that use some variation of `$urandom()`.  You are in the "assignees" list of this issue because git believes you authored one or more of these.

The Verissimo linter flags these as violation [SVTB.29.1.3.1](https://dvteclipse.com/core5verif-verissimo/1/reports/run/1f430b7d805e2459f01c46e6ba58ef1dcbe59419/index.html?N4IgzgrgtlCGBOBPEAuAZrANmApgGhAGMALHQgazABV4ccBlHTMgFwEsB7AO1QFoBGAiTKUadRs0LtuVRAAccqAEwFibFtVgAjZhNaceKEAFYAbCFXrNOnAHl4AExzxUAbVeCQsMIRABdPFcABgJvXz8AkDUNKm1dHAQSVBALECcfAAVYLiY9KQNZBVQQ8BZYFggwADE2TBZnADUsCEUUT2EKatr6-CbMFrcQegaqACEAOiUATnH-cYBmWf8CNFqcLrrG5tbXSMIOKDkEHAAlHDkOeBYazd7t5NT9w-OAYXh1ZzZYG56-gZRdkJaOUDAARco4H5bfqtFKhLhcDhlaRcKF3GFgNy7PwAXyAA), which is a lint error because class members randomized with these calls cannot be constrained.  I also worry that invocations of $urandom() may impair random stability (that is, the ability to reproduce results from the same seed).

### Possible resolutions:
1. Accept the use of $urandom() and create a project-wide waiver for SVTB.29.1.3.1.
2. Enforce the use of a configuration or context class of random members to implement the same functionality.
3. Waive, or not, on a case-by-case basis.

### Example fixes
From [uvma_obi_memory_drv.sv](https://github.com/openhwgroup/core-v-verif/blob/1f430b7d805e2459f01c46e6ba58ef1dcbe59419/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv#L669).  Replace this:
```
UVMA_OBI_MEMORY_DRV_IDLE_RANDOM: begin
   `uvm_info("OBI_MEMORY_DRV", $sformatf("Invalid drv_random: %0d", cfg.drv_idle), UVM_NONE)
   slv_mp.drv_slv_cb.rdata <= $urandom();
   slv_mp.drv_slv_cb.err   <= $urandom();
[...]

PR #1072, opened 2021-12-15
Address multiple lint errors

MikeOpenHWGroup:
Hi @silabs-hfegran, this is a big PR in terms of files, but technically a bit of a softball.  All changes are one of three categories:
1. Fixing lint errors as reported by Verissimo.
2. Removal of unnecessary white-space (continuing an on-going effort started by Steve).
3. Update to the Coding Style guidelines as per the discussion in #926.

### DIFF
--- a/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
+++ b/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
@@ -138,7 +138,7 @@ task uvmt_cv32e40p_firmware_test_c::reset_debug();
     @(negedge env_cntxt.clknrst_cntxt.vif.reset_n);
 
     if (!debug_vseq.randomize()) begin
-        `uvm_error("TEST", "Cannot randomize the debug sequence!")
+        `uvm_fatal("TEST", "Cannot randomize the debug sequence!")
     end
     debug_vseq.start(vsequencer);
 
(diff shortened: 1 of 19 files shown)


cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
```diff
@@ -142,7 +137,9 @@ task uvmt_cv32e40p_firmware_test_c::reset_debug();
     `uvm_info("TEST", "Applying debug_req_i at reset", UVM_NONE);
     @(negedge env_cntxt.clknrst_cntxt.vif.reset_n);
 
-    void'(debug_vseq.randomize());
+    if (!debug_vseq.randomize()) begin
+        `uvm_error("TEST", "Cannot randomize the debug sequence!")
```
silabs-hfegran, line 141: Is uvm_error strong enough in case of randomization failures? (This goes for the other randomizations in this review as well)

I think it may be an idea to decide on _error or _fatal in the guidelines for these cases

silabs-hfegran:
Thanks Mike, in general I think this looks good - the only remark I have is with regards to the error-condition for randomization as I mentioned earlier

cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
MikeOpenHWGroup: Good point.  In general I tend to avoid fatal-ing out of a simulation as I've seen it overused in a past life.  Fortunately, the team on core-v-verif does not do that.  In the case of randomization failures you are right that an error is not strong enough - if randomization fails it either indicates a simple, but catastrophic, bug or (worse) a systematic problem in the environment.  Both need immediate attention so I will push in an update to this PR to call `uvm_fatal()` on randomization errors.

I'll also create a discussion thread on handing randomization failures in a seperate GitHub issue.

MikeOpenHWGroup:
Hi @silabs-hfegran, I've pushed in a set up updates to fatal out when randomization fails.  Thanks again for pointing that out.

After thinking about it some more, I don't think I'll bother opening an issue to discuss it.  What you said makes sense and we should follow that.  So I took the liberty of updating the Coding Style Guidelines as part of this PR ([link](https://github.com/MikeOpenHWGroup/core-v-verif/blob/cv32e40p/dev/docs/CodingStyleGuidelines.md?plain=1#L826)).

silabs-hfegran (approved):

PR #1074, opened 2021-12-15
Fixes for isacov

silabs-robin:
This PR fixes several issues in isacov, particularly for RVC.

Some of the fixes are minor mistakes like using wrong "imm" for a particular instruction format.
A big part of this PR is adding support for RVC; this was fairly incomplete before.
This is not a complete cleanup; I created an issue to document what more I think could be done https://github.com/openhwgroup/core-v-verif/issues/1076.

The biggest idea I try to introduce here is to centralize how information about an instruction is obtained. Instead of each covergroup specifying what to sample (which has shown to be error prone), I would rather have functions in the instr class handle that in one single place and let all covergroups use this single source of truth (only implemented for RVC for now).
By using the "2string" function and logging, one can validate that these functions give correct results by comparing to the rvfi log (I relighted the old isacov_logdiff script and it helped out a lot). Sadly though, some function (e.g. "value_type") are only used by covergroups and can't(?) be validated by the logdiff.
One big challenge here is that RVC, in a way, implicitly has more instruction formats than it claims to have.

----

ps.
I have an idea that i temporarily call "annotated pull request" that I want to try out. If I comment on the diff, then it might help review. Maybe it causes more noise than value? Since this PR is a bit complicated I want to at least try it out now, and you all can give me feedback on whether my added comments where usefull or not.

pps.
Actually, I wonder if isacov could be simplified by doing more of these look up tables and flags.
Could there be "one covergroup to rule them all" if we had functions/LUTs and flags to select applicable cps and bins. Or would that become more unclear instead of being more clean. That is not something to decide upon now, but I must air the idea while I have it.


lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -812,7 +812,7 @@ covergroup cg_btype(
     ignore_bins IGN_OFF = cross_rs1_rs2 with (!reg_crosses_enabled);
   }
 
-  cp_imms_value: coverpoint instr.imms_value_type {
+  cp_immb_value: coverpoint instr.immb_value_type {
```
silabs-robin, line 815: B-type covergroup should use B-type immediate, not S-type.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -855,7 +855,7 @@ covergroup cg_jtype(
 
   cp_rd: coverpoint instr.rd;
 
-  cp_immj_value: coverpoint instr.immu_value_type {
+  cp_immj_value: coverpoint instr.immj_value_type {
```
silabs-robin, line 858: J-type cg should use J-type imm, not U-type.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -997,37 +997,43 @@ covergroup cg_ci(
...
     uvma_isacov_instr_c instr
 );
   option.per_instance = 1;
   option.name = name;
 
   cp_rs1_value: coverpoint instr.rs1_value_type {
-    ignore_bins POS_OFF = {POSITIVE} with (!rs1_is_signed);
-    ignore_bins NEG_OFF = {NEGATIVE} with (!rs1_is_signed);
-    ignore_bins NON_ZERO_OFF = {NON_ZERO} with (rs1_is_signed);
+    ignore_bins  OFF     = cp_rs1_value    with (!has_rs1);
```
silabs-robin, line 1011: Some RVC CI-type instructions don't use the `rs1` of the `rd/rs1` field and would lack coverage if we try to cover it.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -997,37 +997,43 @@ covergroup cg_ci(
...
+    illegal_bins POS_OFF = {POSITIVE}      with (!rs1_is_signed);
+    illegal_bins NEG_OFF = {NEGATIVE}      with (!rs1_is_signed);
+    illegal_bins NON_ZERO_OFF = {NON_ZERO} with ( rs1_is_signed);
```
silabs-robin, line 1014: Made these `illegal` because it would otherwise indicate that sampling of coverage has a bug.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -997,37 +997,43 @@ covergroup cg_ci(
...
   }
 
   cp_imm_value: coverpoint instr.c_imm_value_type {
-    ignore_bins POS_OFF = {POSITIVE} with (!imm_is_signed);
-    ignore_bins NEG_OFF = {NEGATIVE} with (!imm_is_signed);
-    ignore_bins NON_ZERO_OFF = {NON_ZERO} with (imm_is_signed);
+    illegal_bins POS_OFF      = {POSITIVE} with (!imm_is_signed);
+    illegal_bins NEG_OFF      = {NEGATIVE} with (!imm_is_signed);
+    illegal_bins NON_ZERO_OFF = {NON_ZERO} with (imm_is_signed);
+    ignore_bins  ZERO_OFF     = {ZERO}     with (imm_is_nonzero);  // Not illegal, because of HINT instrs
```
silabs-robin, line 1021: Some RVC instructions have `nzimm` or `nzuimm` immediates which (except for HINT instructions) should not be zero. Otherwise, would have 0 coverage on `ZERO`.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -997,37 +997,43 @@ covergroup cg_ci(
...
-    ignore_bins NON_ZERO_OFF = {NON_ZERO} with (rd_is_signed);
+    illegal_bins POS_OFF = {POSITIVE} with (!rd_is_signed);
+    illegal_bins NEG_OFF = {NEGATIVE} with (!rd_is_signed);
+    illegal_bins NON_ZERO_OFF = {NON_ZERO} with (rd_is_signed);
   }
 
   cp_rdrs1: coverpoint instr.c_rdrs1 {
-    ignore_bins RD_NOT_ZERO = {0};
+    illegal_bins RD_NOT_ZERO = {0};
+    illegal_bins NON_X2      = cp_rdrs1 with ((item != 2) && tie_rdrs1_x2);
```
silabs-robin, line 1032: `c.addi16sp`. Cannot cover anything but `x2`.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -997,37 +997,43 @@ covergroup cg_ci(
...
   }
 
   `ISACOV_CP_BITWISE(cp_rd_toggle, instr.rd_value, 1)
-  `ISACOV_CP_BITWISE_5_0(cp_imm_toggle, instr.c_imm, 1)
+  `ISACOV_CP_BITWISE_4_0(cp_imm_toggle, instr.get_field_imm(), 1)
```
silabs-robin, line 1036: Throughout this diff, occurances of `c_imm` is replaced with `get_field_imm()`, because `c_imm` was never implemented correctly and the function call should pick the right one.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1097,7 +1103,8 @@ covergroup cg_ci_lui(
   }
 
   `ISACOV_CP_BITWISE_31_12(cp_rd_toggle, instr.rd_value, 1)
-  `ISACOV_CP_BITWISE_17_12(cp_imm_toggle, instr.c_imm, 1)
+  `ISACOV_CP_BITWISE_17_12(cp_imm_toggle, instr.get_field_imm(), 1)
+  // TODO:ropeders the toggle macros are wrong; must revise
```
silabs-robin, line 1107: `c_imm` was never right, so these toggles never worked.
`get_field_imm()` should now be right, but all the toggles have wrong indexes now (e.g. `_17_12` here).

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1811,9 +1818,9 @@ function void uvma_isacov_cov_model_c::build_phase(uvm_phase phase);
       rv32i_sll_cg    = new("rv32i_sll_cg",
                             .reg_crosses_enabled(cfg.reg_crosses_enabled),
                             .reg_hazards_enabled(cfg.reg_hazards_enabled),
-                            .rs1_is_signed(rs1_is_signed[SUB]),
-                            .rs2_is_signed(rs2_is_signed[SUB]),
-                            .rd_is_signed(rd_is_signed[SUB]));
+                            .rs1_is_signed(rs1_is_signed[SLL]),
+                            .rs2_is_signed(rs2_is_signed[SLL]),
+                            .rd_is_signed(rd_is_signed[SLL]));
```
silabs-robin, line 1823: This looks like a copy-paste error. `sll_cg` should use `SLL`.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -2418,6 +2434,13 @@ function void uvma_isacov_cov_model_c::sample (uvma_isacov_instr_c instr);
...
       DRET:   rv32i_dret_cg.sample(instr);
+
+      default: have_sampled = 0;
+    endcase
+  end else if (!have_sampled && instr.trap && cfg.core_cfg.ext_i_supported) begin
+    have_sampled = 1;
+    case (instr.name)
+      // Ecall and ebreak will trap
       ECALL:  rv32i_ecall_cg.sample(instr);
       EBREAK: rv32i_ebreak_cg.sample(instr);
```
silabs-robin, line 2445: Ebreak instructions are special. We normally don't sample `trap`ed instructions, because they usually contain bogus data. But ebreak/ecall trap by definition and we still want to sample them.

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -119,25 +117,30 @@ class uvma_isacov_instr_c#(int ILEN=DEFAULT_ILEN,
...
 
   string instr_str;
 
   // Printing based on instruction format type
-  if (name inside {LW, LH, LB, LHU, LBU}) begin
-    instr_str = $sformatf("x%0d, %0d(x%0d)", rd, $signed(immi), rs1);
-  end
-  if (name inside {SLLI, SRLI, SRAI}) begin
-    instr_str = $sformatf("x%0d, x%0d, 0x%0x", rd, rs1, rs2);
-  end
```
silabs-robin, line 140: In the past, the return logic of this function was changed, but that left these if-statements in the wrong priority order.

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -148,60 +151,95 @@ function string uvma_isacov_instr_c::convert2string();
     instr_str = $sformatf("x%0d, %0d(x%0d)",  rs2, $signed(imms), rs1);
   end
   if (itype == B_TYPE) begin
-    instr_str = $sformatf("x%0d, x%0d, %0d",  rs1, rs2, $signed({immb, 1'b0}));
+    instr_str = $sformatf("x%0d, x%0d, %0x",  rs1, rs2, ($signed(rvfi.pc_rdata) + $signed({immb, 1'b0})));
```
silabs-robin, line 154: This change, and the following, fixes the 2string formatting to conform with objdump.

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -148,60 +151,95 @@ function string uvma_isacov_instr_c::convert2string();
...
   end
   if (itype == CB_TYPE) begin
-    instr_str = $sformatf("x%0d, x%0d",  rd, rs2);
+    instr_str = $sformatf("x%0d, %0x", this.get_addr_rs1(), ($signed(rvfi.pc_rdata) + this.get_data_imm()));
   end
   if (itype == CJ_TYPE) begin
-    instr_str = $sformatf("x%0d",  c_imm);
+    instr_str = $sformatf("%0x", ($signed(rvfi.pc_rdata) + this.get_data_imm()));
+  end
+  // Special printing for a select few instructions:
```
silabs-robin, line 195: Sadly, there are idiosyncracies for certain asm formattings.  (Maybe there are more common denominators that could be exploited here.)

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -239,6 +277,7 @@ function void uvma_isacov_instr_c::set_valid_flags();
   end
 
   if (itype == CI_TYPE) begin
+    rs1_valid = 1;
     rd_valid = 1;
     return;
```
silabs-robin, line 282: This... is just RVC being inconsistent. In effect, there are more than the 9 instruction formats claimed by the spec.

mk/uvmt/uvmt.mk
```diff
@@ -320,8 +320,8 @@ embench: $(EMBENCH_PKG)
 #   Compare the log against the tracer log.
 #   This checks that sampling went correctly without false positives/negatives.
 
-ISACOV_LOGDIR = $($(SIMULATOR_UC)_RESULTS)/$(CFG)/$(TEST)_$(RUN_INDEX)
-ISACOV_TRACELOG = $(ISACOV_LOGDIR)/trace_core_00000000.log
+ISACOV_LOGDIR = $(SIM_CFG_RESULTS)/$(TEST)/$(RUN_INDEX)
+ISACOV_TRACELOG = $(ISACOV_LOGDIR)/uvm_test_top.env.rvfi_agent.trn.log
 ISACOV_AGENTLOG = $(ISACOV_LOGDIR)/uvm_test_top.env.isacov_agent.trn.log
 
 isacov_logdiff:
```
silabs-robin, line 327: I updated the isacov_logdiff script, now that the old "tracer" has been replaced with the rvfi log.
This has been really useful in checking that coverage is done right, as it uncovered many mistakes. (Using "illegal_bins" also helped in this regard.)
Idea: Could run a "full" regression, but substitute the `test` make-target with `test isacov_logdiff` to check for each test that coverage was sampled correctly. (One blocker is that HINT instructions are not shown with objdump info in the rvfi log.)

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1111,22 +1118,22 @@ covergroup cg_css(
...
+  cp_rs2: coverpoint instr.rs2;
+
   cp_rs2_value: coverpoint instr.rs2_value_type {
     ignore_bins POS_OFF = {POSITIVE} with (!rs2_is_signed);
     ignore_bins NEG_OFF = {NEGATIVE} with (!rs2_is_signed);
     ignore_bins NON_ZERO_OFF = {NON_ZERO} with (rs2_is_signed);
   }
 
-  cp_imm_value: coverpoint instr.c_imm_value_type {
+  cp_imm_value: coverpoint instr.get_imm_value_type() {
```
silabs-robin, line 1129: This is wrong. I'll fix it! But I'd like to wait for more feedback to queue up first.
(Unlike the other new functions, this one was just conceptualized but not implemented.)

silabs-hfegran: I think if something is a known error, then that warrants a comment in the committed code itself - otherwise this may drown in future changes.

mk/uvmt/uvmt.mk
silabs-hfegran: If feasible, I think it would be nice to have the run-index in the filename of the log itself, in case log files are moved around, it would still be obvious to which run they belong.

silabs-hfegran:
Thanks for these fixes @silabs-robin! 
@MikeOpenHWGroup, I would like some input from you on this topic as well - in this PR we have a lot of very small commits that all target fixes in the same general feature. I think maybe a squash here is warranted? Perhaps Robin's suggestion of an annotated review would go well with a squashed version of these commits; In that way the intent is still clear, while we avoid cluttering the git history with too many small commits.  

On the other hand - Having the entire change history is always nice - but as we do not have a time-correlated, linear history on the dev and master branches, this can get somewhat unwieldy. A certain commit does not represent the state of dev/master at any given time, as PRs that are not rebased onto latest before merge backdates git history.

silabs-hfegran:
> pps. Actually, I wonder if isacov could be simplified by doing more of these look up tables and flags. Could there be "one covergroup to rule them all" if we had functions/LUTs and flags to select applicable cps and bins. Or would that become more unclear instead of being more clean. That is not something to decide upon now, but I must air the idea while I have it.

I like this idea - the isacov as of now is somewhat maintenance-intensive

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
```diff
@@ -217,6 +217,9 @@ function string uvma_isacov_instr_c::convert2string();
   if (name == C_ANDI) begin
     instr_str = $sformatf("x%0d, %0d", this.get_addr_rd(), $signed(this.get_data_imm()));
   end
+  if (name == FENCE) begin
+    instr_str = "iorw, iorw";  // Note: This is technically not always true
```
silabs-hfegran, line 221: "This is technically not always true" - I think this comment warrants a more extensive explanation, as it is not obvious to me why.

mk/uvmt/uvmt.mk
```diff
@@ -331,11 +331,15 @@ isacov_logdiff:
 		@ls $(ISACOV_TRACELOG) > /dev/null
 		@ls $(ISACOV_AGENTLOG) > /dev/null
 	@echo filtering logs...
-		@cat $(ISACOV_TRACELOG) \
-			| sed 's/\(.*\)   \(.*\)/\1/' | awk '{$$1=$$2=$$3=$$4=$$5=""; $$0=$$0; $$1=$$1; print $$0}' \
-			| tail -n +2 > trace.tmp
+		@cat $(ISACOV_TRACELOG)              \
```
silabs-hfegran, line 334: Can we have a quick explanation of what the awk/sed-magic here does?

lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
silabs-robin: How about this?
```
// Note: If later found necessary, add support for `fence` arguments other than "iorw"
```
I don't think such support is needed at this time, and I don't think it needs to be a TODO because it would show in isacov_logdiff if this becomes necessary later.

mk/uvmt/uvmt.mk
silabs-robin: Fixed.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
silabs-robin: Fixed.

mk/uvmt/uvmt.mk
silabs-robin: While that might be the case, I don't think I'll dare touching that infrastructure as part of this PR.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
MikeOpenHWGroup: Steve was fond of "TODO" or "FIXME" comments.   Most editors will highlight such strings in a very annoying way to motivate us to "fix them one day".

silabs-hfegran (approved):

MikeOpenHWGroup:
Hi @silabs-hfegran, @silabs-robin.   A few points:

> > pps. Actually, I wonder if isacov could be simplified by doing more of these look up tables and flags. Could there be "one covergroup to rule them all" if we had functions/LUTs and flags to select applicable cps and bins. Or would that become more unclear instead of being more clean. That is not something to decide upon now, but I must air the idea while I have it.
> 
> I like this idea - the isacov as of now is somewhat maintenance-intensive

I like this idea as well.  @silabs-robin can you create a GitHub issue task for this?  (No good idea goes unpunished.)

> @MikeOpenHWGroup, I would like some input from you on this topic as well - in this PR we have a lot of very small commits that all target fixes in the same general feature. I think maybe a squash here is warranted? Perhaps Robin's suggestion of an annotated review would go well with a squashed version of these commits; In that way the intent is still clear, while we avoid cluttering the git history with too many small commits.

This is a good question.  I've moved it to MatterMost to solicit opinions.

silabs-robin:
> > > pps. Actually, I wonder if isacov could be simplified by doing more of these look up tables and flags. Could there be "one covergroup to rule them all" if we had functions/LUTs and flags to select applicable cps and bins. Or would that become more unclear instead of being more clean. That is not something to decide upon now, but I must air the idea while I have it.
> > 
> > 
> > I like this idea - the isacov as of now is somewhat maintenance-intensive
> 
> I like this idea as well. @silabs-robin can you create a GitHub issue task for this? (No good idea goes unpunished.)

Issue https://github.com/openhwgroup/core-v-verif/issues/1083 continues this idea of simplifying isacov cgs.

PR #1077, opened 2021-12-17
Added Zba,Zbb,Zbc,Zbs-extension support to all tests.

silabs-hfegran:
Added Zba,Zbb,Zbc,Zbs-extension support to all tests.
    
    Summary of changes:
      - Updated cv32e40x hash.
      - Updated riscv-dv hash (Now with support for bitmanip v1.0.0).
      - Updated supported march for corev-dv to include Zba,Zbb,Zbc,Zbs.
      - Added support for configuration-wide plusargs.
      - Updated test configurations to use bitmanip by default.
      - Added test configuration (no_bitmanip) without bitmanip support.
      - Expanded full regression to include a few non-bitmanip enabled tests.
      - Removed invocation of old Zbb extension in corev_rand_instr_test.
      - Removed duplicate pma-test entry in regression list.
      - Disabled DLCPTH-warning from xrun as it caused a lot of clutter in the log files.
      - Removed COREV=1 flag from two tests in regression that should have been removed previously.
      - Compliance test still uses the previous config (no_bitmanip) as I am not certain what changes are needed here.

### DIFF
--- a/cv32e40x/regress/cv32e40x_compliance.yaml
+++ b/cv32e40x/regress/cv32e40x_compliance.yaml
@@ -8,10 +8,12 @@ description: Runs all RISCV compliance tests on the CV32E40X
 builds:
   uvmt_cv32e40x:
     cmd: make comp
+    cfg: no_bitmanip
     dir: cv32e40x/sim/uvmt
 
   uvmt_cv32e40x_compliance_build:
     cmd: make all_compliance
+    cfg: no_bitmanip
     dir: cv32e40x/sim/uvmt
 
 # List of tests
--- a/cv32e40x/regress/cv32e40x_full.yaml
+++ b/cv32e40x/regress/cv32e40x_full.yaml
@@ -69,113 +69,136 @@ builds:
     cfg: pma_test_cfg_5
     dir: cv32e40x/sim/uvmt
 
+  uvmt_cv32e40x_no_bitmanip:
+    cmd: make comp_corev-dv comp
+    cfg: no_bitmanip
+    dir: cv32e40x/sim/uvmt
+
 # List of tests
 tests:
   hello-world:
     build: uvmt_cv32e40x
     description: uvm_hello_world_test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=hello-world
 
   csr_instructions:
     build: uvmt_cv32e40x
     description: CSR instruction test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=csr_instructions
 
   generic_exception_test:
     build: uvmt_cv32e40x
     description: Generic exception test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=generic_exception_test
 
   illegal_instr_test:
     build: uvmt_cv32e40x
     description: Illegal instruction test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=illegal_instr_test
 
   branch_zero:
     build: uvmt_cv32e40x
     description: Branch test with zero offsets
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=branch_zero
 
   cv32e40x_csr_access_test:
     build: uvmt_cv32e40x
     description: CSR Access Mode Test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=cv32e40x_csr_access_test
 
   cv32e40x_readonly_csr_access_test:
     build: uvmt_cv32e40x
     description: CSR Read-only Access Mode Test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=cv32e40x_readonly_csr_access_test
 
   requested_csr_por:
     build: uvmt_cv32e40x
     description: CSR PoR test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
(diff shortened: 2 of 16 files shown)


cv32e40x/regress/cv32e40x_compliance.yaml
```diff
@@ -8,10 +8,12 @@ description: Runs all RISCV compliance tests on the CV32E40X
 builds:
   uvmt_cv32e40x:
     cmd: make comp
+    cfg: no_bitmanip
     dir: cv32e40x/sim/uvmt
 
   uvmt_cv32e40x_compliance_build:
     cmd: make all_compliance
+    cfg: no_bitmanip
```
silabs-robin, line 16: AFAIK each compliance test targets specific ISA modules, and since they are written in assembly I would expect them (i.e. "imc" test etc) to not even use anything from B even if the compiler flag for B is on. But I don't know this for sure.

mk/uvmt/xrun.mk
```diff
@@ -206,6 +206,9 @@ XRUN_RUN_FLAGS  += -nowarn RNDNOXCEL
 # Probes
 XRUN_RUN_FLAGS  += -nowarn PRLDYN
 
+# Physical repository related to logical library in a cds.lib does not exist
+XRUN_COMP_FLAGS += -nowarn DLCPTH
```
silabs-robin, line 210: I don't know the causes or consequences of these warnings. Are we sure we want to have the warnings disabled by core-v-verif? I feel a bit uneasy about disabling a warning that I don't understand.

cv32e40x/regress/cv32e40x_full.yaml
```diff
@@ -69,213 +69,237 @@ builds:
...
+    cmd: make comp_corev-dv comp
+    cfg: no_bitmanip
+    dir: cv32e40x/sim/uvmt
+
 # List of tests
 tests:
   hello-world:
     build: uvmt_cv32e40x
     description: uvm_hello_world_test
+    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
```
silabs-robin, line 82: Why does it look like "no_bitmanip" is set for almost every test? Is the intention that all of these shall run with no_bitmanip, or is there a mechanism on top of this which ensures the separation of "daily" and "weekly" regressions?

cv32e40x/tests/cfg/b_ext_all.yaml
```diff
@@ -6,4 +6,8 @@ ovpsim: >
 #    --showoverrides
 #    --trace --tracechange --traceshowicount --monitornets
 cflags: >
+riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+llvm_march:  rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
```
silabs-robin, line 12: I see that "default.yaml" has `+enable_zba_extension=1`, but I don't see that here (nor in b_ext_abs.yaml). Is this right?

cv32e40x/tests/cfg/pma_test_cfg_5.yaml
```diff
@@ -2,9 +2,18 @@ name: pma_test_cfg_5
...
    +dm_halt_addr=0x301000
    +dm_exception_addr=0x301800
+   +enable_zba_extension=1
+   +enable_zbb_extension=1
+   +enable_zbc_extension=1
+   +enable_zbs_extension=1
+riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+llvm_march:  rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
```
silabs-robin, line 19: Does it exist any simple way to let e.g. PMA cfg yamls "include" B-extension cfg yamls? It seems like so many cfgs have have commonalities where it would be nice to just include this from a pre-defined "cfg module".

cv32e40x/regress/cv32e40x_full.yaml
silabs-hfegran: This means run default _and_ no_bitmanip

cv32e40x/tests/cfg/pma_test_cfg_5.yaml
silabs-hfegran: We currently do not have any way to include this in multiple configurations - previously this was on a per-test basis.

cv32e40x/tests/cfg/b_ext_all.yaml
silabs-hfegran: +enable_zbX_extension are directives for riscv-dv, the directed b-tests do not use riscv-dv and as such do not require these flags

mk/uvmt/xrun.mk
silabs-hfegran: Leaving this flag out causes _a lot_ of warnings about missing paths to physical technology-libraries, which are not used for rtl-level verification in core-v-verif, so I propose that we exclude this warning to have compile logs that are humanly readable without too much effort.

cv32e40x/regress/cv32e40x_compliance.yaml
silabs-hfegran: Yes, I would _think_ so, but I also think this should eventually be updated to include the bitmanip extensions.

cv32e40x/regress/cv32e40x_full.yaml
silabs-hfegran: I did a small test run with a full regression, and decided as of this commit to not split out a weekly regression yet, the increase of tests is not that big - this might still be a consideration for the future though

mk/uvmt/xrun.mk
silabs-robin: Then I understand. Agreed, let's suppress this warning.

cv32e40x/regress/cv32e40x_full.yaml
silabs-robin: Ok. I was just confused because the PR description said _"Expanded full regression to include **a few** non-bitmanip enabled tests"_ but this seemed like more than a few. If the impact in run-time was not too bad then I agree that this is fine.

cv32e40x/tests/cfg/pma_test_cfg_5.yaml
silabs-robin: It would be "nice to have", to have a means for including re-usable chunks of configuration. Though what we have is simple enough and seems to be a good-enough solution. At least it is not per-test.

silabs-robin (approved):
Looks good to me.
All my questions about the diff have been answered.

PR #1085, opened 2021-12-24
Makefile update for both core and uvm

MikeOpenHWGroup:
This PR updates the e40p Makefiles so that both the `core` testbench and `uvmt` environment use similar rules for writting compiler outputs and logs to a single location.  I have also included a few other items:
1. Remove unused modules from the uvmt_cv32e40p file-list.
2. Minor edits of READMEs.
3. Renamed `cv32e40p/sim/Common.mk` to `ExternalRepos.mk`.

Item 3 is worth thinking about.  For each of our cores there are two `Common.mk` Makefiles.  This is confusing so at a minimum one of them must be renamed.  I am less attached to the actual name we use: "ExternalRepos" is a good description of the file, but if you have a better idea, let's use that.


bin/ci_check
```diff
@@ -56,7 +56,7 @@ if (sys.version_info < (3,0,0)):
 ################################################################################
 # Globals....
 name_of_ci_check_regression = '<core>_ci_check_dev' # must match the name of one regression list in .metrics.json
-core_tests                  = ['misalign', 'illegal', 'dhrystone', 'fibonacci', 'riscv_ebreak_test_0']
+core_tests                  = ['misalign', 'illegal', 'dhrystone', 'fibonacci'] #, 'riscv_ebreak_test_0'] # FIXME: full pathname of this test is too long for Verilator (seriously!)
```
silabs-robin, line 59: This will only affect ci_check for the "core" testbench using verilator, is that right? It does not affect ci_check for uvmt?

Out of curiousity, what is the problem with the full pathname? 'riscv_ebreak_test_0' doesn't seem much longer than the other items in that list.

cv32e40p/sim/uvmt/Makefile
```diff
@@ -48,5 +48,5 @@ MAKE_PATH            := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))
 export CORE_V_VERIF   = $(abspath $(MAKE_PATH)/../../..)
 export CV_CORE       ?= cv32e40p
 
-include ../Common.mk
+include ../ExternalRepos.mk
```
silabs-robin, line 51: While I didn't mind the "Common.mk" name, I have to agree that "ExternalRepos.mk" reflects the actual contents of that file much better.

Just to brainstorm a little bit, "RepoConfig.mk" or something like that might also work. But I think "ExternalRepos.mk" looks like the best suggestion.

cv32e40p/sim/core/Makefile
```diff
@@ -53,6 +53,19 @@ CV_CORE       ?= CV32E40P
...
+# Source directores
+VERI_CUSTOM = $(CORE_V_VERIF)/$(CV_CORE_LC)/tests/programs/custom
+
+# Common output directories
+RUN_INDEX               ?= 0
+SIM_RESULTS              = simulation_results
+SIM_TEST_RESULTS         = $(SIM_RESULTS)/$(TEST)
+SIM_RUN_RESULTS          = $(SIM_TEST_RESULTS)/$(RUN_INDEX)
+SIM_TEST_PROGRAM_RESULTS = $(SIM_RUN_RESULTS)/test_program
+SIM_BSP_RESULTS          = $(SIM_TEST_PROGRAM_RESULTS)/bsp
```
silabs-robin, line 67: This are now done the same way as is done in "mk/uvmt/uvmt.mk"? That seems nice, to do it all in the same consistent way.

I don't know if maybe it could make sense to extract these common "paths-related" settings into one makefile fragment that all other makefiles could include?
Also I see that the "sim/uvmt/Makefile" just includes "mk/uvmt/uvmt.mk", but the "sim/core/Makefile" does not just include "mk/core/core.mk" (but if it works it works).

silabs-robin (approved):

bin/ci_check
MikeOpenHWGroup: That is correct - this change only affects the core testbench running under Verilator.  (The regression list for the the UVM ci_check comes from the .metrics.json file.)

The issue with riscv_ebreak_test_0 is not necessarily it's filename, it is the full pathname.  I have a _lot_ of working copies of repos (not just OpenHW ones) and keep them organized with a rather tall directory tree.  I've poked around looking for a solution for this that doesn't require a re-compile of Verilator using custom macros without success.  The "best' workaround is to update the Makefile to generate a relative path for Verilator - I've pushed in an update for that to this PR.

cv32e40p/sim/uvmt/Makefile
MikeOpenHWGroup: Almost anything is better than two included Makefiles with the same filename.

cv32e40p/sim/core/Makefile
MikeOpenHWGroup: It is _almost_ the same.

PR #1086, opened 2021-12-27
Fencei asserts update

silabs-robin:
This PR updates the fencei assertions to keep up with RTL.
One SVA bug was fixed, a cover was made conditionally disabled, and a few additional minor checks were added.
It passes ci_check and the handfull of pma configs that I tried in formal.

NB! I "stole" the `bufferable_in_config` function from a [40x SVA file](https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_mpu_sva.sv#L100). Is there any legal issue about this?


cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -537,29 +537,6 @@ module uvmt_cv32e40x_debug_assert
         else `uvm_error(info_tag, "Debug mode not entered correctly while handling illegal instruction!");
 
 
-    // Check that "dm_halt_addr_i" is correct
```
silabs-robin, line 540: Moved these to a new file.

cv32e40x/tb/uvmt/uvmt_cv32e40x_fencei_assert.sv
```diff
@@ -83,10 +108,12 @@ module uvmt_cv32e40x_fencei_assert
...
     |->
     (
-      (instr_req_o && instr_gnt_i) [->1:2]
+      // Normal execution
+      (instr_req_o && instr_gnt_i) [->1:2]  // next req-gnt (or second next, if ongoing req)
       ##0 (instr_addr_o == pc_next)
     ) or (
-      rvfi_valid [->2]
+      // Exception execution
+      rvfi_valid [->2:3]  // retire: fencei, (optionally "rvfi_trap"), interrupt/debug handler
```
silabs-robin, line 116: It was needed to change `->2` into `->2:3` in the exception case, because certain pma configurations revealed in formal that it is (legally) possible to get a trap on the instruction that follows immediately after a fencei. (So, `fencei-trap-handler`, and not only `fencei-handler`.)

cv32e40x/tb/uvmt/uvmt_cv32e40x_fencei_assert.sv
```diff
@@ -138,20 +165,47 @@ module uvmt_cv32e40x_fencei_assert
...
     !data_rvalid_i throughout (
       $fell(wb_valid) [->1]
       ##1 (data_req_o && data_gnt_i) [->1]
-    )
-  ) else `uvm_error(info_tag, "flush req shall not come if rvalid is awaited");
+    );
+  endproperty
+  a_req_wait_bus: assert property (p_req_wait_bus)
+    else `uvm_error(info_tag, "flush req shall not come if rvalid is awaited");
+  if (bufferable_in_config()) begin : gen_c_req_wait_bus
```
silabs-robin, line 178: This cover is only reachable in certain pma configurations, so it was conditionally included to avoid benign unreachability errors.

cv32e40x/tb/uvmt/uvmt_cv32e40x_fencei_assert.sv
```diff
@@ -138,20 +165,47 @@ module uvmt_cv32e40x_fencei_assert
...
 
-  c_req_wait_obi: cover property (
-    $rose(is_fencei_in_wb)
-    ##1 (is_fencei_in_wb throughout ($rose(data_rvalid_i) [->1]))
-    ##0 (is_fencei_in_wb throughout ($rose(fencei_flush_req_o) [->1]))
-  );
+  property p_req_wait_outstanding;
+    fencei_flush_req_o |-> (obi_outstanding == 0);
+  endproperty
+  a_req_wait_outstanding: assert property (p_req_wait_outstanding)
```
silabs-robin, line 193: `a_req_wait_outstanding` should in principle check the same thing as `a_req_wait_bus`, but it is checked via different means because we were in doubt that `a_req_wait_bus` really did check what it is intended to.

`a_outstanding_equivalence` was added as an experiment to track whether the two assertions above remain equivalent as RTL keeps on changing. (It is not a requirement, just a curious check of the asserts themselves.)

cv32e40x/tb/uvmt/uvmt_cv32e40x_integration_assert.sv
```diff
@@ -0,0 +1,92 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+
+
+// This module contains assertions relating to the "Core Integration" chapter
+// of the user manual
+
+module uvmt_cv32e40x_integration_assert
```
silabs-robin, line 22: Created a new assert module because there are more signals than just `dm_halt_addr_i` that is speced to have requirements about stability and alignment. I saw some weird counter examples in formal which prompted me to add these.

MikeOpenHWGroup:
> NB! I "stole" the bufferable_in_config function from a 40x SVA file. Is there any legal issue about this?

There is no legal issue (but thanks for asking).  The CV32E40X is an OpenHW Group "CORE-V" project with the same legal status as the CV32E40P, CV32E40S, CVA6 and CORE-V-VERIF.  That is, all are open-source projects covered by the Sholderpad 2.0 license.  That means _anyone_ may use sources from any of these repostories for _any_ legal purpose, as long as the copywrite is maintained.   So we are good.

silabs-hfegran:
@silabs-robin, thank you for the updates,
Could you please make sure that the properties are disabled during reset? I foresee that the lack of reset-handling may lead to some issues down the line.

silabs-hfegran (changes requested):
```
disable iff (!rst_ni)
```

silabs-hfegran (approved):

PR #1087, opened 2021-12-27
Determine nmi load/store fault

silabs-robin:
This PR makes `uvma_rvvi_ovpsim` choose correct signaling for load/store nmi fault.
Fixes https://github.com/openhwgroup/cv32e40x/issues/339 .
The new `rvfi_nmip[1:0]` RTL signal made this relatively easy.

Tested on the latest cv32e40x/dev branch. It was possible to reproduce the problem reported by the issue without having to use an old hash of core-v-verif.
Tested with `corev_rand_data_obi_err_debug` using different seeds, as well as with `..._data_obi_err` and `..._instr_obi_err`.
Passes ci_check.

Before:
image

After:
image


silabs-hfegran (approved):
Looks good to me @silabs-robin, thanks!

Issue #1052, opened 2021-12-07
Need a simple hello world guide


silabs-robin:
There is a problem that setting up the environment is complicated and we don't have a minimal example.

Currently the main README's "Getting Started" has a link to what OpenHW is and to the verif strategy document which is too technical for someone looking for a simple hello world.
I see some /dev branches have added 40P and A6 links, and that might be sufficient (if 40x/s is similarly updated). But the readmes you find in `mk/` and `cv32e40x/sim/uvmt/` contain far too much information.

To make collaboration easier with either newcomers or 3rd parties, I think we should have a very minimal guide for running hello world.
It is also not clear which of the environment variables are essential for this, and which are extraneous.

**Example:**

To run hello world for e.g. cv32e40x, `cd` into `cv32e40x/sim/uvmt` and run `make test TEST=hello-world`.

The environment needs to be set up correctly. The following is a minimal example (read more in [todo link](...)).
```
setenv CV_CORE cv32e40x
setenv CV_SW_TOOLCHAIN <point to gcc>
setenv CV_SIMULATOR <set to "xrun" or some other simulator>
setenv USE_ISS NO
```

MikeOpenHWGroup:
Ha!   This is a case of either "great minds think alike" or "fools seldom differ".  The state of core-v-verif has been bothering me for a while and so I have been working on an update of the Verification Strategy and included a new [Quick Start](https://github.com/MikeOpenHWGroup/core-v-verif/blob/doc/docs/VerifStrat/source/quick_start.rst) chapter.   The idea is this would be pointed to directly from the top-level README.

This is still on a "doc" branch of my core-v-verif fork and its not yet ready to push up to the master, but your feedback would be welcome.

silabs-robin:
I have read through you "Quick Start" guide. It looks good.
The following are my thoughts about certain parts of the guide, but I have just dumped my thoughts straight out so I hope that they are still ok to read.

What is _"Lots of python3 plug-ins"_? This should either be described, reference some doc, or have an explanatory note by it?

This _"The above will clone a version of the CV32E40P"_ is a bit confusing start of that paragraph, since the commands above have just cloned _core-v-verif_.

Under _"Where is the RTL?"_, I think it should be emphasized that the RTL is not only cloned "as needed" but that it happens "automatically".

[...]

PR #1088, opened 2021-12-28
Add "Quick Start" guide

MikeOpenHWGroup:
Hi @silabs-robin, this is the pull-request to satisfy #1052.  The Verification Strategy still needs a lot of work, so this is just a start.

Most of the changes in this PR fall into one of two categories:
1. Direct updates to the Verification Strategy due to the addition of the Quick Start chapter.
2. Updates to READMEs (that have been long overdue).


cv32e40p/sim/tools/README.md
```diff
@@ -1,7 +1,21 @@
...
+#### dsim
+Code coverage scopes configuration for Metrics dsim.
+
+#### gtk
+Handy configs for bringing up top-level signals on the gtkwave waveform viewer.
+
+#### imperas
 
 #### mcy
 Setup and configuration for [Mutual Coverage with Yosys](https://github.com/YosysHQ/mcy) from [SymbioticEDA](https://www.symbioticeda.com/).
```
silabs-robin, line 14: Seems like an old typo? It should be called _"Mutation Cover with Yosys"_?

https://github.com/YosysHQ/mcy

cv32e40s/sim/tools/README.md
```diff
@@ -1,7 +1,3 @@
 ## TOOLS directory
-Tool-specific sub-dirs for various tools used in the CV32 project.  Please do
-**not** place these in the sim, sim/core or sim/uvmt\_cv32 directories.
-
-#### mcy
-Setup and configuration for [Mutual Coverage with Yosys](https://github.com/YosysHQ/mcy) from [SymbioticEDA](https://www.symbioticeda.com/).
-
+Tool-specific sub-dirs for various tools used in the CV32E40S project.  Please do
+**not** place these in the sim, sim/core or sim/uvmt\_cv32e40s directories.
```
silabs-robin, line 3: Not that it matters (and it doesn't have to be changed), but the corresponding 40p and 40x readmes got the _"Please do [...]"_ sentence starting on its own new line.

cv32e40p/sim/tools/README.md
MikeOpenHWGroup: I've always through "Mutation Cover with Yosys" was bad English.  :-)

cv32e40s/sim/tools/README.md
MikeOpenHWGroup: You are right that it does not matter - the Markdown is rendered the same in both cases - but I made the change anyway.  Just FYI, the reason for the newline is a guideline I read that recommends that each sentence in a Markdown file start and end on its own line.  I do not follow it rigorously, but it does make maintenance of Markdown files easier.

Since I was editing these files I also added a sentence to indicate that the contents of the `tools` directory is not maintained.

MikeOpenHWGroup:
Thanks for the review comments @silabs-robin.  I've responded to your comments and will merge this up to the master branch now as I want to get the Quick Start formally published.

PR #1089, opened 2021-12-28
Resolve or waive linter issues

MikeOpenHWGroup:
This PR resolves and/or waives two distinct linter errors identified by Verissimo:

1. Duplicated code in `lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv`.  Please look at this change carefully. The change is straightforward - two functions have been collapsed into a single function.  Having said that the OBI memory Agent is a critical component in our environments, so we want to be cautious about changes to it.
2. Dealing with `Verissimo SVTB.32.2.0` (Pass strings by reference).  For the most part, these have been waived.

### DIFF
--- a/cv32e40p/tb/uvmt/uvmt_cv32e40p_iss_wrap.sv
+++ b/cv32e40p/tb/uvmt/uvmt_cv32e40p_iss_wrap.sv
@@ -86,7 +86,7 @@ module uvmt_cv32e40p_iss_wrap
        step_compare_if.ovp_cpu_state_cont  = cpu.control.state_cont;
    end
 
-    function void split(input string in_s, output string s1, s2);
+    function automatic void split(ref string in_s, ref string s1, s2);
         automatic int i;
         for (i=0; i<in_s.len(); i++) begin
             if (in_s.getc(i) == ":")
--- a/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
@@ -349,9 +349,10 @@ task uvma_obi_memory_drv_c::prep_req(ref uvma_obi_memory_base_seq_item_c req);
 endtask : prep_req
 
 // Both Master READ and WRITE transactions are handled here because the signalling is almost identical.
+// TODO: this task is yet to be fully tested as the CV32E4 cores are always the OBI bus master.
 task uvma_obi_memory_drv_c::drv_mstr_req(ref uvma_obi_memory_mstr_seq_item_c req);
 
-   if (req.access_type != UVMA_OBI_MEMORY_ACCESS_READ || req.access_type != UVMA_OBI_MEMORY_ACCESS_WRITE) begin
+   if (req.access_type != UVMA_OBI_MEMORY_ACCESS_READ && req.access_type != UVMA_OBI_MEMORY_ACCESS_WRITE) begin
      `uvm_fatal("OBI_MEMORY_DRV", $sformatf("Invalid access_type: %0d", req.access_type))
    end
 
--- a/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
@@ -92,18 +92,9 @@ class uvma_obi_memory_drv_c extends uvm_driver#(
 
    /**
     * Drives the virtual interface's (cntxt.vif) signals using req's contents.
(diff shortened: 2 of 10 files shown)


silabs-hfegran:
Thanks @MikeOpenHWGroup 

While you are at it, would you mind removing the superfluous if on line 395 in uvma_obi_memory_drv? (Technically not part of the diff, there is a while contained within an if, checking same condition)

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
```diff
@@ -357,30 +348,13 @@ task uvma_obi_memory_drv_c::prep_req(ref uvma_obi_memory_base_seq_item_c req);
...
-
-endtask : drv_mstr_req
-
-
-// This task has redundant code with drv_mstr_write_req for the request and
-// address phases.  Rather than create a new method for the common code, the
-// waiver pragmas (@DVT) are placed to warn future maintainers of the situation.
-task uvma_obi_memory_drv_c::drv_mstr_read_req(ref uvma_obi_memory_mstr_seq_item_c req);
+   if (req.access_type != UVMA_OBI_MEMORY_ACCESS_READ || req.access_type != UVMA_OBI_MEMORY_ACCESS_WRITE) begin
+     `uvm_fatal("OBI_MEMORY_DRV", $sformatf("Invalid access_type: %0d", req.access_type))
```
silabs-robin, line 355: I might be reading this wrong but the if-condition looks a bit odd to me.
Wouldn't the "uvm_fatal" always be triggered?

cv32e40p/tb/uvmt/uvmt_cv32e40p_iss_wrap.sv
```diff
@@ -86,7 +86,7 @@ module uvmt_cv32e40p_iss_wrap
        step_compare_if.ovp_cpu_state_cont  = cpu.control.state_cont;
    end
 
-    function void split(input string in_s, output string s1, s2);
+    function automatic void split(ref string in_s, ref string s1, s2);
```
silabs-robin, line 89: If we heed Verissimo's recommendation, what is the benefit?
I am not so familiar with the pros/cons of this.

There is a potential benefit of less resource use for simulations as they don't have to make copies of all of those strings? But how much of an impact does it make and how beneficial is it?
In certain cases the "input"/"output" designation might be of value?
Some lines might become very long when waiver-comments are appended to them?
I am ok with simply waiving these warnings (maybe even globally) but I don't have a complete view of it.

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
silabs-hfegran: Well spotted @silabs-robin

silabs-hfegran (changes requested):

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
MikeOpenHWGroup: Very good eye @silabs-robin, thanks.  I have pushed in a fix.

It turns out, this particular task is _never_ invoked in the CV32E4 environments since the core is always the OBI bus master.

silabs-robin (approved):

cv32e40p/tb/uvmt/uvmt_cv32e40p_iss_wrap.sv
MikeOpenHWGroup: According to the Verissimo documentation on this particular lint check, copying strings in or out of the function's memory space is a significant performance issue.  I have no valid position on that except to observe that copying an entire string will take more time than copying a pointer.  How much of an impact that will be is up for debate, particularly for short strings.

It turns out that in many cases, we pass constant strings to functions and these _cannot_ be passed by reference.  A useless fact about the SV LRM that you can use to impress your friends.  In these cases, I have added a waiver (see [uvmt_cv32e40p_step_compare.sv, line 389](https://github.com/MikeOpenHWGroup/core-v-verif/blob/cv32e40p/dev/cv32e40p/tb/uvmt/uvmt_cv32e40p_step_compare.sv#L389) for an example).

> I am ok with simply waiving these warnings (maybe even globally) but I don't have a complete view of it.

I agree that these should be waived globally.  I will discuss with AMIQ as they often have additional insight to these lint errors.  Having said that, they will need a good argument to not add this check to the core-v-verif list of global waivers.  I'll do that on my next pull-request for linter violations.

silabs-hfegran (approved):

PR #1092, opened 2022-01-04
Refactoring the method for selecting toolchain

MikeOpenHWGroup:
This pull-request implements a prototype of a scheme to refactor how the core-v-verif Makefiles select toolchains.  User-level documentation is available in [mk/uvmt/README.md](https://github.com/MikeOpenHWGroup/core-v-verif/blob/e40x_refractor_toolchain/mk/uvmt/README.md#required-corev-environment-variables) and implementer-level documentation is the common Makefile, (mk/Common.mk [here](https://github.com/MikeOpenHWGroup/core-v-verif/blob/2c4aa196cedb5f87a00a645426933ced26c51842/mk/Common.mk#L194) and [here](https://github.com/MikeOpenHWGroup/core-v-verif/blob/2c4aa196cedb5f87a00a645426933ced26c51842/mk/Common.mk#L244)).

The original code in the Makefile has not yet been deleted, just commented out.

These changes successfully passed a local "ci_check" regression with dsim.


mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-
-DV_OVPM_HOME    = $(CORE_V_VERIF)/vendor_lib/imperas
-DV_OVPM_MODEL   = $(DV_OVPM_HOME)/imperas_DV_COREV
-DV_OVPM_DESIGN  = $(DV_OVPM_HOME)/design
-OVP_MODEL_DPI   = $(DV_OVPM_MODEL)/bin/Linux64/imperas_CV32.dpi.so
-#OVP_CTRL_FILE   = $(DV_OVPM_DESIGN)/riscv_CV32E40P.ic
+###START#REFRACTOR##START#REFRACTOR##START#REFRACTOR##START#REFRACTOR##START###
+#
+#################################################################################
+## "Toolchain" to compile 'test-programs' (either C or RISC-V Assember) for the
```
silabs-robin, line 267: Typo "Assember"

mk/uvmt/README.md
```diff
@@ -15,20 +15,33 @@ To run the UVM environment you will need:
...
+Required COREV Environment Variables
+------------------------------------
+The following environment variables **must** be set for any `make` invocation to run tests or set in the user's environment to customize setting.
+
+| Environment Variable | Description  |
+|----------------------|--------------|
+| CV_SIMULATOR         | The default simulator to use for all tools (dsim, vcs, xrun, vsim, riviera).  Can be overridden on any make invocation. |
+| CV_CORE              | The core to simulate by default.  Can be overridden on any make command line.  Typically specified in the same case as the directory in which the testbench resides (e.g. cv32e40p, cva6).  However makefiles have access to internally defined CV_CORE_LC and CV_CORE_UC variables to get expected cases of the CV_CORE name (e.g. for macros, etc that might typically be capitalized). |
+| CV_SW_TOOLCHAIN      | Points to SW toolchain installation for compiling, assembling, and/or linking test programs. |
+| CV_SW_PREFIX         | Prefix of the SW toolchain installation. |
```
silabs-robin, line 27: This replaces `CV_SW_VENDOR`, right?  (With slightly changed meaning/definition)

mk/uvmt/README.md
```diff
@@ -15,20 +15,33 @@ To run the UVM environment you will need:
...
+| CV_SW_MARCH          | Architecture of tool chain to invoke. The default is `rv32imc`. |
```
silabs-robin, line 28: Since this has a default, could it be in "Optional" instead of in "Required"?

mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-endif
-RISCV            = $(PULP_SW_TOOLCHAIN)
-RISCV_PREFIX     = riscv32-$(PULP_VENDOR)-elf-
-RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
-RISCV_CC         = $(PULP_CC)
-RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_PULP_MARCH),$(PULP_MARCH))
-RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_PULP_CFLAGS),$(PULP_CFLAGS))
-endif
+RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_RISCV_MARCH),$(COREV_SW_MARCH))
+RISCV_MARCH      = $(call RESOLVE_FLAG2,$(CFG_COREV_MARCH),$(TEST_RISCV_MARCH))
```
silabs-robin, line 343: Does this second assignment always override the first one?

I am not quite sure how to write the intended behavior.
Maybe this?:
```
RISCV_MARCH = rv32imc
RISCV_MARCH = $(call RESOLVE_FLAG2,$(CFG_COREV_MARCH),$(RISCV_MARCH))
RISCV_MARCH = $(call RESOLVE_FLAG2,$(TEST_RISCV_MARCH),$(RISCV_MARCH))
RISCV_MARCH = $(call RESOLVE_FLAG2,$(CV_SW_MARCH),$(RISCV_MARCH))
```

mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-$(error test [$(TEST)] does not support the COREV toolchain)
-endif
-RISCV            = $(COREV_SW_TOOLCHAIN)
-RISCV_PREFIX     = riscv32-$(COREV_VENDOR)-elf-
-RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
-RISCV_CC         = $(COREV_CC)
-RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_COREV_MARCH),$(COREV_MARCH))
-RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_COREV_CFLAGS),$(COREV_CFLAGS))
-endif
+RISCV_CC         = gcc
```
silabs-robin, line 339: `gcc` is hard-coded here. But I think LLVM uses `cc`.
Maybe there should be a `CV_SW_CC`, while we still let RISCV_CC default to gcc?

mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-$(error test [$(TEST)] does not support the PULP toolchain)
-endif
-RISCV            = $(PULP_SW_TOOLCHAIN)
-RISCV_PREFIX     = riscv32-$(PULP_VENDOR)-elf-
-RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
-RISCV_CC         = $(PULP_CC)
-RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_PULP_MARCH),$(PULP_MARCH))
-RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_PULP_CFLAGS),$(PULP_CFLAGS))
-endif
+RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_RISCV_MARCH),$(COREV_SW_MARCH))
```
silabs-robin, line 342: I think this "COREV_SW_MARCH" should be "CV_SW_MARCH"?
Same with "COREV_SW_CFLAGS" below.

silabs-hfegran: While we are changing these fundamental parts of the makefile, i would like to propose another change as well, and that is to add a function that resolves more than two alternatives with a prioritized order (e.g. CFG > TEST > ENV > default)

mk/Common.mk
silabs-hfegran: @silabs-robin: no, this is not an override per se, it sets the second if the first is not set.

mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-RISCV            = $(LLVM_SW_TOOLCHAIN)
-RISCV_PREFIX     = riscv32-$(LLVM_VENDOR)-elf-
-RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
-RISCV_CC         = $(LLVM_CC)
-RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_LLVM_MARCH),$(LLVM_MARCH))
-RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_LLVM_CFLAGS),$(LLVM_CFLAGS))
-RISCV_CFLAGS    += -menable-experimental-extensions
-endif
+RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(CFG_RISCV_CFLAGS),$(TEST_RISCV_CFLAGS))
+RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_RISCV_CFLAGS),$(COREV_SW_CFLAGS))
```
silabs-hfegran, line 346: CV_SW_* not COREV_SW_*

mk/uvmt/README.md
silabs-hfegran: I would prefer the defaults to be set in the cfg-yaml files; a user should not need to look into the makefiles themselves unless strictly necessary, thus dropping the default from the makefile, and make it a required argument makes sense to me

mk/Common.mk
```diff
@@ -183,8 +183,32 @@ endif
 # SVLIB repo var end
 
 ###############################################################################
-# Read YAML test specifications
+# Imperas Instruction Set Simulator
```
silabs-hfegran, line 186: I am going to leave this comment here, as the files in question are not changed by this PR - the makefile changes looks good, but it will also lead to some dead code in the various python scripts located in /bin, these will also need some cleanup if these changes are made here. (Not strictly necessary functionality-wise, but we would not want to bloat our codebase with dead code)

mk/uvmt/README.md
silabs-robin: I looked upon this as a list of user-specified variables.
So "required" means "you must set up your environment like this or else nothing will run".
While "optional" means "you may tweak, but defaults already exist".

But I agree that the cfg-yaml can set CFG_COREV_MARCH, and that RISCV_MARCH maybe doesn't need any default in the makefile itself.

mk/Common.mk
silabs-robin: Sorry I didn't specify, I meant that the second _line_ always overrides whatever was done in the line before it.
Is it possible for anything other than CFG_COREV_MARCH and TEST_RISCV_MARCH to be used?
The code looks like it can never use the default value nor COREV_SW_MARCH?
And if neither CFG_COREV_MARCH nor TEST_RISCV_MARCH is set, then RISCV_MARCH itself will be empty or undefined?

silabs-hfegran: That is what I requested with my comment below, to allow for a prioritized list of variables, s.t. the highest prioritized variable will be selected. 

There is also the option of using ?= instead of = to conditionally set it if the env-var is not set.

mk/Common.mk
silabs-robin: I agree that we should have a prioritized order. I just don't know what it should be.

This is what I was trying to do with the code in:
https://github.com/openhwgroup/core-v-verif/pull/1092#discussion_r778016545
```
RISCV_MARCH = rv32imc
RISCV_MARCH = $(call RESOLVE_FLAG2,$(CFG_COREV_MARCH),$(RISCV_MARCH))
RISCV_MARCH = $(call RESOLVE_FLAG2,$(TEST_RISCV_MARCH),$(RISCV_MARCH))
RISCV_MARCH = $(call RESOLVE_FLAG2,$(CV_SW_MARCH),$(RISCV_MARCH))
```
(If that code works then maybe a fixed-arity function isn't needed?)

The order is also not clear to me. Is CFG over TEST?
I thought maybe: ENV > TEST > CFG.
The CFGs serve as "default", and we shouldn't need makefile-given defaults since we already have "default.yaml" (which should be mandatory for any core directory in core-v-verif?).
While TESTs usually don't specify any env variables at all, some tests might need to have particular settings that have to be met regardless of the config in use.
Then ENV can be user-given whenever we need to turn the knobs to debug something specific, or to match our particular computer setup.

mk/Common.mk
MikeOpenHWGroup: Fixed, thanks.

mk/Common.mk
silabs-hfegran: I tend to think that TEST should override other settings, as that is required by that specific test, while the general setting should be in CFG. I am not a big fan of having to set these things in the environment, that makes the entire setup less portable. If there is a common place to create these overrides (my suggestion: CFG), then we would have less of a dependency on system-specific variables. That would also lead to having a common place to configure the entire test-environment. 

The way things are now, some things have to be set by env vars, others in the CFG yamls and finally some things in the test yamls. When I first started out, it was not at all clear to me where certain things were configured, so the more we can confine the configuration to certain spaces, the easier it will be to use. 

Only very system specific things such as compiler locations/naming and simulator location/naming really warrants an environment-setting. Things that are self-contained within core-v-verif should, in my opinion, be located in either CFG or TEST yaml files, depending on scope of the parameter.

mk/uvmt/README.md
silabs-hfegran: Agreed - I prefer the CV_SW_TOOLCHAIN to be the full path to the binary - as long as the /lib, /inc etc. directories does not need to be specified? (do we need RISCV to point to the top level directory of the toolchain?)

mk/uvmt/README.md
MikeOpenHWGroup: @silabs-hfegran, that is a good idea. I will work on that and push in an update to this PR.

@silabs-robin, you have it correct.  I will steal your wording and add it to the README.

mk/uvmt/README.md
silabs-robin: If there is no code that uses the TOOLCHAIN or PREFIX part in isolation then it is fine.

I did a quick search and found that `uvmt.mk` does `PATH=$(RISCV)/bin:$(PATH)` (where "RISCV" comes from "..._TOOLCHAIN"). But I this line should not be necessary.

It is not an easy thing to grep for, so I don't know if this is the whole picture.

mk/Common.mk
MikeOpenHWGroup: Agree that this is confusing. This can be simplified if the defaults are _required_ to be set in the cfg-yaml files.

mk/Common.mk
MikeOpenHWGroup: Sigh, you are right.  Thanks for pointing that out.

mk/Common.mk
MikeOpenHWGroup: @silabs-robin said:
> I thought maybe: ENV > TEST > CFG

Yes, that is the correct order.

MikeOpenHWGroup:
Thanks for the comments @silabs-robin and @silabs-hfegran.  Working to push in another update to address many of these issues.   I knew this would open a big can of worms, but it needs to be done.

Until we are happy with this, please be mindful of that `DO NOT MERGE` label.

MikeOpenHWGroup:
Update: I just pushed in an update, mostly to Common.mk.  It is by no means final, but I think it is on the right track. Please have a look.  If you have shell env variables set you will need to unset these and set _only_ these two:
```
CV_SW_TOOLCHAIN=/opt/riscv
CV_SW_PREFIX=riscv32-unknown-elf-
```
(Yes, I remember that we agreed to merge these into a single variable, but I haven't gotten around to that yet.)

...and you will also need to set `CV_SIMULATOR` and `CV_CORE`.

That big comment block in `mk/Common.mk` is going to need to be moved to a README.

mk/Common.mk
silabs-hfegran: > @silabs-robin said:
> 
> > I thought maybe: ENV > TEST > CFG
> 
> Yes, that is the correct order.

Will this always be the case? I generally agree, but with slight reservations. For compiler paths and the like I agree, but for things like march we may want the test itself to be able to override a particular system-wide configuration, because we lose fine grained control.

mk/Common.mk
```diff
@@ -231,123 +261,157 @@ include $(CFG_FLAGS_MAKE)
...
-RISCV_CC         = $(LLVM_CC)
-RISCV_MARCH      = $(call RESOLVE_FLAG2,$(TEST_LLVM_MARCH),$(LLVM_MARCH))
-RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_LLVM_CFLAGS),$(LLVM_CFLAGS))
-RISCV_CFLAGS    += -menable-experimental-extensions
-endif
+RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(CFG_RISCV_CFLAGS),$(TEST_RISCV_CFLAGS))
+RISCV_CFLAGS     = $(call RESOLVE_FLAG2,$(TEST_RISCV_CFLAGS),$(COREV_SW_CFLAGS))
 
-CFLAGS ?= -Os -g -static -mabi=ilp32 -march=$(RISCV_MARCH) -Wall -pedantic
+CFLAGS ?= -Os -g -static -mabi=ilp32 -march=$(RISCV_MARCH) -Wall -pedantic $(RISCV_CFLAGS)
```
silabs-hfegran, line 444: for the future, we may want to parameterize -mabi as well (e.g. for floating point)

silabs-hfegran: To elaborate a bit further, I think when we allow partial changes to the CFLAGS like we do with these parameters, we may want to break it up into sections s.t. we can modify individual parts, without having to provide a full CFLAGS-argument list. I recon that for example -Os -g may be something we would want to tamper with at certain points

silabs-robin: I don't know what posix dictates should happen when one option is repeated with a different value. Can the CFLAGS variable just overrule all previous arguments?

The gcc man page says this about "-O" options:
> If you use multiple -O options, with or without level numbers, the last such option is the one that is effective.

silabs-hfegran: > The gcc man page says this about "-O" options:
> 
> > If you use multiple -O options, with or without level numbers, the last such option is the one that is effective.

This may be compiler specific - do you know how LLVM treats this situation?

silabs-robin: I couldn't find the equivalent of gcc's "the last such option is the one that is effective" in clang documentation.

Posix has a relevant guideline. "Guideline 11" here:
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/basedefs/V1_chap12.html#tag_12_02
_"If an option that has option-arguments is repeated, the option and option-argument combinations should be interpreted in the order specified on the command line."_

MikeOpenHWGroup: Perhaps we could define (yet) another var, CV_SW_CFLAGS and use what we have now for the default.

mk/Common.mk
silabs-robin: > for things like march we may want the test itself to be able to override a particular system-wide configuration

I agree that tests should be able to make test-specific overrides for special configs that it may need.
With "ENV > TEST > CFG" a test will be able to override the "system-wide" CFG, at least. And usually the ENV has simply not defined most of these variables?
(E.g. a CFG _should_ set march, a TEST _may_ override it, and the user-specified ENV usually never sets it.)

mk/Common.mk
```diff
@@ -232,128 +240,99 @@ endif
...
+RISCV            = $(CV_SW_TOOLCHAIN)
+RISCV_PREFIX     = $(CV_SW_PREFIX)
+RISCV_EXE_PREFIX = $(RISCV)/bin/$(RISCV_PREFIX)
 
-ifeq ($(firstword $(subst _, ,$(TEST))),pulp)
-  CFLAGS = -Os -g -D__riscv__=1 -D__LITTLE_ENDIAN__=1 -march=rv32imcxpulpv2 -Wa,-march=rv32imcxpulpv2 -fdata-sections -ffunction-sections -fdiagnostics-color=always
-endif
+RISCV_MARCH      = $(CV_SW_MARCH)
+RISCV_CC         = $(CV_SW_CC)
+RISCV_FLAGS      = $(CV_SW_FLAGS)
```
silabs-robin, line 314: Should this be `RISCV_CFLAGS = $(CV_SW_CFLAGS)`?
And should there be TEST and CFG variants of it too?

cv32e40x/tests/cfg/default.yaml
```diff
@@ -11,6 +11,7 @@ plusargs: >
     +enable_zbb_extension=1
     +enable_zbc_extension=1
     +enable_zbs_extension=1
+cv_sw_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
```
silabs-robin, line 14: With changes that were simultaneously done in another PR, I think this "cv_sw_march:" line got duplicated.

silabs-robin:
AFAICT, all my previous remarks have been resolved.
The latest diff looks good too.
Hello world ran fine for me.
When my two latest comments have been resolved, I think I will be ready to approve this PR on my side.

cv32e40x/tests/cfg/default.yaml
MikeOpenHWGroup: Good catch, thanks!

mk/Common.mk
MikeOpenHWGroup: You are right.  I have changed all instances of `FLAGS` to `CFLAGS` and also added `CV_SW_CFLAGS` to the table of "Optional COREV Environment Variables"

MikeOpenHWGroup:
> When my two latest comments have been resolved, I think I will be ready to approve this PR on my side.

I believe these have been addressed with my latest push.  If you agree, I will remove the "DO NOT MERGE" label.

mk/Common.mk
```diff
@@ -232,128 +240,99 @@ endif
...
+# Determine the values of the CV_SW_ variables.
+# The priority order is ENV > TEST > CFG.
 
-DV_OVPM_HOME    = $(CORE_V_VERIF)/vendor_lib/imperas
-DV_OVPM_MODEL   = $(DV_OVPM_HOME)/imperas_DV_COREV
-DV_OVPM_DESIGN  = $(DV_OVPM_HOME)/design
-OVP_MODEL_DPI   = $(DV_OVPM_MODEL)/bin/Linux64/imperas_CV32.dpi.so
-#OVP_CTRL_FILE   = $(DV_OVPM_DESIGN)/riscv_CV32E40P.ic
+ifndef CV_SW_TOOLCHAIN
+ifdef  TEST_CV_SW_TOOLCHAIN
```
silabs-hfegran, line 247: This may be a bit pedantic, but for more robustness we may also want to check for an empty, but defined variable. I know it is bad practice to set variables empty when they are not supposed to exist, but this happens.  (e.g. `export CV_SW_TOOLCHAIN=""`  is not the same as  `unsetenv CV_SW_TOOLCHAIN` )

MikeOpenHWGroup: You are right, these are different.  The good news is that the current checks that are in place cannot distinguish between `CV_SW_TOOLCHAIN=""` is not the same as `unsetenv CV_SW_TOOLCHAIN`.  I have tried both and each time receive the follow error message:
```
Common.mk:253: *** CV_SW_TOOLCHAIN not defined in either the shell environment, test.yaml or cfg.yaml.  Stop.
```

MikeOpenHWGroup:
I pushed in one additional change to the READMEs to direct users to the most relevant information.

This refactoring is now complete.  Please review my previous replies and if you agree, I will merge this in and get to work refactoring the e40p and e40s.

silabs-robin:
Hi. The changes look good.
ci_check passes when I use xcelium. But with questa it does not pass.

```
# ** Fatal: (SIGSEGV) Bad handle or reference.
# Time: 78405300 ps Iteration: 8 Process: /uvmt_cv32e40x_tb/iss_wrap/cpu/#ALWAYS#245 File: /work/ropeders/src/core-v-verif-mike/vendor_lib/imperas/imperas_DV_COREV/sv/imperas_CV32.sv
# Fatal error in Module CPU at /work/ropeders/src/core-v-verif-mike/vendor_lib/imperas/imperas_DV_COREV/sv/imperas_CV32.sv line 212
```

However, this seems to be the case on the latest cv32e40x/dev as well (indicating this PR might be in the clear).
Except for one thing...
The number of reported PASSED/FAILED has been reduced.

```
CI Check results:
vsim-hello-world.log : PASSED
vsim-illegal.log : PASSED
vsim-csr_instructions.log : PASSED
vsim-riscv_arithmetic_basic_test_0.log : PASSED
CI Check FAILED: Expected 11 tests to run but found only 4 PASSED or FAILED messages
[core-v-verif-mike](326)>

CI Check results:
vsim-hello-world.log : PASSED
vsim-illegal.log : PASSED
vsim-csr_instructions.log : PASSED
vsim-riscv_arithmetic_basic_test_0.log : PASSED
vsim-corev_rand_arithmetic_base_test.log : PASSED
vsim-corev_rand_instr_test.log : PASSED
vsim-corev_rand_jump_stress_test.log : PASSED
vsim-corev_rand_jump_stress_test.log : PASSED
CI Check FAILED: Expected 11 tests to run but found only 8 PASSED or FAILED messages
[core-v-verif](314)>
```

mk/Common.mk
MikeOpenHWGroup: Whoops, I _thought_ I had done this.  Pushed it in just now.

mk/Common.mk
MikeOpenHWGroup: We now have `CV_SW_CFLAGS` as an optional COREV environment variable.  See `mk/README.md` for more info.

mk/Common.mk
MikeOpenHWGroup: Agreed, but this will not be resolved in this PR.   I will generate a seperate issue for this.

mk/uvmt/README.md
MikeOpenHWGroup: PREFIX is used all over the place, so I have decided to hold my nose and maintain the status quo with both `CV_SW_TOOLCHAIN` and `CV_SW_PREFIX` used to generate the full path to the tools.

One day, I will live to regret this decision.

MikeOpenHWGroup:
Hi @silabs-hfegran and @silabs-robin, thanks for sticking with this one.  I knew it would be a biggy.  I am going to merge this in now so we can make further progress on related items (and also get on with the day-to-day job of verifying the E40X).

I will also generate a set of tasks (github issues) to address the Questa failure and some addition refactoring tasks.

PR #1094, opened 2022-01-04
Isacov: Tweak sampling conditions  (based on multi-bit rvfi_trap)

silabs-robin:
This PR changes the conditions for when isacov samples instructions that are flagged with "rvfi_trap".

Previously, ebreak was not sampled at all, because we excluded sampling of all trapped instructions.
Then a stepping-stone fix was added which allowed ebreak/ecall to be sampled despite of trap.
Now this PR makes the condition more specific, so we don't sample ebreaks/ecalls which are invalid (e.g. that come from a bus error, etc).

RTL hash is also updated, because the sampling conditions relies on the latest extensions to rvfi_trap.

Passes ci_check.


lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -2384,8 +2384,16 @@ endtask : run_phase
...
-
-  if (!have_sampled && !instr.trap && cfg.core_cfg.ext_i_supported) begin
+  logic is_ecall_or_ebreak =
+    ( instr.trap[ 8:3] ==  8)                              ||  // Ecall U-mode
+    ( instr.trap[ 8:3] == 11)                              ||  // Ecall M-mode
+    ((instr.trap[ 8:3] ==  3) && (instr.trap[13:12] == 0)) ||  // Ebreak (ebreakm==0)
+    ( instr.trap[11:9] ==  1);                                 // Ebreak to* or in D-mode (* ebreakm==1)
+  logic is_normal_instr =
+    (instr.trap[0] == 0) ||                              // No rvfi_trap
+    ((instr.trap[11:9] == 4) && (instr.trap[1] == 0));   // Single-step, without any exception
```
silabs-hfegran, line 2394: is instr.trap[1]==0 needed here?

silabs-robin: I think the `instr.trap[1]==0` is necessary because a single-stepped instruction could also come with an exception like obi bus error.
`rvfi_trap[1]` (from [here](https://cv32e40x-user-manual.readthedocs.io/en/latest/rvfi.html?highlight=rvfi_trap#compatibility)) is for _"synchronous traps that do not cause debug entry"_.
Single stepped instructions should IMO be sampled just like any other instruction (please let me know if there is reason not to do so), but they should respect the normal condition that we do not sample "invalid" trapped instructions.

Issue #1075, opened 2021-12-15
Clean up code duplication for riscv-dv extensions


silabs-hfegran:
### Task Outcome
Clean up corev-dv extensions to prevent duplicate instances of identical code.

### Background information
Riscv-dv extensions are implemented both in top level core-v-verif, and in the core-specific env-folders. In some cases, code is copy-pasted between these, where the core-specific variant neither has a unique name nor any changes from code in top level. This leads to confusion and warrants a clean-up. 

### Location Information
`core-v-verif/lib/corev-dv`
`core-v-verif/<core>/env/corev-dv`
including their respective subfolders

### Completion Criteria
Unnecessary duplicate code removed

### Additional context
Specific example: 
The `corev_xori_not_instr` class exists with an identical name, with identical code in 4 different places in the hierarchy, on top level it is contained in its own specific source file, while in the core-specific locations it belongs to a larger instruction library file that contains several other classes (with several other classes also being duplicated in the same manner).
`core-v-verif/lib/corev-dv/instr_lib/corev_xori_not_instr.sv`
`core-v-verif/cv32e40p/env/corev-dv/cv32e40p_misc_instr_lib.sv`
`core-v-verif/cv32e40x/env/corev-dv/cv32e40x_misc_instr_lib.sv`
`core-v-verif/cv32e40s/env/corev-dv/cv32e40s_misc_instr_lib.sv`

Issue #1073, opened 2021-12-15
Remove corev-dv setup_mmode_reg override


silabs-hfegran:
### Task Outcome
Remove \<core\>_privileged_common_seq and its references if it serves no other purpose than to override setup_mmode_reg due to the inifinite loop bug mentioned below.

### Background information
setup_mmode_reg was overriden to prevent the issue where we would reenable interrupts before calling mret within an interrupt handler. Another interrupt between mie and mret would cause the handler to be reentered, with the return address set to mret within the handler itself causing an infinite loop (mret returns to itself). This fix is now upstreamed to google, the code in corev-dv is no longer needed and causes unnecessary clutter.

### Location Information
./\<core\>/env/corev-dv/\<core\>_instr_base_test
./\<core\>/env/corev-dv/\<core\>_privileged_common_seq

### Completion Criteria
Verify that relevant riscv-dv hash includes the fix setting mie to 0.
If this is the case, then remove the \<core\>_privileged_common_seq overrides from all affected cv32-cores
References to this override in \<core\>_instr_base_test removed from all affected cv32-cores

PR #1096, opened 2022-01-05
Removed redundant and unused code from corev-dv (#1075, #1073)

silabs-hfegran:
This commit addresses the redundant code mentioned in #1073 and
the unused code mentioned in #1075 for cv32e40x


silabs-robin (approved):
According to what the two linked issues explain, I agree that these removals make sense.

PR #1097, opened 2022-01-05
Removed redundant and unused code from corev-dv (cv32e40p, #1075, #1073)

silabs-hfegran:
This commit addresses the redundant code mentioned in #1073 and
the unused code mentioned in #1075 for cv32e40p

Signed-off-by: Henrik Fegran Henrik.Fegran@silabs.com


silabs-hfegran:
A note here - if the hash pointed to for riscv-dv on 40p is old, then removal of the privilege sequence may cause issues during simulation.

silabs-hfegran:
This PR needs an updated hash for riscv-dv to function correctly - cv32e40p's riscv-dv hash is about 1 year old.
@MikeOpenHWGroup, are there any good reasons for keeping this old version?

MikeOpenHWGroup:
> This PR needs an updated hash for riscv-dv to function correctly - cv32e40p's riscv-dv hash is about 1 year old. @MikeOpenHWGroup, are there any good reasons for keeping this old version?

No reason.   I will run a few tests and if it looks good will issue a PR to update the riscv-dv hash.

PR #1103, opened 2022-01-10
Fix path for custom programs directory

pcotret:
We can see the path in the Makefile here: https://github.com/openhwgroup/core-v-verif/blob/55deb696e671451f4057d6be3b49c1f4f29c13d1/cv32e40p/sim/core/Makefile#L57

### DIFF
--- a/cv32e40p/sim/core/README.md
+++ b/cv32e40p/sim/core/README.md
@@ -35,7 +35,7 @@ to run the default test (hello_world).
 
 Running your own Assembler programs
 -----------------------------
-If you have a C or assembly program in `../../tests/core/custom`
+If you have a C or assembly program in `../../tests/programs/custom`
 then the following will work with Verilator:<br>
 ```
 make veri-test TEST=dhrystone


MikeOpenHWGroup:
Hi @pcotret, thanks for identifying this issue and creating this pull request.  As per our [contribution rules](https://github.com/openhwgroup/core-v-verif/blob/master/CONTRIBUTING.md#contributor-agreement), you will need to be covered by the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) before I can accept this PR.  This is a painless process and which ensures that all code contributed to OpenHW repositories are truly open-source artifacts.

There are two paths we can take:
1. Follow the links above and register with the [Eclipse Foundation](https://www.eclipse.org/).  Please ensure that the email address you register with Eclipse is the same one used for this pull-request.
2. Close this pull-request and re-create it as an issue.  A member of the OpenHW Group, who is already covered by the Contributor Agreement will resolve the issue.

I would prefer the first option, but the choice is yours.   Thanks again for bringing this issue to our attention.

pcotret:
First step, always read the rules. Sorry for the inconvenience...
I'll get through the Eclipse process a bit later.

PR #1114, opened 2022-01-17
CSR updates for priv_1_12

silabs-hfegran:


cv32e40x/env/corev-dv/cv32e40x_csr_template.yaml
```diff
@@ -753,10 +936,11 @@
       lsb: 0
 
 ###############################################################################
-# mvendorid, marchid, mimpid and mhartid are temporarily excluded from auto-
-# generation of access testing as all bits in these CSRs are RO, so any attempt
-# to write them causes an illegal instruction exception.  Access modes to these
-# CSRs is tested in a separate, manually written test-program.
+# mvendorid, marchid, mimpid, mhartid, tcontrol and mconfigptr are temporarily
```
Silabs-ArjanB, line 939: tcontrol does not belong in this category. It is a MRW CSR. (The same holds for tinfo; both of them are misdocumented in the UM; I will add an issue for that)

Silabs-ArjanB (changes requested):

cv32e40x/env/corev-dv/cv32e40x_csr_template.yaml
silabs-hfegran: Thanks and well spotted!
I will update this accordingly

silabs-hfegran:
This is ready for review 

There are however two requirements for this commit to be merged: 
* https://github.com/google/riscv-dv/pull/843 must be merged to upstream riscv-dv first
* https://github.com/openhwgroup/cv32e40x/pull/396 must be merged to cv32e40x:master

Lacking updated register definitions in riscv-dv will lead to breakage of ___all___ corev-dv tests - so please do not merge until that is in place. Before merge takes place, this PR should be updated with the hash of `riscv-dv` and `cv32e40x:master` that corresponds to having the aforementioned fixes implemented.

With these fixes in place, this passes ci_check, and also resolves the `corev_rand_debug_ebreak` failures observed in regressions.

cv32e40x/env/corev-dv/cv32e40x_csr_template.yaml
```diff
@@ -365,6 +365,79 @@
       reset_val: 1
       msb: 0
       lsb: 0
+
+### Not supported in CV32E40X ###
```
silabs-robin, line 369: Is all of this commented-out code autogenerated?
Where does it come from?

cv32e40x/env/uvme/uvme_cv32e40x_cfg.sv
```diff
@@ -128,6 +129,8 @@ class uvme_cv32e40x_cfg_c extends uvma_core_cntrl_cfg_c;
       unaligned_access_amo_supported == 1;
 
       bitmanip_version        == BITMANIP_VERSION_1P00;
+      priv_spec_version       == PRIV_VERSION_MASTER;
```
silabs-robin, line 132: Do we want to call it "master" version? The ISS does it, but we could (if we want to) call it something like v1.12-beta01 etc? Or do we want "master" to generally map to whatever the ISS "master" means, always?

cv32e40x/env/corev-dv/cv32e40x_privil_reg.sv
```diff
@@ -35,21 +35,27 @@ class cv32e40x_privil_reg extends riscv_privil_reg;
...
+        add_field("MPP",    2, WARL);
+        add_field("WPRI3",  3, WPRI);
+        add_field("MPRV",   2, WARL);
+        add_field("WPRI4", 14, WPRI);
+      end
       // Machine interrupt-enable register
       MIE: begin
         fld.delete();
 
-        add_field("USIE",   1,  WARL);
```
silabs-robin, line 42: Are these fields removed because they are irrelevant to the 40x?

lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
```diff
@@ -173,6 +180,11 @@
       soft trn_log_enabled        == 1;
    }
 
+   constraint riscv_cons_soft {
+     soft priv_spec_version == PRIV_VERSION_1_11;
```
silabs-robin, line 184: Is this chosen as the default to maintain backwards compatibility?

silabs-robin:
I don't have enough knowledge of the new priv spec or of riscv-dv to have a qualified opinion on the correctness of everything in this PR.
But I have read through it all and did not see anything that I think looks odd, apart from my questions above.

cv32e40x/env/corev-dv/cv32e40x_csr_template.yaml
silabs-hfegran: It is _not_ autogenerated, I wrote this manually. The reason why I kept the non-existent registers present, albeit commented out,  was to ease integration with other cores that could potentially use these.

cv32e40x/env/uvme/uvme_cv32e40x_cfg.sv
silabs-hfegran: What ISS calls master points to a certain revision of the spec on which it is based. If we changed the name of this, then it could easily get misrepresented if we update the ISS. That is the reasoning for keeping the original naming.

cv32e40x/env/corev-dv/cv32e40x_privil_reg.sv
silabs-hfegran: They are removed because they are deprecated from the spec - these field definitions no longer exist (changed to WPRI in spec)

lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
silabs-hfegran: Yes, this is chosen to present the exact same version of the privilege spec as before, to keep compatibility with e.g. cv32e40p without having to update code specific to that core. Cores that require another version must set this constraint in its environment

silabs-robin:
The ExternalRepos.mk change looks good to me (even if it is a tiny bit dissatisfying that we must temporarily point to a non-official fork).
Based also on my previous read-through I want to approve.
But the CV_CORE_HASH needs to be updated after the RTL PR is merged, before this PR can be merged?

silabs-hfegran:
> But the CV_CORE_HASH needs to be updated after the RTL PR is merged, before this PR can be merged?

Agreed - core code needs to be merged first, followed by hash update then this PR can be merged

Silabs-ArjanB (commented):
I added some questions

cv32e40x/env/corev-dv/target/cv32e40x/riscv_core_setting.sv
```diff
@@ -94,27 +94,133 @@ parameter int NUM_HARTS = 1;
...
-    MISA,       // ISA and extensions
-    MIE,        // Machine interrupt-enable register
-    MTVEC,      // Machine trap-handler base address
-    MCOUNTEREN, // Machine counter enable
-    MSCRATCH,   // Scratch register for machine trap handlers
-    MEPC,       // Machine exception program counter
-    MCAUSE,     // Machine trap cause
-    MTVAL,      // Machine bad address or instruction
-    MIP         // Machine interrupt pending
+    MSTATUS,        // Machine status
```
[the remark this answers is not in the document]
silabs-hfegran: These have to be added commented out (or updated in riscv-dv) - as adding CSRs here that do not exist in riscv-dv will cause compilation failure.

cv32e40x/env/corev-dv/cv32e40x_privil_reg.sv
```diff
@@ -35,21 +35,27 @@ class cv32e40x_privil_reg extends riscv_privil_reg;
...
+        add_field("WPRI0",  3, WPRI);
+        add_field("MIE",    1, WARL);
+        add_field("WPRI1",  3, WPRI);
+        add_field("MPIE",   1, WARL);
+        add_field("WPRI2",  3, WPRI);
+        add_field("MPP",    2, WARL);
+        add_field("WPRI3",  3, WPRI);
+        add_field("MPRV",   2, WARL);
+        add_field("WPRI4", 14, WPRI);
+      end
```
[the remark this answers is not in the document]
silabs-hfegran: This should be a relatively easy thing to implement, and all required changes are local to corev-dv. To my knowledge, the only use for these bitfield definitions is to set up the initial value for mstatus during initialization, the fields are otherwise unused. 

Similarly, the presence (or lack thereof) of the UIE field here only affects the initial value of mstatus, and it is by default set to 0 which is still the case when this is treated as a non-existing field. (see code below, `setup_mmode_reg()` is the only applicable code that uses this register definition). 

I suggest we add suitable defaults for the VS, FS, XS and SD bits when they are added, there is no need to add them here until they actually need a default reset value.

Silabs-ArjanB (approved):

PR #1118, opened 2022-01-17
Add assembly Hello world test with custom BSP

Gchauvon:
Hello,

I added an "Hello World" test in assembly  that is using a custom BSP.
This test doesn't use the riscv-dv BSP and allows us to build more easily custom tests and modifying trap handlers, syscall functions etc...  

Also fix a typo in smoke-test.sh


JeanRochCoulon (changes requested):
Yes, better to rename hello_word into hello_world ;-)
As the resulting hello_world_asm.S is no more the classical test which prints "HELLO WORLD!", the test name is maybe not adapted.
testlist_custom_bsp.yaml is a testlist dedicated to all tests in custom directory. Renaming it into  testlist_custom.yaml could ease the reuse

MikeOpenHWGroup:
In the hello-world test for the E40 cores, the test is a bit more interesting than simply printing to stdout.  For example:
```
HELLO WORLD!!!
This is the OpenHW Group CV32E40X CORE-V processor core.
CV32E40X is a RISC-V ISA compliant core with the following attributes:
	mvendorid = 0x602
	marchid   = 0x14
	mimpid    = 0x0
	misa      = 0x40001104
	XLEN is 32-bits
	Supported Instructions Extensions: MIC
```
The test-program is self-checking.  That it, it will read the above CSRs and if they are not as expected, the testcase will fail.  (BTW, the value of `mvendorid` is the correct one: it encodes the OpenHW JEDEC Manufacturer ID and the `mimpid` was assigned by RISCV International.

Gchauvon:
Indeed, `hello_world_asm` was not a good name. I renamed it to `custom_test_template`  as it should only be used to create your own test with your own custom BSP.

JeanRochCoulon (approved):

PR #1120, opened 2022-01-17
Cv32e40x/merge dev to rel 57

silabs-robin:
First steps of 2022's first merge procedure.
Passes "rel_check" using xcelium.

Note: We disabled some known failures in "rel_check" to get this passing.
(That might have been unnecessary and we might do it differently the next time.)
The disabled tests will be re-enabled once the master branch is being propagated back onto the dev branches.


MikeOpenHWGroup (approved):
Approving this PR on the basis of:
1. I've seen many of these changes before on merges to cv32e40x/dev.
2. Passing ci_check under Xcelium.

**But**, this PR fails to compile under Metrics DSIM.  I suspect that the issue is with DSIM, not the code, so I will approve this PR and hotfix in a work-around for both the `cv32e40x/dev` and `cv32e40x/release` branches.  A GitHub issue will be created to flag the workaround and a Metrics service issue will also be opened.

MikeOpenHWGroup:
Hotfix has been applied.  See #1122.

PR #1121, opened 2022-01-17
Cv32e40s/merge dev to rel 7

silabs-robin:
First steps of 2022's first merge procedure.
Passes "rel_check" using xcelium [1].

Note: We disabled some known failures in "rel_check" to get this passing.

Note: [1] it sporadically fails on re-runs, but we are ok with this for now.


MikeOpenHWGroup (approved):
Approved based on the previous reviews of these changes going into the dev branch and successful ci_check.

PR #1128, opened 2022-01-20
Fixed wrong cfg used for mhmpcounter29_csr_access tests

silabs-hfegran:
The mhpmcounter29_csr_access_test_* should be run with NUM_MHPMCOUNTERS=29.

Full regression changed to use the num_hpmcounter_29 config for these tests,  
and added support for bitmanip instructions to this cfg.

### DIFF
--- a/cv32e40x/regress/cv32e40x_full.yaml
+++ b/cv32e40x/regress/cv32e40x_full.yaml
@@ -77,126 +77,108 @@ builds:
 # List of tests
 tests:
   hello-world:
-    build: uvmt_cv32e40x
     description: uvm_hello_world_test
     builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=hello-world
 
   csr_instructions:
-    build: uvmt_cv32e40x
     description: CSR instruction test
     builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=csr_instructions
 
   generic_exception_test:
-    build: uvmt_cv32e40x
     description: Generic exception test
     builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=generic_exception_test
 
   illegal_instr_test:
-    build: uvmt_cv32e40x
     description: Illegal instruction test
     builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
     dir: cv32e40x/sim/uvmt
     cmd: make test TEST=illegal_instr_test
 
   branch_zero:
-    build: uvmt_cv32e40x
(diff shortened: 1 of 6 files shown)


cv32e40x/tests/cfg/num_mhpmcounter_29.yaml
```diff
@@ -1,8 +1,12 @@
 name: num_mhpmcounters_29
 description: Configuration for CV32E40X simulations with NUM_MHPMCOUNTER set to 29
+compile_flags:
+    +define+ZBA_ZBB_ZBC_ZBS
 compile_flags:
     +define+SET_NUM_MHPMCOUNTERS=29
-ovpsim: >
-    --override root/cpu/misa_Extensions=0x001104
-    --override root/cpu/marchid=20
```
silabs-robin, line 7: Nice, you got these removed too.

cv32e40x/regress/cv32e40x_full.yaml
```diff
@@ -156,14 +156,14 @@ tests:
   mhpmcounter29_csr_access_test_1:
     build: uvmt_cv32e40x_num_mhpmcounter_29
     description: Hardware performance counter full access coverage test 1
-    builds: [ uvmt_cv32e40x, uvmt_cv32e40x_no_bitmanip]
+    builds: [ uvmt_cv32e40x_num_mhpmcounter_29 ]
```
silabs-robin, line 159: Is this really the only config that this test can run under?

silabs-robin (approved):

cv32e40x/tests/programs/custom/mhpmcounter29_csr_access_test_2/test.yaml
```diff
@@ -1,5 +1,4 @@
 name: mhpmcounter29_csr_access_test_2
 uvm_test: uvmt_$(CV_CORE_LC)_firmware_test_c
 description: >
-    CSR access test with NUM_MHPMCOUNTER = 29 (CANNOT run with ISS)
-
+  CSR access test with NUM_MHPMCOUNTER = 29 (FIXME ISS does not set correct reset value for mcountinhibit)
```
silabs-robin, line 4: Should this `FIXME` be `FIXME:<username>`?

(I have been using `TODO:ropeders`, but maybe I should have used `TODO:silabs-robin`...)

silabs-hfegran: Careful with : in yaml-files, might be interpreted  as a tag rather than its intended purpose. Comments in yamlfiles can also cause issues when placed in the wrong spot. 

But yes, I could tag myself in these fixmes

silabs-robin: Wow. Good point. Certain aspects of YAML seems dangerous.

silabs-robin (approved):
It looks correct to me. I only had that one remark about "FIXME"s, but that is more optional.

MikeOpenHWGroup (approved):
Looks good @silabs-hfegran.  Will merge as soon as my local regression passes.

PR #1132, opened 2022-01-20
Add continuous integration for gitlab-ci

yanicasa:
new CI for gitlab is massively generic to allow each group to be able to use it directly without modifications in the CI source code regardless of their runtime environment. 
Adaptations to the environment are done only in the settings of the projects. (paths of tools, version of tools, urls of repositories to use...)

### DIFF
--- a/.gitlab-ci/cv32.yml
+++ b/.gitlab-ci/cv32.yml
@@ -1,12 +1,94 @@
+# Copyright 2022 Thales DIS design services SAS
+#
+# Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+# You may obtain a copy of the License at https://solderpad.org/licenses/
+#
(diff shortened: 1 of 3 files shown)


.gitlab-ci/cv32.yml
```diff
@@ -0,0 +1,12 @@
+
```
MikeOpenHWGroup, line 1: This file will not be necessary as the original version of it (`.gitlab-ci.yml`) is no longer in use.

MikeOpenHWGroup (approved):
This all looks good @yanicasa.  The `cv32.yml` will not be necessary (but thanks for considering it!).

.gitlab-ci/cv32.yml
yanicasa: would you prefer I remove it or leave it as an entry point to motivate someone to help set up a cv32 regression?

MikeOpenHWGroup: Originally I was thinking that it should be removed.  Your idea to leave it as an entry point for future cv32 regressions on from GitLab is a good one.  Perhaps add a comment to indicate that it is currently not used.  Thanks!

JeanRochCoulon (approved):

PR #1102, opened 2022-01-07
Fixes to match latest RTL signals names; pulled core hash to latest

Silabs-ArjanB:
Signed-off-by: Arjan Bink <Arjan.Bink@silabs.com>

PR #1136, opened 2022-01-21
Fix to match RTL signals

silabs-hfegran:
Implements #1111 and #1102 for CV32E40X onto CV32E40S


silabs-robin:
@silabs-hfegran I think I already did this as part of the merge from 40x->40s.
But it could still be nice to get this PR in as a double check that I got that merging right.

Silabs-ArjanB:
@silabs-hfegran Agreed with Robin, it is weird that this PR is needed if files have already been merged from 40X to 40S

Silabs-ArjanB (approved):
Looks okay, but it is weird that this is a separate PR.

silabs-hfegran:
> Looks okay, but it is weird that this is a separate PR.

Sorry about this - simulation was broken for 40s devs, so I expedited this change to enable devs to be able to run simulations on 40s based on latest hash.

PR #1137, opened 2022-01-21
Excluding the cross with nmi to a_illegal_insn_debug_req

silabs-mateilga:
Corner case taken out of assertion. We are planning on covering this in a different assertion(s) when we look at the debug/nmi cross.

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
@@ -513,9 +513,12 @@ module uvmt_cv32e40x_debug_assert
     // Check that we cover the case where a debug_req_i
     // comes while flushing due to an illegal insn, causing
     // dpc to be set to the exception handler entry addr
+
+    // TODO We have excluded the case where an nmi is taken in the second stage of the antecedent. 
+    //      Make sure this is covered in a debug vs nmi assertion when it is written 
     sequence s_illegal_insn_debug_req_ante;  // Antecedent
         cov_assert_if.wb_illegal && cov_assert_if.wb_valid && !cov_assert_if.debug_mode_q
-        ##1 cov_assert_if.debug_req_i && !cov_assert_if.debug_mode_q;
+        ##1 cov_assert_if.debug_req_i && !cov_assert_if.debug_mode_q && !cov_assert_if.pending_nmi;
     endsequence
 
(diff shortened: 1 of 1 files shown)


cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -513,9 +513,12 @@ module uvmt_cv32e40x_debug_assert
     // Check that we cover the case where a debug_req_i
     // comes while flushing due to an illegal insn, causing
     // dpc to be set to the exception handler entry addr
+
+    // TODO We have excluded the case where an nmi is taken in the second stage of the antecedent. 
+    //      Make sure this is covered in a debug vs nmi assertion when it is written 
     sequence s_illegal_insn_debug_req_ante;  // Antecedent
         cov_assert_if.wb_illegal && cov_assert_if.wb_valid && !cov_assert_if.debug_mode_q
-        ##1 cov_assert_if.debug_req_i && !cov_assert_if.debug_mode_q;
+        ##1 cov_assert_if.debug_req_i && !cov_assert_if.debug_mode_q && !cov_assert_if.pending_nmi;
```
silabs-robin, line 521: This looks good to me.
Limiting the antecedent's scope is a practice we have already done elsewhere (e.g. here https://github.com/openhwgroup/core-v-verif/blob/2789175eea14717bad59e04303569521f281c431/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv#L178).

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -513,9 +513,12 @@ module uvmt_cv32e40x_debug_assert
     // Check that we cover the case where a debug_req_i
     // comes while flushing due to an illegal insn, causing
     // dpc to be set to the exception handler entry addr
+
+    // TODO We have excluded the case where an nmi is taken in the second stage of the antecedent.
```
silabs-robin, line 517: Nice that there is an in-code TODO note to get it done.

In the future, we should consider whether we want `TODO`s to have the author's username next to it, like `TODO:silabs-robin`. But I think this one is fine as it is.

silabs-robin (approved):
Approving based on that we did a live debugging session and I agree that the NMI corner case is just a distraction from what this particular assert is trying to check. As long as we write that new assert (with a coverpoint to observe this scenario) I think we should be covered.

PR #1140, opened 2022-01-24
First version of CVXIF agent

ZElkacimi:
This is a first version of the cvxif agent, a test environment is provided for the cva6 verification
 - support 3 interfaces (issue_interface, commit_interface and result_interface)
 - accept 2 instructions (CUS-ADD, CUS-ADD-RS3)

### DIFF
--- a/lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
+++ b/lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
@@ -1,0 +1,75 @@
+// Copyright 2021 Thales DIS design services SAS
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Zineb EL KACIMI (zineb.el-kacimi@external.thalesgroup.com)
+
+/**
+ * Encapsulates assertions targeting uvma_cvxif_if.
+ */
+module uvma_cvxif_assert #(parameter X_ID_WIDTH = ariane_pkg::TRANS_ID_BITS)
+   (uvma_cvxif_if cvxif_assert,
+    input bit clk
+   );
+
+   import uvm_pkg::*;
+
+   // ---------------------------------------------------------------------------
+  // Local variables
+  // ---------------------------------------------------------------------------
+  string info_tag = "CVXIF_ASSERT";
+
+  // ---------------------------------------------------------------------------
+  // Begin module code
+  // ---------------------------------------------------------------------------
+
+   /**
+    * Issue_interface
+   */
+   property IssueRespNull_when_Naccept;
+      @(posedge clk) (cvxif_assert.cvxif_req_i.x_issue_valid && !cvxif_assert.cvxif_resp_o.x_issue_resp.accept)
+                     |-> (!cvxif_assert.cvxif_resp_o.x_issue_resp.writeback && !cvxif_assert.cvxif_resp_o.x_issue_resp.dualread
+                          && !cvxif_assert.cvxif_resp_o.x_issue_resp.loadstore && !cvxif_assert.cvxif_resp_o.x_issue_resp.exc);
+   endproperty
+
+   ASSERT_IssueRespNull:assert property (IssueRespNull_when_Naccept)
+   else `uvm_error(info_tag, "Failure:  writeback ; dualwrite ; dualread; loadstore; exc; shall be equal to 0");
+   COV_IssueRespNull:cover property (IssueRespNull_when_Naccept);
+
+   /**
+    * Commit_interface
+   */
+   property commit_valid_pulse;
+     bit [X_ID_WIDTH-1:0] id;
+     @(posedge clk) (cvxif_assert.cvxif_req_i.x_commit_valid, id=cvxif_assert.cvxif_req_i.x_commit.id)
+                    |=> !(cvxif_assert.cvxif_req_i.x_commit_valid && cvxif_assert.cvxif_req_i.x_commit.id==id);
+   endproperty
+   ASSERT_COMMIT_VALID_PULSE:assert property (commit_valid_pulse)
+   else `uvm_error(info_tag, "Failure: commit_valid is asserted more than 1 cycle");
+
+    property issue_commit ;
+      bit [X_ID_WIDTH-1:0] id;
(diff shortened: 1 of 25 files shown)


MikeOpenHWGroup (changes requested):
Hi @ZElkacimi, this all looks really good.  It is clear that you took the time to review the structure and coding style used by core-v-verif and this is very much appreciated.

Speaking of structure, can you update this pull-request to conform to the directory structure used by other core-v-verif agents?   If looks like this:
```
uvma_cvxif/
        +--------- docs/
        +--------- bin/
        +--------- examples/
        +--------- src/
                    +------ comps/
                    +-------obj/
                    +------- seq/
```
* `comps` is for the "components" of an Agent, such as the Driver, Monitor, etc.
* `obj` is for the "objects" used by the Agent (configuration and context).
* `seq` is for the sequence items and sequences

A good example is at `lib/uvm_agents/uvma_obi_memory`.  You will notice that in many cases the `docs`, `bin` and `examples` directories are empty.

By the way, you may not know that we have written a specific set of Coding Style Guidelines for core-v-verif and this maintained on the core-v-verif repo ([link](https://github.com/openhwgroup/core-v-verif/blob/master/docs/CodingStyleGuidelines.md)).  This document will always be a "work in progress", so if you have any opinions about it, please do create a Issue and we can discuss.

ZElkacimi:
Agent files are re-organized as requested.

MikeCovrado (approved):

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @ZElkacimi!

Please disregard the review from @MikeCovrado.  That _is_ me, but it is my private account and I should not have used that to review an OpenHW pull-request.   The review from @MikeOpenHWGroup is the "real" one.

@JeanRochCoulon, this looks good to me (actually, it looks great), but I think it is important that you also have a chance to comment.  Unless you have specific comments, please feel free to merge.

JeanRochCoulon (approved):

JeanRochCoulon:
Thanks @ZElkacimi and also @ASintzoff for this great contribution. 
This agent has been aligned on cve4 agent template to be re-used in other projects. @Silabs-ArjanB, @davideschiavone,... Do not hesitate to reuse it for your own verifications !  Feedbacks are welcome

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
```diff
@@ -0,0 +1,75 @@
...
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Zineb EL KACIMI (zineb.el-kacimi@external.thalesgroup.com)
+
+/**
+ * Encapsulates assertions targeting uvma_cvxif_if.
+ */
+module uvma_cvxif_assert #(parameter X_ID_WIDTH = ariane_pkg::TRANS_ID_BITS)
```
silabs-robin, line 13: This `ariane`-specific package usage should likely not be core-specific, since this is in `lib/`.
I don't know which package is appropriate, but something like `uvma_core_cntrl_pkg` or similar might be better?

(I just noticed while skimming this diff and thought to mention it so it might get updated in future commits.)

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 Thales DIS design services SAS
```
silabs-robin, line 1: I see some of the other agents include `Copyright 2021 OpenHW Group` as well. @MikeOpenHWGroup do you know if there are any OpenHW/Eclipse policies governing this?

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
ZElkacimi: Thank you @silabs-robin for your feedback, I will take care of it in the next commit.
I am looking forward for other remarks from you, thanks.

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_assert.sv
MikeOpenHWGroup: Eclipse does not place any limitations on the copyright holder, as long as there is one and that a header for one of the accepted open-source licenses is used.  The OpenHW Group is pleased if a member contribution also assigns copyright to OpenHW, but in the end it does not matter.

Silabs-ArjanB:
@ZElkacimi Thanks for your contribution! Here is some initial feedback (from a non-UVM-expert):

- Can all of the code made to support either master or slave behavior right from the start (not added in later as afterthought)?
- Can support for random waits (both for issuing valid signals and for issueing ready signals) be added in? (Now a hard coded number of cycles seems to be used)
- It seems to me that the result handshake does not wait for the related result_ready signal (the same might be true for other handshakes; I did not check; it could also be that this question just reflects my lack of experience in this area)
- Can you refer to a git hash of the XIF repos so that we can know to what version the code is trying to be compatible (as the spec is not frozen yet)?
- Many values are hardcoded that should depend on configuration parameters (id width, data width, etc.)
- How is this supposed to work together with the ISS lockstep check-and-compare mechanism (i.e. how will the ISS be guaranteed to perform the same actions for user defined instructions)? 

The last question is quite fundamental; @MikeOpenHWGroup any ideas on that because without a link to the ISS nothing is really getting checked here.

ZElkacimi:
@Silabs-ArjanB thank you for your comment

1- This agent is first developed to verify a CPU and not a Copro, that's why it's coded as a slave. I can specify in my functions that they are made for a slave, so that if, in the future, someone wants to add a master behavior, it will be clear.
2-  Random waits for issuing valid signals and ready signals are ongoing for implementation, it will be in the next version of the agent.
3- The wait for the related result_ready signal, is done in the sequence before sending the response to driver. For other handshakes, they concern interfaces which are not yet supported.
(since the result_ready is always 1 (for cva6), I couldn't check if my implementation is correct but I reviewed my code after your question, and I discover that with the current implementation we will lose results were result_ready is 0, so I will correct that).
4- Okay, I will add the git hash of the XIF repos in README file.
5- You are right, the values are hardcoded in the sequence items, I will correct that, thank you.
6- I didn't look at the ISS lockstep check-and-compare mechanism, CVA6 use the ISS Spike, and the implementation of the custom instructions on Spike is ongoing. For now, we only have assertion checks for 4 protocol features, you will find other assertions in next coming versions.

Thank you again @Silabs-ArjanB for your relevant comment, I am looking forward for other remarks from you, thanks.

PR #1108, opened 2022-01-13
Add code coverage and riscv-dv random tests

ASintzoff:
- to support code coverage, test output files need to have a unique name (add timestamp to the filename)
- provide a way to exclude some modules from coverage (e.g. rvfi_tracer module)
- to support riscv-dv random tests, use unique name for program file to be loaded (no more Mem_init.txt)

PR #1141, opened 2022-01-25
Rename Mem_init.txt to expected name for vcs_core and vcs_gate target

Gchauvon:
Just a small fix : Mem_init.txt has been renamed to $(TESTNAME).txt in PR #1108. 
I have updated vcs_core and vcs_gate target so it would work with this PR.

### DIFF
--- a/cva6/sim/Makefile
+++ b/cva6/sim/Makefile
@@ -61,7 +61,9 @@ vcs_core_comp:
 	cd $(VCS_WORK_DIR) && vcs $(ALL_VCS_FLAGS) work.cva6_core_only_tb
 
 vcs_core_run:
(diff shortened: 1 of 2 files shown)


JeanRochCoulon (commented):
Hello Guillaume, just a little question. Why elf-bin option is added when making vcs_core_run ?

Gchauvon:
`elf-bin` option is added because the variable `TESTNAME` takes it as its input. Otherwise the value of `TESTNAME ` is empty and Mem_init.txt can not be created.

JeanRochCoulon (approved):
Thanks for the explanations. It is ok for me !

JeanRochCoulon:
@MikeOpenHWGroup As this PR corresponds to a Fix, I took the liberty of merging it.

PR #1145, opened 2022-01-25
40X changing Makefile name in line 66 as is in ../ directory

Nicolas-Gaudin:
##Change makefile name called in another makefile for 40x

https://github.com/Nicolas-Gaudin/core-v-verif/blob/a13787838c108f9e6d55af308a2bc1b70acd6419/cv32e40x/sim/core/Makefile#L66

Change in l66 of cv32e40x/sim/core/Makefile

include ../Common.mk

into

include ../ExternalRepos.mk

As is in ../ directory


MikeOpenHWGroup (approved):
Thanks for this @Nicolas-Gaudin!

Issue #1110, opened 2022-01-14
RVVI and RVFI agent incompatibility across CV32E40X/E40S


MikeOpenHWGroup:
### Type
Compile error of the CV32E40S in the cv32e40x/dev branch.

As we were discussing this morning, I wanted to look into whether all cores would run in a specific core's "dev" branch.   In this case, the cv32e40x/dev branch.  If you try compiling the sanity test of the CV32E40X on the `cv32e40x/dev` branch of `git@github.com:openhwgroup/core-v-verif.git`, it will compile and run correctly, but the e40s will not compile at all.   In an attempt to rectify this, I found what seems to be an incompatibility of RVVI and RVFI across both the E40X and E40S.

### Steps to Reproduce
On my fork of core-v-verif there is a branch `cv32e40x_dev_incompatible_with_e40s` that illustrates the point.  This branch updates to the RVVI and RVFI agents to be compatible with the E40S environment.
```
$ git clone -b cv32e40x_dev_incompatible_with_e40s git@github.com:MikeOpenHWGroup/core-v-verif.git tryit
$ cd tryit/cv32e40s/sim/uvmt
$ make sanity SIMULATION=<your_fav_simulator>
```
The above will work.  But if you try to run the e40x, you will get compile-time failures:
```
$ cd ../../../cv32e40x/sim/uvmt
$ make sanity SIMULATION=<your_fav_simulator>
```
If you try the above on the `cv32e40x/dev` branch of `git@github.com:openhwgroup/core-v-verif.git`, you will get the opposite results: the 40x will compile and run correctly, but the e40s will not compile.

### Additional context
Affected files are:
```
lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
lib/uvm_agents/uvma_rvfi/uvma_rvfi_cfg.sv
lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
lib/uvm_agents/uvma_rvfi/uvma_rvfi_mon_trn_logger.sv
lib/uvm_agents/uvma_rvvi/uvma_rvvi_cfg.sv
lib/uvm_agents/uvma_rvvi_ovpsim/seq/uvma_rvvi_ovpsim_control_seq.sv
lib/uvm_agents/uvma_rvvi_ovpsim/seq/uvma_rvvi_ovpsim_control_seq_item.sv
[...]

PR #1146, opened 2022-01-25
Cv32e40x/merge master to rel 18

silabs-robin:
This PR is the "master_to_rel" step of the merge procedure.
The regression is green with xcelium (with certain known failures temporarily disabled).


MikeOpenHWGroup:
The two cv32e40p failures are "known" which I am prepared to waive (for now).   @silabs-robin, @silabs-hfegran are the three e40s failures waviable?

silabs-robin:
> The two cv32e40p failures are "known" which I am prepared to waive (for now). @ silabs-robin, @ silabs-hfegran are the three e40s failures waviable?

That relates to [this](https://github.com/openhwgroup/core-v-verif/issues/1110) issue.
Let's remember to discuss that at our meeting tomorrow.
My initial thoughts are that these 40p/40s failures should not block the merge to 40x's release branch, but we can wait til the meeting gives us a consensus.

silabs-robin:
For future reference:
I asked Henrik to merge this PR because I don't think the "40p working on 40x branch" should be a blocker for the merge procedure.
The 40p and 40s have some issues that are being tracked and fixed separately from this.
I think it will be easier to keep the common source in sync and fix potential future conflicts if we finish this merge instead of letting the sources deviate.

Issue #1123, opened 2022-01-18
Merge conflicts: "40p_release" -> "master"


silabs-robin:
As part of the merge procedure I merged "40p_release" into "master" (on my own fork), but there was merge conflicts.
@MikeOpenHWGroup could you have a look to see if I did my manual conflict resolution correctly?
Here is my branch https://github.com/openhwgroup/core-v-verif/compare/master...silabs-robin:silabs-robin/pr/master_merge_220118_40p
And the latest commit (the merge commit) is the one of interest https://github.com/openhwgroup/core-v-verif/commit/09b938b059681e332700d4516741b404743fa426.

--------

Here is a summary of the files I had to resolve conflicts for:

Problem: "docs/CodingStyleGuidelines.md" modified in "master" and "40p".
Which is correct?
I chose the latest of those two edits.
(Shall/Should/May/Can part.)

Problem: "README.md" modified in "master" and "40p".
Which is correct?
I did a best guess.
("## Getting Started" part.)

Problem: "cv32e40p/sim/README.md" modified in "master" and "40p".
Which is correct?
I chose the latest one.
("Directories" part.)

Problem: "cv32e40p/sim/core/.gitignore" modified in "master" and "40p".
Which is correct?
I chose the latest one.
("*_results" part.)

Problem: "cv32e40p/sim/core/Makefile" modified in "master" and "40p".
Which is correct?
I did a best guess, trying to not lose anything, trying to use latest commits.
(Several parts changed.)

silabs-robin:
I am continuing the `silabs-robin/pr/master_merge_220118_40p` merging in `silabs-robin/pr/master_merge_220118_40p_40x`, where I shall pull in from `cv32e40x/release`.
So if you do see something you want to change, I don't think it matters whether you PR to my `..._40p` or `..._40p_40x` branch because I can plan to continue on `..._40p_40x_40s` and I can just pull in changes from whichever of the previous ones AFAICT.

MikeOpenHWGroup:
Hello @silabs-robin, good work on this.

> Problem: "docs/CodingStyleGuidelines.md" modified in "master" and "40p".
> Which is correct?
> I chose the latest of those two edits.
> (Shall/Should/May/Can part.)

Yes, that is correct.

> Problem: "README.md" modified in "master" and "40p".
> Which is correct?
> I did a best guess.
> ("## Getting Started" part.)

I would call your "best guess" an "educated guess" as you chose the proper ones.

[...]

PR #1148, opened 2022-01-25
Cv32e40p/merge master to rel 18

silabs-robin:
This PR is the "master_to_rel" step of the merge procedure.
Note: I made this PR as I am doing the 40x/40s PRs already, but there are still known problems with the 40p regressions. So if we want, we could decide to close this PR and wait until that one is fixed on "master" first? (It is trivial to re-run the master_to_rel script so I can do it again later if desired.)


MikeOpenHWGroup:
As identified in #1123, there is an issue with this pull-request related to the core testbench.   It will be easier to fix it as a hotfix, so I will merge this in and fix it immediately following.  (A little unorthodox.)

Issue #410, opened 2020-11-30
Please add "compliance" target to riviera.mk


MikeOpenHWGroup:
Hi @dawidzim .  As of #406 the compliance regression has the ability to dump compliance test signatures and compare them to the reference signatures provided in compliance test-suite.   More information is available in the [UVM README](https://github.com/openhwgroup/core-v-verif/tree/master/cv32/sim/uvmt_cv32#risc-v-compliance-test-suite-and-regressions).

This update requires a simulator-specific `compliance` target to be added to each \<sim\>.mk   Can you add this for riviera.mk?   Check out xrun, dsim or vsim for a working example.

Thanks!

dawidzim:
Hey @MikeOpenHWGroup . Sure, I will do this

MikeOpenHWGroup:
Thanks @dawidzim.  #419 is merged in.

PR #1153, opened 2022-01-28
debug_assert: add checks for havereset/running/halted

silabs-robin:
This PR adds 3 checks to see that [this](https://cv32e40x-user-manual.readthedocs.io/en/latest/debug.html#debug-state) part of the user manual holds true.
I initially just wanted to see for my self, but I think it makes sense to include in the debug_assert assertion set.
@silabs-oysteink do you agree with these assertions?
(I realized the first assert is redundant vs the core SVAs, so I can remove it if we want, but for completeness of the debug_assert module we could keep it.)


silabs-oysteink:
This makes sense. Note that I just created issue #410 for the cv32e40x, which address the issue with the arc from debug_havereset to debug_halted.

PR #1155, opened 2022-01-31
Cv32e40s/merge rel to dev 4 manual

silabs-robin:
This is the 40s merge from "release" to "dev".

Passes "rel_check" on xcelium except for 2 failures (csr access tests).
I had to resolve part of this merge manually because there were conflicts, {uvmt_cv32e40s_tb_ifs.sv, uvmt_cv32e40s_tb.sv, uvmt_cv32e40s_fencei_assert.sv, ExternalRepos.mk}.
In my opinion, the csr access tests should not block this, because things are in development and those tests has been known to be fragile already and I think the merge is valuable and the csr tests can be regenerated and fixed in dev.

@silabs-hfegran


silabs-hfegran:
Overall this looks good to me - but hard to get an overview with the amount of commits and changes

MikeOpenHWGroup:
Yeah, these rel2dev merges _look_ big, but are not really.  Local regression is clean, so I will merge.

PR #1133, opened 2022-01-20
Silabs robin/pr/master merge 220118

silabs-robin:
Here is the merge of {`cv32e40p/release`, `cv32e40x/release`, `cv32e40s/release`, 40x->40s manual merge} into `master`.

It does not pass all `rel_check`s yet.
40x passes, but sporadically don't
40s had 2 failures in the last run (`cv32e40s_csr_access_test`, `cv32e40s_readonly_csr_access_test`).
40p has (consistently) 2 failures (`corev_rand_debug_single_step` twice).
The 40x/40s failures are ok as these are both WIP.
The 40p I would like to request some help with debugging.

Before merge, we should fix the 40p failures, review the 40x->40s manual merge, review that we accept the current 40x/40s failures, and review the overall diff/commitlog of the whole merge.
@MikeOpenHWGroup @silabs-hfegran @silabs-mateilga 

If it helps review, use github's compare feature on the step-wise [branches](https://github.com/silabs-robin/core-v-verif/branches/all) I used for this.
That is, branch `silabs-robin/pr/master_merge_220118`, `.../master_merge_220118_40p`, ..., `.../master_merge_220118_40p_40x_40s_x2s`.
For example, [here](https://github.com/silabs-robin/core-v-verif/compare/silabs-robin/pr/master_merge_220118_40p_40x_40s...silabs-robin:silabs-robin/pr/master_merge_220118) is the diff of my manual 40x->40s merging.

PR #1160, opened 2022-02-02
Cv32e40p/merge rel to dev 12

silabs-robin:
This is the 40p merge from "release" to "dev" branch, following the core-v-verif merge procedure that Steve wrote.
It was done by our Jenkins automated scripts and tested for the "rel_check" regression, where it passed all tests except corev_rand_debug_single_step.

--------

@MikeOpenHWGroup this one problem is why I submit this PR as a draft.

On the PR to "master" (https://github.com/openhwgroup/core-v-verif/pull/1133), there was reported 2 failures for the 40p "rel_check":
> 40p has (consistently) 2 failures (corev_rand_debug_single_step twice).

I think the latest update on it was this:
> I've determined that it is an environment, not DUT error and suspect that it is caused by the introducing of an updated toolchain. In my opinion it is not worth gating this merge for that as it will not impede the 40pv2 and I can resolve it on the master branch.

The reported problems with the merge conflicts were fixed https://github.com/openhwgroup/core-v-verif/issues/1123 .
But the failing test mentioned above has not been fixed yet.
Our main focus now is developing the 40x and 40s, but we should be careful about leaving the 40p in an unstable state.
@MikeOpenHWGroup do you have any proposals for how to solve this?
Are you not seeing this failure on your end, and we need to update the environment that we run this test in?


silabs-robin:
No point in keeping this open anymore. The linked issues/PRs will show the history for anyone (me?) digging this up in the future.

PR #1162, opened 2022-02-03
Pr/fix 1140

ZElkacimi:
following comments from @Silabs-ArjanB and @silabs-robin 
https://github.com/openhwgroup/core-v-verif/pull/1140#discussion_r792471796
https://github.com/openhwgroup/core-v-verif/pull/1140#issuecomment-1022975299


lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
```diff
@@ -74,24 +74,37 @@ endtask
...
-      join_any
-
-      // 3. Tell sequencer we're ready for the next sequence item
-      seq_item_port.item_done();
-      `uvm_info(info_tag, $sformatf("item_done"), UVM_LOW);
-    end
+   fork
+      begin
+         forever begin
+            repeat($urandom_range(uvma_cvxif_issue_ready_min,uvma_cvxif_issue_ready_max)) @(posedge cvxif_vif.clk);
```
MikeOpenHWGroup, line 80: Hi @ZElkacimi, I suggest we consider a different approach here as there are a couple of concerns:
1. The upper and lower bounds are constants, so they cannot be easily changed.
2. Using $urandom_range() in this way means that the cycle delay is not constrainable.  See Clause 5.18 of the [core-v-verif coding style guidelines](https://github.com/openhwgroup/core-v-verif/blob/master/docs/CodingStyleGuidelines.md).

I would recommend we do this "the UVM way":
1. Create a configuration class for the package.
2. Add random variables for anything you need to randomize.
3. Add soft constraints to define "reasonable values".

This will allow for a testcase to apply different issue_ready delays as necessary and (perhaps more importantly) preserve random stability across simulations.  A really good introduction to this is available [here](http://cluelogic.com/2012/01/uvm-tutorial-for-candy-lovers-configurations/).

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_constants.sv
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Zineb EL KACIMI (zineb.el-kacimi@external.thalesgroup.com)
+
+
+`ifndef __UVMA_CVXIF_CONSTANTS_SV__
+`define __UVMA_CVXIF_CONSTANTS_SV__
+
+
```
MikeOpenHWGroup, line 14: As stated above, I would recommend moving these to a configuration class for the CVXIF agent and applying soft constraints:

```
class uvma_cvxif_cfg_c extends uvm_object;

rand int unsigned uvma_cvxif_issue_ready;
rand int unsigned uvma_cvxif_issue_not_ready;

constraint reasonable_values {
    soft uvma_cvxif_issue_ready inside     {[4:10]};
    soft uvma_cvxif_issue_not_ready inside {[1:2]};
}
...
```

MikeOpenHWGroup (changes requested):
Looks good @ZElkacimi.  I've requested a couple of changes to better align with our coding style guidelines and for better run-time control.

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_constants.sv
ZElkacimi: thanks @MikeOpenHWGroup for your comment, I have pushed a new commit to add uvma_cvxif_cfg.sv and make issue_ready randomization configurable.

lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
```diff
@@ -74,24 +89,45 @@ endtask
...
-      join_any
-
-      // 3. Tell sequencer we're ready for the next sequence item
-      seq_item_port.item_done();
-      `uvm_info(info_tag, $sformatf("item_done"), UVM_LOW);
-    end
+   fork
+      begin
+         forever begin
+            int unsigned effective_ready = cfg.calc_random_ready();
```
MikeOpenHWGroup, line 95: These two lines could be replaced with `cfg.randomize();` if you make the change I suggest for `uvma_cvxif_cfg.sv`.

lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
```diff
@@ -74,24 +89,45 @@ endtask
...
+            int unsigned effective_not_ready = cfg.calc_random_not_ready();
+            if (effective_not_ready==0) begin
+               @(posedge cvxif_vif.clk);
+               break;
+            end
+            else begin
+                repeat(effective_ready) @(posedge cvxif_vif.clk);
```
MikeOpenHWGroup, line 102: You could use the values of the (randomized) configuration object directly:
```
begin
     repeat(cfg.effective_ready) @(posedge cvxif_vif.clk);
     cvxif_vif.cvxif_resp_o.x_issue_ready <= 0;
     repeat(cfg.effective_not_ready) @(posedge cvxif_vif.clk);
     cvxif_vif.cvxif_resp_o.x_issue_ready <= 1;
end
```
This eliminates the need to have local and cfg copies of the same member.

MikeOpenHWGroup (commented):
Thanks for making this update @ZElkacimi.  The changes you make should work, but actually create extra work for you.  I made a few suggestions that are closer to the "UVM way".  Following that structure is both easier and makes it a lot simpler to apply new constraints from a testcase.

lib/uvm_agents/uvma_cvxif/src/obj/uvma_cvxif_cfg.sv
```diff
@@ -0,0 +1,98 @@
...
+
+function int unsigned uvma_cvxif_cfg_c::calc_random_ready();
+
+   int unsigned effective_ready;
+   int unsigned effective_not_ready;
+
+
+   case (ready_mode)
+      UVMA_CVXIF_READY_FIX      : begin
+         effective_not_ready = 0;
```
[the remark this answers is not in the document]
ZElkacimi: No, it's effective_not_ready, when ready is fix that means that ready is always 1, in other words, effective_not_ready is 1 for "0 cycle"

lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
ZElkacimi: I thought about this solution, it works but the inconvenience of it is that if we add other variables to the cfg class that we want to be randomized only one time, in this case it will be also randomized many times in this forever begin. Please tell me if that's clear I search for this "cfg.randomize()" in the other agents and it's never used I think?
You can take a look to the uvma_obi_memory, search for calc_random_gnt_latency(), it's a perfect example that implements something very close.

ZElkacimi:
thank you @MikeOpenHWGroup for your comments, please tell me if my reasons to do it like this are convincing to you otherwise the cfg_randomize() is working and much simpler 'for now'.

ZElkacimi:
@JeanRochCoulon

lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
MikeOpenHWGroup: Typically this is handled by having multiple configuration objects.  These can be randomized independently.

MikeOpenHWGroup:
> okay @MikeOpenHWGroup I got your point: avoid using **$urandom()** and use **randomize()** instead :)

Thanks.  Much appreciated.

> do you prefer doing it like this "cfg.randomize(var);" or "cfg.randomize()" with the possibility to have many config objects ?

Good question.  I have a slight preference for the first option.  Our [coding style guidelines](https://github.com/openhwgroup/core-v-verif/blob/master/docs/CodingStyleGuidelines.md) do not provide guidance on this and I would be interested in your opinion.

ZElkacimi:
> do not provide guidance on this and I would be interested in your opinion.
For me, I don't know which one is uvm best practice but I go for the first option as I don't like having many config objects at the same agent and also because we know exactly what we randomize without referring to the config class.

ZElkacimi:
I have forced the push to edit the last commit with the requested changes, @MikeOpenHWGroup is this the right way to do it ?
I appreciate if you review my changes again, Thanks.

lib/uvm_agents/uvma_cvxif/src/comps/uvma_cvxif_drv.sv
```diff
@@ -74,24 +92,49 @@ endtask
...
-      `uvm_info(info_tag, $sformatf("item_done"), UVM_LOW);
-    end
+   fork
+      begin
+         forever begin
+            if (cfg.ready_mode == UVMA_CVXIF_ISSUE_READY_FIX) begin
+               @(posedge cvxif_vif.clk);
+               break;
+            end
+            else gen_random_ready();
```
MikeOpenHWGroup, line 102: If cfg.ready_mode != UVMA_CVXIF_ISSUE_READY_FIX and `gen_random_ready()` is not a time-consuming method then this will be zero-delay infinite loop.  It might use a good idea to have a hard constraint that ensure that `cfg.uvma_cvxif_issue_ready` and `cfg.uvma_cvxif_issue_not_ready` cannot be both set to 0.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for sticking with this @ZElkacimi.  Looks good.  I have made one minor comment, but have approved the PR.

I will leave it to @JeanRochCoulon to give the final review and merge.

ZElkacimi:
@JeanRochCoulon I have added the hard constraint requested by @MikeOpenHWGroup , you can now give your final review, thanks

JeanRochCoulon:
@MikeOpenHWGroup, your review allows to increase the UVM quality, and that is very good for everybody!   Great

JeanRochCoulon (commented):
When executing uvm, plenty of cvxif agent messages are displayed in the console (while my test does not exercize this agent). Do you know a solution to display them only when needed?

JeanRochCoulon (approved):
ok, I have verify, the cvxif messages were coming from my simulation environment. So no problem, I approve and merge. Thanks Zineb

PR #1165, opened 2022-02-04
Cv32e40s/merge dev to rel 8

silabs-robin:
Starting the next round of the merge procedure.
This PR is 40s "dev"->"release".
Passes "rel_check" on xcelium, except(!) for the two known problems in "csr_access_test" and "readonly_csr_access_test".

I think we should do this merge despite the two known problems.
The 40s is WIP.
And I think frequent merges has a higher value than waiting for everything to be fixed and pristine (potentially inviting new failures in the meantime).
We can wait, if anyone think it is better to have the "rel_check" passing 100% first.
@silabs-hfegran @silabs-mateilga @MikeOpenHWGroup 

--------

Note.
The 40p didn't completely finish the previous round (see https://github.com/openhwgroup/core-v-verif/pull/1160).
But I think the 40s and 40x can do one round just the two of them.
The "rel_check"s for all 3 cores will be done on "master", so the 40p will still be tested.
Doing the merges frequently like this should make each diff smaller and much easier to handle.


silabs-robin:
Protip: When viewing the diff in the "Files changed" tab, hold the "Alt" key while pressing the "v"-shaped hide button on one of the files, to collapse all of the files.

(I learned this last week and it made it easier to see which files where touched by these big merges.)

MikeOpenHWGroup (approved):
Thanks @silabs-robin.   The known failing tests notwithstanding, I am going to merge this in.  The `release` branches are actually "staging" branches, and it is far better to keep the size of these staging events small.

PR #1166, opened 2022-02-04
Cv32e40x/merge dev to rel 58

silabs-robin:
Starting the next round of the merge procedure.
This PR is 40x "dev"->"release".
Passes "rel_check" on xcelium (two tests are still commented out, but will be back in in later merges).


MikeOpenHWGroup (approved):
Wow, only 16 files in a `dev` to `rel` review.  Almost too easy.   Thanks @silabs-robin

PR #1167, opened 2022-02-04
uncomment temporarily disabled tests in 40x rel_check

silabs-robin:
This PR re-enables some tests that were commented out in relation to the merge procedure and known problems.

The tests actually did pass now, but that should not be a criteria for re-enabling them anyway.
Passes ci_check (this change shouldn't impact ci_check anyway, but for good measures).


silabs-robin:
I noticed I had forgotten the commit signature.
So I did `git commit --amend -s` and `git push --force`.

MikeOpenHWGroup (approved):

PR #1175, opened 2022-02-09
Improvement of dret_dpc assert

silabs-mateilga:
split dret_dpc assert to increase coverage, converted to rvfi for readability

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
@@ -146,7 +146,7 @@ interface uvmt_cv32e40x_debug_cov_assert_if
     input         rvfi_valid,
     input  [31:0] rvfi_insn,
     input         rvfi_intr,
-    input         rvfi_dbg,
+    input  [2:0]  rvfi_dbg,
     input         rvfi_dbg_mode,
     input  [31:0] rvfi_pc_wdata,
     input  [31:0] rvfi_pc_rdata,


cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
```diff
@@ -144,8 +144,15 @@ interface uvmt_cv32e40x_debug_cov_assert_if
     input         fetch_enable_i,
 
     input         rvfi_valid,
+    input  [31:0] rvfi_insn,
+    input         rvfi_intr,
+    input         rvfi_dbg,
```
silabs-robin, line 149: Should this signal be [2:0]?

It seems like it isn't used, so maybe we could either extend the width or remove it.

cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -394,28 +394,36 @@ module uvmt_cv32e40x_debug_assert
...
+
+    property p_dmode_dret_pc_int;
+        int dpc; (1, dpc =cov_assert_if.rvfi_csr_dpc_rdata)
+        ##0(cov_assert_if.rvfi_valid && cov_assert_if.rvfi_dbg_mode && cov_assert_if.rvfi_insn == DRET_INSTR_OPCODE)
+
+        ##1 cov_assert_if.rvfi_valid[->1] 
+        ##0 (cov_assert_if.rvfi_intr && !cov_assert_if.rvfi_dbg_mode)
+        |->
+
+        (cov_assert_if.rvfi_csr_mepc_wdata & cov_assert_if.rvfi_csr_mepc_wmask) == dpc;
```
silabs-robin, line 421: This split of "int" and "!int" looks very clean.
Looks good.

It might be that Verissimo is able to spot the similarities of the two properties.

silabs-robin (changes requested):
The only thing is the "rvfi_dbg" width. Everything else looks good.

cv32e40x/tb/uvmt/uvmt_cv32e40x_tb_ifs.sv
silabs-mateilga: fix commited

silabs-robin (approved):

PR #1179, opened 2022-02-10
uvmt_cva6_dut_wrap.sv, uvmt_cva6_tb.sv: fix parameter declaration

JeanRochCoulon:
Fix the uvm testbench parameter declaration. Down size the number from 2**25 downto 2**24 because verilog array cannot be greater than 2**24 

Signed-off-by: Jean-Roch Coulon <jean-roch.coulon@thalesgroup.com>

### DIFF
--- a/cva6/tb/uvmt/uvmt_cva6_dut_wrap.sv
+++ b/cva6/tb/uvmt/uvmt_cva6_dut_wrap.sv
@@ -27,10 +27,11 @@ module uvmt_cva6_dut_wrap # ( parameter int unsigned AXI_USER_WIDTH    = 1,
                             output ariane_rvfi_pkg::rvfi_port_t rvfi_o
                            );
 
-    cva6_tb_wrapper #(     .AXI_USER_WIDTH    (1),
-     .AXI_ADDRESS_WIDTH (64),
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
@ZElkacimi, this PR will fix the UVM testbench, could you have a look and if possible approve it ?

MikeOpenHWGroup:
> verilog array cannot be greater than 2**24

I have not seen that in the SystemVerilog [LRM](https://ieeexplore.ieee.org/document/8299595). (Of course, the LRM is _huge_, so maybe I missed it).  But it is clear that the [sram model](https://github.com/JeanRochCoulon/core-v-verif/blob/feature/parameters/cva6/tb/core/tb_components/cva6_core_tb_sram.sv#L100) in the testbench is getting big.   I suggest using a sparse array (SystemVerilog associative array) to model the SRAM.
```
// 1 GiB memory modelled as an associative array (aka sparse memory):
// 2**28, 32-bit words
bit [32:0] sparse_memory [bit[27:0]];
```

ZElkacimi:
Hi @JeanRochCoulon, I took a look to the commit and now it seems great.

ZElkacimi (approved):

PR #1181, opened 2022-02-14
Cv32e40x/merge rel to dev 15

silabs-robin:
Merge procedure "rel->dev" 40x.

Passes "rel_check" w/ Xcelium, except for the 4 failures below.

There are known failures:
`cv32e40x_csr_access_test`, `corev_rand_debug_ebreak` (both fail in "full" regression on "dev" already).
`debug_test`, `corev_rand_interrupt_debug` (both fail on the known debug_assert problem).

@silabs-hfegran , @silabs-mateilga , I want to point out the known failures above, so you are aware (and in case you want to object to the merge).


silabs-hfegran (approved):
Looks good to me

PR #1183, opened 2022-02-14
uvma_cvxif_cfg.sv: Do not randomize the issue_ready, by default

ZElkacimi:
Fix the issue_ready by default in config class, because its randomization is not yet supported by cva6.


JeanRochCoulon (changes requested):
As the configuration could be fine-tuned by user, would it be possible to add a comment to inform which functionality is randomized when configuring this parameter?

JeanRochCoulon (approved):
Thanks for the comment

PR #1187, opened 2022-02-17
Buserr scoreboard bug fixes

silabs-mateilga:
Changes made to make buserr regression pass:

- Changed the rvfi instruction monitor to use mcause rdata in stead of "get_csr_retirement_data()" when determining interrupt cause.
- Allowed buserr_sb to discount buserror instruction matches during debug, as we are not able to observe entries in to exception handler while in debug. This will be changed when RVFI extends intr signal to include cause.
- Temporarily disabled a_dmode_dret_pc_int to remove regression noise until RVFI accounts for kill signals.


cv32e40x/tests/programs/corev-dv/corev_rand_instr_obi_err/test.yaml
```diff
@@ -10,4 +10,4 @@ plusargs: >
     +obi_memory_instr_random_err_enabled
     +obi_memory_instr_one_shot_err_enabled
     +gen_irq_noise
-    +buserr_sb_disabled
+
```
silabs-hfegran, line 13: not a blocker, but you might want to change your editor configuration to not give these additions to PRs

silabs-robin (approved):
Discounting the errmatches in debug was a smart solution!
(When multi-bit "rvfi_intr" comes up, maybe we want to keep both checks?)

I agree that "mcause" should be checked on "rdata" in the case of interrupts, because interrupts happen in between rvfi retires and hence we want to look at the before-state.
Both the sb and rvfi monitor should indeed have this.
(The comment in the monitor actually already says this, I didn't notice before now.)

The SB is also re-enabled in the yaml files, so that looks good.

(I don't fully like that "cnt_rvfi_errmatch_debug" has another signal name that is a subset of this name, but that is just my subjective preference.)

This PR looks good to me.

cv32e40x/tests/programs/corev-dv/corev_rand_instr_obi_err/test.yaml
silabs-mateilga: fixed in last commit

PR #1194, opened 2022-02-18
Removed atomics support for cv32e40s cvverif

silabs-hfegran:


cv32e40s/env/corev-dv/cv32e40s_pma_instr_lib.sv
```diff
@@ -720,311 +720,3 @@ class corev_jalr_pma_undefined_region_instr extends corev_jalr_pma_instr;
   endfunction : post_randomize
 endclass : corev_jalr_pma_undefined_region_instr
 
-// #############################################################################
```
silabs-robin, line 723: Wrt the merge procedure...
Should we maybe add a comment like `// NB! 40s doesn't have the atomic PMA bit` at this location?
During manual 40x->40s merge, this would clearly show up in a diff, so manual diff-review and merge would get a good hint to not accept the whole diff.

silabs-robin (approved):
I had just that one comment.
Otherwise, I recognize the lines that were blocking "make comp_corev-dv" and I agree with the fix.

PR #1198, opened 2022-02-21
Fix Verissimo integration

gabrielrad:
Hi @MikeOpenHWGroup ! These are the changes needed in order to make Verissimo work again. Please let me know what you think or have better ideas.

### DIFF
--- a/cv32e40p/sim/uvmt/Makefile
+++ b/cv32e40p/sim/uvmt/Makefile
@@ -48,5 +48,9 @@ MAKE_PATH            := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))
 export CORE_V_VERIF   = $(abspath $(MAKE_PATH)/../../..)
 export CV_CORE       ?= cv32e40p
 
+DPI_DASM_SPIKE_REPO  ?= https://github.com/riscv/riscv-isa-sim.git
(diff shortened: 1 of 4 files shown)


cv32e40p/sim/uvmt/Makefile
```diff
@@ -48,5 +48,9 @@ MAKE_PATH            := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))
 export CORE_V_VERIF   = $(abspath $(MAKE_PATH)/../../..)
 export CV_CORE       ?= cv32e40p
 
+DPI_DASM_SPIKE_REPO  ?= https://github.com/riscv/riscv-isa-sim.git
+CV_SW_TOOLCHAIN      ?= /opt/riscv
+CV_SW_PREFIX         ?= " "
```
MikeOpenHWGroup, line 53: I'm not sure about this one, but will bring it in for now.  It will be fine for users who have set up their shell environment properly.  A few experiments will be needed to see how it affects users who have not set `CV_SW_PREFIX`.

MikeOpenHWGroup (approved):

cv32e40p/sim/uvmt/Makefile
gabrielrad: Yes, I am hopping for a better alternative in the future.

PR #1201, opened 2022-02-22
Fixed outdated core interfaces

silabs-hfegran:
bs.com>


silabs-robin (approved):
If num mhpmcounters is gone on the 40s, then this diff looks like standard procedure.

PR #1202, opened 2022-02-22
Update e40x hash and tb after RTL renaming, pma_region_t -> pma_cfg_t

silabs-oivind:


silabs-hfegran:
Thanks @silabs-oivind, In general search/replace only in core specific folder for core-v-verif will not suffice - there are top level dependencies that  needs to be checked as well. 

Without these changes, corev-dv fails to compile. I've added a commit that fixes these issues.

Issue #1193, opened 2022-02-18
OVPSIM mret does not set MIE = MPIE in MSTATUS CSR


silabs-mateilga:
For a few select seeds when running the "corev_rand_interrupt_wfi_mem_stress" test, what seems to be a bug in the reference model can be observed. The scenario is that when a core is woken from a wfi instruction by a disabled interrupt, and the next instruction is an mret instruction, the MIE bit is not set to MPIE in the MSTATUS CSR, which is in breach with the 1.12 privileged spec.

Steps to Reproduce:
Branch: cv32e40x/dev  https://github.com/openhwgroup/core-v-verif/tree/cv32e40x/dev
Hash: fe9bdeff9178adbff54ca5591e71e08c757bffae

Command line path: 
core-v-verif/cv32e40x/sim/uvmt/

Run commands:
make comp_corev-dv gen_corev-dv TEST=corev_rand_interrupt_wfi_mem_stress CFG=default SIMULATOR=xrun USE_ISS=YES COV=YES RUN_INDEX=27 GEN_START_INDEX=27 RNDSEED=-844432648

make test TEST=corev_rand_interrupt_wfi_mem_stress CFG=default SIMULATOR=xrun USE_ISS=YES COV=YES RUN_INDEX=27 GEN_START_INDEX=27 RNDSEED=-844432648

Waveform showing the difference in the MSTATUS register between RVFI and RVVI:
![image](https://user-images.githubusercontent.com/59358563/154660714-2cb9c087-3645-4178-9212-377b3152f448.png)

@eroom1966 Could you take a look at this?

eroom1966:
@silabs-mateilga 
Having trouble reproducing your steps, this is what I did after setting my environment to run xcelium

$ git clone https://github.com/openhwgroup/core-v-verif
$ cd core-v-verif
$ git checkout cv32e40x/dev
$ git reset fe9bdef --hard
$ make comp_corev-dv gen_corev-dv TEST=corev_rand_interrupt_wfi_mem_stress CFG=default SIMULATOR=xrun USE_ISS=YES COV=YES RUN_INDEX=27 GEN_START_INDEX=27 RNDSEED=-844432648

make: *** No rule to make target 'comp_corev-dv'. Stop.

Please advise
Thx
Lee

silabs-mateilga:
Ah, apologies, I sometimes forget how much of the environment is "just there". 

[...]

PR #1215, opened 2022-03-01
update to add rvvi.sleep

eroom1966:
@MikeOpenHWGroup @silabs-mateilga 
Hi Mike, Marton

As promised I have a PR which includes a sleep signal in the RVVI interface, lets see if this compared with the one generated in the RVFI when implemented


vendor_lib/imperas/imperas_DV_COREV/ChangeLog.md
```diff
@@ -1,6 +1,6 @@
 CV32E40P Change Log
```
MikeOpenHWGroup, line 1: This changelog mentions the E40P, but I'm guessing it applies to all E40 cores.   You can fix that in a future PR if you like.

MikeOpenHWGroup:
AFAICT, this is not intended to fully resolve #1193, but i did a local merge and ran both a ci_check (passed) and 100 runs of `corev_rand_interrupt_wfi_mem_stress`.  There were three failures, all with the same signature: after a number of instructions (43,545, 162,359 and 7,454) the tests hang and eventually is terminated by the Global timeout.   I have not investigated to see what caused the hang.  There were no miscompares in any of the tests.

eroom1966:
@MikeOpenHWGroup I have also added the deferint fix into the haltreq code that drives the signals into the REF model

silabs-mateilga (approved):
The changes looks good, and no new or unknown issues on a full regression run with xrun.

PR #1216, opened 2022-03-02
Added PMP assertions

silabs-hfegran:


silabs-robin (approved):
Approving based on that we reviewed and discussed yesterday.

PR #1221, opened 2022-03-03
Fixed for addition of PMA integrity bit. Fixed for mimpid pin change

Silabs-ArjanB:

### DIFF
--- a/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
+++ b/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
@@ -94,6 +94,9 @@
    rand bit [MAX_XLEN-1:0]       mhartid;
    bit                           mhartid_plusarg_valid;
 
+   rand bit [MAX_XLEN-1:0]       mimpid;
+   bit                           mimpid_plusarg_valid;
+
    rand bit [MAX_XLEN-1:0]       mimpid_patch;
    bit                           mimpid_patch_plusarg_valid;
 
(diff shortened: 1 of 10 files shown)


Silabs-ArjanB:
See also https://github.com/openhwgroup/core-v-verif/issues/1222

lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_agent.sv
```diff
@@ -160,7 +160,7 @@ function void uvma_rvvi_ovpsim_agent_c::configure_iss();
    // Boot strap pins
    // -------------------------------------------------------------------------------------
    $fwrite(fh, $sformatf("--override root/cpu/mhartid=%0d\n", cfg.core_cfg.mhartid));
-   $fwrite(fh, $sformatf("--override root/cpu/mimpid=%0d\n", cfg.core_cfg.mimpid));
+   $fwrite(fh, $sformatf("--override root/cpu/mimpid=%0d\n", cfg.core_cfg.mimpid_patch));
```
silabs-hfegran, line 163: This is the expected mimpid-value we provide to the ISS, thus it should be a _full_ 32-bit mimpid - in its current form the 4-bit value is cast to a 32 bit value. If the other bits are non-zero this will cause mismatches with what the ISS expects.

In other words, the configuration object  needs to be aware of the full mimpid - ideally we should be able to set that correctly based on the bits that may be overridden combined with what the core is hard coded to. 

-> configuration should set mimpid as a combination of the patch + major +minor-fields.

lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
```diff
@@ -159,7 +159,7 @@
       `uvm_field_int(                          num_mhpmcounters               , UVM_DEFAULT          )
       `uvm_field_array_object(                 pma_regions                    , UVM_DEFAULT          )
       `uvm_field_int(                          mhartid                        , UVM_DEFAULT          )
-      `uvm_field_int(                          mimpid                         , UVM_DEFAULT          )
+      `uvm_field_int(                          mimpid_patch                   , UVM_DEFAULT          )
```
silabs-hfegran, line 162: This reports the current settings to the log/stdout, thus I think it would be better to retain the full mimpid-field here (and optionally add the patch-field to explicitly show what the user set)

silabs-hfegran (changes requested):

MikeOpenHWGroup (approved):
Looks good to me.  @silabs-hfegran, you should have final sign-off.

silabs-hfegran (approved):

PR #1245, opened 2022-04-04
initial additions for pulp

eroom1966:
@MikeCovrado 
This PR contains the new core module which will contain the pulp instruction additions.
I needed to change the way the testbench schedules the deferint signal, to ensure it is used for both debug and interrupt assignments
I can only run the ci_check examples, so I am unsure if there are any side effects


MikeOpenHWGroup (approved):
Passes a release regression (two know failures).

Issue #1252, opened 2022-04-18
Update core-v-verif to support latest riscv-arch-tests infrastructure


M-Ijaz-10x:
Hi,

core-v-verif verification environment needs up-gradation in order to support the latest [riscv-arch-tests](https://github.com/riscv-non-isa/riscv-arch-test) infrastructure as core-v-verif is working at an older hash of riscv-arch-tests

**Why do we need to update**
The framework of the latest [riscv-arch-tests](https://github.com/riscv-non-isa/riscv-arch-test) is completely changed that is updated directory structure, updated macros, updated tests structure, etc. For testing any newly added extension in any of the core which uses core-v-verif as its verification environment, the latest compliance tests need to be run to test the functionality, (e.g. while adding support for bitmanip in the CVA6 core I needed bitmanip tests, which are present in updated riscv-arch-tests framework, but the core-v-verif CVA6 environment was not compatible with those tests).

**NOTE:**
As the current default branch [master] of [riscv-arch-tests](https://github.com/riscv-non-isa/riscv-arch-test) will be completely replaced with branch [riscof-dev](https://github.com/riscv-non-isa/riscv-arch-test/tree/riscof-dev) on **1st MAY 2022**. This transition essentially migrates the current framework to the RISCOF framework. This probably will break the core-v-verif environment and it will have to be updated. cc @MikeOpenHWGroup

M-Ijaz-10x:
I have made the following changes in order to use the latest riscv-arch-tests(master branch) in core-v-verif CVA6 environment

- Updated the hash of the riscv-arch-tests in the core-v-verif CVA6 environment
- Updated the testlist.yaml file with the current directory structure of the riscv-arch-tests
- While running tests on the core-v-verif environment the spike ISS simulation is malfunctioning, or more specifically the spike is using the pc for DATA_BEGIN section instead of CODE_BEGIN section, right now I am using a hack i.e --pc=’h8000_0000 but this needs to be fixed properly
- While logging the results of the spike and verilated model of CVA6, there is a mismatch of one instruction, i.e. the spike starts logging from pc value of 0x80000000 and the verilated model of CVA6 starts logging from the pc value of 0x80000004, hence the tests are failing in comparison of spike and model results, so either spike or verilator trace python script of CVA6 have to be updated - which needs to be fixed
- Updated the regex for ISA check in the cva6.py to identify bitmanip as a legal ISA extension
[...]

PR #1254, opened 2022-04-21
[Issue-1252] Added the support to run latest riscv-arch-tests in core-v-verif CVA6 infrastructure

fatimasaleem:
This PR solves the issue mentioned in [1252](https://github.com/openhwgroup/core-v-verif/issues/1252). 
I have updated the cva6 environment to make it compatible to run the latest riscv-arch-tests with the riscof-dev infrastructure.

### What was Changed

The following major changes were made while setting up the cva6 environment

- Created a new [testlist yaml](https://github.com/10x-Engineers/core-v-verif/blob/3ca88a5fda05424e4221b5c66d2596ac059f3bdb/cva6/tests/testlist_riscv-arch-tests-cv64a6_imafdc_sv39.yaml) file and a new install script for riscv-arch-test as  @ASintzoff [suggested](https://github.com/openhwgroup/core-v-verif/issues/1252#issuecomment-1103530833)
        - [dv-riscv-arch-tests.sh ](https://github.com/10x-Engineers/core-v-verif/blob/3ca88a5fda05424e4221b5c66d2596ac059f3bdb/cva6/regress/dv-riscv-arch-tests.sh) sets all the environment variables and then calls all required install scripts
        - Created an [install-riscv-isa-sim.sh ](https://github.com/10x-Engineers/core-v-verif/blob/3ca88a5fda05424e4221b5c66d2596ac059f3bdb/cva6/regress/install-riscv-isa-sim.sh) script which is installing riscv-isa-sim as in riscof-dev the riscv-target has been removed, and for using the spike simulator as an architectural test model, we needed the required dependency files (e.g model_test.h)
        - Created a separate [testlist_riscv-arch-tests-cv64a6_imafdc_sv39.yaml](https://github.com/10x-Engineers/core-v-verif/blob/3ca88a5fda05424e4221b5c66d2596ac059f3bdb/cva6/tests/testlist_riscv-arch-tests-cv64a6_imafdc_sv39.yaml) file for running the riscv-arch-tests on cva6 with proper path and env variable setup, major additions are as follows
                     1. Added compile-time macros, XLEN, TEST_CASE_1 - as suggested by @neelgala
                     2. Updated the path of the env according to the riscof-dev dir structure
                     3. Updated the arch-test target to use the locally installed spike
 
- Updated the [cva6_spike_log_to_trace_csv.py](https://github.com/10x-Engineers/core-v-verif/blob/8fc3143c3c14973e123bdb52a6c92f1ce9b1f115/cva6/sim/cva6_spike_log_to_trace_csv.py) so that spike logging is aligned with CVA6 verilated model logging
- Replaced the [cva6/sim/link.ld](https://github.com/10x-Engineers/core-v-verif/blob/8870dc8e4e7bc2f0f3b32485b8e72cb6dc552751/cva6/sim/link.ld) with the latest one

### Environment Variables
```
export RISCV=/path/to/tootchain
export RISCV_GCC=$RISCV/bin/riscv64-unknown-elf-gcc
export RISCV_OBJCOPY=$RISCV/riscv64-unknown-elf/bin/objcopy
export SPIKE_ROOT=$RISCV
```
### Steps to Run
After all these changes the CVA6 core-v-verilf environment is running all the current riscv-arch-tests on the riscof-dev branch.
Steps to run:
```
git clone git@github.com:10x-Engineers/core-v-verif.git
cd core-v-verif
git checkout issue-1252-fix
source cva6/regress/dv-riscv-arch-test.sh

### CODE REVIEW

File: cva6/regress/dv-riscv-arch-tests.sh [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+
+export RISCV_ISA_SIM=https://github.com/riscv-software-src/riscv-isa-sim.git
+export RISCV_ISA_SIM_BRANCH=master
+export RISCV_ISA_SIM_HASH=b9fc8e4e9087a6064dfcc627efabbe3fd4bdc309
+
+# install the required tools
+source ./cva6/regress/install-cva6.sh
+source ./cva6/regress/install-riscv-dv.sh
+source ./cva6/regress/install-riscv-isa-sim.sh
+source ./cva6/regress/install-riscv-compliance.sh
```
ASintzoff, line 41: Please create install-riscv-arch-test.sh

cva6/regress/dv-riscv-arch-tests.sh
```diff
@@ -0,0 +1,53 @@
...
+if ! [ -n "$DV_TARGET" ]; then
+  DV_TARGET=cv64a6_imafdc_sv39
+fi
+
+if ! [ -n "$DV_SIMULATORS" ]; then
+  DV_SIMULATORS=veri-testharness,spike
+fi
+
+cd cva6/sim
+python3 cva6.py --testlist=../tests/testlist_riscv-arch-tests-$DV_TARGET.yaml --target $DV_TARGET --iss_yaml=cva6.yaml --iss=$DV_SIMULATORS $DV_OPTS
```
ASintzoff, line 52: Please replace testlist_riscv-arch-tests by testlist_riscv-arch-test (as in the test suite repository name).

cva6/regress/dv-riscv-arch-tests.sh
```diff
@@ -0,0 +1,53 @@
...
+export COMPLIANCE_BRANCH=riscof-dev
+export COMPLIANCE_HASH=7907c462c700279c5d75ec5e6042f762dcb95a25
+
+export DV_REPO=https://github.com/google/riscv-dv.git
+export DV_BRANCH=master
+export DV_HASH=0ce85d3187689855cd2b3b9e3fae21ca32de2248
+
+export RISCV_ISA_SIM=https://github.com/riscv-software-src/riscv-isa-sim.git
+export RISCV_ISA_SIM_BRANCH=master
+export RISCV_ISA_SIM_HASH=b9fc8e4e9087a6064dfcc627efabbe3fd4bdc309
```
ASintzoff, line 35: For the other test suites, environment variables are not defined in such script. Default values are provided in the installation scripts and can be overwritten by defining them before sourcing this file.

cva6/tests/testlist_riscv-arch-tests-cv64a6_imafdc_sv39.yaml
```diff
@@ -0,0 +1,736 @@
...
+# --------------------------------------------------------------------------------
+
+#- import: <riscv_dv_root>/target/rv64imc/testlist.yaml
+
+- test: rv64im-add-01
+  iterations: 1
+  path_var: TESTS_PATH
+  path_root: ROOT_PROJECT
+  gcc_opts: "-DXLEN=64 -DTEST_CASE_1=True -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -I<path_var>/riscv-compliance/riscv-test-suite/env/ -I<path_var>/riscv-isa-sim/arch_test_target/spike/"
+  asm_tests: <path_var>/riscv-compliance/riscv-test-suite/rv64i_m/I/src/add-01.S
```
ASintzoff, line 40: Please replace riscv-compliance by riscv-arch-test (accordingly with creation of install-riscv-arch-test.sh) in the whole file.

ASintzoff (changes requested):
Please rename dv-riscv-arch-tests.sh in dv-riscv-arch-test.sh (as the test suite repository name).

ASintzoff (changes requested):
Why are you changing the end address of trampoline in spike?

ASintzoff (changes requested):
As link.ld is used by all tests for cva6, I'm wondering if they are still buildable and executable.

Are you facing any trouble with the previous version?
What is the source of the version you propose?

cva6/sim/.gitignore
```diff
@@ -0,0 +1 @@
+*.txt
```
ASintzoff, line 1: Any reason to create this ignore file?

ASintzoff (changes requested):
Last year, we faced lot of issues with verilator versions greater than 4.110. This is the reason to use it. If a recent version is running riscv-tests, riscv-compliance, and riscv-arch-test test suite, I would be pleased to switch.

For the moment, I prefer to not modify install-cva6.sh and if required for riscv-arch-test, install a recent version of verilator and to use it, modify VERILATOR_ROOT variable outside the scripts.

ASintzoff (changes requested):
As said on previous commit, could you confirm that all the existing tests are runnable with this newest verilator?

cva6/regress/dv-riscv-arch-tests.sh
fatimasaleem: Sure, I will create an `install-riscv-arch-test.sh`. But is there any specific reason for which we are going to need `install-riscv-compliance.sh`? As the same tests have been modified in the `riscv-arch-test`, and the older version of the test infrastructure is no longer compatible with the spike current infrastructure.

fatimasaleem:
> Why are you changing the end address of trampoline in spike?

As also mentioned in the issue,

> While logging the results of the spike and verilated model of CVA6, there is a mismatch of one instruction, i.e. the spike starts logging from pc value of 0x80000000 and the verilated model of CVA6 starts logging from the pc value of 0x80000004, hence the tests are failing in comparison of spike and model results, so either spike or verilator trace python script of CVA6 have to be updated -

That's why I have updated the spike end address of trampoline according to the end address of the [verilator trampoline](https://github.com/openhwgroup/core-v-verif/blob/8fc3143c3c14973e123bdb52a6c92f1ce9b1f115/cva6/sim/verilator_log_to_trace_csv.py#L117)

fatimasaleem:
> As said on previous commit, could you confirm that all the existing tests are runnable with this newest verilator?

All the tests are passing with the new version of the verilator. But if required, we can keep the older version(4.110) as it is for the time being.

ASintzoff:
> > Why are you changing the end address of trampoline in spike?
> 
> As also mentioned in the issue,
> 
> > While logging the results of the spike and verilated model of CVA6, there is a mismatch of one instruction, i.e. the spike starts logging from pc value of 0x80000000 and the verilated model of CVA6 starts logging from the pc value of 0x80000004, hence the tests are failing in comparison of spike and model results, so either spike or verilator trace python script of CVA6 have to be updated -
> 
> That's why I have updated the spike end address of trampoline according to the end address of the [verilator trampoline](https://github.com/openhwgroup/core-v-verif/blob/8fc3143c3c14973e123bdb52a6c92f1ce9b1f115/cva6/sim/verilator_log_to_trace_csv.py#L117)

Thanks to highlight this issue in the conversion scripts. After looking at them, I think the root cause is in the verilator script. This script is not writing executed instruction at 0x8000_0000 from .log file into .csv one. It will be better to fix the verilator script to be sure we don't drop 0x8000_0000 address.

cva6/regress/dv-riscv-arch-tests.sh
ASintzoff: I would prefer to keep scripts for riscv-compliance for a few weeks/months to give people time to move all existing continuous integration for riscv-compliance to riscv-arch-test.
Regarding test infrastructure, as it is also used for other test suites and test environment, therefore, any change to it may not break them.

fatimasaleem:
> As link.ld is used by all tests for cva6, I'm wondering if they are still buildable and executable.
> 
> Are you facing any trouble with the previous version? What is the source of the version you propose?

Yes, I was facing a problem with compiling the test with the spike using the older version of link.ld, it was mainly because the spike linker has been updated with the [new entry point ](https://github.com/riscv-software-src/riscv-isa-sim/blob/b9fc8e4e9087a6064dfcc627efabbe3fd4bdc309/arch_test_target/spike/link.ld#L2) but riscv-dv linker is an outdated version.

fatimasaleem:
@ASintzoff I have made all the required changes, are there still some changes you would like to mention; which are blocking this PR? Or can we merge this PR in cv6/dev? cc @MikeOpenHWGroup

ASintzoff:
> @ASintzoff I have made all the required changes, are there still some changes you would like to mention; which are blocking this PR? Or can we merge this PR in cv6/dev? cc @MikeOpenHWGroup

Some comments:
- It seems Spike is not built when using dv-riscv-arch-test.sh script (only git clone is done by install-riscv-isa-sim.sh script).
- The new linker script prevents correct builds of riscv-tests test cases as this test suite has no rvtest_entry_point. A binary is built but as the entry point is incorrect, therefore execution does not start at the correct place.

fatimasaleem:
>      It seems Spike is not built when using dv-riscv-arch-test.sh script (only git clone is done by install-riscv-isa-sim.sh script).

Yes, I added install-riscv-isa-sim.sh script to only clone riscv-isa-sim directory to use the latest model_test.h source file, which was previously being used from the riscv-target folder of riscv-arch-tests repo, but in riscof-dev the riscv-target has been removed, and for using the spike simulator as an architectural test model we needed the required dependency file.

I can also invoke the build of spike as well in the same script when the $SPIKE env variable is not set.

fatimasaleem:
>     The new linker script prevents correct builds of riscv-tests test cases as this test suite has no rvtest_entry_point. A binary is built but as the entry point is incorrect, therefore execution does not start at the correct place.

As, I already have mentioned above that

>  the same tests have been modified in the riscv-arch-test, and the older version of the test infrastructure is no longer compatible with the spike current infrastructure.

And the new test suite (riscof-dev infrastructure) is using the `rvtest_entry_point` instead of `_start`, without getting the proper entry point the spike cannot differentiate between the pc for `DATA_BEGIN` section and `CODE_BEGIN` section.

We clearly cannot maintain two different linkers for spike at the same time. If we are to migrate to newer (riscof) infrastructure then I think we need to use the latest linker entry point as it is being used by the riscof framework, it will replace the current framework of riscv-arch-test soon. I am open to any other suggestions, that how can we maintain the older linker version.

MikeOpenHWGroup:
> If we are to migrate to newer (riscof) infrastructure then I think we need to use the latest linker entry point as it is being used by the riscof framework, it will replace the current framework of riscv-arch-test soon. I am open to any other suggestions, that how can we maintain the older linker version.

I am not in favour of changing our testbench infrastructure to use the RISCOF framework.  The proper solution here is to create a framework in which we can select things like the linker entry points on a _per test-program_ basis.  The CV32E4 testbenches can do this, and CVA6 should adopt something like that (see slide 15 of the [Writing Tests for CORE-V-VERIF](https://docs.google.com/presentation/d/18IHmCRNLtPp16tTtUlwtVCwZwDvlgtmu/edit#slide=id.g1212ca9b51e_0_0) presentation I presented a couple of weeks ago).

fatimasaleem:
> The new linker script prevents correct builds of riscv-tests test cases as this test suite has no rvtest_entry_point. A binary is built but as the entry point is incorrect, therefore execution does not start at the correct place

@ASintzoff please check that the riscv-tests test cases are now working correctly with the latest changes.

fatimasaleem:
@ASintzoff I am not changing the existing linker anymore but added an argument in the cva6.py through which I have an option to select a different linker and by default, things can be run as they were running before.

Please let me know if still there is anything that needs more modification or addition. cc @MikeOpenHWGroup

fatimasaleem:
Pinging for merge approval.

ASintzoff (changes requested):
Thanks for the PR and all the modifications you did based on my comments.
I have just a new one on your latest commit.

cva6/regress/dv-riscv-arch-test.sh
```diff
@@ -0,0 +1,32 @@
...
+if ! [ -n "$DV_TARGET" ]; then
+  DV_TARGET=cv64a6_imafdc_sv39
+fi
+
+if ! [ -n "$DV_SIMULATORS" ]; then
+  DV_SIMULATORS=veri-testharness,spike
+fi
+
+cd cva6/sim
+python3 cva6.py --testlist=../tests/testlist_riscv-arch-test-$DV_TARGET.yaml --target $DV_TARGET --iss_yaml=cva6.yaml --issrun_opts="+time_out=40000000" --iss=$DV_SIMULATORS $DV_OPTS --linker=../tests/riscv-isa-sim/arch_test_target/spike/link.ld
```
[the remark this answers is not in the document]
JeanRochCoulon: Can I ask you to make this minor modification?  After, in the case the CI is PASS, I will merge the PR

JeanRochCoulon:
Thank you again @fatimasaleem for this contribution. For your information, we plan to upgrade the GCC version to version 12 to support the Bit Manip  extension. And the arch-test will be inserted in the Thales CI. Very nice !!

fatimasaleem:
> Thank you again @fatimasaleem for this contribution. For your information, we plan to upgrade the GCC version to version 12 to support the Bit Manip extension. And the arch-test will be inserted in the Thales CI. Very nice !!

Great! That's good to hear, thank you @JeanRochCoulon.

PR #1264, opened 2022-05-10
cva6.yml, benchmark.sh: add benchmark job from riscv-tests and wb job

Gchauvon:
Hello,

I added two job to the gitlab CI : 

-  pub_benchmark to run benchmarks from riscv-test
-  pub_wb_dache to run some ISA tests with wb_dcache

Both jobs are manual and optional. It will help gitlab users to check those features without having to trigger github CI with a PR.

@JeanRochCoulon can you look into this ?

Guillaume


JeanRochCoulon (commented):
in cva6.yml, DV_TARGET is set, but benchmark.sh force it to cv64a6_imafdc_sv39. Could you make it configurable to have the possibility to run benchmarks in 64 and 32 bits?

Gchauvon:
Unfortunately it is not yet possible to run 32 bits benchmarks from our environment since we do not use the same tools and makefile to compile the tests. However, I have let the possibility to use 32 bits in case someone find a way to make it work.

JeanRochCoulon (approved):

PR #1266, opened 2022-05-11
CI Generated configuration

Gchauvon:
Related to https://github.com/openhwgroup/cva6/pull/862#issue-1211045880

Hello,

I have integrated the use of custom configurations in our core-v-verif flow. It allow us to create CI on features that we don't check with our ISA tests. It also ease the creation and the use of configurations different from the 4 already existing configs in CVA6. 

I've also added a testlist and a script to be runned in CI. `testlist_config.yaml` has a field `needs` to check that the configuration we use can run the test.

@JeanRochCoulon @ASintzoff can you look at this ?

### DIFF
--- a/.gitlab-ci/cva6.yml
+++ b/.gitlab-ci/cva6.yml
@@ -147,6 +147,22 @@ pub_smoke:
     paths:
       - artifacts/reports/*.yml
 
+pub_hwconfig:
+  stage: two
+  extends:
+    - .template_job_short_ci
+  needs:
+    - job: pub_smoke
+      artifacts: false
+  parallel:
+    matrix:
+      - DV_SIMULATORS: ["vcs-uvm,spike", "veri-testharness,spike"]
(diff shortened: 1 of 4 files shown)


JeanRochCoulon (commented):
1. Why not defining only one job with more configurations. As we will have a lot of hardware configuration to tests (22 I think), it is better to minimize the number of jobs. And vcs-uvm could be removed to reduce the license use.
2. Could you condition the job to the smoketest success.
3. (This point can be discussed) The scripts define "custom" and "config_opts" new variables. Could you call them "hwconfig" and "hwconfig_opts" ?    config_tests.sh into hwconfig_tests.sh, pub_config into pub_hwconfig, testlist_config.yaml into testlist_hwconfig.yaml, CONFIG_OPTS into DV_HWCONFIG_OPTS,...

JeanRochCoulon (approved):

Issue #1268, opened 2022-05-13
uvma_rfvi: non compliant LRM SystemVerilog code


ZElkacimi:
With VCS (which is picky regarding LRM SystemVerilog), I get the following compilation error
 “ Following verilog source has syntax error :
Select on function call uvma_rvfi_instr_mon.sv
 token is ']' 
mon_trn.csrs["dcsr"].get_csr_retirement_data()[3]);

file with the syntaxe error is at: https://github.com/openhwgroup/core-v-verif/blob/master/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv.

To reproduce the error, compile the following example:
```
class base_seq;
    
 function bit [5:0] testsyntx();
           return (5'b00100);
 endfunction

endclass

class seq_c extends base_seq;
   bit zi;
function new(string name="seq_c");
      super.new(name);
   endfunction : new
   
task test_sytax();
      zi = testsyntx()[2];
   endtask

endclass
```
use the following command: vcs  -sverilog file.sv

Error message: "Error-[SE] Syntax error
  Following verilog source has syntax error :
  	Select on function call
  "not_lrm_compliant.sv", 28: token is ';'
        zi = testsyntx()[2];"

Same error with the option :  -xlrm

**_Possible solution:_**
Use an intermediate variable:
 bit [5:0] inter_var = testsyntx();
 zi = inter_var[2];

MikeOpenHWGroup:
Hi @ZElkacimi.  VCS is doing the right thing - this code construct is not compliant with the LRM.  Did it compile/run with another simulator?  In any case, it must be fixed.  Please create a pull request.

ZElkacimi:
Hello @MikeOpenHWGroup

I don't have the opportunity to  compile/run with another simulator, but since it's in lib/ directory then I suppose that it works with another simulator. I have created a PR for a simple fix, please check it to see if it's the best one to do.

Thanks

MikeOpenHWGroup:
Hi @ZElkacimi.  As this Issue has not received any comments in almost 2 years and we do not believe there is an issue with the uvma_rvfi, I am closing this issue.  Please feel free to reopen if you have further questions.

PR #1270, opened 2022-05-17
uvma_rvfi_instr_mon.sv: fix a non compliant LRM SystemVerilog code

ZElkacimi:
#https://github.com/openhwgroup/core-v-verif/issues/1268#issue-1235352499
Signed-off-by: Zineb El Kacimi <zineb.el-kacimi@external.thalesgroup.com>

### DIFF
--- a/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
+++ b/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
@@ -38,6 +38,7 @@ class uvma_rvfi_instr_mon_c#(int ILEN=DEFAULT_ILEN,
 
    // State of monitor
    bit                last_dcsr_nmip = 0;
+   bit [uvma_rvfi_csr_seq_item_c::XLEN-1:0] dcsr_ret_data;
 
    string log_tag = "RVFIMONLOG";
 
@@ -204,11 +205,12 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
             end
          end
 
+         dcsr_ret_data = mon_trn.csrs["dcsr"].get_csr_retirement_data();
          // In debug mode, detect NMIP event for a data bus error
          if (mon_trn.dbg_mode &&
              !last_dcsr_nmip &&
              mon_trn.nmip[0] &&
-             mon_trn.csrs["dcsr"].get_csr_retirement_data()[3])
+             dcsr_ret_data[3])
          begin
             `uvm_info("RVFIMON", $sformatf("Debug NMIP"), UVM_LOW);
 
@@ -228,7 +230,7 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
          end
 
          // Latch the last DCSR NMIP bit to detect positive assertion of dcsr.nmip
-         last_dcsr_nmip = mon_trn.csrs["dcsr"].get_csr_retirement_data()[3];
+         last_dcsr_nmip = dcsr_ret_data[3];
 
          `uvm_info(log_tag, $sformatf("%s", mon_trn.convert2string()), UVM_HIGH);
 


MikeOpenHWGroup (commented):
Hi @ZElkacimi, this fix looks good to me.  Before merging I will ask @silabs-robin to review as he was one of the authors of this agent and it is used extensively at his organization (Silicon Labs).

silabs-robin:
> I will ask @silabs-robin to review

Thanks for considering me regarding this change.
I tried to review it but I didn't fully understand it, so I just have that one question above if either of you could help explain.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
```diff
@@ -204,11 +205,12 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
             end
          end
 
+         csr_ret_data = get_csr_retirement_data();
```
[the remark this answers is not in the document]
silabs-robin: (This might possibly be one of those cases where I again wish we were using `default_nettype none`. But I'm not sure how that works on the oop side of systemverilog.)

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
```diff
@@ -204,11 +205,12 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
...
          // In debug mode, detect NMIP event for a data bus error
          if (mon_trn.dbg_mode &&
              !last_dcsr_nmip &&
              mon_trn.nmip[0] &&
-             mon_trn.csrs["dcsr"].get_csr_retirement_data()[3])
+             mon_trn.csrs["dcsr"].csr_ret_data[3])
```
silabs-robin, line 213: Once we are at it, maybe we should add a named constant instead of the magic number 3?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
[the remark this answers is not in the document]
MikeOpenHWGroup: > (I don't have a way of easily running this myself.)

This is a library component, so if you clone @ZElkacimi's fork/branch and run any test of the E40X/S that uses the RVFI instruction monitor (virtually every test!), you should be able to see it in action.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
silabs-robin: I am not surprised 😄 Sorry about that. I shall do my best to motivate my future self to utmost diligence.

silabs-robin: Hi. I have my comments above. I did not understand how the code would work, and it seems Xcelium also could not compile it.

silabs-robin: Is it possible for you to add a breakpoint in the code when running simulation? What happens if you add a breakpoint on this line?: https://github.com/openhwgroup/core-v-verif/blob/611850728c0f6fa4a8e5da0a60a3bf4d7270c0c1/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv#L208
To me, it seems like it is defined in a different scope and should not be visible from that scope, and it seems Xcelium agrees with me. @MikeOpenHWGroup , could you try this on e.g. dsim or replicate on one of the simulators we use?

ZElkacimi: Trying to set the BP, I realized that I did not create the rvfi agent correctly, that's why the compilator didn't detect these errors before. Now, I have same errors you mentioned above;  "Identifier 'get_csr_retirement_data' has not been declared yet ..." 

I think this agent was functional, one day, since it's in lib/ , maybe you can debug, if if you have any project where you used it for the first time. Or what do you think?

silabs-robin: The agent should be functional. I think it compiled on the commit right before the one in this PR.
The fix itself is also a good idea (if the existing code is truly not LRM compliant), so I think what needs to be done is to work on the code a bit to get the fix working.

ZElkacimi:
Hello @silabs-robin
Sorry, It was me who proposed a fix that not make sense. Please try this one, and let me know if it works better.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
```diff
@@ -228,7 +230,7 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
          end
 
          // Latch the last DCSR NMIP bit to detect positive assertion of dcsr.nmip
-         last_dcsr_nmip = mon_trn.csrs["dcsr"].get_csr_retirement_data()[3];
+         last_dcsr_nmip = csr_ret_data[3];
```
silabs-robin, line 233: Could we name it `dcsr_ret_data` so that usage of the variable makes it clear which CSR is being referenced?

ZElkacimi: > Could we name it `dcsr_ret_data` so that usage of the variable makes it clear which CSR is being referenced?

Done

silabs-robin (approved):
Approving on the basis of: The fix makes sense to me and hello-world runs without problem.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks all.

PR #1272, opened 2022-05-19
Cvvdev/dev/several c tests

JeanRochCoulon:
@ASintzoff, could you review this PR.

The goal is to :
- extend the capability of iss*_opts (previous commit on master branch)
- add the possibility to simulate several c_tests by cva6.py execution

Thanks


JeanRochCoulon:
@ASintzoff 
For information, the "" have been added to allow spaces in the iss*_opts options, when we need to pass several different options for instance.

ASintzoff (approved):
LGTM

Issue #1256, opened 2022-04-27
User-mode vplan review, cv32e40s


silabs-robin:
### Task Outcome
The attached 40s vplan for user-mode is reviewed, revised, accepted, and pushed to the repo.

### Location Information
The spreadsheet is uploaded in this PR: https://github.com/openhwgroup/core-v-verif/pull/1287.
~~The [pre-review] vplan is downloadable from here: [CV32E40S_UserMode.xlsx](https://github.com/openhwgroup/core-v-verif/files/8571732/CV32E40S_UserMode.xlsx).~~
~~A PR will later place the final file at `core-v-verif/cv32e40s/docs/VerifPlans/Simulation/privileged_spec`.~~

### Completion Criteria
The review procedure is completed (as per [here](https://github.com/openhwgroup/core-v-verif/blob/master/docs/VerifPlans/VerificationPlanning101.md#verification-plan-reviews)) and the finished file is in the repo.

silabs-robin:
_"Arrange for a review call for the verification plan"_
This can wait a bit, as we have more vplans coming up (PMP, CLIC, Zc), and it could be(?) more efficient to do them all at once or in close proximity.

Silabs-ArjanB:
[CV32E40S_UserMode-Arjan.xlsx](https://github.com/openhwgroup/core-v-verif/files/8812069/CV32E40S_UserMode-Arjan.xlsx)

MikeOpenHWGroup:
Is that #1287?

PR #1280, opened 2022-05-29
User-mode Asserts

silabs-robin:
SVAs for checking user-mode, according to the vplan https://github.com/openhwgroup/core-v-verif/issues/1256.

Note, rtl is not ready to pass everything yet, and rvfi updates are needed. (But our test stimuli doesn't exercise user-mode yet so it is still expected that these asserts will not cause failures in simulation.)

Note, the vplan review is not done, so these asserts might see future updates. (Which would not be unlikely otherwise anyway.)

This does not pass ci_check yet, because the core hash was updated and this necessitates iss-related updates.
Hence, the pr is a "draft".
After a separate pull request has fixed the core-updating issue, then I will merge from 40s/dev into this pr branch, make sure ci_check passes, and remove the "draft" status.


cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -0,0 +1,619 @@
...
+  localparam int CY_POS      =  0;
+  localparam int CY_LEN      =  1;
+  localparam int IR_POS      =  2;
+  localparam int MPRVEN_POS  =  4;
+  localparam int MPRVEN_LEN  =  1;
+
+  localparam int MODE_U = 2'b 00;
+  localparam int MODE_M = 2'b 11;
+
+  localparam int MRET_IDATA    = 32'b 0011000_00010_00000_000_00000_1110011;
```
MikeOpenHWGroup, line 91: Maybe a comment about where these numbers come from.

MikeOpenHWGroup:
Great stuff @silabs-robin.  General question: are these assertions intented to be used for both formal and simulation?  If so, which set of assumptions are required to use formal?

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: Ok. Fixed. (I will use a second branch for queuing up fixes, since a previous pr showed that updates can cause an onslaught of pings.

silabs-robin:
> are these assertions intented to be used for both formal and simulation? If so, which set of assumptions are required to use formal?

No additional assumptions should be needed for these user-mode asserts.
For all the assertion files we have, only general assumes are required (obi protocol, no "scan_cg_en", no Xes from blackboxed multiplier, input addresses stable after "fetch_enable").

silabs-robin:
Just a small update on this:
It was pending on some fixed in "cv32e40s/dev" which are now in, but also the vplan review meeting led to some vplan fixes that will be incorporated in the assertions as well, and it is also pending on some rvfi fixes, etc.
So this is on halt for a while longer.

silabs-robin:
I am back on this branch. Will close and re-open when ready, in case people are "subscribed" to this PR.

PR #1287, opened 2022-06-07
add 40s umode vplan

silabs-robin:
Adding a spreadsheet with the vplan for user-mode for 40s.


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
In the interests of getting this DVplan completed, I've approved this PR and will merge it.  Having said that, is there anything in this DVplan that is specific to the CV32E40S?  Could this DVplan be used as a generic priviledge-spec/user-mode DVplan for RV32 compliant cores?

silabs-robin:
> Having said that, is there anything in this DVplan that is specific to the CV32E40S? Could this DVplan be used as a generic priviledge-spec/user-mode DVplan for RV32 compliant cores?

It has some 40s-specific elements.
So it is almost generic enough to be general, but not quite, I think.

PR #1292, opened 2022-06-10
Initial commit of updates required for core-v-verif to match state of cv32e40s rtl

silabs-hfegran:
This commit pulls in 0.2.0 version of the ISS and updates simulation environment to match changes in rtl.

Note to reviewers: This **_will_** fail ci_check:
* hello_world: misa mismatch with iss, iss does not set bit 23(X) to 1.
* illegal_instr_test: test requires update, pending iss update to enable proper feature set
* Any of the interrupt tests: known error, being addressed in parallel effort.


MikeOpenHWGroup (approved):
Approving and merging on @silabs-hfegran's request: it is known that this will (temporarily) break CI.

PR #1300, opened 2022-06-14
[skip ci] Fix curly braces

MikeOpenHWGroup:
This was caught by a prospective member of the OpenHW Group who is using Synopsys VCS.  Apparently both dsim and xrun accept both curley braces and round braces around variable names in manifests and VCS does not.  Who knew?


silabs-hfegran (approved):
LGTM, Thanks for the fix @MikeOpenHWGroup

PR #1312, opened 2022-06-23
gitlab-ci: show test fail in dashboard

yanicasa:
Related to gitlab-ci dashboard management

Currently, the report of each test is created at the end of the script of each job.
The problem is that if one of the commands of the job returns a non-zero error code, then no report exists for the job and nothing mentions in the dashboard that the job is failed.

Ideally, jobs should not return a zero error code even in case of fail. The fault should be seen and generated by the script that generates the report for the job.

To fix the problem, this PR adds a script ran at the beginning of each job. That generates a "fail" report, and if the job is OK, it will be overwritten by the real report at the end.

Some descriptions in the CI has also been changed

### DIFF
--- a/.gitlab-ci/cva6.yml
+++ b/.gitlab-ci/cva6.yml
@@ -139,11 +139,13 @@ pub_smoke:
     matrix:
       - DV_SIMULATORS: ["veri-testharness,spike","vcs-testharness,spike","vcs-uvm,spike" ]
   script:
+    - mkdir -p artifacts/reports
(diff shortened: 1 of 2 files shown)


yanicasa:
@JeanRochCoulon, please check [https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_core-v-verif.html](https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_core-v-verif.html) before merge

Gchauvon (approved):
LGTM

yanicasa:
> @JeanRochCoulon, please check https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_core-v-verif.html before merge

Dashboard OK, Can be merged

JeanRochCoulon (approved):

PR #1313, opened 2022-06-24
cva6.py: save cva6/sim path to go back to it after generating config

Gchauvon:
Small fix in order to use hwconfig regression from cva6 repository via Thales-CI

@JeanRochCoulon or @yanicasa can you review it ?

Guillaume


yanicasa (approved):
ok

JeanRochCoulon (approved):

PR #1338, opened 2022-07-13
Fixed xcelium warning: Frequent false starts for some debug assertions

silabs-hfegran:
Assertions appeared to start evaluation and sample a local variable value unconditionally - moved sampling to only sample on actual match


MikeOpenHWGroup:
Hi @silabs-hfegran, why was this PR closed?

silabs-hfegran:
> Hi @silabs-hfegran, why was this PR closed?

@MikeOpenHWGroup, I initially closed it because I believe some further investigation is needed here - I do not want to end up just masking away the warning, but rather make sure that I actually resolved the problem before this is merged.

silabs-hfegran:
I believe these should be fine to merge now

MikeOpenHWGroup (approved):

PR #1342, opened 2022-07-19
Set end_trampoline back to 80000000 and process 1st instruction

Gchauvon:
Hello, 

I set back end_trampoline back to 0x80000000 but suppress the continue statement to process instruction at 80000000.
We had issue with the UVM testbench which has no bootrom and start at 80000000. 
The script was not able to find instruction at address 10014 ( it doesn't exist in UVM tb) and didn't process the log correctly resulting in an empty csv.

The solution I propose solves this issue and still process the 1st instruction to match spike trace.

### DIFF
--- a/cva6/sim/verilator_log_to_trace_csv.py
+++ b/cva6/sim/verilator_log_to_trace_csv.py
@@ -114,7 +114,7 @@ def read_verilator_trace(path, full_trace):
   # true. Otherwise, we are in state EFFECT if instr is not None, otherwise we
   # are in state INSTR.
 
-  end_trampoline_re = re.compile(r'core.*: 0x0000000000010014 ')
+  end_trampoline_re = re.compile(r'core.*: 0x0000000080000000 ')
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
This will work for now, but it is not really a proper solution.  The CV32E4 cores have implemented a process to control the testbench memory model addressing, test-program compile flags and the core configuation at compile and run-time.  This allows for per-test-case control of such things as the location of the "bootrom".

I strongly recommend that the CVA6 adopt this flow - it will be difficult to close coverage with a static set of parameters.

I will leave it to @JeanRochCoulon to determine if this PR should be merged in for a short-term fix.  Either way, we need a long term solution for this issue.

JeanRochCoulon:
Thank @MikeOpenHWGroup for challenging our methodology. As our CI is broken for the time being, I would propose to commit the PR and in parallel I opened a task thread to find a proper solution: https://github.com/openhwgroup/core-v-verif/issues/1343

PR #1345, opened 2022-07-20
[HOTFIX] put back "continue" in else statement

Gchauvon:
Hello,

PR #1342 has been merge too quickly. 
testharness testbench trace to csv does not work with the changes I made in #1342. My bad ! It worked locally on a slightly different test from the ones in the regression. 
It should work well now. 

@JeanRochCoulon make sure to check dashboard before merging this PR.
@MikeOpenHWGroup I have seen the remark about not using static parameters (in this case addresses). I agree with you and it is a lot of work to do. Meanwhile this PR should fix the regression until we update our workflow for verification.

Guillaume


JeanRochCoulon:
The CI is PASSED. Let's commit in master branch

PR #1348, opened 2022-07-22
Fixes for mtvec-related issues in regressions

silabs-hfegran:
Previous implementation for trap handler generation and nmis did not take vectored/non-vectored mode into account and used the wrong symbol as mtvec handler reference. 

This PR aims to solve these issues by:
- nmi is always located at mtvec + 4*15 (actual configured mtvec, not the address of mtvec_handler)
- vectored and direct mode interrupt should have different handlers, one using trampoline code to get to the right handler (leaving room for nmi at the correct offset), the other places the jump-instructions directly at mtvec (and includes a jump to the nmi handler at the correct location)

These changes should resolve most timeouts due to interrupt-related runaway code.

### DIFF
--- a/lib/corev-dv/corev_asm_program_gen.sv
+++ b/lib/corev-dv/corev_asm_program_gen.sv
@@ -36,9 +36,16 @@ class corev_asm_program_gen extends riscv_asm_program_gen;
     instr_stream.push_back(".section .text.start");
     instr_stream.push_back("");
 
-    instr_stream.push_back(".section .mtvec_bootstrap, \"ax\"");
-    instr_stream.push_back(".globl _mtvec_bootstrap");
-    instr_stream.push_back("    j mtvec_handler");
+    if (cfg.mtvec_mode == DIRECT) begin
+      instr_stream.push_back(".section .mtvec_bootstrap, \"ax\"");
+      instr_stream.push_back(".globl _mtvec_bootstrap");
+      instr_stream.push_back("    j mtvec_handler");
+    end else begin
+      instr_stream.push_back(".globl vectored_mode");
+      instr_stream.push_back(".section .mtvec_handler, \"ax\"");
+      instr_stream.push_back(".globl mtvec_handler");
+      instr_stream.push_back(".type mtvec_handler, @function");
+    end
     instr_stream.push_back("");
 
     instr_stream.push_back(".globl _start");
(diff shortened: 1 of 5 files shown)


silabs-hfegran:
I have marked this as DO NOT MERGE as I just discovered an error in my code - will remove the label when I get the error sorted out and pushed through regressions.

silabs-hfegran:
Should be OK now - there are still timeout issues with corev_pma_test, but I believe that to be an unrelated issue

silabs-hfegran:
Added a fixup commit for the pma test here as well, seeing that this is not closed yet

lib/corev-dv/corev_asm_program_gen.sv
```diff
@@ -36,9 +36,19 @@ class corev_asm_program_gen extends riscv_asm_program_gen;
...
-    instr_stream.push_back(".globl _mtvec_bootstrap");
-    instr_stream.push_back("    j mtvec_handler");
+    if (cfg.mtvec_mode == DIRECT) begin
+      instr_stream.push_back(".section .mtvec_bootstrap, \"ax\"");
+      instr_stream.push_back(".globl _mtvec_bootstrap");
+      instr_stream.push_back("    j mtvec_handler");
+      instr_stream.push_back(".section .nmi_bootstrap, \"ax\"");
+      instr_stream.push_back(".globl _nmi_bootstrap");
+      instr_stream.push_back("    j nmi_handler");
+    end else begin
```
silabs-robin, line 46: I don't understand how these sections are managed: `mtvec_bootstrap`, `nmi_bootstrap`, `mtvec_handler`.
Some of them are declared in assembly right here and someare declared elsewhere, some of them are used by assembly here but some are used elsewhere, but they are all used by the linker scripts?
All of these pieces fits together more complexly than I can grok atm, but if they are known to cooperate correctly then I will blindly trust it.

silabs-robin (approved):
Approving based on this:
From what I know of how nmi-related specs have changed, the PR description makes sense.
From what I can understand of the code diff, it seems to match the PR description, and I cannot see anything that seems off.

If this code is more critical than I have thought, then I can make a more thorough effort to get a deeper understanding of the code, if desired.

lib/corev-dv/corev_asm_program_gen.sv
silabs-hfegran: bootstrap-sections are used to provide a section for a handler-jump instruction. we want mtvec at address 0 and nmi-handler at offset15(mtvec.address) - if the handlers themselves were to be placed at these exact locations, there would not be space for the nmi-handler in case of the non-vectored variant, thus the bootstrap sections implement trampolines to the actual handler code. The idea is that we either create sections for mtvec and nmi jump instructions - OR in the case of vectored interrupts, create a jump table where the nmi_handler jump is located at offset 15 inside the mtvec jump table

PR #1350, opened 2022-07-25
Added several new privilege tests

silabs-anvesten:
Signed-off-by: Anders Vestengen <Anders.Vestengen@silabs.com>


cv32e40s/tests/cfg/pmp.yaml
```diff
@@ -0,0 +1,10 @@
+name: pmp
+description: Default pmp configuration for CV32E40S simulations
+compile_flags: >
+   +define+PMP_ENABLE_64
+ovpsim: >
+    --override root/cpu/PMP_registers=64
+    --showoverrides
```
silabs-hfegran, line 7: The --showoverrides line should be commented out or removed before merge

MikeOpenHWGroup:
Also, please add the appropriate license header to the top of each new file.

silabs-hfegran:
Can we change this to a draft PR until it is ready for proper review?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
```diff
@@ -0,0 +1,80 @@
...
+0xFC0-0xFFF
+"""
+
+def generator():
+    """  
+    This is the function which generates the csr_commands. It attempts to Read, Write, Clear and Set for every CSR in the list above. 
+    """
+    num_lines = 0 # printed later to help debugging, and assertion checks in C.
+    string_split = (reg_str.split("\n"))
+    string_split = string_split[1:-1]
```
silabs-robin, line 53: Could you add an in-line comment about what `[1:-1]` means?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+input_string = "// Start of generated code." # start string, this will move the wrie HEAD
+pointer = 0 # file pointer 
+filename = "csr_privilege_loop.S" # file which will be written to 
+
+
+# register list fetched manually from the spec (V20211203), contains all U-, S-, R-, and M-mode CSR registers
+reg_str = """
+0x100-0x1FF
```
silabs-robin, line 13: This starts at `0x100`, are the user level CSRs missing?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
```diff
@@ -0,0 +1,80 @@
...
+
+"""
+
+input_string = "// Start of generated code." # start string, this will move the wrie HEAD
+pointer = 0 # file pointer 
+filename = "csr_privilege_loop.S" # file which will be written to 
+
+
+# register list fetched manually from the spec (V20211203), contains all U-, S-, R-, and M-mode CSR registers
+reg_str = """
```
silabs-robin, line 12: Why do we need a list of individual ranges?
Is it not contiguously covering all of 0x000-0xFFF (or 0x100-0xFFF), or is there some differentiation in how the different ranges are tested?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
silabs-robin: From discussions in chat: 0x000 etc are deliberately left out because this test intends to do U-mode accesses to "MRW" CSRs etc.

Could we add this in a comment too?

I read "contains all U-, S-, R-, and M-mode CSR registers" and wondered why U was missing.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
silabs-robin: From discussions in chat: The reason is to exclude User-Level CSRs, as per the purpose of the test.

It would also be possible to loop over `0x000-0xFFF` and filter out `[9:8] == 0`, but these are just 30 lines so I think it is good as it is now.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
silabs-anvesten: Yeah, I can change this locally now before pushing the new change. It should be just S-, R-, and M-mode CSR's

silabs-anvesten: This is fixed now

MikeOpenHWGroup:
@silabs-anvesten, @silabs-hfegran, @silabs-robin, @silabs-mateilga
(I'm not sure who to ask about this, so I am spamming all of you.)

I am not sure how this `illegal_access_loop_tests` works.  I have a couple of comments/questions:
1. The directory `cv32e40s/tests/programs/custom/illegal_access_loop_tests` seems to contain the sources for multiple test-programs.  Is that the case, or are all these sources individual files of a single test-program?  In core-v-verif we want to maintain one test-program per directory.
2. It is not clear to me how the python script generators are lauched.  These scripts appear to generate source file for test-programs.  What it is the mechanism that launches the scripts to generate the test-program?  AFAIK, the current Makefile(s) do not support this and we should think carefully before introducing a new compilation flow to core-v-verif.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
silabs-robin: Thanks. If it is fixed then it should be ok to push the "Resolve conversation" button if it exists.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/csr_privilege_gen.py
silabs-anvesten: Should I mark this as resolved as well ?

silabs-robin: Sure. The answer made sense and the code seems good.

silabs-anvesten:
> @silabs-anvesten, @silabs-hfegran, @silabs-robin, @silabs-mateilga (I'm not sure who to ask about this, so I am spamming all of you.)
> 
> I am not sure how this `illegal_access_loop_tests` works. I have a couple of comments/questions:
> 
> 1. The directory `cv32e40s/tests/programs/custom/illegal_access_loop_tests` seems to contain the sources for multiple test-programs.  Is that the case, or are all these sources individual files of a single test-program?  In core-v-verif we want to maintain one test-program per directory.
> 2. It is not clear to me how the python script generators are lauched.  These scripts appear to generate source file for test-programs.  What it is the mechanism that launches the scripts to generate the test-program?  AFAIK, the current Makefile(s) do not support this and we should think carefully before introducing a new compilation flow to core-v-verif.

1. The directory Contains three tests, as well as the source files for running them. I can change this, but they are also closely related to one another as they're all generated tests which tries to access different CSR and custom registers while in U-mode. They also all run at the same time (in order) when one calls make test on the directory, but again I can change this to the one-test-per-directory if that's more desirable.

2. The Python scripts have to be run manually, and it is not intended that they run along the c. and S. files during a test. They run when called directly from the command line and they change the generated parts of the assembly files in the same directory. This is done so it should be easier to maintain the generated commands in the S. files, and also to have some reference as to how those commands were created.

cv32e40s/tests/cfg/pmp.yaml
silabs-anvesten: Removed the --showoverrides line

silabs-robin:
>     1. The directory `cv32e40s/tests/programs/custom/illegal_access_loop_tests` seems to contain the sources for multiple test-programs.  Is that the case, or are all these sources individual files of a single test-program?  In core-v-verif we want to maintain one test-program per directory.
> 
>     2. It is not clear to me how the python script generators are lauched.  These scripts appear to generate source file for test-programs.  What it is the mechanism that launches the scripts to generate the test-program?  AFAIK, the current Makefile(s) do not support this and we should think carefully before introducing a new compilation flow to core-v-verif.

Just one little meta comment first: If we make comments in-line in the code, then each topic will be self-contained organized and threaded, and each completed discussion can be marked as resolved to more easily scroll through a PR to see if all feedback has been addressed.

There is a difference between "test" and "test" (what terminology differentiates them?).
The debug tests has lots of smaller tests within them.
As long as everything in a directory `mytest` is runable with `make test TEST=mytest` then I think it is ok.
Each test-directory should have one test-program, and it is ok to have multiple test-sourcefiles, and AFAIK that is how it is right now.

It looks like the python scripts are run manually before PR, and the generated files are included within the PR.
Someone that runs the test already has the generated files when they checkout the repo.
(I have one other concern about this, but I'll start another thread on it.)

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
```diff
@@ -0,0 +1,65 @@
+"""
```
silabs-robin, line 1: The file generated by this script is `~ 130 000` lines long.
That is so big that github refuses to show its content.
I am a bit skeptical to merge such a big file.

silabs-anvesten: This is because there are 130K possible custom instructions at the moment. The size of the file is 2.3MB. I don't know how one would deal with this otherwise, I could remove the generated lines but then anyone running the test would have to manually generate them again.

silabs-anvesten: I'll also add that the test-time lies between 22 and 36 minutes.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/helpers.S
```diff
@@ -0,0 +1,57 @@
...
+	csrrw	x0, pmpcfg0, t0
+
+	// Return to caller
+	jalr	x0, 0(ra)
+
+set_u_mode: // puts the core in usermode.
+	// setup the return adress to stack.
+	addi	sp, sp, -4
+	// store the ra to the stack pointer.
+	sw	ra, 0(sp)
```
silabs-robin, line 45: `ra` is saved on the stack, and later restored.
But is it ever changed in between?
It looks like the stack manipulation is not necessary, or I am missing something.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/helpers.S
```diff
@@ -0,0 +1,57 @@
...
+
+	// Zero "mstatus" to set MPP=umode
+	li	t0, 0x1800
+	csrrc   x0, mstatus, t0 // clear the mstatus (mpp -> User mode).
+	la t0, empty_symb
```
silabs-robin, line 50: `ret` will jump to `ra`, so would it be easier to just set `mepc` directly to `ra` instead of jumping via the `empty_symb` label?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/helpers.S
silabs-anvesten: No, I think this is old and also happens in the privilege_tests.S file. I think this is something I added when I had just started, but I agree, I don't understand what it does here and I don't think the ra gets changed either.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
```diff
@@ -0,0 +1,89 @@
...
+  if (actual != expect) {
+    printf(msg);
+    printf("expected = 0x%lx (%ld), got = 0x%lx (%ld)\n", expect, (int32_t)expect, actual, (int32_t)actual);
+    exit(EXIT_FAILURE);
+  }
+}
+
+void change_m_mode(void) {
+  //s4 is x20
+  __asm__ volatile("addi s4, x0, 0x3"); // add 3 to the s4 register, which will be read by the trap handler late
```
silabs-robin, line 45: `s4` is a preserved register, so the function that called this one might already be using it.
If we randomly change it here, then the rest of the program might get corrupted.

For communicating with a trap handler, I would rather recommend using global variables. `gbl_mysignaltothehandler`.

silabs-anvesten: I agree, I'll change this and update the PR today.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
```diff
@@ -0,0 +1,89 @@
...
+  __asm__ volatile("ecall");
+  __asm__ volatile("li s4, 0"); // Reset the value after entering machine mode
+}
+
+void illegal_custom_loop(void) {
+    /* 
+    Tests U-mode access to various custom functions which are not yet implemented. Should all trap.
```
silabs-robin, line 52: What are "custom functions"?

And, from "not yet implemented" I wonder, are these expected to be implemented soon?
If the spec say we should have some functionality, then the tests should expect them to work, and either error out on parts that aren't implemented or we can leave a TODO comment to enable them once the RTL is ready for it.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-robin: I shall read the generation script and see if I can think of any other way to do this.

One thought that I have already:
For the other generation script (the one created that 12k program), self-modifying code _might_ be a solution.
(Though, warning, self-modifying code is probably a bit tricky.)
We know that a CSR instruction uses bit `[31:20]` for the CSR address...
So if you have e.g. 4 different instructions (read/write/set/clear), then we can add labels to them and let the surrounding code overwrite those bits.
Instead of several thousands of lines of code, we'd have a loop that does the same thing.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/helpers.S
silabs-robin: If this function is used by the rest of the code then it would be nice to clean it up. (If it is not used then it should be removed.)

cv32e40s/tests/programs/custom/illegal_access_loop_tests/helpers.S
silabs-robin: When the function starts, `ra` contains the return address.

In the current code, `mepc/mret` takes us to `empty_symb`, and then the `ret` jumps to `ra` (to the return address).

Alternatively, I think we could let `mepc/mret` take us directly to `ra` (to the return address).

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
```diff
@@ -0,0 +1,89 @@
...
+    */
+    change_m_mode();
+    volatile unsigned int epp = 0;
+    //setup_pmp();
+    epp = illegal_full();
+    assert_or_die(epp, 131072, "error: not all illegal custom instructions triggered the trap handler\n");
```
silabs-robin, line 58: What is `131072`?
For so called "magic numbers" it is nice for readability to define a named constant.
(Although, admittedly we don't always.)
Alternatively an explanatory comment is also a nice way to show what the number comes from.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
```diff
@@ -0,0 +1,89 @@
...
+
+void csr_privilege_loop(void) {
+    /*  
+    Checks Read, Write, Set and Clear privielege on all implemented CSR-registers according to the list in 'csr_privilege_gen.py'
+    */
+    change_m_mode();
+    volatile unsigned int ecc = 0;
+    //setup_pmp();
+    ecc = csr_loop();
+    assert_or_die(ecc, 12288, "error: not all illegal csr instructions triggered the trap handler\n");
```
silabs-robin, line 69: Same as the comment above.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-robin: @silabs-hfegran and @MikeOpenHWGroup, this big file is difficult.
Do you have any opinions, insights, or solutions?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
```diff
@@ -0,0 +1,65 @@
...
+*******************************************************************************
+**
+** Script which generates the usermode custom instructions, to check illegal access. Instructions are written directly to the 'illegal_custom_loop.S' file after the 'input_string' declared below.
+**
+*******************************************************************************
+"""
+
+func6 = {35, 51} # usermode unprivileged function codes
+outer_loop = 2048 # 11-bit custom instruction
+inner_loop = 32 # 5-bit custom instruction
```
silabs-robin, line 26: What is "5-bit custom instruction"?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
```diff
@@ -0,0 +1,65 @@
...
+** limitations under the License.
+**
+*******************************************************************************
+**
+** Script which generates the usermode custom instructions, to check illegal access. Instructions are written directly to the 'illegal_custom_loop.S' file after the 'input_string' declared below.
+**
+*******************************************************************************
+"""
+
+func6 = {35, 51} # usermode unprivileged function codes
```
silabs-robin, line 24: What is "usermode unprivileged function codes"?

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-anvesten: > I shall read the generation script and see if I can think of any other way to do this.
> 
> One thought that I have already: For the other generation script (the one created that 12k program), self-modifying code _might_ be a solution. (Though, warning, self-modifying code is probably a bit tricky.) We know that a CSR instruction uses bit `[31:20]` for the CSR address... So if you have e.g. 4 different instructions (read/write/set/clear), then we can add labels to them and let the surrounding code overwrite those bits. Instead of several thousands of lines of code, we'd have a loop that does the same thing.

Maybe, but I believe this was discussed and (I might be mistaken here) the conclusion was that the commands need to be loaded into memory at compile time, and you can't (I believe) change the asm command while running. Although, @silabs-hfegran has a better understanding of it.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
silabs-anvesten: Vplan says this:

Feature descr: ""The subspace of the SYSTEM major opcode shown in Figure 3.30 is designated for custom use."

"Unprivileged or User-Level"

(40s has no such thing.)"

Verification Goal: "Execute custom SYSTEM "Unprivileged or User-Level" instructions, ensure they are illegal instructions."

MikeOpenHWGroup (changes requested):
@all, I am adding a "Request changes" state to my review for several reasons:
1. This seems to be a big PR needing careful review.
2. Multiple test-programs per directory is not in keeping with our current process and we need a very strong reason to deviate from this.
3. Requiring a user to invoke a specific python script to run a specific test is also not part of our current process for running tests.

Changes like this affect everyone and should only be done after careful consideration.

Please do not merge this PR until we'd had time to discuss.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
silabs-anvesten: 131072 is the number of illegal custom instructions. The Python file will also print the number of lines (the relevant number for the assert function) after writing.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-anvesten: Added more comments to expand on this. The 11-bit and 5-bit refers to the bit-fields that hold the custom instructions.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-anvesten: I'll add comments to this as well, this refers to the two possible unprivileged codes in the 'funct6' bit-field. This is referenced in fig. 3.30 p.49 of the privileged spec.

silabs-robin:
> I am adding a "Request changes" state to my review

Agreed, that changes and more discussion is necessary.


>     1. This seems to be a big PR needing careful review.

Agreed


>     2. Multiple test-programs per directory is not in keeping with our current process and we need a very strong reason to deviate from this.

What makes this "multiple test-programs"?
The `debug_test` has [24 individual sub-tests within it](https://github.com/openhwgroup/core-v-verif/blob/master/cv32e40s/tests/programs/custom/debug_test/debug_test.c#L539) and I can't see that this is any different.
 
 
>     3. Requiring a user to invoke a specific python script to run a specific test is also not part of our current process for running tests.

The user is not required to invoke a specific python script.
The generated file is checked in with this PR.
This is the same as how a small part of Spike is used together with Isacov and the generated shared object file is checked in to the repo, and directed CSR tests are also generated and have the result checked in to the repo.


> Changes like this affect everyone and should only be done after careful consideration.

Agreed


> Please do not merge this PR until we'd had time to discuss.

Agreed.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-robin: Pseudo assembly to demonstrate some of the idea:
```
my_csr_instr:
  csrrs t0, 0x0, x0
  
  la t1, my_csr_instr
  lw t2, 0(t1)
  t2 &= 0x000F_FFFF
  t2 |= 0x3A00_0000
  sw t2 0(t1)
  j my_csr_instr
```

But as we discussed, we also have the option of using some random instructions instead of every hypothetical one. (Maybe combine these two ideas.)

cv32e40s/tests/programs/custom/umode/umode.c
```diff
@@ -0,0 +1,73 @@
...
+extern volatile int  myfunc(void);
+//extern volatile void run_in_umode(int (*funptr)(void));
+extern volatile int  run_in_umode(int (*funptr)(void));
+extern volatile void setup_pmp(void);
+extern volatile void set_mmode(void);
+
+extern volatile uint32_t gbl_request_mmode = 0;
+
+
+__attribute__ ((interrupt ("machine")))
```
silabs-robin, line 17: `umode/` is some example code that I wrote before. Do we want to include it here?

Do you know what state it is in and if it works well?
If this code is good, we might want to (add a license header and) keep it as a simple example of using user mode.
But it is not otherwise so useful for testing purposes now that we have these new tests.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_access_loop_tests.c
silabs-robin: Ok, it refers to custom instructions. Then it's fine.

cv32e40s/tests/programs/custom/illegal_access_loop_tests/illegal_custom_loop_gen.py
silabs-robin: We agreed that the file size is not too big, and we are keeping the current method. Resolved.

cv32e40s/tests/programs/custom/umode/umode.c
silabs-anvesten: No I believe this was just for reference, nice catch. Can we delete this here or should I push the change from local?

silabs-robin: Whether github supports it in-browser, I don't know. Pushing the deletion as a commit is probably good.

silabs-robin:
Hi @MikeOpenHWGroup.
This PR should be ready to merge now.

Please have a look and see if you would be willing to merge it.
We have done a live walkthrough meeting and all feedback has been addressed.

You had some concerns before, but I think there is no threat.
There is only one single test-program per directory (it is just organized in multiple files).
A user does not have to invoke python (the test writer has already commited all generated code, and the scripts are merely included for maintainability).

MikeOpenHWGroup (changes requested):
Hi @silabs-anvesten, thanks for this contribution.  It is clear a lot of thought and work went into it.  I've made a few comments directly in three headers files, please have a look.

I also have two global requests before approving this PR:
1. Please update your license headers to Solderpad 2.1 and add a one-liner "SPDX license header" comment to all files:
```
// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```
2. The use of a "generator script" to generate code for the test-program is smart, but atypical for test-programs in core-v-verif. Please add a `README.md` providing an explaination for the existance and usage of the python scripts.

cv32e40s/tests/programs/custom/csr_priv_gen_test/csr_priv_gen_test.h
```diff
@@ -0,0 +1,8 @@
+#ifndef csr_priv_gen_test_h
+# define csr_priv_gen_test_h
```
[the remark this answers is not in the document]
silabs-anvesten: Thanks for spotting this. I'm working on addressing the header issue(s) right now. I'll also be doing some slight work on restructuring the python files for readability and then writing a readme.md to go in all the test-directories which use a python generator

silabs-anvesten:
@MikeOpenHWGroup Added the changes you requested. New Readmes in all directories with .py scripts. Restructured python files for better readability. Fixed header file gaffe's, and re-ran all tests and scripts successfully.

silabs-robin (approved):
(Forgot to mark my approval, in addition to stating it in text.)

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for sticking with this @silabs-anvesten.  Changes look good.  Merging...

PR #1352, opened 2022-07-25
New version of dsim

MikeOpenHWGroup:
Solves the so-called [stale dedupe](https://support.metrics.ca/hc/en-us/requests/1239) issue.


MikeOpenHWGroup:
Merging this in as a hot-fix without review.  (Relatively safe as it only affects MetricsCI.)

PR #1353, opened 2022-07-26
Ported regression timeout fixes from 40S to 40X

silabs-hfegran:
Implements same fixes as PR #1348 for cv32e40s


silabs-robin (approved):
Seems to be exactly the same as the 40s change (plus core hash update).
Approving based on having approved the 40s change and the expectation that these two cores should be the same when it comes to these aspects of NMIs.

The 40s has more reasons for NMIs (checksum err), but this should either not have an impact on the code in question or it is subtly handled within this diff. Am I correct in believing that?

silabs-hfegran:
> The 40s has more reasons for NMIs (checksum err), but this should either not have an impact on the code in question or it is subtly handled within this diff. Am I correct in believing that?

Correct - this code should cover _any_ nmi, as all should be caught by the nmi handler

PR #1355, opened 2022-07-28
Hash update + pmp path update

silabs-hfegran:


silabs-robin (approved):
Looks as expected.

PR #1359, opened 2022-08-04
Use vcs-uvm Makefile target in vcs-gate iss

Gchauvon:
Hello,

This PR fixes the vcs-gate target which was broken since the removal of the core only testbench. 
vcs-gate now uses vcs-uvm testbench to run post synthesis simulation. 

We can directly use vcs-uvm Makefile targets and remove old vcs-gate targets.

Guillaume

### DIFF
--- a/cva6/sim/Makefile
+++ b/cva6/sim/Makefile
@@ -160,23 +160,9 @@ vcs-uvm:
 generate_cov_dash:
 	urg -dir $(VCS_WORK_DIR)/simv.vdb
 
-vcs_gate_comp:
-	@echo "[VCS] Building Model"
-	mkdir -p $(VCS_WORK_DIR)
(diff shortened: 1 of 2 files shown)


MikeOpenHWGroup (commented):
This looks good to me @Gchauvon.  However, I currently lack the ability to run with VCS so I will defer to @JeanRochCoulon for a final approval.

ASintzoff (changes requested):
Is there any need to have configurable `gate_opts` at the moment?
I would suggest to remove them as if we keep it, it will clutter uselessly scripts to configure it.

Gchauvon:
> Is there any need to have configurable `gate_opts` at the moment? I would suggest to remove them as if we keep it, it will clutter uselessly scripts to configure it.

No there is, indeed, no need for those options for the moment. I removed it. 

@MikeOpenHWGroup, Jean-Roch is in vacation. Since André reviewed the PR. I think you can merge it.

ASintzoff (commented):
LGTM

PR #1362, opened 2022-08-08
CVA6: Add riscv-dv generated test

AyoubJalali:


MikeOpenHWGroup:
Hi @AyoubJalali, thanks for your Contribution.  Before this PR can be approved and merged, you need to get covered by an [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php).  A member of the Thales team can help you with that.  (Please let me know if you have any problems.)

Also, I am not sure about the file `cva6/sim/desc_gen-tests.yaml`.  It seems to be a direct copy of the same file from riscv-dv.  If that is true, why do we need it in core-v-verif?

AyoubJalali:
Hello, that's what i did i sign the agreement but still have the same problem.
And for the file cva6/sim/desc_gen-tests.yaml is not the same, because i add more test description used to approve the code coverage of the cva6, i keep some usual tests like riscv_arithmetic_basic_test.

MikeOpenHWGroup:
There _is_ a valid ECA for `ayoub.jalali@external.thalesgroup.com`.  However, this commit is signed off by `ayoub.jalali@external.thalesgroup` (the ".com" is missing).  Perhaps Eclipse cannot interpret an email domain with more than two levels of hierarchy (that is, "external.thalesgroup.com" gets truncated to "external.thalesgroup").  I will check an get back to you.

AyoubJalali:
Ok, thanks

AyoubJalali:
Hi, @MikeOpenHWGroup the ECA problem has fixed.
Also, for what i re-use from the YAML description in the riscv-dv.I can keep my file, but only with what i create as a description, and for the descriptions that i re-used from the riscv-dv using the script dv-generated-tests.sh by creating another list for reusing what in the YAML file in riscv-dv.
Just to avoid duplication.

cva6/sim/desc_gen-tests.yaml
```diff
@@ -0,0 +1,511 @@
+# Copyright Google LLC
```
MikeOpenHWGroup, line 1: This file _looks_ like, but is not, a direct copy of a test-program generator description from riscv-dv (e.g. `riscv-dv/yaml/base_testlist.yaml`), so it may confuse new uses of core-v-verif/cva6.  I would recommend changing the filename to something like `cva6_base_testlist.yaml` and add a section to the top of the comment header to make it distinguishable from a testlist in riscv-dv/yaml.  Perhaps something like this:
```
# Copyright 2022 Thales
#
# CVA6-specific testlist.  Written in the same format as riscv-dv testlists.
#
```

MikeOpenHWGroup (changes requested):

cva6/sim/desc_gen-tests.yaml
AyoubJalali: Ok it's done for the YAML description

AyoubJalali:
Hi @MikeOpenHWGroup, I change the file name to cva6_base_testlist.yaml as you recommend, and now in the script (dv-generated-tests.sh) we can use new test description in the cva6_base_testlist.yaml, and also tests in the riscv-dv, simply to give the list_num variable =0, to re-use tests in the base_testlist.yaml in riscv-dv.
Now the cva6_base_testlist.yaml don't contain any test from the base_testlist file in riscv-dv, but we can re-used these tests using the script

cva6/sim/desc_gen-tests.yaml
AyoubJalali: Hi @MikeOpenHWGroup, I change the file name to cva6_base_testlist.yaml as you recommend, and now in the script (dv-generated-tests.sh) we can use new test description in the cva6_base_testlist.yaml, and also tests in the riscv-dv, simply to give the list_num variable =0, to re-use tests in the base_testlist.yaml in riscv-dv.
Now the cva6_base_testlist.yaml don't contain any test from the base_testlist file in riscv-dv, but we can re-used these tests using the script

MikeOpenHWGroup (approved):

PR #1363, opened 2022-08-09
Formal

silabs-robin:
This PR adds scripts and other files to run formal property checking on cv32e40x, using JasperGold.

There is room for improvement, but I don't have too much time to fix small details on this, so I think it is better to merge it as it is now and make new PRs in the future for improving it.

The main part is `fv.tcl`, everything else is just helpers.
It works by compiling ("analyzing") source files for the core, for the core-v-verif "tb", and some additional helper files.
Then it elaborates "tb" as top level, and you can run proofs for core-v-verif assertions.
However, certain packages/interfaces/etc have (temporary) "dummy" definitions because core-v-verif isn't formal friendly (yet).


MikeOpenHWGroup:
This is tremendous @silabs-robin, thanks for this. Can we move it  from `cv32e40x/sim/fv` to `cv32e40x/fv`?

silabs-robin:
> Can we move it from cv32e40x/sim/fv to cv32e40x/fv?

Certainly.
I also considered it before but wasn't sure.

Fixed.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
> This PR adds scripts and other files to run formal property checking on cv32e40x, using JasperGold.

This is huge @silabs-robin, thanks for this.

PR #1367, opened 2022-08-10
Debug privilege tests

silabs-anvesten:
Suite of 6 test-cases focused on verifying correct behavior when switching between user and debug mode.

### DIFF
--- a/cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
+++ b/cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
@@ -159,6 +159,7 @@ int main(void){// TODO: test will failed until issue #278 in core-v-verif/cv32e4
   /*
   Have dcsr.ebreaku=0, be in U-mode, execute ebreak, ensure "normal" ebreak behavior and no debug entry.
   */
+  assert_or_die(glb_debug_status, 0, "Error: Core was in debug mode before test start!");
   set_u_mode();
   asm volatile("ebreak");
   assert_or_die(wmcause, 0x3, "Error: Illegal 'ebreak' did not trigger breakpoint exception!");
--- a/cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
+++ b/cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
@@ -1,0 +1,260 @@
+/*
+** Copyright 2022 OpenHW Group
+**
+** SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+** Licensed under the Solderpad Hardware License v 2.1 (the "License"); you may not use this file except in compliance
+** with the License, or, at your option, the Apache License version 2.0.  You may obtain a copy of the License at
+**                                        https://solderpad.org/licenses/SHL-2.1/
+** Unless required by applicable law or agreed to in writing, any work distributed under the License is distributed on
+** an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the License for the
+** specific language governing permissions and limitations under the License.
+*******************************************************************************
+**
+** Several smaller directed test-cases asserting proper execution between debug and user mode.
+**
+*******************************************************************************
+*/
+
+
+#include <stdio.h>
+#include <stdlib.h>
+#include "corev_uvmt.h"
+
+
+// Global values to set test_cases for the assembly debug code, or write to global values which the .c code can read and assert
+
+// Written by assembly debug code to signal .c code that debug is exiting.
+volatile uint8_t glb_debug_status               = 0;
+// Written by main code to set test case in assembly debug code 
+volatile uint8_t glb_expect_ebreaku_exc         = 0;
+// Written by main code to set test case in assembly debug code 
+volatile uint8_t glb_setmprv_test               = 0; 
+// Written by main code to set test case in assembly debug code
+volatile uint8_t glb_check_prv_test             = 0; 
+// Written by main code to set test case in assembly debug code
+volatile uint8_t glb_check_prv_test_2           = 0; 
+// Written by main code to set test case in assembly debug code
+volatile uint8_t glb_unsupported_mode_test      = 0;
+// Written by main code to set test case in assembly debug code
+volatile uint8_t glb_ebreaku_one_test           = 0;
+// Written by main code to set test case in assembly debug code
+volatile uint8_t glb_check_previous_prv_test    = 0;
+// Written by assembly debug code. Asserts test case failure
+volatile uint8_t glb_unsupported_check_failed   = 0;
+// Written by assembly debug code. Asserts test case failure 
+volatile uint8_t glb_ebreaku_one_check          = 0;
+// Written by assembly debug code. Contains dcsr register value
+volatile uint32_t glb_check_prv_test_val        = 0;
+// Written by assembly debug code. Contains dcsr register value
+volatile uint32_t glb_dcsr_register             = 0;
+
+// global variable written while in trap handler. Used to check mcause and mstatus registers in assertions
+volatile uint32_t wmcause, wmstatus;
+// standard value for the mstatus register
+#define MSTATUS_STD_VAL 0x1800
+// mstatus.MPRV bit
+#define MPRV_BIT 17
+// Debug define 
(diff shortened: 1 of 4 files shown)


silabs-anvesten:
@silabs-robin

silabs-anvesten:
@silabs-robin Please add a 'do not merge' label to this. There is some commit issue which will be fixed, then it should be ready for review. All debug tests pass or are otherwise in the states we agreeed.

silabs-anvesten:
@silabs-robin Fixed the commit issue, this should be ready for review. Beware you cannot run the tests without the cfg/pmp.yaml file found in #1350 as I didn't want duplicates of merge issues

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
```diff
@@ -0,0 +1,260 @@
...
+  DEBUG_REQ_CONTROL_REG = debug_req_control.bits; 
+  // waits for debug to finish
+  while(glb_debug_status != 1){
+      ;
+  }
+  // resets the debug status bit
+  glb_debug_status = 0;
+}
+
+int main(void){// TODO: test will failed until issue #278 in core-v-verif/cv32e40s is fixed
```
silabs-robin, line 152: Will the test fail when run, or are the expected-to-fail parts just commented out?
(I'd prefer the latter, with a todo at the place where code is commented out.)

silabs-anvesten: The actual tests will pass, but it will print FAILED at the end because of the assert issue in #278

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
```diff
@@ -0,0 +1,260 @@
...
+
+
+
+  // Test start:
+  /*
+  Have dcsr.ebreaku=0, be in U-mode, execute ebreak, ensure "normal" ebreak behavior and no debug entry.
+  */
+  set_u_mode();
+  asm volatile("ebreak");
+  assert_or_die(wmcause, 0x3, "Error: Illegal 'ebreak' did not trigger breakpoint exception!");
```
silabs-robin, line 164: Would it make sense to also check `glb_debug_status`?

silabs-anvesten: It could. Although, `glb_debug_status` should be 0 at all times, except the time in-between debug exit and before `run_debug_mode()` exits. The only other case I see is if for some other reason the core is in debug mode before main() executes, but I don't know if this is applicable? And if it is then I don't know if my version of the `debugger_start:` symbol has loaded as the core would already be using some version of this symbol when it entered debug mode. 

In any case I don't see any issues with adding an `assert_or_die` case checking that the `glb_debug_status` but is low.

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
```diff
@@ -0,0 +1,260 @@
...
+  /* 
+  Transition out of D-mode, ensure that executions starts in the same privilege mode as was indicated in dcsr.prv (dcsr.prv=M and dcsr.prv=U).
+  */
+  // First check machine mode.
+  glb_check_prv_test = 1;
+  run_debug_mode();
+  asm volatile("ecall");
+  uint32_t check_MPP_bit = get_field(wmstatus, MPP_FIELD[0], MPP_FIELD[1]);
+  uint32_t check_prv_bit = get_field(glb_check_prv_test_val, PRV_FIELD[0], PRV_FIELD[1]);
+  assert_or_die(check_MPP_bit, check_prv_bit, "Error: previous privilege mode did not match the one set in dcsr.prv before exiting debug mode!\n");
```
silabs-robin, line 197: Could we add a check that the mode is indeed M/U as expected? Just to be safe.

cv32e40s/tests/programs/custom/debug_priv_test/debugger.S
```diff
@@ -0,0 +1,225 @@
...
+/*
+Write unsupported modes to dcsr.prv, ensure the value read back is unchanged from the previous value.
+*/
+Check_unsupported_modes:
+	li t0, 0x3
+	csrrs x0, dcsr, t0 // preload supported M-mode to get a reference
+
+	// Check S-mode
+	li t1, 0x1
+	csrrs x0, dcsr, t1
```
silabs-robin, line 170: Will this attempt to set `'b 1` and the previous value is `'b 11`, so it tries to change from `'b 11` to `'b 11`?

silabs-anvesten: This is probably a legitimate bug. I'll change this to read, change and write back the register.

cv32e40s/tests/programs/custom/debug_priv_test/debugger.S
```diff
@@ -0,0 +1,225 @@
...
+	j end_debugger
+
+
+
+/*
+Have dcsr.ebreaku=1, be in U-mode, execute ebreak, ensure debug entry happens instead of "normal" ebreak behavior.
+*/
+Set_ebreaku_bit:
+	lw t0, 0(x0)
+	lui t0, 1
```
silabs-robin, line 201: While I like to avoid pseudo instructions, I think `li` would be easier to read here. Because I can't quickly determine if the correct bit gets set.

silabs-anvesten: I think It would make sense to add a li with some #define value specifying what the value is. I can do that

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
silabs-robin: Right. That is alright then.

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
silabs-robin: > Although, glb_debug_status should be 0 at all times, except the time in-between debug exit and before run_debug_mode() exits.

If the ebreak mistakenly caused debug mode to be entered, would `glb_debug_status` be set then?

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
silabs-robin: Does this trust that "dcsr.prv" was set to M/U? And does the first test trust that preceding tests left the mode in M?
It is probably correct, but it seems like a place where a small slip could cause the test to do something else than what we think it does.

cv32e40s/tests/programs/custom/debug_priv_test/debugger.S
silabs-robin: That would be nice.

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
silabs-anvesten: In a way, but only until the debug assembly code exits, and right after the `run_debug_mode()` function will set `glb_debug_status` to `0` again so that it has something to compare to next time its called. The way I've rewritten it `glb_debug_status` isn't meant to be a check for whether debug mode is enabled, its there to tell the `run_debug_mode()` that the debug assembly code has finished executing and that the .c code can now move on with testing. This exists because it seemed to be the case for the other custom test in the cv32e40s called "debug_test" where the executive .c code seems to wait for the assembly program to finish before moving on.

cv32e40s/tests/programs/custom/debug_priv_test/debug_priv_test.c
silabs-anvesten: Both of the preceeding tests trap and the trap handler always puts the privilege mode back to machine mode. Therefore, I believe it is safe to assume that the privilege mode is machine mode at the start of the test in question. The test does not trust dcsr.prv functionality, it first sets dcsr.prv to the intended mode (M or U) and then once debug mode is finished it immediately does an ecall where the trap handler records mstatus. I believe that mstatus would then have to hold the immediate privilege mode that runs when coming out of debug mode and right before the ecall. If the dcsr.prv and mstatus privilege modes do not match then I believe that means the dcsr.prv is not functioning correctly. However, the test passes so then the assertions must also have passed for lines 198 and 207.

PR #1370, opened 2022-08-12
Interrupt test finished and passes

silabs-anvesten:
Four test-cases checking privilege mode functionality between user and machine mode trap/interrupt handlers.


silabs-anvesten:
@silabs-robin Pushed!

MikeOpenHWGroup (approved):

PR #1373, opened 2022-08-17
Zc assertion set, common support logic and rvfi_if functions

silabs-mateilga:
- Added Zc assertion set
- Created a common module for support logic and accompanying interface, opening up for shared support logic between assertion sets
- Added functions to the rvfi_if for use with assertion sets


cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -0,0 +1,72 @@
+//
+// Copyright 2020 OpenHW Group
+// Copyright 2020 Datum Technology Corporation
```
silabs-robin, line 3: Date and owner.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -0,0 +1,72 @@
...
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
```
silabs-robin, line 5: _(Btw, I see some files use "2.1". Maybe we should change our default in the future?)_

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -134,6 +134,97 @@ interface uvma_rvfi_instr_if
...
+  end
+  else begin
+    return 0;
+  end
+
+  endfunction : match_instr
+
+// Return wdata of register "gpr"
+function bit [ DEFAULT_XLEN-1:0] get_gpr_wdata( int gpr);
+  return rvfi_gpr_wdata[gpr* DEFAULT_XLEN +: DEFAULT_XLEN];
```
silabs-robin, line 162: Should it use `rvfi_gpr_wmask`?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -134,6 +134,97 @@ interface uvma_rvfi_instr_if
...
+endfunction : get_gpr_rdata
+
+// Return valid data of memory transaction "txn"
+function bit [ DEFAULT_XLEN-1:0] get_mem_data_word( int txn);
+  bit [ DEFAULT_XLEN-1:0] ret;
+
+  for (int i = 0; i <  DEFAULT_XLEN/8; i++) begin
+    if (rvfi_mem_wmask[(txn* DEFAULT_XLEN/8) + i]) begin
+      ret[i*8 +:8] = rvfi_mem_wdata[((txn* DEFAULT_XLEN) + (i*8)) +:8];
+    end else begin
```
silabs-robin, line 177: If wmask is `0011`, will it assume at rmask is `1100` and that there is valid data om rdata?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: This function is a replica of the one that exists for the uvm sequence object, and in that context this unqualified function is necessary. We could also implement a "get_gpr_retirement_data" like the one that exists for csr's, that would return wdata if wmask is high and rdata if not, but I do think this function should remain as is.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: That is indeed the assumption. The check_mem_act() function can be used to qualify this if necessary.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -0,0 +1,72 @@
...
+    // ---------------------------------------------------------------------------
+    // Local parameters
+    // ---------------------------------------------------------------------------
+
+
+
+    // ---------------------------------------------------------------------------
+    // Local variables
+    // ---------------------------------------------------------------------------
+    logic exception_active;
```
silabs-robin, line 36: What timing does this have? Could we have a prefix or comment to indicate what this signal means?

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -0,0 +1,72 @@
...
+    // Used to verify exception timing with multiop instruction
+    always @(posedge support_if.clk_i or negedge support_if.rst_ni) begin
+      if (support_if.rst_ni) begin
+        support_if.req_after_exception_o <= 0;
+        exception_active <= 0;
+      end else  begin
+
+        if (rvfi.rvfi_valid) begin
+          support_if.req_after_exception_o <= 0;
+          exception_active <= 0;
```
silabs-robin, line 55: Why can there be no exception_active during rvfi_valid? The "set" logic below uses internal signals, and those signals could be true at the same time as rvfi_valid.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
```diff
@@ -263,4 +263,51 @@ interface uvmt_cv32e40s_debug_cov_assert_if
...
+interface uvmt_cv32e40s_support_logic_if
+   import cv32e40s_pkg::*;
+   import uvma_rvfi_pkg::*;
+   (
+      input    clk_i,
+      input    rst_ni,
+
+      // core signals
+      input ctrl_fsm_t  ctrl_fsm_o_i,
+      input             data_bus_req_i
```
silabs-robin, line 275: Is this req `_i` because it is an input to this interface? The signal above uses `_o_i` to say that it is an output from where it originates but an input to this interface?

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-mateilga: The signal and support logic is only relevant for multiop instructions, And is evaluated when an exception is encountered during a suboperation. As such, it is not relevant for the next instruction after an rvfi_valid, neither is is relevant for single cycle instructions.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
silabs-mateilga: Yes, all input signals on the interface uses the naming from where it originates in the core with the _i suffix. outputs have the _o suffix. I agree that it can become a bit difficult, and am open to suggestions for other naming conventions.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin: Ok, got it. This is just for getting the `wdata` signal itself (among the array of multi-writes), not the "data written".

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin: I think rdata is allowed to contain garbage data as long as rmask is 0.
And check_mem_act() doesn't have any per-byte mask.
Does that mean one would later have to apply something like get_mem_data_wmask?
Seems it is theoretically possible to do a halfword write and get upper word data from remnant rdata.

silabs-mateilga: That depends on how this list is populated from the core, maybe @silabs-oysteink could comment?

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-robin: If an ADD is reported on rvfi_valid, while the core is currently detecting exception on a multiop instruction, then the logic will not detect the exception?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-oysteink: As far as I can recall without digging to deep into the rvfi, the rmask and wmask are mutually exclusive depending on the type (load or store) of the memory operation.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-mateilga: Hmm, I think you are right, I will modify this before progressing.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
MikeOpenHWGroup: Yes, Solderpad 2.1 is preferable, although if the file was originally published under 2.0 (or, in some cases 0.5.1), that is fine.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
MikeOpenHWGroup: > As far as I can recall without digging to deep into the rvfi, the rmask and wmask are mutually exclusive depending on the type (load or store) of the memory operation.

Intuitively, I think this would be true in the CV32E40 cores, but I do not see anything in the RVFI specs (either in the SymbioticEDA or CV32E40X versions) that explicitly states this as a rule for the interface.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-mateilga: comment added

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: Unsure if there is any action here, I guess the question if we get a partial write with nothing in rmask if we will get the unwritten(and unread) bytes as potentially garbage data(with no qualification) or 0's, which is also garbage data. Does it really make a difference?

silabs-robin: > Unsure if there is any action here

It might be fine as it is now.

silabs-robin (approved):

MikeOpenHWGroup (approved):

PR #1377, opened 2022-08-22
Formal updates

silabs-robin:
This PR improves on the formal scripts for 40x.
(But please read the description below and let us discuss a little bit before merging.)

* "git clone" is now automatic
* Started "mk/fv/" directory for re-usability
* Using defined names instead of hardcoding "../../" paths
* Slightly clearer names for vendor-specific code
* ".flist" for core-specific files

Is this going in the right direction?
Some parts of this I feel is definitely good, but other parts I think might be good but lack certainty.
I feel confident in reusing "mk/Common.mk" for "CLONE_CV_CORE_CMD".
Using an ".flist" also seems good, because then each core can have its own file list while re-using other shared scriptware.
Is it really a good thing to replace the "../../../" with "${CV_UVMT_PATH}" etc?
Some of the structure is just mimicking the "uvmt.mk" etc, but I'm not sure if those are already pristine.

In the future I'd like to make clean separation between tool vendors, like "mk/fv/jg.mk" or smt.
I also want to port this to the 40s.


MikeOpenHWGroup:
> Is this going in the right direction? Some parts of this I feel is definitely good, but other parts I think might be good but lack certainty.

I think this PR is going in the right direction and we can work together to improve our "certainty".

> I feel confident in reusing "mk/Common.mk" for "CLONE_CV_CORE_CMD".

Agreed.

> Using an ".flist" also seems good, because then each core can have its own file list while re-using other shared scriptware.

I would like to have one-and-only-one manifest (flist), although I recognize that won't happen without significant effort even if adopt something like [edalize](https://edalize.readthedocs.io/en/latest/).  Is there a way we can re-use an existing manifest rather than create a new one?

> Is it really a good thing to replace the "../../../" with "${CV_UVMT_PATH}" etc?

Yes.  We will need to document that in the README, but can come later.

> Some of the structure is just mimicking the "uvmt.mk" etc, but I'm not sure if those are already pristine.
:rofl: 

In the future I'd like to make clean separation between tool vendors, like "mk/fv/jg.mk" or smt.

silabs-robin:
> > Using an ".flist" also seems good, because then each core can have its own file list while re-using other shared scriptware.
>
> I would like to have one-and-only-one manifest (flist), although I recognize that won't happen without significant effort even if adopt something like edalize. Is there a way we can re-use an existing manifest rather than create a new one?

Do you mean, to avoid creating a new manifest file for each core? Or for each tool? Or... you maybe mean to re-use the manifest files that we already have for simulation?

The reason for not re-using `tb/uvmt/uvmt_cv32e40s.flist` is the same reason that we need the [dummy_pkg.sv](https://github.com/openhwgroup/core-v-verif/blob/29b6869465b1a1a29108293d6e379611c294fc18/cv32e40x/fv/dummy_pkg.sv#L1).
There is too much uvm and other non-formal-friendly constructs in the files that are included by those manifests.
But that is a good point, that duplicating all that information is a bit annoying.
(Though, it might speed up "compilation" to omit unneeded files?)
I don't know what overhead and worth edalize would bring, and if it would work, and I see jasper gold is not in their list.


> > Is it really a good thing to replace the "../../../" with "${CV_UVMT_PATH}" etc?
>
> Yes. We will need to document that in the README, but can come later.

Ok, nice to hear an extra opinion on whether that was a smart move or not.
One _could_ change "cv32e40x" with one of those defined names too, but there isn't much need for it(?), because this file list is specific to 40x and can't be shared anyway, afaict.

I agree about the readme. We should really have a more detailed one. But I also don't want to write it while the structure is still volatile and undecided.

silabs-robin:
> We could have a very simple manifest that looks something like this:
> 
> ```
> dummy_pkg.sv
> -f b/uvmt/uvmt_cv32e40s.flist
> ```

Yes, that is nice. And if/when we get rid of "dummy_pkg.sv" we could just use "tb/uvmt/uvmt_cv32e40s.flist" directly.
I'll keep this in mind when doing future updates.

If you are ok with everything else in the PR then I think it is ready to merge.

MikeOpenHWGroup:
>  I'll keep this in mind when doing future updates.

Can we do it now?  Once a duplicate manifest creeps into the code base, it's hard to find and fix them.

MikeOpenHWGroup:
After an in-person discussion with @silabs-robin and the SiLabs team, it was decided to merge this in as-is and resolve the manifest issue at a later date.

MikeOpenHWGroup (approved):

PR #1380, opened 2022-08-24
rvfi_trap changed to struct

silabs-mateilga:
Switched rvfi_trap in the tb if definition to the struct matching the core rvfi"


silabs-hfegran (approved):
Thanks, LGTM @silabs-mateilga

PR #1383, opened 2022-08-24
Workaround for known DSIM issue

MikeOpenHWGroup:
As of version 20220328.14.0 or earlier, DSIM cannot accept localparams as arguments of properties.   Tracking as Metrics service request 1253.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_zc_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_zc_assert.sv
@@ -29,6 +29,7 @@ module uvmt_cv32e40s_zc_assert
   // ---------------------------------------------------------------------------
   // Local parameters
   // ---------------------------------------------------------------------------
+`ifndef DSIM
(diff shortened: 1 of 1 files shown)


silabs-mateilga (approved):
This is fine for now, but should there be an issue to revert when DSIM supports this? This kind of workaround has a tendency to "hang around" far beyond their usefulness if not monitored.

silabs-hfegran (approved):

MikeOpenHWGroup:
> This is fine for now, but should there be an issue to revert when DSIM supports this? This kind of workaround has a tendency to "hang around" far beyond their usefulness if not monitored.

Good point.  If I am reading the Metrics DSIM User Manual correctly, this appears to be a known issue.  As such, it may not be resolved for a long time.  I had intended to wait for Metrics to respond before creating the issue, but you are right - there is no need to wait.

MikeOpenHWGroup:
Update: Metrics has acknowledged the problem and are able to reproduce it at their end.  I will use this issue to keep everyone informed.

MikeOpenHWGroup:
Based on @silabs-hfegran's and @silabs-mateilga's approval, I am merging this in.

PR #1385, opened 2022-08-25
Cv32e40x/merge dev to rel 61

silabs-robin:
Merge `40x/dev -> 40x/rel`.

Merge procedure. Following the same old way of doing it.
This merge was done automatically by scripts on Jenkins, then `rel_check` ran and it passed 100%, then the branch was automatically pushed to my repo.
After 40x and 40s has done "dev-to-rel", the next step is merging both "rel" branches to "master".

PR #1361, opened 2022-08-05
Added tests for PMP

silabs-kun:
Signed-off-by: Kun Zhu <Kun.Zhu@silabs.com>

PR #1386, opened 2022-08-25
Cv32e40s/merge dev to rel 23

silabs-robin:
Same description as https://github.com/openhwgroup/core-v-verif/pull/1385

Note, "rel_check" for this 40s merge did not pass.
We have 10 failures:
* Counters:
    * perf_counters_instructions
    * hpmcounter_basic_test
* CSRs:
    * cv32e40s_csr_access_test
    * cv32e40s_readonly_csr_access_test
* Illegal:
    * corev_rand_illegal_instr_test
    * corev_rand_illegal_instr_test
* Debug:
    * debug_test
    * debug_test_reset
    * corev_rand_debug_ebreak
    * corev_rand_debug_ebreak

Some of these are known, but not all.
We are, however, prepared to do the merge anyway, because the 40s has been in a developmental volatile state on the past previous merges too and we see it as more worth to get progress on the merging between x and s.


cv32e40p/sim/ExternalRepos.mk
```diff
@@ -31,10 +31,13 @@ EMBENCH_REPO    ?= https://github.com/embench/embench-iot.git
 EMBENCH_BRANCH  ?= master
 EMBENCH_HASH    ?= 6934ddd1ff445245ee032d4258fdeb9828b72af4
 
-COMPLIANCE_REPO   ?= https://github.com/riscv/riscv-compliance
-COMPLIANCE_BRANCH ?= master
-# 2020-08-19
-COMPLIANCE_HASH   ?= c21a2e86afa3f7d4292a2dd26b759f3f29cde497
+# TODO: silabs-hfegran: Temporary fork compliance suite to support bitmanip and
```
silabs-robin, line 34: @silabs-hfegran, is this change to 40p deliberate?

cv32e40x/tb/core/cv32e40x_tb_wrapper.sv
```diff
@@ -22,7 +22,7 @@ module cv32e40x_tb_wrapper
                 BOOT_ADDR         = 'h80,
```
silabs-robin, line 22: The reason that so many X files have changed in this S merge (IIRC) is that we did a "cheat merge" from "x->s", to quickly get some changes over without going via "master".

cv32e40p/sim/ExternalRepos.mk
silabs-hfegran: I believe so, changes were made to support (at the time) latest compliance suite, but bitmanip had yet to be included. This required changes to top-level core-v-verif files, and as not to break compatibility, this was added to 40p as well.

cv32e40s/tests/cfg/pmp.yaml
```diff
@@ -0,0 +1,9 @@
+name: pmp
+description: Default pmp configuration for CV32E40S simulations
+compile_flags: >
+   +define+PMP_ENABLE_64
+ovpsim: >
+    --override root/cpu/PMP_registers=64
+cflags: >
+plusargs: >
+cv_sw_march: rv32imc
```
silabs-hfegran, line 9: While not a result of the merge - perhaps we should enable the zb*-instructions for the compiler as well here?

silabs-hfegran (approved):
Looks good to me

cv32e40s/tests/cfg/pmp.yaml
silabs-robin: I think Kun's PR will fix that? https://github.com/openhwgroup/core-v-verif/pull/1361/files#diff-3754c84dacc70eb0061c4ff9a4d8c746d662c63032ae35af064dc6aaf6ebf997R11

MikeOpenHWGroup:
Could we get #1383 included in this PR?  DSIM will not compile the e40s environment without that workaround which kills the automated CI flow driven by this repo.

Otherwise, this PR LGTM.

silabs-robin:
> Could we get #1383 included in this PR?

Yes. (Not technically "this" PR, but yes.) When #1383 is merged, I can restart the "dev-to-rel" scripts and substitute this PR with a new one that includes that change.

MikeOpenHWGroup:
I was hoping to get #1383 merged into `cv32e40s/dev` first, and then update this PR to include the update (only one file).  Otherwise the cv32e40s dev and rel branches will both be unable to run a CI under dsim and we run the risk of propagating that to master.

I just now noticed that #1383 has been approved by @silabs-hfegran and @silabs-mateilga, so I took the liberty of merging it.  Can you add that to this PR?

silabs-robin:
This PR is replaced with a new PR https://github.com/openhwgroup/core-v-verif/pull/1393 which includes the requested inclusion of #1383.

PR #1388, opened 2022-08-25
Add initial 'regression' custom test suite.

zchamski:
The 'cva6/custom/issues' area is intended to hold tests of issues identified during CVA6 development.

Files added:

* cva6/regress/issue-tests.sh: New.
* cva6/tests/custom/issues/compressed-fpreg-commits-rv32.S: Ditto.
* cva6/tests/custom/issues/compressed-fpreg-commits-rv64.S: Ditto.
* cva6/tests/testlist_issues.yaml: Ditto.

Signed-off-by: Zbigniew Chamski <zbigniew.chamski@thalesgroup.com>

PR #1392, opened 2022-08-30
Add test suite for issues found during CVA6 development.

zchamski:
The `cva6/tests/custom/issues` test family is intended to exercise any issues found during CVA6 development - as opposed to upstream test suites.   The initial two tests exercise the proper formatting of writes into FP registers triggered by compressed instructions whereby an issue in the RTL tracer caused discrepancies between RTL logs and the output of Spike.

This PR supersedes #1388. 
 
Files changed/added:

* cva6/regress/issue-tests.sh: New.
* cva6/tests/custom/issues/compressed-fpreg-commits-rv32.S: Ditto.
* cva6/tests/custom/issues/compressed-fpreg-commits-rv64.S: Ditto.
* cva6/tests/testlist_issues.yaml: Ditto.

Signed-off-by: Zbigniew Chamski <zbigniew.chamski@thalesgroup.com>

### DIFF
--- a/cva6/regress/issue-tests.sh
+++ b/cva6/regress/issue-tests.sh
@@ -1,0 +1,34 @@
+# Copyright 2022 Thales DIS France
+#
+# Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+# You may obtain a copy of the License at https://solderpad.org/licenses/
+#
+# Original Author: Zbigniew CHAMSKI (zbigniew.chamski@thalesgroup.fr)
+
+# where are the tools
+if ! [ -n "$RISCV" ]; then
+  echo "Error: RISCV variable undefined"
+  return
+fi
+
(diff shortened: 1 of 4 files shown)


MikeOpenHWGroup:
Hi @zchamski, I am not sure I understand the purpose of the `cva6/tests/custom/issues` test suite.  Is it to create a dedicated test-program for _every_ RTL issue found during verification, or just for test-programs that find bugs that were not included in the DV plan?

In any case, I would suggest that it is equally important to have specific functional coverage for all bugs that were not on features identified in the DV plan.  In many cases, it is better to have the coverage than to have the test.

zchamski:
Hi @MikeOpenHWGroup , I agree that coverage is the primary way to go.  However, coverage implies the availability of a spec and a corresponding DV plan, and for CVA6 they are "in the making".

The idea behind the `cva6/tests/custom/issues` suite is to capture issues that are identified now but are not associated (yet...) with a specific entry in the specification, or are beyond the scope of CVA6 core specification, e.g., the testbench.  This is exactly the case here: the initial two tests check the correct generation of RTL logs in the TB.

Issue #1389, opened 2022-08-25
Need longer-term fix for DSIM workaround (see pull-request #1383)


MikeOpenHWGroup:
Pull Request #1383 implements a workaround for a known Metrics DSIM issue.  This core-v-verif issue exists to ensure the workaround is removed once Metrics resolves their issue.

### Type
This is a tool issue that (to the best of my knowledge) only affects Metrics DSIM.

### Steps to Reproduce
Running any simulation of the CV32E40S on the cv32e40s/dev branch, hash [35f8f26](https://github.com/openhwgroup/core-v-verif/commit/35f8f26cb4dbfcc3b66d04f132639f1ab7569fce), will illustrate whether your simulator is affected by #1383.

PR #1393, opened 2022-08-31
Cv32e40s/merge dev to rel 24

silabs-robin:
Same as the https://github.com/openhwgroup/core-v-verif/pull/1386 PR, but including the https://github.com/openhwgroup/core-v-verif/pull/1383 changes.


MikeOpenHWGroup:
This PR LGTM with a few caveats:
1. We will need to back out the DSIM workaround once #1389 is resolved.
2. The `debug_test` from the ci_check regression fails (I believe this is a "known failure").
3. 10 of 55 tests in the cv32e40s/release CI regression fail:
   - perf_counters_instructions
   - hpmcounter_basic_test
   - debug_test (same test as in the ci_check regression)
   - debug_test_reset
   - cv32e40s_csr_access_test
   - cv32e40s_readonly_csr_access_test
   - corev_rand_illegal_instr_test (two failures of the same pseudo-random testcase)
   - corev_rand_debug_ebreak  (two failures of the same pseudo-random testcase)

I believe that all of the above are known failures.  @silabs-hfegran, if that is the case, please go ahead and merge this in.

silabs-robin:
>     1. We will need to back out the DSIM workaround once [Need longer-term fix for workaround (see pull-request #1383) #1389](https://github.com/openhwgroup/core-v-verif/issues/1389) is resolved.

Yes.

--------

>     2. The `debug_test` from the ci_check regression fails (I believe this is a "known failure").

This is "known". We don't know the cause, but we know it fails.

--------

>     3. 10 of 55 tests in the cv32e40s/release CI regression fail:

Here is what I wrote in the previous PR:

> Note, "rel_check" for this 40s merge did not pass.
> We have 10 failures:
> [...]
> Some of these are known, but not all.
> We are, however, prepared to do the merge anyway, because the 40s has been in a developmental volatile state on the past previous merges too and we see it as more worth to get progress on the merging between x and s.

PR #1399, opened 2022-09-02
PMP Asserts Addendum

silabs-robin:
This PR adds additional PMP assertions in accordance with the vplan.

Passes ci_check (except for the pre-existing known "debug_test" issue).

Most notable differences:
* New asserts to check more of the vplan
* Split of `pmp_assert` / `pmprvfi_assert` / `pmp_model`

NBs:
* Note(!), this does not encompass _all_ of the vplan's ambitiously designated "assertion check" items, but the ones deemed to be of high enough priority and importance.
* Note(!), there are still some counter examples, and 1 or 2 RTL/RVFI problems are known to be the cause of this, so _if_ the assertion set needs updates correspondingly then it can be done later.

@silabs-hfegran @MikeOpenHWGroup


cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
```diff
@@ -63,10 +63,12 @@ ${TBSRC_HOME}/core/tb_riscv/riscv_gnt_stall.sv
 ${TBSRC_HOME}/core/tb_riscv/riscv_rvalid_stall.sv
 ${TBSRC_HOME}/core/tb_riscv/riscv_random_interrupt_generator.sv
```
MikeOpenHWGroup, line 64: I think we can remove these "core" verification components from this flist as they are no longer used in the E40S UVM environment.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
```diff
@@ -63,10 +63,12 @@ ${TBSRC_HOME}/core/tb_riscv/riscv_gnt_stall.sv
...
 
-${DV_UVMT_PATH}/uvmt_cv32e40s_interrupt_assert.sv
```
MikeOpenHWGroup, line 66: We are creating a large set of Assertion files.  This is great!  Does it make sense to package them separately?

MikeOpenHWGroup (commented):
LGTM.  I have a couple of minor comments that need not gate merging.  Will defer to @silabs-hfegran on that.

cv32e40s/tests/cfg/pmp_g3r3.yaml
```diff
@@ -0,0 +1,11 @@
...
+description: Default pmp configuration for CV32E40S simulations
+compile_flags:
+   +define+PMP_G3R3
+ovpsim: >
+    --override root/cpu/PMP_registers=3
+    --override root/cpu/PMP_grain=3
+    #--showoverrides
+cflags: >
+plusargs: >
+cv_sw_march: rv32imc
```
silabs-hfegran, line 11: Why is this config not running with bitmanip?

cv32e40s/tests/cfg/pmp.yaml
```diff
@@ -6,4 +6,4 @@ ovpsim: >
     --override root/cpu/PMP_registers=64
 cflags: >
 plusargs: >
-cv_sw_march: rv32imc
\ No newline at end of file
+cv_sw_march: rv32imc
```
silabs-hfegran, line 9: same comment as below, why no bitmanip?

silabs-hfegran:
Thanks @silabs-robin, In general looks good, apart for a couple of config comments.

I have a question for @MikeOpenHWGroup - did we agree on a policy for squashing/not squashing large commit histories like this? Although I very much like that this has been rebased, there are quite a lot of commits here that makes sense during development, but are perhaps not as useful for the upstream branch? Any thoughts on this?

cv32e40s/tests/cfg/pmp.yaml
silabs-robin: Should be fixed now.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-robin: I've thought the same.

How do you envision it?
Simply create a new directory within "tb/uvmt/"?
Create a new directory like e.g. "tb/asserts/"?
Separate these assertion files into a new ".flist" file?

Can we do it in a separate PR?

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-robin: Agreed. If we do "package [the assertion files] separately" in a new PR, then we can maybe remove these "core" components from the uvmt flist at the same time?

silabs-robin:
> did we agree on a policy for squashing/not

I think it is destructive. But it could be nice to try it out and experiment to see if it pans out as I think.

It is gonna be some hours before Mike awakens, so I'm fine with either of 1) wait til he rises and a discussion is held, 2) squash this PR, 3) regular-merge this PR.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
MikeOpenHWGroup: I was thinking something like `tb/asserts`, because they are not really a "UVM-thing" and ideally these assertions can be (are being???) used in both simulation and formal.

Yes, this can be a separate PR.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
MikeOpenHWGroup: Agreed.

MikeOpenHWGroup:
> > did we agree on a policy for squashing/not

> I think it is destructive. But it could be nice to try it out and experiment to see if it pans out as I think.

AFAIK, in core-v-verif we do not have a formal policy on squashing.  @silabs-robin is right that squashing destroys commit information.  On the other hand, it is not clear to me how often users of this repo actually require commit-by-commit data (as opposed to merge-by-merge), and it is easy to get "lost in the commits" of a big PR.  So while I prefer to keep all commit data, it should not be at the cost of clarity.

For now I prefer to leave it to the discretion of the Committer.  If we need a policy in the future, that will make itself apparent.

silabs-hfegran (approved):

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-hfegran: I agree that it makes sense to put these in a common location, tb/asserts makes sense to me

PR #1401, opened 2022-09-05
S/dev/xsecure assertions

silabs-krdosvik:
Xsecure assertion for the following features:
- minor alert (lfsr lockup and exception, but not nmi errors)
- data independent timing, 
- dummy instr
uctions, 
- reduction of profiling infrastructure

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
@@ -107,38 +107,81 @@ endinterface : uvmt_cv32e40s_core_status_if
 
 
 // Interface to xsecure assertions and covergroups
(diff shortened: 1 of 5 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
```diff
@@ -104,6 +104,86 @@ interface uvmt_cv32e40s_core_status_if (
...
+    import cv32e40s_rvfi_pkg::*;
+    (
+    input logic  clk_i,
+    input logic  rst_ni,
+
+    // CORE:
+    input logic core_clk_gated,
+
+    // CSR
+    //input logic core_alert_major_o,
```
silabs-mateilga, line 121: Commented out code should not be PRed unless it is temporarily disabled, and preferably marked with a TODO for when it is to be commented back in.

PR #1405, opened 2022-09-06
Isacov Agent: fix miss assign signals in cover group declaration

AyoubJalali:


silabs-robin (approved):
Seems correct to me.
Nice catch.

JeanRochCoulon:
Thanks @AyoubJalali for this !

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @AyoubJalali!  I also tested this against the cv32e40x and these changes work there as well (as expected - it would be very confusing if they did not).

Although it is not directly relate to this PR, while looking at this I found a bug in `uvma_rvfi_instr_mon.sv` that only seems to affect Metrics DSIM.  See #1408.

PR #1411, opened 2022-09-07
Cv32e40x/merge dev to rel 69

silabs-robin:
Did this step of the merge procedure again, because it was requested that the "rvfi_trap struct" changes were included.
You can see from the 7 commits that there is not much that has changed.
Passes rel_check 100%, as the previous one did before.

(The next step after merging this is to do the manual merge of x and s into master, etc.)


silabs-hfegran (approved):
LGTM

PR #1413, opened 2022-09-08
Add more pmp configs

silabs-robin:
As the title says, this PR adds more PMP configs.
Adding these configs because we wanted more variation in regressions (specifically G = {0, 1, >1} to stimulate generate blocks).
The other configs are picked semi-arbitrarily, with regards to PMA configs and the nature of PMP behavior and CSRs etc.


MikeOpenHWGroup:
Thanks @silabs-robin, this LGTM at the implementation level.  I've requested a review from @silabs-hfegran because he will be in a position to review the actual definitions of the PMP configs.

silabs-hfegran (approved):
LGTM - and no impact on existing setup as far as I can see.

Issue #1414, opened 2022-09-08
ISACOV : incorrect passed arguments in get_instr_value_type() method


AyoubJalali:
Hello @JeanRochCoulon @ASintzoff ,
So I have notice some incorrect argument value in get_instr_value_type(bit[XLEN-1:0] value, int unsigned width, bit is_signed) for some C extension instruction,starting from https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L232 to https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L279 , the problem is in width argument, because the length given don't match with the length of the imm field extract using the get_field_imm() method https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L374 , but I think it match perfectly with get_data_imm(), so i propose to change the get_field_imm() with get_data_imm() in the ISACOV monitor, or we correct the width values to match with the get_field_imm().
Some of this instruction :
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L234
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L238
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L244
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L250 ...

MikeOpenHWGroup:
Hi @silabs-robin, can you have a look?  AFAICT, you are the author.

silabs-robin:
[...]

PR #1416, opened 2022-09-13
ISACOV: Fix issue #1414

AyoubJalali:
I replace the **get_field_imm** by **get_data_imm** method, and instead of passing constant argument of the **imm** field length (we can make some error in counting the field length),I used **$bits** to return automatically the length field. 
We can use the same logic for the rs1 & rs2, because in CVA6 rs1 & rs2 are 64bits length, but in CV32 are 32bits length.
I change also the signed field, instead of passing directlly the value, i used the type definition (**c_imm_is_signed**).

### DIFF
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
@@ -213,70 +213,70 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
 
   // Set enumerations for each immediate value (if applicable)
   if (mon_trn.instr.itype == B_TYPE)
-    mon_trn.instr.immb_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immb, 12, 1);
+    mon_trn.instr.immb_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immb, $bits(mon_trn.instr.immb), 1);
 
   if (mon_trn.instr.itype == S_TYPE)
-    mon_trn.instr.imms_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.imms, 12, 1);
+    mon_trn.instr.imms_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.imms, $bits(mon_trn.instr.imms), 1);
 
   if (mon_trn.instr.itype == U_TYPE)
-    mon_trn.instr.immu_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immu, 20, 0);
+    mon_trn.instr.immu_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immu, $bits(mon_trn.instr.immu), 0);
 
   if (mon_trn.instr.itype == I_TYPE)
-    mon_trn.instr.immi_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immi, 12, immi_is_signed[mon_trn.instr.name]);
+    mon_trn.instr.immi_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immi, $bits(mon_trn.instr.immi), immi_is_signed[mon_trn.instr.name]);
 
   if (mon_trn.instr.itype == J_TYPE)
-    mon_trn.instr.immj_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immj, 20, 1);
+    mon_trn.instr.immj_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immj, $bits(mon_trn.instr.immj), 1);
 
   if (mon_trn.instr.itype == CI_TYPE) begin
     case (mon_trn.instr.name)
-      C_ADDI:      mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 1);
-      C_NOP:       mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 1);
-      C_ADDI16SP:  mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 10, 1);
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Hi @silabs-robin, can you have a look?  Much of this code is more than a year old and was originally developed by Steve Richmond.  I notice that you have touched at least some of it in the past six months (no good deed goes unpunished).

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -213,70 +213,70 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
...
   if (mon_trn.instr.itype == CI_TYPE) begin
     case (mon_trn.instr.name)
-      C_ADDI:      mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 1);
-      C_NOP:       mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 1);
-      C_ADDI16SP:  mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 10, 1);
-      C_LWSP:      mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 0);
-      C_SLLI:      mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 0);
-      C_LI:        mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 6, 1);
-      C_LUI:       mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_field_imm(), 18, 0);
+      C_ADDI:      mon_trn.instr.c_imm_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.get_data_imm(), $bits(mon_trn.instr.get_data_imm()), c_imm_is_signed[mon_trn.instr.name]);
```
silabs-robin, line 232: Have you done some "printf debugging" to see what `$bits()` returns for these?
I might expect that `$bits(mon_trn.instr.get_data_imm())` would always return `32`.
If it always returns 32 then these compressed instructions would not be covered correctly.
The function returns an int, `function  int  uvma_isacov_instr_c::get_data_imm();`.

For the non-compressed instructions I think it looks nice. Good idea to use `$bits()` to avoid an extra step of potential human error.

We could consider adding fields for the various C-extension immediates, similar to how there exists fields like "immj" etc. (I don't remember why this was not done originally, and I don't know if it is a good idea, but I think it is worth considering.)

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -213,70 +213,70 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
 
   // Set enumerations for each immediate value (if applicable)
   if (mon_trn.instr.itype == B_TYPE)
-    mon_trn.instr.immb_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immb, 12, 1);
+    mon_trn.instr.immb_value_type = mon_trn.instr.get_instr_value_type(mon_trn.instr.immb, $bits(mon_trn.instr.immb), 1);
```
silabs-robin, line 216: Should we make `immb_is_signed` etc? I don't know the pros/cons of it, I just see that some of these lines are missing `*_is_signed`.

AyoubJalali: @silabs-robin in the https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv file, we have only immi_is_signed definition, we don't have for other type like B_type or J_type.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
AyoubJalali: I try size not working either, i think because of the **get_field/data_imm** method return int type, so if we want to use $size, we should change the return type of **get_field/data_imm** method.

AyoubJalali: I have an idea to create a method, called get_imm_size() that return the size of the field imm of different C extension.

AyoubJalali:
@silabs-robin I did some changes :
1- For compressed instructions, I used the **get_field_imm** and passed directly the length (easy to find in the spec),just for now (I have an idea but it will take some time).
2- For non-compressed instructions I keep the $bits function, and i used also for the rs1 & rs2 & rd, to improve the agent and make it more flexible to reused for other core with different XLEN (we can use also XLEN instead of $bits);
3- I notice there are some instructions like C_ANDI, C_SRLI also C_SRAI https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L270
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L271
https://github.com/openhwgroup/core-v-verif/blob/9962aebe36b339bf88ca7e1d8cb3143e10edbda4/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv#L272

are define in the B_TYPE, so there blown to I_TYPE so i correct that also.
Please review when you have the chance.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-robin: I was thinking about maybe adding it. But it might be that we don't need it.

silabs-robin:
Hi. Sorry for my late reply. I added just a few more comments. Everything else looks good.

AyoubJalali:
@silabs-robin take your time that's fine

silabs-robin (approved):
Looks good, afaict :)

MikeOpenHWGroup:
Thanks for the review @silabs-robin, much appreciated.  @JeanRochCoulon, this PR looks good to me also, but you should be the one to give the final approval.

JeanRochCoulon (approved):

PR #1420, opened 2022-09-20
Zc directed test

silabs-mateilga:
Directed test checking atomicity of zcmp instructions

### DIFF
--- a/cv32e40s/regress/cv32e40s_full.yaml
+++ b/cv32e40s/regress/cv32e40s_full.yaml
@@ -428,4 +428,12 @@ tests:
     builds: [ uvmt_cv32e40s_b_ext_abs, uvmt_cv32e40s_b_ext_all]
     dir: cv32e40s/sim/uvmt
     cmd: make test TEST=b_ext_test
+
+#TODO:MT Comment back in when Zc is enabled in the core
+#  zc_test:
+#    description: Zc directed test
+#    builds: [ uvmt_cv32e40s, uvmt_cv32e40s_pma_1]
+#    dir: cv32e40s/sim/uvmt
+#    cmd: make test TEST=illegal
+#    num: 5
 #
--- a/cv32e40s/tests/programs/custom/b_ext_test/test.yaml
+++ b/cv32e40s/tests/programs/custom/b_ext_test/test.yaml
@@ -3,9 +3,4 @@ uvm_test: uvmt_$(CV_CORE_LC)_firmware_test_c
 description: >
     Simple sanity check for B extension instructions
 
-# Toolchain configurations
-riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
-gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
-corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
 pulp_not_supported: 1
-llvm_march:  rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
--- a/cv32e40s/tests/programs/custom/b_ext_test/test.yaml
+++ b/cv32e40s/tests/programs/custom/b_ext_test/test.yaml
@@ -4,8 +4,8 @@ description: >
     Simple sanity check for B extension instructions
 
 # Toolchain configurations
-riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
-gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
-corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
+gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
+corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
 pulp_not_supported: 1
-llvm_march:  rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+llvm_march:  rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
--- a/cv32e40s/tests/programs/custom/zc_test/zc_test.S
+++ b/cv32e40s/tests/programs/custom/zc_test/zc_test.S
@@ -145,6 +145,39 @@ trigger_irq:
   //return to caller
   jalr	x0, 0(ra)
 
+  /*
+  ** The rest of this file consist of generated functions that work in the following manner:
+  ** interrupt_push_pop:
+  **    input value determines which rlist to test (case statement)
+  **    jump to interrupt trigger routine
+  **      - sets an interrupt to hit the atomic part of the following instruction,
+  **      - set up from C test by global variables
+  **    run push instruction to test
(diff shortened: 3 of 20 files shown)


silabs-mateilga:
@silabs-hfegran @MikeOpenHWGroup Note that the configs have been updated to include Zc, even though the core has not enabled this feature by default yet. Please take this in to consideration before merging.

One solution to this is to now let the core enable the Zc extension, and deal with any resulting incompatibilities after the merge. As an example, the illegal instruction test will need to be updated, but there might be other problems as well.

cv32e40s/regress/cv32e40s_full.yaml
```diff
@@ -428,4 +428,12 @@ tests:
     builds: [ uvmt_cv32e40s_b_ext_abs, uvmt_cv32e40s_b_ext_all]
     dir: cv32e40s/sim/uvmt
     cmd: make test TEST=b_ext_test
+
+#TODO:MT Comment back in when Zc is enabled in the core
```
silabs-hfegran, line 432: should this only be run with pma cfg 1?

cv32e40s/tests/programs/custom/zc_test/zc_test.S
```diff
@@ -0,0 +1,1165 @@
...
+  sw ra, 0(t0)
+
+
+  pp_case_4:
+    li t1, 4
+    bne a0, t1, pp_case_5
+    jal ra, trigger_irq
+  push_0:
+    cm.push {ra}, -64
+    nop
```
silabs-hfegran, line 162: what are the purpose of these nops?

cv32e40s/regress/cv32e40s_full.yaml
silabs-mateilga: I intend to revisit this when the regression is commented back in.

cv32e40s/tests/programs/custom/zc_test/zc_test.S
silabs-mateilga: To let the interrupts be taken in an easy to estimate(and debug) program flow.

cv32e40s/tests/programs/custom/zc_test/zc_test.c
```diff
@@ -0,0 +1,214 @@
...
+
+//trigger interrupt
+void vp_assert_irq(uint32_t mask, uint32_t cycle_delay) {
+    *TIMER_REG_ADDR = mask;
+    *TIMER_VAL_ADDR = 1 + cycle_delay;
+}
+
+uint32_t vp_random_num(uint32_t upper_bound, uint32_t lower_bound) {
+    uint32_t random_num = *((volatile uint32_t *) CV_VP_RANDOM_NUM_BASE);
+    uint32_t num = (random_num  % (upper_bound - lower_bound + 1)) + lower_bound;
```
silabs-hfegran, line 42: I would like an assertion of some kind to stop compilation and give a meaningful error message that the combination of upper_bound = 0xffff_ffff and lower_bound = 0x0 is not supported, or a fix to the potential overflow leading to a divide-by-zero error here, otherwise future modifications might accidentally modify the input parameters to this function such that things break.
The compiler detects this overflow condition and inserts a trap to avoid performing illegal arithmetic, and this can be somewhat complicated to debug.

cv32e40s/tests/programs/custom/zc_test/zc_test.S
```diff
@@ -0,0 +1,1165 @@
...
+
+    jal ra, trigger_irq
+  pop_0:
+    cm.pop {ra}, 64
+    nop
+    nop
+
+
+
```
silabs-hfegran, line 172: Some sort of comment what the various code sections do, or a general explanation on top that encompasses the generated functionality here would make this much more readable,

lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_agent.sv
```diff
@@ -141,14 +141,12 @@ function void uvma_rvvi_ovpsim_agent_c::configure_iss();
      $fwrite(fh, $sformatf("--override root/cpu/scontext_undefined=1\n"));
      $fwrite(fh, $sformatf("--override root/cpu/ecode_mask=0x7ff\n"));
      $fwrite(fh, $sformatf("--override root/cpu/mtvec_mask=0xffffff81\n"));
-      /* Comment back in when we activate Zc
      //TODO:MT make these part of the configuration
      $fwrite(fh, $sformatf("--override root/cpu/Zca=1\n"));
      $fwrite(fh, $sformatf("--override root/cpu/Zcb=1\n"));
-     $fwrite(fh, $sformatf("--override root/cpu/Zcmb=1\n"));
```
silabs-hfegran, line 148: should this still be commented out?

lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_agent.sv
```diff
@@ -141,14 +141,12 @@ function void uvma_rvvi_ovpsim_agent_c::configure_iss();
      $fwrite(fh, $sformatf("--override root/cpu/scontext_undefined=1\n"));
      $fwrite(fh, $sformatf("--override root/cpu/ecode_mask=0x7ff\n"));
      $fwrite(fh, $sformatf("--override root/cpu/mtvec_mask=0xffffff81\n"));
-      /* Comment back in when we activate Zc
```
silabs-hfegran, line 144: looks like it is commented back in, remove comment perhaps?

cv32e40s/tests/programs/custom/b_ext_test/test.yaml
```diff
@@ -4,8 +4,8 @@ description: >
     Simple sanity check for B extension instructions
 
 # Toolchain configurations
-riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
-gnu_march:   rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
-corev_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00
+riscv_march: rv32imc_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zcmp
```
silabs-hfegran, line 7: remove all the *march except for cv_sw_march, they are deprecated

silabs-hfegran (changes requested):

lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_agent.sv
silabs-mateilga: No, I don't think so, as the core now support Zc.

cv32e40s/tests/programs/custom/zc_test/zc_test.c
silabs-mateilga: Fix that stops execution will be added.

cv32e40s/tests/programs/custom/zc_test/zc_test.S
silabs-mateilga: To avoid the file becoming even longer than the 1600+ lines it is today, a comment section will be added rather than inline comments

lib/uvm_agents/uvma_rvvi_ovpsim/uvma_rvvi_ovpsim_agent.sv
silabs-mateilga: agreed

cv32e40s/tests/programs/custom/b_ext_test/test.yaml
silabs-mateilga: agreed

silabs-hfegran (approved):

PR #1422, opened 2022-09-21
ISACOV: Make the isacov agent more flexible for different XLEN value

AyoubJalali:

### DIFF
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
@@ -566,7 +566,7 @@ typedef enum {
   ZERO,     // For signed and unsigned values
   NON_ZERO, // For unsigned values
   POSITIVE, // For signed values
-  NEGATIVE  // For signed value
+  NEGATIVE  // For signed values
 } instr_value_t;  // TODO:ropeders should be "value_type_t"?
 
 // Package level methods to map instruction to extension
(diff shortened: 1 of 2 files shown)


lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
```diff
@@ -725,7 +725,7 @@ function instr_type_t get_instr_type(instr_name_t name);
 endfunction : get_instr_type
 
 // Package level methods to map instruction to type
-function instr_group_t get_instr_group(instr_name_t name, bit[31:0] mem_addr);
```
silabs-robin, line 728: This uses XLEN for memory addresses. Is that functionally correct? I know of the names XLEN and ILEN, but I can't remember if there is a similar thing for the width of memory addresses, or if that is implicitly inherited from e.g. XLEN.

However, I see that the function only looks at the 2 LSBs. So maybe we could change that argument to something like `bit[1:0] mem_addr_lsbs`?

I only gave it a brief look so I don't know if what I propose is a good idea. What do you think?

silabs-robin:
Overall, this diff seems like a reasonable adjustment to me.

(Note that I am only familiar with 32-bit RISC-V, so I am unable to foresee implications of how the isacov structure will affect 64-bit coverage collection.)

AyoubJalali:
> Overall, this diff seems like a reasonable adjustment to me.
> 
> (Note that I am only familiar with 32-bit RISC-V, so I am unable to foresee implications of how the isacov structure will affect 64-bit coverage collection.)

I try to make the ISACOV more flexible, because for CVA6 we want to verify the 2 configuration (CV32A6 & CV64A6), for the fist configuration it seems like the CV32E (just modifying rvfi to manage 2 retire instructions), and for the CV64A6, we need to make some changes (Add some instructions & different XLEN & modifying the macro definition for toggling a 32 bits register to 64 bits register).

MikeOpenHWGroup:
Thanks for the discussion @AyoubJalali and @silabs-robin.  Based on Robin's comment (Overall, this diff seems like a reasonable adjustment to me) and that this PR only affects the CVA6, I will approve and merge.

We will need to look at this more closely when we merge up to the master branch.

PR #1423, opened 2022-09-22
Clic assertions and necessary infrastructure updates

silabs-hfegran:

Includes: 
- clic assertions (some minor work and bugfixes remains - some are pending rtl fixes for completion)
- uvm clic interrupt agent
- infrastructure updates to support clic agent in sim

Lacks:
- ISS setup for CLIC (lack of documentation for old interface, awaiting ImperasDV update)
- Full core_v_dv support (notably interrupt handler code and constraints for edge case instructions (e.g. mscratchcs* accesses not defined for anything but CSRRW)

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
@@ -208,7 +208,7 @@ module uvmt_cv32e40s_clic_interrupt_assert
     ECALL_U_MODE        = 11'd8,
     ECALL_M_MODE        = 11'd11,
     INSTR_BUS_FAULT     = 11'd48,
-    INSTR_PARITY_FAULT  = 11'd49,
+    INSTR_PARITY_FAULT  = 11'd25,
     NMI_LOAD            = 11'd1024,
     NMI_STORE           = 11'd1025,
     NMI_LOAD_PARITY     = 11'd1026,
@@ -543,20 +543,23 @@ module uvmt_cv32e40s_clic_interrupt_assert
   logic is_invalid_instr_word;
   logic is_cause_nmi;
   logic is_cause_interrupt;
+  logic is_cause_instr_access_fault;
+  logic is_cause_instr_bus_fault;
   logic is_trap_exception;
   logic is_intr_ecall_ebreak;
   logic is_intr_exception;
 
-  assign is_load_bus_fault     = (rvfi_intr.cause == NMI_LOAD           && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
-  assign is_store_bus_fault    = (rvfi_intr.cause == NMI_STORE          && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
-  assign is_load_parity_fault  = (rvfi_intr.cause == NMI_LOAD_PARITY    && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
-  assign is_store_parity_fault = (rvfi_intr.cause == NMI_STORE_PARITY   && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
-  assign is_cause_nmi          = (rvfi_intr.cause inside { NMI_LOAD, NMI_STORE, NMI_LOAD_PARITY, NMI_STORE_PARITY })
-                                  && rvfi_intr.intr
-                                  && rvfi_intr.interrupt;
-  assign is_cause_interrupt    = !( rvfi_intr.cause inside {NMI_LOAD, NMI_STORE, NMI_LOAD_PARITY, NMI_STORE_PARITY})
-                                 && rvfi_intr.intr
-                                 && rvfi_intr.interrupt;
+  assign is_load_bus_fault           = (rvfi_intr.cause == NMI_LOAD           && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
+  assign is_store_bus_fault          = (rvfi_intr.cause == NMI_STORE          && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
+  assign is_load_parity_fault        = (rvfi_intr.cause == NMI_LOAD_PARITY    && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
+  assign is_store_parity_fault       = (rvfi_intr.cause == NMI_STORE_PARITY   && rvfi_intr.intr == 1 && rvfi_intr.interrupt == 1);
+  assign is_cause_instr_access_fault = (rvfi_intr.cause == INSTR_ACCESS_FAULT && rvfi_intr.intr == 1 && rvfi_intr.exception == 1);
+  assign is_cause_instr_bus_fault    = (rvfi_intr.cause == INSTR_BUS_FAULT    && rvfi_intr.intr == 1 && rvfi_intr.exception == 1);
+  assign is_cause_instr_parity_fault = (rvfi_intr.cause == INSTR_PARITY_FAULT && rvfi_intr.intr == 1 && rvfi_intr.exception == 1);
+
+  assign is_cause_nmi                = (rvfi_intr.cause inside { NMI_LOAD, NMI_STORE, NMI_LOAD_PARITY, NMI_STORE_PARITY }) && rvfi_intr.intr && rvfi_intr.interrupt;
+  assign is_cause_interrupt          = !(rvfi_intr.cause inside { NMI_LOAD, NMI_STORE, NMI_LOAD_PARITY, NMI_STORE_PARITY }) && rvfi_intr.intr && rvfi_intr.interrupt;
+
   assign is_interrupt_taken    = (rvfi_intr.intr == 1'b1 && rvfi_intr.interrupt == 1'b1);
 
   assign is_instr_access_fault = (rvfi_trap.exception_cause == INSTR_ACCESS_FAULT && rvfi_trap.exception == 1 && rvfi_trap.trap == 1);
@@ -928,7 +931,10 @@ module uvmt_cv32e40s_clic_interrupt_assert
         || rvfi_intr.exception
         || rvfi_trap.exception
         || is_cause_nmi
-        || !$stable(clic, @(posedge clk_i))
+        || $changed(clic, @(posedge clk_i))
+        || $changed(mstatus_fields.mie, @(posedge clk_i))
+        || $changed(mintthresh_fields.th, @(posedge clk_i))
+        || $changed(mintstatus_fields.mil, @(posedge clk_i))
                 )
       seq_irq_pend(1'b1)
       implies
@@ -1057,12 +1063,17 @@ module uvmt_cv32e40s_clic_interrupt_assert
             irq_ack
         ##1 rvfi_valid[->1]
       |->
-            rvfi_intr.intr
-         && rvfi_intr.interrupt
+            // all the following are rvfi_intr-based signals
+            is_cause_interrupt
       or
-            rvfi_trap.debug
+            is_cause_nmi
       or
-            rvfi_trap.exception;
+            is_cause_instr_access_fault
+      or
+            is_cause_instr_bus_fault
+      or
+            is_cause_instr_parity_fault
+      ;
     endproperty : p_every_ack_followed_by_rvfi_intr
 
     a_every_ack_followed_by_rvfi_intr: assert property (p_every_ack_followed_by_rvfi_intr)
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
@@ -113,6 +113,7 @@ module uvmt_cv32e40s_clic_interrupt_assert
     input logic                       rvfi_dbg_mode,
     input logic [2:0]                 rvfi_dbg,
 
+    input logic                       wu_wfe,
     input logic                       core_sleep_o
   );
 
@@ -554,7 +555,7 @@ module uvmt_cv32e40s_clic_interrupt_assert
   logic is_intr_exception;
 
   logic is_wfe_wakeup_event;
-  assign is_wfe_wakeup_event = dut_wrap.cv32e40s_wrapper_i.wu_wfe_i; // TODO: pass through if
+  assign is_wfe_wakeup_event = wu_wfe;
 
   assign is_wfi_instr = is_instr(rvfi_insn, WFI);
   assign is_wfe_instr = is_instr(rvfi_insn, WFE);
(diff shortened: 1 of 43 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+    endproperty : p_mstatus_mpp_neq_mcause_mpp
+
+    property p_mstatus_mpie_neq_mcause_mpie;
+      mstatus_fields.mpie == mcause_fields.mpie;
+    endproperty : p_mstatus_mpie_neq_mcause_mpie
+
+    p_mpp: assert property (p_mstatus_mpp_neq_mcause_mpp);
+    p_mpie: assert property (p_mstatus_mpie_neq_mcause_mpie);
+
+    weird_csr_access_cover_1 : cover property (
```
silabs-hfegran, line 1572: note to self> This should have been removed

cv32e40s/tests/uvmt/compliance-tests/uvmt_cv32e40s_firmware_test.sv
```diff
@@ -116,6 +121,7 @@ task uvmt_cv32e40s_firmware_test_c::run_phase(uvm_phase phase);
 
    if ($test$plusargs("gen_irq_noise")) begin
     fork
+      clic_noise();
       irq_noise();
```
silabs-robin, line 125: Will both `clic_noise` and `irq_noise` always run simultaneously, regardless of config enables? Or will they get stopped by something else?

cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
```diff
@@ -18,6 +18,13 @@
 `ifndef __UVMT_CV32E40S_CONSTANTS_SV__
 `define __UVMT_CV32E40S_CONSTANTS_SV__
 
+   `ifdef SMCLIC_EN
+      parameter int SMCLIC = 1;
+   `else
+      parameter int SMCLIC = 0;
```
silabs-robin, line 24: Most other params in this file have a `CORE_PARAM_` prefix. It doesn't matter to me, but I am pointing it out in case it was accidentally omitted while actually having some significance worth preserving.

cv32e40s/tests/uvmt/compliance-tests/uvmt_cv32e40s_firmware_test.sv
silabs-hfegran: both run but have their outputs gated if not enabled, such as in lib/uvm_agents/uvma_interrupt/uvma_interrupt_drv.sv  <- e.g. gated by the cfg.enable - might be better to disable here for performance reasons, but I left both in to keep the changes as small as possible.

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
+//
+// Copyright 2020 OpenHW Group
```
silabs-robin, line 2:
```
// Copyright 2022 Silicon Labs
// Copyright 2022 OpenHW Group
```

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+  import uvm_pkg::*;
+  import cv32e40s_pkg::*;
+  import uvma_rvfi_pkg::*;
+  #(
+    parameter int   SMCLIC          = 0,
+    parameter int   SMCLIC_ID_WIDTH = 5,
+    parameter int   NUM_IRQ         = 32
+  )(
+    input logic                       clk,
+    input logic                       clk_i,
```
silabs-robin, line 28: Could we have a comment (here or in the clocking block) stating which of these two clocks are the gated one?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -449,6 +455,96 @@ module uvmt_cv32e40s_tb;
...
+      .mie                 (cs_registers_i.mie_rdata),
+      .mnxti               (cs_registers_i.mnxti_rdata),
+      .mscratch            (cs_registers_i.mscratch_rdata),
+      .mscratchcsw         (cs_registers_i.mscratchcsw_rdata),
+      .mscratchcswl        (cs_registers_i.mscratchcswl_rdata),
+      .dcsr                (cs_registers_i.dcsr_rdata),
+
+      .rvfi_mepc_wdata     (rvfi_i.rvfi_csr_mepc_wdata),
+      .rvfi_mepc_wmask     (rvfi_i.rvfi_csr_mepc_wmask),
+      .rvfi_mepc_we        (rvfi_i.csr_mepc_we_i),
```
silabs-robin, line 484: This is not part of rvfi. What does it do that wmask doesn't?

silabs-hfegran: Should be removed, not used

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+    logic [11:11] stepie;
+    logic [10:10] stopcount;
+    logic [9:9]   stoptime;
+    logic [8:6]   cause;
+    logic [5:5]   v;
+    logic [4:4]   mprven;
+    logic [3:3]   nmip;
+    logic [2:2]   step;
+    logic [1:0]   prv;
+  } dcsr_t;
```
silabs-robin, line 252: These definitions, and the exception codes etc, they could be quite useful to have in some shared support file.

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+
+  generate
+  if (SMCLIC) begin : gen_clic_assertions
+
+    // ------------------------------------------------------------------------
+    // mintstatus.mil resets to 0
+    // ------------------------------------------------------------------------
+
+    property p_mintstatus_mil_reset_to_zero;
+      $rose(rst_ni) |-> mintstatus_fields.mil == '0; // TODO add bitwidth
```
silabs-robin, line 788: Could something happen to this register on the way from reset, via fetch_enable, and to the first retirement?

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+         && (mapped_instr.opcode[1:0] != 2'b11)
+      or
+            is_dret_insn
+        ##1 rvfi_valid[->1]
+        ##0 rvfi_pc_rdata == ({$past(mtvec_fields.base_31_7), $past(mtvec_fields.base_6_2), 2'b00} + NMI_OFFSET)
+         && !rvfi_dbg_mode
+      or
+        //    rvfi_valid[->1]
+        ##0 rvfi_pc_rdata == debug_halt_addr
+         && rvfi_dbg_mode
```
silabs-robin, line 860: Could one expect `dpc == fifteenth...` here?

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+        );
+    end
+
+    // ------------------------------------------------------------------------
+    // Enabled and pending interrupts should eventually be taken,
+    // assuming that the request is not retracted
+    // ------------------------------------------------------------------------
+
+    assign incr_data_obi = obi_data_req && obi_data_gnt && !obi_data_rvalid;
+    assign decr_data_obi = !(obi_data_req && obi_data_gnt) && obi_data_rvalid && (obi_data_in_flight > 0);
```
silabs-robin, line 908: Would it be safer to drop `&& (obi_data_in_flight > 0)`? That's a scenario that should not happen, so if it did then the havoc would uncover potential problems?

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+
+
+    always @(posedge clk_i) begin
+      if (!rst_ni) begin
+        obi_data_in_flight <= 0;
+      end else begin
+        obi_data_in_flight <= obi_data_in_flight + incr_data_obi;
```
silabs-robin, line 915: Wouldn't this line always get ignored?

silabs-hfegran: deprecated code, shall remove

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+        $sformatf("irq ack prerequisites not met and ack occurred"));
+
+    // ------------------------------------------------------------------------
+    // There should be no irq_ack on taken nmi
+    // ------------------------------------------------------------------------
+
+    property p_nmi_taken_no_ack;
+        irq_ack
+      |->
+        (rvfi_valid && !is_cause_nmi)[->1];
```
silabs-robin, line 1035: Wouldn't this always be true? On every cycle before `(rvfi_valid && !is_cause_nmi)` then the goto is true, and on the cycle of `(rvfi_valid && !is_cause_nmi)` (if any) then it is true because it found it (and it is not strong, afaik, so if it never finds it then no problem).

Should it be something like: `(rvfi_valid [->1])  ##0 !is_cause_nmi`?

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
silabs-hfegran: Removed and resolved

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
silabs-hfegran: Plan is to move this over to the common support logic (eventually, as time permits)

silabs-hfegran: Will fix this in a future PR

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
silabs-hfegran: I believe that check should be in the debug assertions, not here.

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2693 @@
...
+    else
+      `uvm_error(info_tag,
+        $sformatf("mscratchcswl value not as expected"));
+
+    // ------------------------------------------------------------------------
+    // Formal verification constraints
+    // Gated due to some simulators not ignoring restrict keyword
+    // ------------------------------------------------------------------------
+
+    `ifdef FORMAL
```
silabs-hfegran, line 2674: note to self - these should not be enabled by default for all runs, fix

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
silabs-hfegran: Yes, interrupts could happen if the external clic presents a proper pending and enabled interrupt prior to first retirement,  the value would then no longer represent the reset value

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
```diff
@@ -0,0 +1,2687 @@
...
+    end
+
+    always @* begin
+      obi_data_in_flight_n <= obi_data_in_flight - decr_data_obi + incr_data_obi;
+    end
+
+    localparam MAX_STALL_CYCLES = 20;
+
+    property p_obi_instr_max_load_stalls;
+      obi_instr_req && obi_instr_gnt |-> !obi_instr_rvalid [*0:MAX_STALL_CYCLES] ##1 obi_instr_rvalid;
```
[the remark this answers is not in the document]
silabs-hfegran: This property is only used for restrict/assume, and the case you mentioned above would constrain more than necessary. What needs to be changed though, is the inclusion of these assumes on a test-specific basis, and not globally.

lib/uvm_agents/uvma_clic/uvma_clic_macros.sv
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020 OpenHW Group
+// Copyright 2020 Datum Technology Corporation
+// Copyright 2020 Silicon Labs, Inc.
```
silabs-robin, line 3: On several of these files:
```
// Copyright 2022 Silicon Labs, Inc.
// Copyright 2022 OpenHW Group
```

silabs-robin (approved):
I'm giving an early approval on this. My feedback has been addressed, so when the rest is ready there will most likely be nothing to object to. (Ping me if some new part needs special attention.)

silabs-hfegran:
added do not merge to ensure that I do not add overconstraining assumes globally - pending those fixes this should be more or less ready to go

cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
silabs-hfegran: This has been removed - comment in source code explains the reasoning behind the removal

lib/uvm_agents/uvma_clic/uvma_clic_drv.sv
```diff
@@ -215,17 +215,17 @@ endtask : deassert_irq
...
+      if (cntxt.vif.mon_cb.irq_ack && cfg.enabled) begin
          // Try to get the semaphore for the irq_id,
          // If we can't get it, then this irq is managed by assert_irq_until_ack and we will ignore this ack
          // Otherwise deassert the interrupt
          int unsigned irq_id;
 
          irq_id  = cntxt.vif.mon_cb.clic_irq_id_drv;
 
-         `uvm_info("IRQDRV", $sformatf("irq_ack_clear: ack for IRQ: %0d", irq_id), UVM_DEBUG);
+         `uvm_info("IRQDRV", $sformatf("irq_ack_clear: ack for IRQ: %0d", irq_id), UVM_LOW);
```
silabs-hfegran, line 226: this should not be included

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -332,6 +332,7 @@ module uvmt_cv32e40s_interrupt_assert
     else if (wb_stage_instr_valid_i) begin
       last_instr_rdata <= wb_stage_instr_rdata_i;
     end
+  end
```
silabs-robin, line 335: How is this possible, i see two `end` being added? This commit's diff of `*interrupt_assert.sv` shows 2 additions and they are just `end`s. Did it not compile before?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -699,15 +793,16 @@ module uvmt_cv32e40s_tb;
       .*
     );
 
-  bind cv32e40s_wrapper
-    uvmt_cv32e40s_support_logic_if support_logic_if (
 
-      .ctrl_fsm_o_i(core_i.controller_i.controller_fsm_i.ctrl_fsm_o),
-      .data_bus_req_i(obi_data_if_i.req),
-      .data_bus_gnt_i(obi_data_if_i.gnt),
+    bind cv32e40s_wrapper
```
silabs-robin, line 797: Maybe the support logic interface doesn't need to be "bound" anymore? Where it's instantiation lives seems less important. (But it doesn't have to be part of this PR, can be done later, as it isn't a "CLIC" issue.)

silabs-robin (approved):
I see no problems with the newest additions either.

PR #1428, opened 2022-09-28
cva6.py : Disable compressed instruction when 'disable_compressed_instr=1'

AyoubJalali:
The cva6.py script use the YAML description to generate random tests, the problem here in the compressed instructions, actually these instructions are active by default, so to disable these instructions we should add an option in the YAML description (disable_compressed_instr=1), but in the **cva6.py** disable the compressed instructions when it detect the string **disable_compressed_instr** only, but what if we passed to disable_compressed_instr=0, that's means the C extension still active, so we need to modify the condition to disable compered inst only when **disable_compressed_instr=1**.


ASintzoff (approved):
LGTM

Thank you

PR #1429, opened 2022-09-28
Generated tests script: Delete clean_all target to keep the vcs_results for CC report in CI

AyoubJalali:


AyoubJalali:
Hi @JeanRochCoulon, I delete the clean_all target, because it clear the vcs_results, so for the continuous Integration (CI), I can't generate the CC report.

PR #1440, opened 2022-10-11
User mode coverage

silabs-robin:
This PR adds functional covers for umode/mmode checking.

The asserts in umode_assert already provides a lot of coverage, but these new covers are to fill in for the rest of what the vplan asks for.
In sim they help check stimuli comprehensiveness, in formal they help check overconstraints.

* Passes ci_check (except for known "debug_test" issue).
    * (I don't know when "debug_test" started failing in cv32e40s/dev but it has been like that for at least a month.)
* Formal works.


cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
```diff
@@ -68,9 +68,11 @@ ${DV_UVMT_PATH}/uvmt_cv32e40s_fencei_assert.sv
...
-${DV_UVMT_PATH}/uvmt_cv32e40s_pmp_model.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40s_pmprvfi_assert.sv
-${DV_UVMT_PATH}/uvmt_cv32e40s_support_logic.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40s_umode_assert.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40s_xsecure_assert.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40s_zc_assert.sv
+
+${DV_UVMT_PATH}/uvmt_cv32e40s_pmp_model.sv
+${DV_UVMT_PATH}/uvmt_cv32e40s_umode_cov.sv
+${DV_UVMT_PATH}/uvmt_cv32e40s_support_logic.sv
```
silabs-robin, line 78: Grouped these so all `*_assert.sv` are together.

cv32e40s/tb/uvmt/uvmt_cv32e40s_dut_wrap.sv
```diff
@@ -91,30 +91,11 @@ module uvmt_cv32e40s_dut_wrap
     assign debug_if.clk      = clknrst_if.clk;
     assign debug_if.reset_n  = clknrst_if.reset_n;
 
-    // --------------------------------------------
+
```
silabs-robin, line 94: Did not intend to remove this line, it happened while cleaning up the multiple drivers issue. But the block above also don't have this line so it is inconsistent either way.

cv32e40s/tb/uvmt/uvmt_cv32e40s_dut_wrap.sv
```diff
@@ -91,30 +91,11 @@ module uvmt_cv32e40s_dut_wrap
...
-    assign obi_instr_if_i.wuser     = 'b0;
-    assign obi_instr_if_i.aid       = 'b0;
-    assign obi_instr_if_i.wdata     = 'b0;
-    assign obi_instr_if_i.reqpar    = ~obi_instr_if_i.req;
-    assign obi_instr_if_i.achk      = 'b0;
-    assign obi_instr_if_i.rchk      = 'b0;
-    assign obi_instr_if_i.rready    = 1'b1;
-    assign obi_instr_if_i.rreadypar = 1'b0;
+    assign obi_instr_if_i.we = 'd 0;
+    assign obi_instr_if_i.be = 'h F;  // Assume 32-bit full bus reads
```
silabs-robin, line 97: All of these assigns where removed because they showed as "multiple drivers" in formal, since the interface is connected to the dut and is driven from there.
`we` is not used on instr-side, and apparently `be` also needs to be specified, so these two were kept.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -52,9 +52,20 @@ module uvmt_cv32e40s_tb;
    uvma_clknrst_if              clknrst_if_iss();
    uvma_debug_if                debug_if();
    uvma_interrupt_if            interrupt_if();
-   uvma_obi_memory_if           obi_instr_if_i(
-     .clk(clknrst_if.clk),
-     .reset_n(clknrst_if.reset_n)
+   uvma_obi_memory_if #(
+    /* TODO:silabs-robin  Correct widths sat via named definitions, everywhere!!! Important!
```
silabs-robin, line 56: The default parameters leads to **significant** overconstraints in formal. The commented-out code does fix the overconstraint problem, but it conflicts with the simulation environment and that one is pretty in-grown and rooted.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -402,7 +413,7 @@ module uvmt_cv32e40s_tb;
       .RUSER_WIDTH(0),
       .ID_WIDTH(0),
       .ACHK_WIDTH(0),
-      .RCHK_WIDTH(0),
+      .RCHK_WIDTH(0),  // TODO:silabs-robin  These (and similars) must be specified
```
silabs-robin, line 416: This is about the same issue as the obi-if-default-parameters-overconstraints problem above.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -479,12 +491,47 @@ module uvmt_cv32e40s_tb;
       .rvfi_csr_mstatus_wmask    (rvfi_i.rvfi_csr_mstatus_wmask),
       .rvfi_csr_mstateen0_rdata  (rvfi_i.rvfi_csr_mstateen0_rdata),
 
-      .impu_valid (core_i.if_stage_i.mpu_i.core_trans_valid_i),
-      .impu_addr  (core_i.if_stage_i.mpu_i.core_trans_i.addr),
+      .mpu_iside_valid (core_i.if_stage_i.mpu_i.core_trans_valid_i),
+      .mpu_iside_addr  (core_i.if_stage_i.mpu_i.core_trans_i.addr),
```
silabs-robin, line 495: Cleared up a mystical naming, `impu`.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -479,12 +491,47 @@ module uvmt_cv32e40s_tb;
...
+
+      .obi_iside_prot (core_i.instr_prot_o),
+      .obi_dside_prot (core_i.data_prot_o),
```
silabs-robin, line 498: Minor additions to umode_assert while developing umode_cov.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -792,18 +839,18 @@ module uvmt_cv32e40s_tb;
...
-     uvm_config_db#(virtual uvma_fencei_if              )::set(.cntxt(null), .inst_name("*.env.fencei"),     .field_name("vif"), .value(fencei_if_i));
-     uvm_config_db#(virtual uvma_rvfi_instr_if          )::set(.cntxt(null), .inst_name("*.env.rvfi_agent"), .field_name("instr_vif0"),.value(dut_wrap.cv32e40s_wrapper_i.rvfi_instr_if_0_i));
-     uvm_config_db#(virtual uvma_fencei_if              )::set(.cntxt(null), .inst_name("*.env.fencei_agent"), .field_name("fencei_vif"),     .value(fencei_if_i)  );
-     uvm_config_db#(virtual uvmt_cv32e40s_vp_status_if  )::set(.cntxt(null), .inst_name("*"),                .field_name("vp_status_vif"),    .value(vp_status_if) );
-     uvm_config_db#(virtual uvma_interrupt_if           )::set(.cntxt(null), .inst_name("*.env"),            .field_name("intr_vif"),         .value(interrupt_if) );
-     uvm_config_db#(virtual uvma_debug_if               )::set(.cntxt(null), .inst_name("*.env"),            .field_name("debug_vif"),        .value(debug_if)     );
+     uvm_config_db#(virtual uvma_isacov_if             )::set( .cntxt(null), .inst_name("*.env.isacov_agent"),           .field_name("vif"),           .value(isacov_if)      );
```
silabs-robin, line 842: Just aligned these because it was really headache-inducing to try to read them without that. Was part of an attempt to clean up the obi if widths problem.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -326,7 +329,6 @@ module  uvmt_cv32e40s_umode_assert
 
   a_trap_mpp_exception: assert property (
     p_trap_mpp_exception
-    // TODO:silabs-robin  Cov cross U/X and Exc/Int
```
silabs-robin, line 329: These "Cov" TODOs where implemented in the functional coverage.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -876,6 +876,14 @@ module  uvmt_cv32e40s_umode_assert
...
 */
 
+  a_prot_iside_legal: assert property (
+    obi_iside_prot  inside  {3'b 000, 3'b 110}
+  ) else `uvm_error(info_tag, "the prot on fetch must be legal");
+
+  a_prot_dside_legal: assert property (
+    obi_dside_prot  inside  {3'b 001, 3'b 111}
+  ) else `uvm_error(info_tag, "the prot on loadstore must be legal");
+
```
silabs-robin, line 886: Minor additions to umode_assert while developing umode_cov.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_cov.sv
```diff
@@ -0,0 +1,425 @@
...
+// limitations under the License.
+
+
+`default_nettype none
+
+
+module  uvmt_cv32e40s_umode_cov
+  import cv32e40s_rvfi_pkg::*;
+(
+  input wire  clk_i,
```
silabs-hfegran, line 26: I am really not fond of "wire" apart from the cases where it is strictly necessary. It also deviates from how most of the other code in core-v-verif is written.

silabs-robin: For the port list there is a different reason than for the helper signals below.
`default_nettype none` is added because there have been several cases in the past where implicit signal creation has caused problems and made problems harder to debug.
And without default nettype, then the portlist needs to have the nettype specified ("logic" is not a net type).

For the helper signals below, I can change it if there are more votes in favor of changing it, so that it changes like this:
```
// From this:
wire a = x;
wire b = y;
wire c = z;

// To this:
logic a;
assign a = x;
logic b;
assign b = y;
logic c;
assign c = z;
```

silabs-robin: I really like the lines in https://github.com/openhwgroup/core-v-verif/blob/29f167b4d745d1a0f98425479333cfe70be7e9e3/cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_cov.sv#L91
Because 1) each declaration reads like a title/header 2) there is no duplication of code, 3) it is impossible to accidentally use a nonblocking/clocked assignment on these signals (even if that would be unlikely to happen).

MikeOpenHWGroup:
Oh boy a coding style discussion!  :rofl:  FWIW, I too do not like `wire`.  Having said that, our [verification coding style guidelines](https://github.com/openhwgroup/core-v-verif/blob/master/docs/CodingStyleGuidelines.md) are silent on the topic and the [lowRISC RTL coding style guidelines](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#use-logic-for-synthesis) only say that `logic` is preferred over `wire`.  Having said that, they also provide an example of when wire is preferable of logic:
```
It is permissible to use wire as a short-hand to both declare a net and perform continuous assignment.
Take care not to confuse continuous assignment with initialization. For example:

wire  [7:0] sum = a + b;  // Continuous assignment
logic [7:0] sum = a + b;  // Initialization (not synthesizable)
```
Of course, we are not constrained by synthesis, but it seems that between the verification and design coding guidelines, there is a slight tilt in Robin's favour.

silabs-hfegran:
Coding style discussions are always fun 🤣 

why shorthand continuous assignment can be a bad idea, consider the following: 
> continuous assignments: 
wire [7:0] sum = a + b;
wire logic [7:0] sum = a + b;

> Initializations: 
logic [7:0]  sum = a + b;
reg [7:0] sum = a + b;
var reg [7:0]  sum= a + b;
var logic [7:0] sum = a + b;

which ones of the above should be used? In my opinion, far too much room for varying styles of code.

vs: 
> initialization
logic [7:0] sum = a + b;

> assignment:
logic [7:0] sum;
assign sum = a + b;


The above relation easily gets confusing for many that are not used to the old pre-systemverilog data-types. Furthermore, with the "new" logic data type, syntax is identical to C. Without this discussion in mind, I could easily read these statements as an initialization and not a continuous assign. ```assign statement a = b``` is unambiguous in that respect. 
One thing that puzzles me in this discussion is the aversion to having an extra line of code with an assign, when the removal of default nettype mandates adding nettypes to all signals, which is not the norm in the rest of the codebase, nor in the RTL repos. 

My two cent: SystemVerilog `logic` was introduced for a reason

MikeOpenHWGroup:
@silabs-hfegran makes a good point.   One of the complaints I have heard regarding coding style guidelines is that they inevitably create work for _someone_.  So it is in this case:
- @silabs-hfegran can you issue a pull-request to the coding style guidelines to add this rule?
- @silabs-robin can you update this pull request to use continuous assignments into a `logic` variable?

silabs-robin:
I am back. Wires are now `logic` with assign. (There are still `wire`s in the port list, in order to reap the benefits of default_nettype none.)

silabs-hfegran (approved):

PR #1450, opened 2022-10-18
Fixed warnings

silabs-hfegran:
Warning fixes:
1. Removed one unadorned reference
2. Added compile guards around imperasDV code

### DIFF
--- a/vendor_lib/imperas/design/monitor.sv
+++ b/vendor_lib/imperas/design/monitor.sv
@@ -15,10 +15,12 @@
  * for the location of the open source models.
  *
  */
- 
+`ifndef __IMPERAS_MONITOR_SV__
+`define __IMPERAS_MONITOR_SV__
+
  `include "typedefs.sv"
 
-module MONITOR 
+module MONITOR
 (
     RVVI_bus bus,
     RVVI_io  io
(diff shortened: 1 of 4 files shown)


vendor_lib/imperas/design/monitor.sv
```diff
@@ -15,10 +15,12 @@
  * for the location of the open source models.
  *
  */
- 
+`ifndef __IMPERAS_MONITOR_SV__
+`define __IMPERAS_MONITOR_SV__
```
silabs-robin, line 19: Why is it `MONITOR_SV` and not just `MONITOR`? (Doesn't matter much.)

vendor_lib/imperas/design/ram.sv
```diff
@@ -15,9 +15,11 @@
  * for the location of the open source models.
  *
  */
- 
+`ifndef __IMPERAS_RVVI_MEMORY__
+`define __IMPERAS_RVVI_MEMORY__
```
silabs-robin, line 19: It seems the file is about `module RAM` so maybe it should be `__IMPERAS_RAM__`? (Doesn't matter much.)

silabs-robin (approved):
Looks sensible to me. I only have two cosmetic questions that are not blocking.

PR #1442, opened 2022-10-11
ISACOV : Fix #1437 Issue

AyoubJalali:
Signed-off-by: Ayoub Jalali <ayoub.jalali@external.thalesgroup.com>

PR #1464, opened 2022-11-01
ISACOV : Re-fix issue #1442

AyoubJalali:


AyoubJalali:
@silabs-robin, So for this PR i try to re-fix the issue 1442 related to disable rs1 covergoup for c.mv instruction (because don't has), so I think i miss that we should also use has_rs1 signal, to disable the rs1_toogle (because it's always enable), please checkout this, and let me know if you have any suggestions.

Thanks !!

silabs-robin (approved):
Approving based on this reasoning:

I remember that issue. It made sense to filter out `cp_rs1_value` based on `has_rs1`.
Now, since `cp_rs1_toggle` has a near identical purpose as that previous cp, it makes sense to filter it too based on `has_rs1`.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for the quick turn-around on this @AyoubJalali and @silabs-robin.

Issue #1453, opened 2022-10-25
ISACOV : wrong arguments for the toggle macro


AyoubJalali:
Hi, 
I think we have another problem related to difference between get_field_imm & get_data_imm, because while looking for some missing coverage, I notice that in some instructions we're trying to toggle a 6 bits imm using a macro to toggle a 8 bits imm, here lengths did not match and let us with some uncover bins, that's because in the macros we passed the get_filed_imm (return always 6 bits length) instead of get_data_imm (add some bits to the real length of imm, which i don't know why).
So I think we should modify some macros to match the length, because if we want to passed the get_data_imm, we're going to fall also in some uncover bins, because we add 0 bits, so we gonna have uncover bins related to 1 bits.
@silabs-robin what do you think here ?

silabs-robin:
I think this is one of the points on my list https://github.com/openhwgroup/core-v-verif/issues/1076, _"Fix indices of toggle macros (the "imm"s are now aligned downto 0)"_.

Using `get_data_imm` doesn't make sense, right? Because it will in many cases just append a zero to a number, so that zero cannot get toggled.
IIRC, I switched the coverpoints to using `get_field_imm` before, and I started fixing the toggles _(but I didn't get very far because I got directed to working on other tasks)_.

> So I think we should modify some macros to match the length

I think I very much agree with what you wrote.

AyoubJalali:
> Using `get_data_imm` doesn't make sense, right?

Yes , I think less we use  get_data_imm , better for coverage

PR #1465, opened 2022-11-02
ISACOV : Fix macros (issue #1453), and fixing CG definition of compressed shifting instructions

AyoubJalali:


AyoubJalali:
So for changing some macros, as we discuss
For changing some CG definition for cg_ci_shift & cg_cb_shift, a noticed that for the cp SHAMT we create only bins for (0 to 31),  and I agree because the MSB bit should always be ZERO, but I think it's best to create bins for all possible values, and then we define bins 32 to 63 as illegal, that's gonna let ant one that reading the code that SHAMT (or imm) is a 6 bits length and not 5 bits length, and also I delete macro to toggle the imm filed because I think the SHAMT CG do the Job.

Let me know if I said anything wrong @silabs-robin.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1062,15 +1062,15 @@ covergroup cg_ci_shift(
   }
 
   cp_shamt: coverpoint instr.get_field_imm() {
-    bins SHAMT[] = {[0:31]};
+    bins SHAMT[] = {[0:63]};
+    illegal_bins ILLEGAL_SHAMT[] = {[32:63]};                                  // MSB of the immediate value should be always zero
```
silabs-robin, line 1066: > I think it's best to create bins for all possible values, and then we define bins 32 to 63 as illegal, that's gonna let ant one that reading the code that SHAMT (or imm) is a 6 bits length and not 5 bits

Wouldn't a bin for [0:31] followed by a bin for [32:63] also inform a reader that a 6-bit value is involved?

What happens when the illegal bins overlap the normal bins? Do the normal bins shrink? Do the normal bins have a big unreachable chunk? (I can't remember having tried this before so I don't know what would happen.)

Do we also need to differentiate between 32-bit and 64-bit processors here?

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1062,15 +1062,15 @@ covergroup cg_ci_shift(
   }
 
   cp_shamt: coverpoint instr.get_field_imm() {
-    bins SHAMT[] = {[0:31]};
+    bins SHAMT[] = {[0:63]};
```
silabs-robin, line 1065: Maybe we shouldn't use `[]`?
The default auto_bin_max in SystemVerilog is indeed big enough (64), but specifically hitting all of those might not be necessary?
Other "imm_value" CPs have pos/neg/zero/nonzero to be practical and feasible to hit...

Maybe we could choose some reasonable subset?
(Just playing with the idea: e.g. `SHAMT[6] = {0, 1, [2:30], 31}`, or several more specific bins.)

@MikeOpenHWGroup do you have any opinion on what is most sensible to do here?

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1062,15 +1062,15 @@ covergroup cg_ci_shift(
...
+    illegal_bins ILLEGAL_SHAMT[] = {[32:63]};                                  // MSB of the immediate value should be always zero
   }
 
   cp_rd: coverpoint instr.rd {
     ignore_bins RD_NOT_ZERO = {0};
   }
 
-  `ISACOV_CP_BITWISE(cp_rd_toggle, instr.rd_value, 1)
-  `ISACOV_CP_BITWISE_5_0(cp_imm_toggle, instr.get_field_imm(), 1)
+  `ISACOV_CP_BITWISE(cp_rd_toggle, instr.rd_value, 1)							// No need to toggle imm again because cp_shamt did the job
```
silabs-robin, line 1073: This depends on if we go for the full 64 bin SHAMT CP.
If it stays as it is then we don't need this indeed, if we limit the CP then we might need this after all.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1344,13 +1343,13 @@ covergroup cg_cb_shift(
...
-    bins SHAMT[] = {[0:31]};
+    bins SHAMT[] = {[0:63]};
+    illegal_bins ILLEGAL_SHAMT[] = {[32:63]};                                  // MSB of the immediate value should be always zero
   }
 
   cp_rs1: coverpoint instr.rs1;
 
-  `ISACOV_CP_BITWISE(cp_rs1_toggle, instr.rs1_value, 1)
-  `ISACOV_CP_BITWISE_8_1(cp_imm_toggle, instr.get_field_imm(), 1)
+  `ISACOV_CP_BITWISE(cp_rs1_toggle, instr.rs1_value, 1)							// No need to toggle imm again because cp_shamt did the job
```
silabs-robin, line 1352: These comments have a bit weird alignment. What is the pattern for why they are shifted so much to the right?

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: > What happens when the illegal bins overlap the normal bins?

It works :) 
shamt

> Do we also need to differentiate between 32-bit and 64-bit processors here?

I don't think so, anyway for now cva6 team try to verify the 32bits configuration (cv32a6), for the 64bits it's another story to tell.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: I think you are right, to hit or try to cover all possible values of imm, it's like toggling the imm, for me it's just an extra line code if we let this with imm togging.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: Yes it's depends of SHAMT CP .

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1105,8 +1105,7 @@ covergroup cg_ci_lui(
   }
 
   `ISACOV_CP_BITWISE_31_12(cp_rd_toggle, instr.rd_value, 1)
-  `ISACOV_CP_BITWISE_17_12(cp_imm_toggle, instr.get_field_imm(), 1)
-  // TODO:ropeders the toggle macros are wrong; must revise
+  `ISACOV_CP_BITWISE_5_0(cp_imm_toggle, instr.get_field_imm(), 1)
```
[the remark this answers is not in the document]
AyoubJalali: Thanks for the feedback @silabs-robin !!

MikeOpenHWGroup:
@silabs-robin, @AyoubJalali, it _appears_ as if this PR is ready to merge.  Can you confirm?

silabs-robin (approved):
I would prefer if the discussion about `[]` is resolved before we merge this.

silabs-robin:
I accidentally posted a comment that I was only deliberating. But to be clear, yes I think this can be merged now.

PR #1466, opened 2022-11-02
RVFI Assert

silabs-robin:
This PR starts an assertion set for checking RVFI details.

Reasons for including this new assertion file:
* This set of asserts has caught real bugs https://github.com/openhwgroup/cv32e40s/issues/236
* We must be able to rely on RVFI reporting the truth
* There are many aspects of the RVFI-implementation that could be checked
* It is a decently good idea to have an assertion set for the above purpose
* I sporadically come up with things to test but have nowhere to write asserts for it
    * (I have a long list of RVFI things I want to check)

Passes formal (except for 2 slow-converging ones that I didn't wait til the end for).
Passes ci_check (except for the known "debug_test" problem).


MikeOpenHWGroup:
Hi @silabs-robin, could the assertions module, `cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv`, be placed in a common area (maybe `lib/assertions`)?  Are they usable by any module that implements RVFI or are they specific-purpose for the cv32e40s?

silabs-robin:
@MikeOpenHWGroup, here is how I would describe this:

_(Maybe this should be included in a readme or header comment.)_


> The original "vanilla" RVFI is defined at https://github.com/SymbioticEDA/riscv-formal and the original RVFI assertion set can be found there.
> This assertion set does not try to replace the vanilla RVFI assertions, but is meant to check the new signals added in the extended RVFI as specified in https://docs.openhwgroup.org/projects/cv32e40s-user-manual/en/latest/rvfi.html .
> It is still meant to be RVFI-generic and _not_ core-specific.

However... there is 1 problem: The `p_rs_resetvalue` asserts are not strictly found in the RVFI specs, they are just common sense IMO.

silabs-robin:
When that is said, I have to make a disclaimer: I can't _promise_ that I haven't read the specs wearing 40s-tinted glasses...

silabs-robin:
As I claimed initially, _"I have a long list of RVFI things I want to check"_.
Here it is:

```
rvfi_trap |-> !(|rvfi_mem_rmask || |rvfi_mem_wmask)

rvfi_mem_*mask |-> have seen / will see obi activity

write_bufir_i.valid_i |-> goto rvfi_mem_wmask |-> (rvfi_mem_wmask == old(write_buffer_i.trans_i.addr))

!(rvfi_intr.interrupt && rvfi_dbg_mode && !rvfi_dbg)

num_irq_ack == num_rvfi_trap_interrupt

rvfi_intr.exception -> had_rvfi_trap

(pc_wdata && !trap) -> pc_rdata

trap -> umode

assert "rvfi_mem_*mask" reaches obi.
https://github.com/openhwgroup/cv32e40s/issues/281#issuecomment-1220605783

assert "pipe.illegal ##0 ... rvfi_valid |-> trap"

assert that exception_cause "always" matches mcause?

Prop(prot?) and rvfi, check that accesses to mpu is as rvfi advertised on fetch/loadstore

assert, if no trap/dbg/intr/nothing, then order=1 rvfi_valid must be boot_addr_i?
(And, never "!halt && !kill && !wb_valid"?)

assert, at order1, rs1/rs2 always read 0?

todo
assert
rvfi_intr -> rvfi_intr.intr
(same for rvfi_trap)

todo
assert
core no deadlock

Assert "rvfi_gpr" matches "rvfi_rs1/rs2".

Assert, rvfi_gpr "pre-data" never updated by the instr itself.

Csr wdata -> rdata
```

Note, these are very "scribbly" (just taking quick notes), and some of it is 40s-spesific so we can just ignore that,

MikeOpenHWGroup:
Hmmm.  The properties that reference OBI will make this PR CV32E4*-specific at least, although the CVE2 will eventually be updated to support both the RVVI and OBI.  Having said that, none of the properties in this pull-request include the OBI, so we could find a common location for this.

A few ideas:
- new directory `lib/assertions/rvfi/`
- integrate them into the RVFI interface (`lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv`)

Whatever we do, we absolutely do not want to disrupt the E40S simulation and formal testbenches.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
FWIW I approve this PR.  However, I would like to see this moved to a common area (see above), either in this PR or a subsequent PR.

@silabs-robin, @silabs-hfegran please comment.

silabs-hfegran:
> @silabs-robin, @silabs-hfegran please comment.

I agree with having common things in a common place, but for now I think that should not be our main focus - There are far more pressing things to get in place.

silabs-robin:
We could let this one stay in 40s, then anyone can copy it to a common place if they desire the asserts, and if migration is easy then we can change to using that common file.

PR #1467, opened 2022-11-02
Added missing stdint.h includes

silabs-hfegran:


cv32e40s/tests/programs/custom/b_ext_test/b_ext_test.c
```diff
@@ -1,5 +1,6 @@
 
 #include <stdio.h>
+#include <stdint.h>
```
silabs-robin, line 3: I remember thinking before "why did people say that you need stdint.h, it totally works without it"

PR #1473, opened 2022-11-04
Xsecure direct tests and PC hardening assertion

silabs-krdosvik:
Two comments for this PR:
1) The assertions a_xsecure_pc_hardening_branch_set_alert_major and a_xsecure_pc_hardening_jump_set_alert_major have not converged yet (but have a bound of 36 and has run for almost a day). For now I assume that it fine, so that you at least can review my work. I will update you when they have finally passed or faild.

2) The RTL for PC_hardening = 0 is not yet implemented. This means that the assertions that assumes pc_hardening =0 can be faulty and must be checked when the rtl is implemented.

The assertion passes ci check (except from the debug test). The direct test pass as well.


cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
+#ifndef __XSECURE_TEST_H__
+#define __XSECURE_TEST_H__
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+//TODO: functions are fetched from the privilege test code. If these functions end up in a common function file, we should adopt this file to use the common functions.
+
+#define CPUADDR_CPUCTRL     0xBF0
```
silabs-robin, line 10: Should it be `CSRADDR_*`? Or maybe there is some meaning that I'm not getting.

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.S
```diff
@@ -0,0 +1,31 @@
+
+.section .text
+
+
+// imported variable
+.global reg_input
```
silabs-robin, line 6: Is this unused?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+  if (actual != expect) {
+    printf(msg);
+    printf("expected = 0x%lx (%ld), got = 0x%lx (%ld)\n", expect, (int32_t)expect, actual, (int32_t)actual);
+    exit(EXIT_FAILURE);
+  }
+}
+
+/* Retuns specific bit-field from [bit_indx_low : bit_indx_high] in register x */
+unsigned int get_field(unsigned int x, int bit_indx_low, int bit_indx_high){
+    int field = ( 1 << ( (bit_indx_high - bit_indx_low) + 1) )  - 1;
```
silabs-robin, line 56: `field` is acting as a `mask`, isn't it?
Since this function gets reused a lot I think it could be made clearer to read.

`(bit_indx_high - bit_indx_low) + 1` is acting as `field_length`?

`x` (when used as the output) is actually the real `field`?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+
+typedef enum {
+  M_MODE_BEH,
+  TRAP_INCR_BEH
+} trap_behavior_t;
+
+// trap handler behavior definitions
+volatile trap_behavior_t trap_handler_beh;
+
+volatile uint32_t NUM_TRAP_EXECUTIONS;
```
silabs-robin, line 42: Conventionally, this would be lowercase, I think.

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+      __asm__ volatile("csrrs x0, mstatus, %0" :: "r"(MSTATUS_STD_VAL));
+      increment_mepc();
+      break;
+
+    // csr_privilege_loop behavior
+    case TRAP_INCR_BEH :
+      NUM_TRAP_EXECUTIONS += 1;
+      increment_mepc();
+      break;
+  }
```
silabs-robin, line 107: Would it be a good idea to add a default case to catch unintended program flow?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+  trap_handler_beh = M_MODE_BEH;
+  __asm__ volatile("ecall");
+}
+
+
+/* Checks the mepc for compressed instructions and increments appropriately */
+void increment_mepc(void){
+  volatile unsigned int insn, mepc;
+
+    // read the mepc
```
silabs-robin, line 73: The indentation level changed?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+// assembly function to set privilege-mode to user-mode
+extern volatile void set_u_mode();
+
+// standard value for the mstatus register
+#define MSTATUS_STD_VAL 0x1800
+
+typedef enum {
+  M_MODE_BEH,
+  TRAP_INCR_BEH
+} trap_behavior_t;
```
silabs-robin, line 37: Could we add something like `UNEXPECTED_ERROR_BEH` and let it be set at the end of the handler function every time the handler is invoked?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////
+/////////////////////////////////////////////////////////////////////////////////////////
+/////////////////////////////////////////////////////////////////////////////////////////
+
+
+void test_read_ctrlcpu_and_secureseeds_in_machine_mode(void){
+  uint32_t csr_read;
+
+  // setting the trap handler behaviour
+  trap_handler_beh = TRAP_INCR_BEH;
```
silabs-robin, line 121: Even if the trap handler behavior is set here, is it not reverted again on the call to set_m_mode() below?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -0,0 +1,185 @@
...
+
+  // resetting the trap handler count
+  NUM_TRAP_EXECUTIONS = 0;
```
silabs-robin, line 124: Should it be checked at the end of this test that this number stayed at 0?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+
+  ////////// PC HARDENING SEQUENTIAL INSTRUCTION: NORMAL BEHAVIOUR WHEN THERE ARE NO GLITCHES //////////
+
+  sequence seq_dummy_if_id;
+
+    //Generate a dummy instruction
+    xsecure_if.core_if_stage_instr_meta_n_dummy
+
+    //Make sure the PC in the ID stage is equal to the PC in the IF stage as long as there is a dummy instruction present
+    ##1 (xsecure_if.core_i_if_stage_i_pc_if_o == xsecure_if.core_i_id_stage_i_if_id_pipe_i_pc)[*1:$];
```
silabs-robin, line 1103: What is meant by _"as long as there is a dummy instruction present"_?
In the sequence it looks more like _"for at least one cycle on the cycle after a dummy instruction was present"_, but I might be missing something.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+    && xsecure_if.core_i_if_stage_i_pc_if_o != xsecure_if.core_i_id_stage_i_if_id_pipe_i_pc + 2
+
+    |->
+    //Abnormal behaviours are due to:
+
+    //Initalization after reset
+    xsecure_if.core_i_if_stage_i_pc_if_o == 0 && xsecure_if.core_i_id_stage_i_if_id_pipe_i_pc == 0
+
+    //Insertion of dummy instruction
+    or seq_dummy_if_id.triggered
```
silabs-robin, line 1136: With the antecedent above and this particular case of a consequent, am I right in reading it as: _"The IF/ID PC was not an increment, and it is ok because there was a dummy instr in IF and ever since then IF/ID PC were equal"_?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+  a_xsecure_pc_hardening_secuential_normal_behaviour: assert property (
+    @(posedge xsecure_if.core_clk)
+
+    //Make sure we look at valid cycles
+    core_clock_cycles
+
+    //Make sure the PC hardening setting is on
+    && xsecure_if.core_xsecure_ctrl_cpuctrl_pc_hardening
+
+    //Checkout the abnormal behaviour
```
silabs-robin, line 1125: Does _"abnormal behavior"_ mean _"non-incrementing IF/ID PC"_?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+
+    //Set the PC value to a given address
+    xsecure_if.core_i_if_stage_i_prefetch_unit_i_alignment_buffer_i_ctrl_fsm_i_pc_set
+
+    //Make sure the PC in if stage remains the same untill it is forwarded to the id stage (and then either incremented or set of a new valid PC jump)
+    ##2 $stable(xsecure_if.core_i_if_stage_i_pc_if_o)[*1:$];
+  endsequence
+
+
+  a_xsecure_pc_hardening_secuential_normal_behaviour: assert property (
```
silabs-robin, line 1116: "secuential" typo (repeated two times more below)

silabs-robin, line 1116: What does it mean that the assert is called `normal_behavior` but everything in it talks about _"abnormal behaviour"_?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+    ##1 (xsecure_if.core_i_if_stage_i_pc_if_o == xsecure_if.core_i_id_stage_i_if_id_pipe_i_pc)[*1:$];
+  endsequence
+
+  sequence seq_pc_set_stable;
+
+    //Set the PC value to a given address
+    xsecure_if.core_i_if_stage_i_prefetch_unit_i_alignment_buffer_i_ctrl_fsm_i_pc_set
+
+    //Make sure the PC in if stage remains the same untill it is forwarded to the id stage (and then either incremented or set of a new valid PC jump)
+    ##2 $stable(xsecure_if.core_i_if_stage_i_pc_if_o)[*1:$];
```
silabs-robin, line 1112: Is it `##2` because: On the first `##1` then the FSM signal has "reached" IF, and on the next one its stability can be checked?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-robin: Is it like this?:
* "normal behavior" means "no glitches"
* "abnormal behavior" means "non-incrementing if/id pc"?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+    xsecure_if.core_xsecure_ctrl_cpuctrl_pc_hardening
+
+    //Make sure the fsm is in branching state
+    && xsecure_if.core_i_if_stage_i_pc_check_i_ctrl_fsm_i_pc_mux == BRANCH_STATE
+
+    //Make sure the branch decision differs in the hardened cycles
+    ##1 xsecure_if.core_i_ex_stage_i_alu_i_cmp_result_o != $past(xsecure_if.core_i_ex_stage_i_alu_i_cmp_result_o)
+
+    //Make sure the branch decision is to always take the branch
+    && !xsecure_if.core_xsecure_ctrl_cpuctrl_dataindtiming;
```
silabs-robin, line 1283: Isn't it `dataindtiming== 1` that behaves as if the branch is always taken? Maybe I got that wrong, but it seems like the `!` and the comment is opposite to each other?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1076,264 @@ module uvmt_cv32e40s_xsecure_assert
...
+  ////////// PC HARDENING OFF NON-SEQUENTIAL INSTRUCTION: DONT SET MAJOR ALERT IF GLITCH IN PC TARGET //////////
+
+  //TODO: recheck property when rtl for PC_hardnine == 0 is imoplemented
+  property p_xsecure_hardened_pc_non_sequential_dont_set_major_alert(pc_hardening, fsm_state, calculated_signal);
+
+    seq_pc_hardening_non_sequential_instruction_with_glitch(pc_hardening, fsm_state, calculated_signal)
+
+    |=>
+    //Make sure alert major is set
+    !xsecure_if.core_alert_major_o;
```
silabs-robin, line 1309: Should the comment say "not"?

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: A good question! Didnt notice it myself. I have copied the code from xsecure_csr.c, so if it is supposed to be so in the xsecure_csr file, it is probably also supposed to be so in this file.

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Actually the function isnt used at all. So ill just remove it

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.S
silabs-krdosvik: Yes, ill remove it

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Agreed

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Good idea!

silabs-krdosvik: But we shouldnt really use it in the function, so Ill rather check the unexpected behavior variable (which you recommended to add). Thanks for noticing!

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Yess, it would!

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Shouldnt be used in this function, so ill remove it. Thanks for noticing!

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: Aaa, good catch. It is a bug. The code should have trap_handler_heh=M_MODE_BEH

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: First thought: it would be a good idea. Second thought: After the gloabal variable trap_handler_beh is set for the first time the handler will never execute a default case because it would go into the case that was previously set. So i dont think it will work that well with default case.

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: As far as i understand this is almost the same as adding a default case, which I believe will not work that well.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: It should be a not there yess.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: You are right! The comment is not correct, will fix it. Thanks for noticing.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: Yess, that is correct! I should explain it better :-)

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-robin: Yes, but calls like M_MODE_BEH are a one time thing, so shouldn't they reset the behavior to a default when they are done so that they aren't accidentally invoked in the future? That would also catch the case (mentioned in another comment) where the intended behavior setting was overwritten.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: Yess, your interpretation is correct. Will try to make it clearer in the comments

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: Yess. As it can be confusing ill change to using the word non-incremental behavior instead.

cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
silabs-krdosvik: True! Okei, ill do that then

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: Yeah, I see how this can be confusing. It will be changed!

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: That is my intention yess.

silabs-robin: Ok, that is good, then I didn't misunderstand too bad.
Would it make sense to name the sequence something like `seq_pc_equal_since_dummy` or `seq_if_id_pc_equal_since_if_dummy`?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1069,4 +1097,265 @@ module uvmt_cv32e40s_xsecure_assert
...
+    @(posedge xsecure_if.core_clk)
+
+    //Make sure we look at valid cycles
+    core_clock_cycles
+
+    //Make sure the PC hardening setting is on
+    && xsecure_if.core_xsecure_ctrl_cpuctrl_pc_hardening
+
+    //Checkout the non-incremental behaviour
+    && xsecure_if.core_i_if_stage_i_pc_if_o != xsecure_if.core_i_id_stage_i_if_id_pipe_i_pc + 4
```
silabs-hfegran, line 1149: should this +2 or +4 be conditional wrt. opcode[1:0]?

silabs-hfegran: (There are more examples in the code, I am only commenting here)

silabs-krdosvik: Good idea! Will add it

silabs-robin (approved):
Seems fine to me, and I have no new comments. Approving.

silabs-hfegran (approved):

PR #1474, opened 2022-11-04
Workaround for the iss's pmp fault

silabs-krdosvik:
Signed-off-by: Kristine Dosvik <krdosvik@silabs.com>

There is a mismatch between the core and the ISS when NUM_PMP_REGIONS is zero. The ISS reports illegal instruction, while the core access fault. Because it is is a pmp fault the correct error code is access fault. This PR makes a workaround for this scenario.

Passes the ci check.


cv32e40s/tests/cfg/pmp_g3r3.yaml
```diff
@@ -4,8 +4,170 @@ compile_flags:
...
 ovpsim: >
-    --override root/cpu/PMP_grain=3
-    --override root/cpu/PMP_registers=3
+    --override cpu/PMP_grain=3
+    --override cpu/PMP_registers=64
+    --override cpu/PMP_initialparams=T
+    --override cpu/PMP_maskparams=T
+    --override cpu/pmpaddr0=0
+    --override cpu/pmpaddr1=0
+    --override cpu/pmpaddr2=0
```
silabs-robin, line 13: If there should be 3 regions in this config, should the first 3 overrides here not exist?

silabs-robin: Or is that handled by the masks below?

cv32e40s/tests/cfg/clic_default.yaml
```diff
@@ -4,17 +4,170 @@ compile_flags:
...
-    --override cpu/CLICXCSW=T
-    --override cpu/CLICXNXTI=T
-    --override cpu/CLIC_version=master
-    --override cpu/externalCLIC=T
-    --override cpu/Zcmb=f
-    --override cpu/Zicsr=T
-    --override cpu/noinhibit_mask=0xfffffffa
-    --override cpu/trigger_num=0
-    --override cpu/extension/mcountinhibit_reset=0x5
-    --override cpu/mtvec_mask=0xffffff80
```
silabs-hfegran, line 17: These should not have been removed

cv32e40s/tests/cfg/pmp_g3r3.yaml
silabs-hfegran: pmpaddrX sets default value, mask_pmpaddrX sets writeable bitmask. The RTL expects all 64 registers to always be present, even when PMP number of regions is set to 0 - this fix works around that in the ISS, which completely disables the PMP regions if number of regions = 0

silabs-hfegran (changes requested):

cv32e40s/tests/cfg/pmp_g3r3.yaml
silabs-robin: Good

cv32e40s/tests/cfg/clic_default.yaml
silabs-krdosvik: True, reinserted!

silabs-hfegran (approved):
I am approving this for now - should be moved into the common configuration object when things stabilize

PR #1475, opened 2022-11-04
ImpDV Formal Fix

silabs-robin:
@silabs-hfegran does this make sense?

This PR excludes everything `rvvi` from formal.

Lets formal run again.
Passes ci_check **except** for `debug_test` (which has been known to fail for a while), `interrupt_test` and `illegal` also fails (but these also fail in vanilla 40s/dev now).

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
@@ -23,7 +23,9 @@
 `define __UVMT_CV32E40S_TB_SV__
 
 // Import the Imperas-DV RVVI API
-`include "rvvi/rvvi-api.svh"
+`ifndef FORMAL
+  `include "rvvi/rvvi-api.svh"
+`endif
 
 /**
  * Module encapsulating the CV32E40S DUT wrapper, and associated SV interfaces.
@@ -36,7 +38,9 @@ module uvmt_cv32e40s_tb;
    import cv32e40s_pkg::*;
    import uvmt_cv32e40s_pkg::*;
    import uvme_cv32e40s_pkg::*;
-   import rvviApi::*;
+  `ifndef FORMAL
+     import rvviApi::*;
+  `endif
 
    // ENV (testbench) parameters
    parameter int ENV_PARAM_INSTR_ADDR_WIDTH  = 32;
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Do we need compile-time macros for this (ifdef)?  Can the same thing be done with conditional instantiation using genvars?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -23,7 +23,9 @@
 `define __UVMT_CV32E40S_TB_SV__
 
 // Import the Imperas-DV RVVI API
-`include "rvvi/rvvi-api.svh"
+`ifndef FORMAL
+  `include "rvvi/rvvi-api.svh"
```
silabs-robin, line 27: >  Can the same thing be done with conditional instantiation using genvars?

Not sure how this `include` could be done differently.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -36,7 +38,9 @@ module uvmt_cv32e40s_tb;
    import cv32e40s_pkg::*;
    import uvmt_cv32e40s_pkg::*;
    import uvme_cv32e40s_pkg::*;
-   import rvviApi::*;
+  `ifndef FORMAL
+     import rvviApi::*;
```
silabs-robin, line 42: >  Can the same thing be done with conditional instantiation using genvars?

Also not sure how this `import` could be done differently.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -90,7 +94,9 @@ module uvmt_cv32e40s_tb;
                                                    .sec_lvl());     // Core status outputs
 
    // RVVI SystemVerilog Interface
-   rvviTrace #( .NHART(1), .RETIRE(1)) rvvi_if();
+   `ifndef FORMAL
+     rvviTrace #( .NHART(1), .RETIRE(1)) rvvi_if();
```
silabs-robin, line 98: >  Can the same thing be done with conditional instantiation using genvars?

This instantiation could be done with a generate block, yet.
But... then the `dv_wrap` instantiation below would probably need to be in the same generate block, to have access to the same scope?
And... I'm not sure what else would be implied by this (e.g. would several other things need to be shifted around?).

@MikeOpenHWGroup

MikeOpenHWGroup: Sigh.  I _hate_ ifdefs...

Issue #1472, opened 2022-11-04
remove references to core-v-verif/vendor_lib/imperas 


eroom1966:
@MikeOpenHWGroup @silabs-hfegran @silabs-mateilga 

I have a blocking issue due to the incorporation of libraries into the running simulator executable
This is due to the previous generation ISS integration loading an out of date library
in order to move forward I need any references removing, could somebody take the lead on this ?

MikeOpenHWGroup:
@eroom1966 can you provide more information?

eroom1966:
I think @silabs-hfegran is working on this but ...
There is a shared object in the vendor_lib/imperas which gets loaded 
This conflicts with a newer version.

The requirement here, is to remove all references to the code in vendor_lib/imperas - this can then be deleted and the problem goes away

eroom1966:
Looks like this is now resolved thanks @silabs-hfegran

PR #1478, opened 2022-11-07
Fix for #1472

silabs-hfegran:
Moved iss configuration to core_cntrl_agent and disabled deprecated rvvi_agents 
 
 This also reintroduces correct functionality to the "USE_ISS"-flag (i.e. no longer needs to be set to 0/NO for ImperasDV)
 I have not removed the shared objects in the vendor_lib-folder, but there should be no inclusion of any shared libraries from that location in 40S anymore.


silabs-hfegran:
@eroom1966 Could you have a look and see if this PR resolves your issue?

eroom1966:
@silabs-hfegran I ran with this and this all looks good to me
switching between USE_ISS=YES/NO seems to work as expected

I needed to make the modification to not compare the DCSR

Can we sync up Tuesday Morning and get your environment setup to reproduce my environment

silabs-robin (approved):
Seems like a good way out of the problem

PR #1481, opened 2022-11-08
Update core-v-verif to support latest 40s ISS

silabs-hfegran:


cv32e40s/sim/tools/xrun/covfile.tcl
```diff
@@ -51,4 +51,4 @@ set_covergroup -new_instance_reporting
 # Instances/modules to remove from coverage
 # For performance and to avoid spurious warnings, remove these modules from code coverage collection
 # ----------------------------------------------------------------------------------
-deselect_coverage -all -instance uvmt_cv32e40s_tb.iss_wrap...
\ No newline at end of file
+deselect_coverage -all -instance uvmt_cv32e40s_tb.imperas_dv...
```
silabs-robin, line 54: Could this interfere with the new soon-to-be coverage model?
If yes, then it will be very apparent when we can't find it, and then we can look back to this file?
If no, then there is no problem.

cv32e40s/env/uvme/uvma_cv32e40s_core_cntrl_agent.sv
```diff
@@ -146,24 +146,8 @@ function void uvma_cv32e40s_core_cntrl_agent_c::configure_iss();
   // -------------------------------------------------------------------------------------
 
   $fwrite(fh, $sformatf("--override %s/misa_Extensions=0x%06x\n", refpath, cfg.get_misa()));
-  $fwrite(fh, $sformatf("--override %s/tcontrol_undefined=0\n", refpath));
```
silabs-robin, line 149: Does these "_undefined" options not need a replacement?
(Same with "ecode"?)

silabs-robin (approved):
I've read through the diff and found answers to most of my questions myself. The changes seem to make sense IMO.
Approving, anticipating that my remaining questions are not blocking.

cv32e40s/env/uvme/uvma_cv32e40s_core_cntrl_agent.sv
silabs-hfegran: These are supposed to (and verified to be) set by default by the ISS internal configuration.

cv32e40s/sim/tools/xrun/covfile.tcl
silabs-hfegran: This turns is to not gather coverage on the sv module that encompasses the ISS itself - its internal coverage mechanisms should not be affected. This behavior is the same as before, but the old module is removed and replaced by the new and thus this setting would have no effect and generate a warning in its old state.

PR #1487, opened 2022-11-11
Xsecure bus hardening protocol

silabs-krdosvik:
Passes ci check, except debug test and interrupt test which never stops running.


cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1132,5 +1146,119 @@ module uvmt_cv32e40s_xsecure_assert
 
   end endgenerate
 
+  //////////////////////////////////////////////////////////////////////////
+  ///////////////////////// BUS PROTOCOL HARDENING /////////////////////////
+  //////////////////////////////////////////////////////////////////////////
+
+  //TODO: comment
```
silabs-mateilga, line 1153: Seems to be lacking a comment :)

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -80,3 +80,77 @@ module uvmt_cv32e40s_support_logic
 
 
 endmodule : uvmt_cv32e40s_support_logic
+
+
+module uvmt_cv32e40s_obi_phases_monitor
```
silabs-mateilga, line 85: The intention with this file was to have all the support logic instantiated in the "uvmt_cv32e40s_support_logic" module, and allowing assertion set modules to integrate the support_logic interface and just read the signals.

I am not opposed to this method of defining modules and instantiating them in the assertion set modules themselves, although I'm not sure I see an advantage to adding an approach.

silabs-mateilga (approved):
Looks good.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-krdosvik: I see, and agree!. We need to module instances of the obi_phase_monitor (one for instruction memory and one for data memory), so Ill just make two sumbodules to the support module.

silabs-krdosvik: Update: will not use interface into the submodule in the support_logic file because I want the signal in the submodule to be able to operate on different modports (obi_data_Slave and obi_instr_Slave), and not only one (which would be the alternative).

silabs-krdosvik:
This PR is no longer valid! See PR #1494

PR #1496, opened 2022-11-17
Update hash, fix overflow condition, updated changed param, added inv…

silabs-hfegran:
- Update hash
- Fix overflow condition
- Updated changed param
- Added invalid instruction print suppression


silabs-krdosvik:
The changes to the xsecure modules looks good!

silabs-robin (approved):
Approving based on:
Nothing looks weird. Seems to make sense.
You fixed 2 things that I too had been negatively affected by. Nice.

PR #1497, opened 2022-11-17
Formal Script Repaired

silabs-robin:
Hi @MikeOpenHWGroup. I fixed the broken formal script. So if you want to try it out it should be functional again with this PR.

```
$ cd cv32e40s/fv/
$ make
```

_(Note, there are no assumes yet, so there will be CEXes.)_
_(Note, also, the tcl command `cvfv_rerun` can be used in Jasper's GUI. Very convenient.)_


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
WFM!  Thanks @silabs-robin

PR #1500, opened 2022-11-21
FenceiVplan - Link to Coverage

silabs-robin:
This PR adds "link to coverage" for the 40s fencei vplan.

Note, the linkage was not 100% (but close).

Incidentally also added comments and small tweaks to the assert module.


silabs-robin:
Thanks for reviewing this with critical eyes, @MikeOpenHWGroup 

> My questions:
> How do we get riscv-dv to generate self-modifying code like that? Does it do it out-of-the-box, or is that a corev-dv extension?
> How do we know that the modified instruction was executed?

Would you agree that the scenario in question is handled by these two vplan items (under "StoresVisible")? Or is it not specific enough?
If not, then I could add more details to the vplan item to make it more specific, or add a new vplan item if that is better.

image

It might be more suitable to add it as a functional coverage requirement for the "Fetching" item?

image

I want to make sure that, when the time comes to complete the missing ("unlinked") vplan items (hopefully soon), then the issue you raised will not be forgotten and neglected, and that it will be covered fully.

MikeOpenHWGroup:
I think between the two `fencei` verification goals the requirement can be adequately covered.  Having said that, I see no reason to remove the `Fetching`goal.  One could argue that the fetching goal is all one needs, but it would be dangerous to remove the word "fence" from all goals in the DVplan.

To me, the question is more about the coverage goal.  What, _exactly_ can we measure to know that the fencei instruction was encountered by the pipeline and acted upon?  In the E40S this is somewhat simplified because of the external fence handshake. The first assertion should be to validate that the fencei_flush_req_o is asserted correctly.  That is:
- When a fencei instruction is retired, and
- Any possible earlier store instructions have finished. (You'll need coverage of this scenario as well.)

I would then expect a couple of things to be observed by coverage:
1. The IF prefetch buffer should be flushed, and
2. The IF should fetch the address immediately following the address of the fencei instruction.

It may be my understanding of how fencei is intended to work in E40S is flawed, so it will be necessary to sit down with the owner of the RTL to validate that understanding and determine the best probe-points for the above coverage.

silabs-robin:
I have reviewed the vplan to see if your points have been handled, and I think they have.

--------

> that the fencei_flush_req_o is asserted correctly
>    When a fencei instruction is retired

✅ The "ReqHigh" item has an assert, that fencei in WB can either 1) not retire, or 2) must have a req_o.

--------


> that the fencei_flush_req_o is asserted correctly
>    Any possible earlier store instructions have finished

✅ The several "ReqWait*" items ensures that req_o must not come before any outgoing/incoming transactions. (And "ReqHigh" above ensures req_o must come sometime.)

--------

> a couple of things to be observed by coverage:
>    The IF prefetch buffer should be flushed

❌ This is actually not checked directly.
The vplan covers it, _"pipeline stages are flushed, prefetcher is flushed, write buffer is flushed, and data_req_o is eventually supressed"_ (in "Flush").
But the linkage is actually wrong... It does not adequately handle that scenario.
It would practically be checked by the assert under "StoresVisible", but 1) I never managed to get that assert right, and 2) it would not handled cases were memory changed by itself and not from a write by the core itself (e.g. common cache).

I will have to think more about how to solve this.
For now, this PR is about linking to coverage, so I have just added a "TODO missing assert" comment with an explanation of what is missing.

--------

> a couple of things to be observed by coverage:
>    The IF should fetch the address immediately following the address of the fencei instruction

✅ The two "*after_retire" asserts (items "BranchInitiated" and "Fetching") checks this in two different ways.

silabs-mateilga (approved):

MikeOpenHWGroup:
>For now, this PR is about linking to coverage, so I have just added a "TODO missing assert" comment with an explanation of what is missing.

WFM.  Let's merge in what we have no,

Issue #1244, opened 2022-04-02
Syntax Error


mattvlsi:
Reproduce
_$ cd $CORE_V_VERIF/bin
$ pip install -r requirements.txt_

**I'm getting some errors during installation requirements before starting to verification.
The error is:**
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-kqdo817d/constraint/setup.py'"'"'; __file__='"'"'/tmp/pip-install-kqdo817d/constraint/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-kqdo817d/constraint/pip-egg-info
         cwd: /tmp/pip-install-kqdo817d/constraint/
    Complete output (6 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-kqdo817d/constraint/setup.py", line 143
        except OSError, ex:
                      ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

guojiazhuxi:
Excuse me, how is this solved?

MikeOpenHWGroup:
@ghost can you comment?

guojiazhuxi:
> @ghost can you comment?


The Constraint package in the Requirements.txt file is a Python2 package, not the Python3 package，Did you write wrong?

MikeOpenHWGroup:
@guojiazhuxi and @ghost can you have a look at pull-request #1505?  Thanks.

PR #1505, opened 2022-11-23
Fix requirements

MikeOpenHWGroup:
This PR in response to #1244.

The `constraint` package was unable to be installed using pip3.  Error message is below.  Manual installation of the latest python constraint works fine. AFAIK, none of the scriptware in core-v-verif uses the python constraint package so it has been removed (commented out) of the requirements file for now.

```
Collecting constraint==0.4.1
  Using cached constraint-0.4.1.tar.gz (48 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [7 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-oia56g2q/constraint_b991aa459eea49aaad63010590661ea1/setup.py", line 143
          except OSError, ex:
                 ^^^^^^^^^^^
      SyntaxError: multiple exception types must be parenthesized
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```


bin/README.md
```diff
@@ -11,14 +11,20 @@ determine their own directory based on the implementation langugage hooks availa
 For example from a bash-type shell:<br>
```
silabs-robin, line 11: This PR also deletes `RV32Zifencei_Extension_Instruction copy.xlsx`.
Can you confirm that this is intentional?

That ISA fencei vplan has _more_ info in it than the non-"copy" ISA fencei vplan.
But I agree that this PR makes the right move, in keeping the simpler ISA fencei vplan and deferring the fencei semantics details to core-specific vplans.

bin/requirements.txt
```diff
@@ -1,7 +1,7 @@
 bitstring==3.1.9
-constraint==0.4.1
+#constraint==0.4.1
 dv==1.0.10
-Jinja2==2.11.3
+Jinja2==3
```
silabs-robin, line 4: I have momentarily seen an error about jinja2 myself recently, but then I had not configured my shell properly.
Since our IT setup is different, I don't anticipate that this change will affect us.

Is the usage of jinja2 within the python scripts still functioning after upgrading to major version 3 like this?
If yes, then I do not oppose the change.

docs/VerifStrat/source/quick_start.rst
```diff
@@ -40,7 +40,7 @@ An easy way to get the Python plug-ins installed on your machine is::
 
    $ git clone https://github.com/openhwgroup/core-v-verif.git $CORE_V_VERIF
    $ cd $CORE_V_VERIF/bin
-   $ pip install -r requirements.txt
+   $ pip3 install -r requirements.txt
```
zchamski, line 43: LGTM.  We should be systematic in using Python3 and pruning any possibility of using Python2 flavors.

bin/requirements.txt
zchamski: I'm checking the status of Jinja changes on the CVA6 side, will report the feedback here when available.

zchamski: > I'm checking the status of Jinja changes on the CVA6 side, will report the feedback here when available.

No adverse impact in CVA6 (already using `Jinja2 v3.x`, not using `constraint`) ==> LGTM.

bin/README.md
MikeOpenHWGroup: Yes, this was intentional.  I assumed that including it was a fat-finger-issue.  Let me know if you wish to keep it.

docs/VerifStrat/source/quick_start.rst
MikeOpenHWGroup: Agreed!

MikeOpenHWGroup:
Hi @JeanRochCoulon, can you have a look as well?  I need a Committer to review/approve before merging.  Thanks.

JeanRochCoulon (approved):
LGTM !

bin/README.md
silabs-robin: I too assumed it was a fat-finger-issue. I tip in favor of purging it like this PR does.

However, even if "ISA-general" vplans are more superficial than "core-specific" vplans (which might include micro architecture and implementation-specific decicions), the Vol I spec _does_ say that _"A FENCE.I instruction ensures that
a subsequent instruction fetch on a RISC-V hart will see any previous data stores already visible to the same RISC-V hart"_.
So it _could_ be that the non-"copy" vplan is a bit too thin.
But that is a methodology/philosophy question that is a bit on the side of the Python fixing of this PR.

PR #1509, opened 2022-11-25
Updated ISS config to support Zc* extensions

silabs-halfdan:
Signed-off-by: Halfdan Bechmann <halfdan.bechmann@silabs.com>


silabs-robin (approved):
Those extensions should be on by default indeed, now that the RTL defaults to enabling Zc.

Approving.

This one should be good to merge IMO.

PR #1514, opened 2022-11-30
FenceiAssert - Check Handshake vs RVFI

silabs-robin:
This PR adds one assert to the fencei assertion set.

As mentioned by @silabs-oysteink, there is a problem that the RTL treats `fence` as `fencei`.
The original fencei asserts probed some internal signals and did not double check against RVFI!

@silabs-mateilga, @silabs-krdosvik, @MikeOpenHWGroup, I would appreciate a review and merge. (Very little code to review.)

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
@@ -48,23 +48,37 @@ module uvmt_cv32e40s_fencei_assert
   input wire         data_gnt_i,
   input wire         data_rvalid_i,
 
-  input wire  rvfi_valid,
-  input wire  rvfi_intr,
-  input wire  rvfi_dbg_mode
+  input wire         rvfi_valid,
+  input wire         rvfi_intr,
+  input wire         rvfi_dbg_mode,
+  input wire [31:0]  rvfi_insn
 );
 
   default clocking @(posedge clk_i); endclocking
   default disable iff !rst_ni;
 
-  localparam int CYCLE_COUNT = 6;
-
   string info_tag = "CV32E40S_FENCEI_ASSERT";
 
+  localparam int CYCLE_COUNT  = 6;
+  localparam int FENCEI_IDATA = 32'b 000000000000_00000_001_00000_0001111;
+  localparam int FENCEI_IMASK = 32'b 000000000000_00000_111_00000_1111111;
+  localparam int FENCE_IDATA  = 32'b 000000000000_00000_000_00000_0001111;
+  localparam int FENCE_IMASK  = 32'b 000000000000_00000_111_00000_1111111;
 
   // Helper Signals/Functions
 
   logic  is_fencei_in_wb;
   assign is_fencei_in_wb = wb_sys_fencei_insn && wb_sys_en && wb_instr_valid;
+
+  logic  is_rvfiinstr_fencei;
+  assign is_rvfiinstr_fencei = (
+    ((rvfi_insn & FENCEI_IMASK) == FENCEI_IDATA)
+  );
+
+  logic  is_rvfiinstr_fence;
+  assign is_rvfiinstr_fence = (
+    ((rvfi_insn & FENCE_IMASK) == FENCE_IDATA)
+  );
 
   int obi_outstanding;
   always @(posedge clk_i, negedge rst_ni) begin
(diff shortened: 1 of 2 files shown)


silabs-mateilga (approved):
Changes seem correct. Only note I have is that I think the assertion modules should take in the rvfi_instr_if instead of select signals, but I won't make it a requirement for approval.

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
```diff
@@ -142,6 +156,16 @@ module uvmt_cv32e40s_fencei_assert
...
 
+  // vplan:ShadowingBranch  (TODO:silabs-robin  New vplan item instead)
+
+  a_req_must_rvfi_fencei: assert property (
+    fencei_flush_req_o
+    |=>
+    (rvfi_valid [->1])   ##0
+    is_rvfiinstr_fencei
+  ) else `uvm_error(info_tag, "TODO");
+
```
silabs-oysteink, line 167: This would correctly check that a fencei_flush_req_o was caused by a fence.i instruction, but it will not check that a fence.i always causes a fencei_flush_req_o ?

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
```diff
@@ -48,24 +48,38 @@ module uvmt_cv32e40s_fencei_assert
...
   default disable iff !rst_ni;
 
-  localparam int CYCLE_COUNT = 6;
-
   string info_tag = "CV32E40S_FENCEI_ASSERT";
 
+  localparam int CYCLE_COUNT  = 6;
+  localparam int FENCEI_IDATA = 32'b 000000000000_00000_001_00000_0001111;
+  localparam int FENCEI_IMASK = 32'b 000000000000_00000_111_00000_1111111;
+  localparam int FENCE_IDATA  = 32'b 000000000000_00000_000_00000_0001111;
```
Silabs-ArjanB, line 65: Where are the FENCE_IDATA and FENCE_IMASK signals being used? I think it should be checked that the fencei handshake does not occurfor fence instructions (as we know there currently is a bug related to that)

Silabs-ArjanB (changes requested):

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
silabs-robin: > Where are the FENCE_IDATA and FENCE_IMASK signals being used?

_(I wondered if anyone would ask about this...)_

They were used by me manually in formal, because I needed a temporary constraint to disregard a known-problem CEX.

> I think it should be checked that the fencei handshake does not occurfor fence instructions (as we know there currently is a bug related to that)

I think it is implicitly checked now. If there is a handshake, a `fencei` must retire next (i.e. can't be a `fence`).
But I could add an explicit check too (in case there is some potential path that my human reasoning has overlooked).

The assert catches that bug btw.
A counter-example pops up very fast with exactly that problem.

Silabs-ArjanB: I would say then to either remove dead code or to use these signals as an extra sanity check for an issue that we know exists in the RTL today.

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
```diff
@@ -48,24 +48,38 @@ module uvmt_cv32e40s_fencei_assert
   input wire         data_gnt_i,
```
silabs-robin, line 48: (Btw, if we comment in-code, then concluded discussions can be closed as "resolved" and it is easier for "the merge person" to see if anything outstanding was overlooked.)

> Changes seem correct. Only note I have is that I think the assertion modules should take in the rvfi_instr_if instead of select signals, but I won't make it a requirement for approval.

Sure.
When https://github.com/openhwgroup/core-v-verif/pull/1515/files is merged I can (as suggested before) migrate the new helper signals from this file into that shared location and convert to using the interface.

silabs-robin: @silabs-mateilga

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
silabs-robin: > use these signals as an extra sanity check

Done.

silabs-robin:
> the assertion modules should take in the rvfi_instr_if instead of select signals

I don't think I can prioritize this now. So it would be better to merge.

silabs-robin:
If there is anything more I should fix, then let me know and I will fix it. (I don't oppose blocking this on rvfi_instr_if, but I will save the energy and not do it until someone hits "request changes" for it.)

If there is nothing more I should fix then please hit "approve" or "merge".

silabs-mateilga:
> 

As I said in my first comment, I won't block approval for this

cv32e40s/tb/uvmt/uvmt_cv32e40s_fencei_assert.sv
silabs-robin: @Silabs-ArjanB if you are happy with the changes, can you update your "requested changes"?

Silabs-ArjanB (approved):

Issue #1406, opened 2022-09-06
e40x/s - Lump assert files together


silabs-robin:
From discussion in https://github.com/openhwgroup/core-v-verif/pull/1399#discussion_r962960940

--------

We have many assertion files and they are all just dumped into the `tb/uvmt/` directory.
Let's create `tb/asserts/` or something similar and have the files more cleanly organized that way.

Additionally, remove the `core/` stuff from the uvmt ".flist" file.

Consider also, whether we might want a separate ".flist" file for the assertion files (to help share code between sim/fv).

silabs-robin:
_(The inference is "lumped > dumped")_

MikeOpenHWGroup:
Thanks for creating this issue @silabs-robin.  I think it will be of interest to the E40P team as well (@pascalgouedo, @YoannPruvost).  I like the idea of moving all assertion files from `<core>/tb/uvmt/uvmt_<core>_*_assert.sv` to `<core>/tb/assert/<core>_*_assert.sv`.

Having a separate manifest (flist) is also a good idea.  Hopefully we can create a unified one that can be used for both simulation and formal verification.

The `<core>/tb/assert` directory should also have a README listing the purpose of each file and which FV tool(s) have been used with the assertions.

PR #1519, opened 2022-12-01
Xsecure interface integrety

silabs-krdosvik:
Passes formal and simulation except debug_test


silabs-krdosvik:
@silabs-robin @silabs-mateilga When the other PR (bus hardening) is merged (which is not ready yet) I will move the support logic to the support logic module.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -222,3 +293,107 @@ module uvmt_cv32e40s_obi_phases_monitor
     end
   end
 endmodule : uvmt_cv32e40s_obi_phases_monitor
+
+
+module sl_gnt_error_in_respons
```
MikeOpenHWGroup, line 298: Having more than one module per file is not recommended (see section 1.4 of the [Coding Style Guidelines](https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/docs/CodingStyleGuidelines.md)).  Unless there is a very good reason, please break this module out into its own file (of the same name as the module).

silabs-mateilga: I actually think we might have a very good reason here Mike, let me explain.

This support logic module is meant to contain support logic for several, maybe even all, assertion sets over time. The nature of support logic is small blocks of code, and Kristine has started to organize them in modules, a method of which I approve. If we stick with the Coding Style Guideline of one module per file, we will have a quickly expanding set of files that essentially is submodules to the support_logic module. This might very quickly start cluttering up the /uvmt/ directory, with small modules that essentially all belong in one module. 

I think the best here would be to make an exception to the guideline in one of two ways:
1. Keep the submodules in the uvmt_cv32e0s_support_logic.sv file as is now
2. Gather them in a file called uvmt_cv32e0s_support_logic_submodules.sv

Third option is of course to follow the guideline with one module per file.

I think I'm leaning towards option two here, as the support_logic file might grow in to a massive and confusing beast if it's all in one file.

silabs-robin: What about a directory?

```
...tb/uvmt/support_logic/main.sv
...tb/uvmt/support_logic/abc.sv
...tb/uvmt/support_logic/xyz.sv
```

There will be one file per module.
There will be small blocks of code organized in modules.
There will be no clutter in the /uvmt/ directory.

silabs-mateilga: A directory sounds like probably the best option. Honestly, the /uvmt/ directory could do with a few subdirectories in general, at least one for /assert_modules/.

silabs-robin: For an assertion modules directory, see https://github.com/openhwgroup/core-v-verif/issues/1406

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -222,3 +292,106 @@ module uvmt_cv32e40s_obi_phases_monitor
     end
   end
 endmodule : uvmt_cv32e40s_obi_phases_monitor
+
+module uvmt_cv32e40s_sl_req_setting_in_respons
```
silabs-mateilga, line 296: Could this have a more descriptive name, closer to what it does more than what it's used for? I suggest something like  `uvmt_cv32e40s_addr_ph_attribute_fifo`, as the functionality is to track an attribute of a address phase for checking with the corresponding response phase.

Also, these modules should have a good description in comments. This is especially important in support logic, as the goal is to make reuse easy.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -222,3 +292,106 @@ module uvmt_cv32e40s_obi_phases_monitor
...
+        fifo_req_settings[req_setting_pointer] <= req_setting;
+        fifo_req_settings <= {fifo_req_settings[1:0], 1'b0};
+
+      end
+    end
+  end
+
+endmodule : uvmt_cv32e40s_sl_req_setting_in_respons
+
+module uvmt_cv32e40s_sl_req_gntpar_error_in_resp
```
silabs-mateilga, line 343: Try to reuse the module above, as this seems like just a special case of that module.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
```diff
@@ -222,3 +292,106 @@ module uvmt_cv32e40s_obi_phases_monitor
...
+  (
+    input logic rst_ni,
+    input logic clk_i,
+
+    input logic gnt,
+    input logic req,
+    input logic rvalid,
+    input logic setting_i,
+
+    output logic setting_in_respons
```
silabs-mateilga, line 308: As with the name change suggested, this name should probably be changed as well, as the actual stored data shown is what was recorded in the address phase. Maybe "attr_in_addrph"(attr for attribute), or maybe "attr_in_outstanding_txn"?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
```diff
@@ -519,7 +556,22 @@ interface uvmt_cv32e40s_support_logic_for_assert_coverage_modules_if;
    integer lml_bus_v_addr_ph_cnt;
    integer lrfodi_bus_v_addr_ph_cnt;
 
-   modport master (
+   //Signal deterimends if the respons is related to a store request
+   logic is_store_in_respons_data;
```
silabs-mateilga, line 560: I see you are using the keyword "respons" to mean "response phase" quite a bit. I think this is a bit Norwegian sounding, and may be a bit confusing for international users. I suggest "resp_ph" or "resp_phase" as alternatives.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1157,6 +1164,543 @@ module uvmt_cv32e40s_xsecure_assert
...
+    //Make sure major alert is not or has not been set
+    && !alert_major_was_set && !xsecure_if.core_alert_major_o
+
+    //Make sure we recive a respons packet
+    && xsecure_if.core_i_m_c_obi_data_if_s_rvalid_rvalid
+
+    //Make sure the packet has integrity
+    && support_if.integrity_in_respons_data
+
+    //Make sure the checksum dont matches the content
```
silabs-mateilga, line 1460: Comment should probably indicate "checksum correctness", not that it "doesn't match the content"

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1157,6 +1164,543 @@ module uvmt_cv32e40s_xsecure_assert
...
+    |=>
+    //Verify that major alert is not set
+    !xsecure_if.core_alert_major_o
+
+  ) else `uvm_error(info_tag, "Response phase checksum error for data set major alert even though integrity setting is off.\n");
+
+
+ ////////// INTERFACE INTEGRITY ADDRESS CHECKSUM FOR INSTRUCTIONS IS GENERATED CORRECTLY //////////
+
+  sequence seq_address_phase_checksum(achk, addr, prot, memtype, be, we, dbg, atop, wdata);
```
silabs-mateilga, line 1477: As this isn't a sequence of events, I'd suggest assigning this value to a signal instead.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1157,6 +1164,543 @@ module uvmt_cv32e40s_xsecure_assert
...
+    //Make sure the packet has integrity
+    && support_if.integrity_in_respons_data
+    )
+
+    |->
+    //Verify that the data packet's integrity error is set
+    xsecure_if.core_i_load_store_unit_i_bus_resp.integrity_err
+
+  ) else `uvm_error(info_tag, "The response phase checksum for data is generated wrongly.\n");
+
```
silabs-mateilga, line 1704: General comment to this file: A few of these asserts, especially related to alerts and checksum errors, are situations that are only possible through cutpoints and driving signals to values they logically can't reach. These asserts will give "cover not met" in formal for tests that don't force the error, if the rtl is clean. Each of these asserts should have a clear warning in the comments, to avoid unaware engineers wasting time debugging them.

silabs-mateilga (changes requested):

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: That is a good point! It might be a bit messy, but Ill fix this in the PR: https://github.com/openhwgroup/core-v-verif/pull/1538, as I have already done a general cleanup there. (By doing it in the other PR, I hope that the merge will be less stressful)

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: True! Will fix it! :-)

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-krdosvik: WIll change! And make descriptive comments!

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-krdosvik: Changed! :-)

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
silabs-krdosvik: respons is changed to the correct English word: response. I use response and not response phase, as the response is the specific packet, while response phase is all the clock cycles from a valid response is formed untill it is accepted.

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-krdosvik: Yess! I will try :-)

cv32e40s/tb/uvmt/uvmt_cv32e40s_support_logic.sv
silabs-krdosvik: Agreed! Ill make a folder then! :-)

silabs-krdosvik:
I made a new PR (https://github.com/openhwgroup/core-v-verif/pull/1542) because of some git rebase trouble.

PR #1521, opened 2022-12-02
fixes to debug_assert for simulation

silabs-mateilga:
PR #1515 had syntax errors that wasn't picked up in formal, but crashes sim. This PR fixes that.


silabs-oysteink (approved):

silabs-oysteink:
Looks like good changes to me, approving and merging.

PR #1532, opened 2022-12-09
InterruptAssert - Updates

silabs-robin:
This PR updates some failing/outdated asserts in `interrupt_assert`.

Additionally, it adds some new assertions to check if the updates were good.

* Passes ci_check (except known debug_test).
* Passes in formal (except non-convergent).
* (Caveat is that 2 RTL changes will be necessary and are on the way.)


cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +358,422 @@ module uvmt_cv32e40s_interrupt_assert
...
 
+  logic  wb_wfi_wfe_invalidated;
+  assign wb_wfi_wfe_invalidated = (
+    !wb_stage_instr_valid_i                  ||
+    (wb_stage_instr_mpu_status != MPU_OK)    ||
+    wb_stage_instr_err_i                     ||
+    wb_kill                                  ||  // TODO:silabs-robin  Remove? Have more isa-specific reason?
+    debug_mode_q                             ||
+    ((priv_lvl == PRIV_LVL_U) && mstatus_tw) ||
+    dcsr_step  // Arbitrary uarch decision
```
silabs-robin, line 420: @silabs-oysteink @Silabs-ArjanB , please comment if you see anything to object on here.

Single-stepping treats WFI/WFE as NOPs.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +358,422 @@ module uvmt_cv32e40s_interrupt_assert
...
+    obi_iside_outstanding_d1 <= obi_iside_outstanding;
+    obi_dside_outstanding_d1 <= obi_dside_outstanding;
+  end
+
+  logic  is_wfi_wfe_blocked;
+  assign is_wfi_wfe_blocked = (
+    |obi_iside_outstanding        ||
+    |obi_iside_outstanding_d1     ||  // Arbitrary uarch decision
+    |obi_dside_outstanding        ||
+    |obi_dside_outstanding_d1     ||  // Arbitrary uarch decision
```
silabs-robin, line 499: @silabs-oysteink @Silabs-ArjanB , please comment if you see anything to object on here.

Not only must sleep mode be blocked until outstanding transactions are done, but because of clocking it must wait one additional "unnecessary" cycle.

silabs-krdosvik:
There are some TODOs and comment out lines. Is that something to fix, or should it just stay like that?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-oysteink: That's correct, the debug spec states that WFI are treated as NOPs during step. Our custom WFE behaves exactly like a WFI with the only exception that it can also wake up to the wu_wfe_i pin. Your comment says "arbitrary uarch decision" though, but this is really mandated by the debug spec.

silabs-robin: > Your comment says "arbitrary uarch decision" though, but this is really mandated by the debug spec.

True that. Thanks.
Fixed.

image

MikeOpenHWGroup:
>  In conclusion, I personally think the PR should be merged with the todo comments

Agreed.

silabs-hfegran:
Are the RTL updates in? and should this be merged prior to that change?

silabs-robin:
Status: Waiting for debug asserts to solve "sticky non-sticky debug". Will need to mimic that logic to get correct models for some of these interrupt asserts.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
   assign is_wfi = wb_stage_instr_valid_i                     &&
                   (wb_stage_instr_rdata_i == WFI_INSTR_DATA) &&
-                  !branch_taken_ex                           &&
                   !wb_stage_instr_err_i                      &&
                   !((priv_lvl == PRIV_LVL_U) && mstatus_tw)  &&
-                  (wb_stage_instr_mpu_status == MPU_OK);
-    assign is_wfe = wb_stage_instr_valid_i                   &&
+                  (wb_stage_instr_mpu_status == MPU_OK)      &&
+                  !wb_kill                                   &&
+                  !debug_mode_q;
```
silabs-robin, line 373: These asserts are (similarly to xsecure etc) a bit close to the core, as rvfi only reports wfi/wfe events after the fact.

silabs-robin:
This PR should be ready for review now.

Formal passes everything that is not slow/non-converging.
ci_check passes as much as seems possible today.
Sadly the "interrupt_test" is failing on self-checking, so these asserts cannot be checked with it.

The original PR description is still valid.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
-                  (wb_stage_instr_mpu_status == MPU_OK);
+                  (wb_stage_instr_mpu_status == MPU_OK)      &&
+                  !wb_kill                                   &&
+                  !debug_mode_q;
   always @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
       in_wfi_wfe <= 1'b0;
     end
     else begin
       if ((is_wfi || is_wfe) && !in_wfi_wfe) //
```
silabs-krdosvik, line 386: Just wondering if there is a reason why you (I see this is not "green" so maybe someone else I guess) have included "&& !in_wfi_wfe"  in the if statement. I might be mistaken, but since we dont have a general else here, I think this will happened for the different situations: 
1) if  ((is_wfi || is_wfe) && **!in_wfi_wfe**), next cycle: in_wfi_wfe = 1 (unless !reset)
2) if ((is_wfi || is_wfe) && **in_wfi_wfe**), next cycle: in_wfi_wfe = 1 (because in_wfi_wfe is already 1) (unless !reset).

Maybe the reason is that adding it will make it easier to read/understand?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -68,8 +72,30 @@ module uvmt_cv32e40s_interrupt_assert
...
     input branch_taken_ex,
 
-    // WFI Interface
-    input core_sleep_o
+    // WFI/WFE Interface
+    input core_sleep_o,
+    input  wu_wfe_i,
+
+    // OBI
+    input mpu_iside_req,
```
silabs-krdosvik, line 80: Idk, if this is a noob question but, what du you mean by iside and dside?
If it is obvious (and I just simply dont know yet), it is ok,
if not, maybe add a comment to explain?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+    dcsr_step
+  );
+
+  logic  is_wfi_wfe_in_wb;
+  assign is_wfi_wfe_in_wb = (
+    (wb_stage_instr_rdata_i  inside  {WFI_INSTR_DATA, WFE_INSTR_DATA})  &&
+    !wb_wfi_wfe_invalidated
+  );
+
+  logic is_wfi_wfe_in_wb_d1;
```
silabs-krdosvik, line 412: I use _qX instead of _dX in the xsecure assertions to denote clock cycle delay. (The code shown in the figure is outdated but illustrate that I have used _q1 (which I believe I have used in the rest of the file too))
image

I guess it is not that important, but I think it would be best if we use a standard (and maybe add it to the coding style guide), so either I should change the _qX to _dX, or you should change this.

I think it would be best to use the name that makes the most sense of _qX and _dX.
My understanding is that
q: implies flop output, and 
d: implies delay

What do you think is best?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+  assign obi_iside_receiving = mpu_iside_rvalid;
+  assign obi_dside_receiving = obi_dside_rvalid;
+
+  always @(posedge clk_i or negedge rst_ni) begin
+    if (!rst_ni) begin
+      obi_iside_outstanding <= 1'b 0;
+      obi_dside_outstanding <= 1'b 0;
+    end else begin
+      if ( obi_iside_initiating && !obi_iside_receiving) begin
+        obi_iside_outstanding <= obi_iside_outstanding + 1;
```
silabs-mateilga, line 451: Does this support logic differ from what's already in the support_logic module, like `out_support_if.instr_bus_v_addr_ph_cnt`?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
       `uvm_error(info_tag,
                  $sformatf("Assertion of core_sleep_o did not occur within %0d clocks", WFI_TO_CORE_SLEEP_LATENCY))
+
   c_wfi_assert_core_sleep_o: cover property(p_wfi_assert_core_sleep_o);
 
+  c_wfi_assert_core_sleep_long: cover property(
+    (
+      p_wfi_assert_core_sleep_o
+    ) and (
+      //((is_wfi_wfe_in_wb == 0) && (!is_wfi_wfe_blocked == 0) && (core_sleep_o == 0)) [*1:$]  ##1
```
silabs-mateilga, line 542: As mentioned, commented out code should be marked with a TODO or removed

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+  );
+
+
+  // Check expectations for sleep mode
+
+  a_wfi_assert_sleepmode_expected: assert property (
+    model_sleepmode == core_sleep_o
+  ) else `uvm_error(info_tag, "core_sleep_o must matchexpectations");
+
+  a_wfi_assert_sleepmode_nodbg: assert property (
```
silabs-mateilga, line 556: Note: The relation between wfi/wfe and debug mode is also covered in debug_asserts. I do not mind the redundancy, just wanted to raise awarenes

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+
+    implies
+
+    !wb_valid  until (
+      $rose(core_sleep_o)  // core_sleep_o must come...
+      ||
+      (is_wfi_wfe_in_wb && is_wfi_wfe_wake)
+    )
+    ;
+    // TODO:INFO:silabs-robin  Idea: packed struct (like pmp reasons), cover several onehots
```
silabs-mateilga, line 752: Sounds like an interesting idea :)

silabs-mateilga (commented):
Generally good assertions, left a few small comments.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -68,8 +72,30 @@ module uvmt_cv32e40s_interrupt_assert
     // Determine whether to cancel instruction if branch taken
     input branch_taken_ex,
 
-    // WFI Interface
-    input core_sleep_o
+    // WFI/WFE Interface
+    input core_sleep_o,
+    input  wu_wfe_i,
```
silabs-krdosvik, line 77: Question: 
what does wu stand for?

General comment: 
I dont really know why I believe this, but I assume wu_wfe_i is the signal inputted to the core that make the core go out of the wfe state, correct?
 
In my opinion making the purpose of the signal more clear (add a comment or change the name), will increase the code quality

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+
+  logic  model_sleepmode;
+  always_latch begin
+    if (!rst_ni) begin
+      model_sleepmode <= 1'b 0;
+    end
+
+    if (
+      is_wfi_wfe_in_wb  &&
+      is_wfi_wfe_in_wb_d2)  // Arbitrary uarch decision (2 cycles)
```
silabs-krdosvik, line 503: Hehe, "The noob" (me) dont understand what the u in uarch means, hehe... So unless it is very clear for everybody else, maybe explain it?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
     |->
     core_sleep_o;
   endproperty
+
   a_wfi_assert_core_sleep_o: assert property(p_wfi_assert_core_sleep_o)
     else
       `uvm_error(info_tag,
                  $sformatf("Assertion of core_sleep_o did not occur within %0d clocks", WFI_TO_CORE_SLEEP_LATENCY))
+
   c_wfi_assert_core_sleep_o: cover property(p_wfi_assert_core_sleep_o);
```
silabs-krdosvik, line 537: I see there are both an assert and a cover for the property p_wfi_assert_core_sleep_o.
Is that nessessary? Wouldn't the assertion automatically generate a "related" cover? Or is this important for simulation purposes?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+  // Blocked wfi/wfe stay in wb (unless excused)
+
+  a_wfi_assert_sleepmode_wait: assert property (
+    is_wfi_wfe_in_wb    &&
+    is_wfi_wfe_blocked  &&
+    !is_wfi_wfe_wake
+    |=>
+    is_wfi_wfe_in_wb
+    or
+    $past(support_if.recorded_dbg_req)
```
silabs-krdosvik, line 601: Is recorded_dbg_req the same as next DBG_CAUSE_HALTREQ?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-krdosvik: The same goes for p_wfi_assert_core_sleep_o_cond

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+
+    if (is_wfi_wfe_wake) begin
+      model_sleepmode <= 1'b 0;
+    end
+  end
+
+
   // WFI assertion will assert core_sleep_o (in WFI_TO_CORE_SLEEP_LATENCY cycles after wb, given ideal conditions)
+
   property p_wfi_assert_core_sleep_o;
```
silabs-krdosvik, line 520: Is there an important difference between this property and p_wfi_assert_core_sleep_o_cond? Or is checking p_wfi_assert_core_sleep_o_cond sufficient?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-mateilga: This is a new support signal that records activity on the pin debug_req_i, which causes debug entry with cause haltreq. Due to the debug_req signal becoming unstickied, there is no longer a guarantee that the core will enter debug if the debug_req signal isn't kept high until the core is in debug, but we the core might recognize a short high pulse and enter debug. This signal is a way of determining if there has been a high debug_req within a reasonable amount of time that the core _could_ take. Essentially, if recorded_dbg_req is high, the core is allowed to enter debug with cause haltreq, but it does not have to.

silabs-krdosvik: Oo, complicated, but I think I understand! Thank you! :-D

silabs-krdosvik:
As we all know, there are three types of feedback, positive, constructive and mean.
I hope you feel that the comment above falls under the category constructive.

I also want to express my positive comments too:
It is interesting to see how you have structured your file. It was easy to follow, making it easy to review, which is nice!

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
```diff
@@ -334,68 +360,403 @@ module uvmt_cv32e40s_interrupt_assert
...
+  logic  is_wfi_wfe_blocked;
+  assign is_wfi_wfe_blocked = (
+    |obi_iside_outstanding        ||
+    |obi_iside_outstanding_d1     ||  // Arbitrary uarch decision
+    |obi_dside_outstanding        ||
+    |obi_dside_outstanding_d1     ||  // Arbitrary uarch decision
+    (writebufstate != WBUF_EMPTY)
+  );
+
+  logic  is_wfi_wfe_wake;
```
silabs-krdosvik, line 486: One more comment: when reviewing your file I thought that wake meant that core was awake, but it is supposed denote being woken up from sleep-state, right?

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-krdosvik: Wait, u as in unratified, right?

silabs-oysteink: Micro, as in microarchitecture :)

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: Hm, let's see. It was introduced [5 months ago](https://github.com/openhwgroup/core-v-verif/commit/792d2605182b9e348715e356c2ad9e24ea80f92a#diff-6577a4a85f81658856f01986b48c167d3a7e0fef34bcd1dd97c906d37946cd7eR356).

--------

Considering the code to be just:
```
      if (is_wfi || is_wfe)
        in_wfi_wfe <= 1'b1;
      else if (|pending_enabled_irq || debug_req_i || pending_nmi || debug_mode_q)
        in_wfi_wfe <= 1'b0;
```

What about this scenario:
`is_wfi && pending_nmi`

I think the pending nmi should switch `in_wfi_wfe` off.
So the `&& !in_wfi_wfe` seems to disable the first if-clause, so that the second clause gets a chance to run.

--------

_(It could arguably have been written differently, but it seems to follow a principle of changing the existing source code as little as possible to avoid unforeseen consequences.)_

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: > I think it would be best if we use a standard

I agree. It makes reading code fast and distraction free, so we can develop without hiccups.

image

I shall change it to q.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: > maybe add a comment to explain?

Will do. (Commiting in after I've fixed more of the feedback.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: > I assume wu_wfe_i is the signal inputted to the core that make the core go out of the wfe state, correct?

Correct. It is the wake-up signal.

> making the purpose of the signal more clear (add a comment or change the name), will increase the code quality

That is a good point. In this case, I think the context of the WFE part of the user manual makes it sufficiently clear, so I don't know if anything needs to change.

silabs-krdosvik: Okei, I see now that wu can be an abbreviation for wake up, so it is indeed indicated in the signal name.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-krdosvik: Aaa, yess, good point!

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: It might have been added to get more observability in formal. But I agree that it seems superfluous and I don't see that it is needed on more than a temporary basis so I am removing them.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: I think the latter might cover a superset of the former, but it is difficult to reason about with absolute certainty so I think it is safer to let the old variant of them keep on living.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: I am changing it from `is_wfi_wfe_wake` to `should_wfi_wfe_awaken`.

cv32e40s/tb/uvmt/uvmt_cv32e40s_interrupt_assert.sv
silabs-robin: I can have a look

silabs-robin: Seems it is close enough to `abiim` and `lrfodi` to be substituted. Fixed.

silabs-mateilga (approved):
Looks good, I have nothing more to note here. Approving.

PR #1537, opened 2022-12-12
Remove remains of hpmcounters

silabs-halfdan:
Removed hpmcounter testing from performance counter tests (currently named hpmcounter_*_test, will rename in separate PR).
Removed NUM_MHPMCOUNTER parameter which does not exist in RTL for 40s.


silabs-mateilga (approved):
Did not spot any problems here, approved.

PR #1538, opened 2022-12-12
Xsecure hint

silabs-krdosvik:
Hint instructions are almost the same as dummy instructions, so I reused the dummy instruction assertions.
I have also corrected (hopefully all) English mistakes, and reduced the number of todos.


silabs-halfdan (changes requested):
Some minor updates needed, will approve when fixed.

silabs-halfdan (approved):

PR #1540, opened 2022-12-14
Temp fix unstickied debug req

silabs-mateilga:


silabs-robin (approved):
With the todo-comments and the plans to fix this properly on the S, this PR looks like a reasonable temporary hotfix.

PR #1541, opened 2022-12-14
Update mcycle and minstret assertions after design implemented dcsr.s…

silabs-oivind:
…topcount

### DIFF
--- a/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
+++ b/cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
@@ -467,9 +467,10 @@ module uvmt_cv32e40x_debug_assert
     // Check that mcycle works as expected when not sleeping
     // Counter can be written an arbitrary value, check that
     // it changed only when not being written to
+    // Counter should not increment when in debug mode with dcsr.stopcount set
 
     property p_mcycle_count;
-        !cov_assert_if.mcountinhibit_q[0] && !cov_assert_if.core_sleep_o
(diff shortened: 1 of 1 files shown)


cv32e40x/tb/uvmt/uvmt_cv32e40x_debug_assert.sv
```diff
@@ -467,9 +467,10 @@ module uvmt_cv32e40x_debug_assert
     // Check that mcycle works as expected when not sleeping
     // Counter can be written an arbitrary value, check that
     // it changed only when not being written to
+    // Counter should not increment when in debug mode with dcsr.stopcount set
```
silabs-robin, line 470: This updates expectations for `mcycle` and `minstret` counting, to heed `dcsr.stopcount`?

I have two questions:
1. How does this capture the _"or on ebreak instructions that cause entry into Debug Mode"_ part?
2. Should this also apply to the mhpmcounters, but we simply lack asserts for it?

Otherwise the change looks good.

silabs-oivind: 1. We simplified the implementation of stopcount a bit, and it will only stop counters when we have entered debug mode. Most counter events were already gated when we have an ebreak, the only exceptions are cycle and wb_invalid.
2. This applies to all counters, but I only noticed assertions for minstret and mcycle.

silabs-robin: Ok. Thanks for the explanation.

silabs-robin (approved):
Approving based on: The diff+rationale+explanation seems consistent and sensible.

PR #1543, opened 2022-12-14
PmpAssert - Update exception codes

silabs-robin:
As the title says.

Passes ci_check, as far as that is possible right now (known debug test, known fencei assert).
Formal still runnable.

Incidentally also adds a few more comments, etc.


silabs-halfdan (approved):
Looks good

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
If its good enough for @silabs-halfdan ...

PR #1544, opened 2022-12-14
temporary fix to keep simulation functional after unstickying debug__req

silabs-mateilga:
Quick fix to keep things moving for everyone else while I'm working on updating debug_assert file.


silabs-halfdan (approved):
Thanks, looks good. Approving.

PR #1547, opened 2022-12-16
CSR hardening update, because rtl is updated

silabs-krdosvik:
RTL is updated, so must the assertions.
Change: take into account the reset value of a CSR register when shadowing the CSR.

PR #1548, opened 2022-12-19
Revert "CSR hardening update, because rtl is updated"

Silabs-ArjanB:
Reverts openhwgroup/core-v-verif#1547


MikeOpenHWGroup:
Thanks for this @Silabs-ArjanB.  In future, please allow a member of the verif'n team to review before merging.  It is important that _someone_ on the verif team knows in abvance that these are coming.   Thanks.

silabs-krdosvik:
I made a mistake in my PR related to a CSR hardening assertion fix (which I didnt notice before after the merge), so Arjan simply undid the Merge. I think the communication here were internal. (A new PR for CSR hardening fix is out but needs to be reviewed before merging.)

Issue #1384, opened 2022-08-25
Remove duplicate manifests


silabs-robin:
### Task Outcome
Remove ".flist" files (or .flist contents) which duplicates other .flists within a core.

### Background information
When adding formal scripts to 40x, a new manifest file was created to overcome sim/fv differences, but it is better to have a single source of truth instead of having to update two manifests whenever something changes.

### Location Information
core-v-verif/cv32e40x/fv/fv.flist

### Completion Criteria
There should be only one manifest file, or just one "base" manifest file which others can include, and there should be no duplication in content among manifest files.

### Additional context
(I don't have permissions to assign this task to myself, but I am responsible for adding the duplicates and can take responsibility for cleaning it up.)

silabs-robin:
Ref this PR https://github.com/openhwgroup/core-v-verif/pull/1377

silabs-robin:
One (bad) alternative:

Yet another dummy package.

```
`define  uvm_info(ID, MSG, VERBOSITY)  ;
`define  uvm_warning(ID, MSG)  ;
`define  uvm_error(ID, MSG)  ;
`define  uvm_fatal(ID, MSG)  ;

`define  uvm_component_utils_begin(TYPE)  ;
`define  uvm_component_utils_end  ;

`define  uvm_field_int(ARG,FLAG)  ;

ad nauseum
```

```
package uvm_pkg;

  class  uvm_component;
  endclass

  class uvm_phase;
  endclass
  
  ad nauseum

endpackage

`include "uvm_macros.svh"
```

This will slow down compilation, is a pain to write, and might be yet another thing to maintain.

silabs-robin:
Status:
This has been partially fixed, but there still remains https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/fv/fv.flist

PR #1554, opened 2022-12-27
Fv - Update Flist

silabs-robin:
Just the periodic sync of the formal flist.

We still have this https://github.com/openhwgroup/core-v-verif/issues/1384 ticket to properly fix it.
But I don't know how to optimally solve it...

### DIFF
--- a/cv32e40s/fv/fv.flist
+++ b/cv32e40s/fv/fv.flist
@@ -24,37 +24,18 @@
 
 -f  ${CV_CORE_PKG}/cv32e40s_manifest.flist
 
-+incdir+${CV_UVMT_PATH}
-+incdir+${CV_UVME_PATH}
-+incdir+${CV_UVMA_PATH}/uvma_rvfi
-+incdir+${CV_UVMA_PATH}/uvma_fencei
-+incdir+${CV_UVMA_PATH}/uvma_clic
-+incdir+${CV_UVMA_PATH}/uvma_obi_memory/src
++incdir+${DV_UVMT_PATH}
++incdir+${DV_UVME_PATH}
++incdir+${DV_UVMA_PATH}/uvma_rvfi
++incdir+${DV_UVMA_PATH}/uvma_fencei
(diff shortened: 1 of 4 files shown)


silabs-robin:
@MikeOpenHWGroup, you previously said: _"I would like to have one-and-only-one manifest (flist)"_

* Current flists:
    1. The main 40s flist: https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
    2. Two Imperas flists: (right next to the main 40s flist)
    3. And, one special formal flist: (ref this PR) (duplicates a subset of the main 40s flist)

Problem:
Formal cannot handle all the UVM stuff from the main 40s flist.

Proposal:
Take whatever is common between "the main 40s flist" and the "special formal flist" and separate it out into one (or more) common flist(s).
_(One alternative is to dedicating an effort to creating a dummy uvm library, hope that that will solve it, and suffer unnecessarily long compilation times as non-used files are compiled.)_

MikeOpenHWGroup:
> I would like to have one-and-only-one manifest (flist)

True, I said that.  But I wasn't entirely clear.  The "one manifest" refered to the RTL.  Currently we have a manifest for RTL simulation and another for synthesis.  Bad.  I would dearly love to fix that and suggest that using FuseSoC could be a solution.

As for the verification environment, we can have heirarchical netlist.

silabs-robin:
@MikeOpenHWGroup, I moved the part from `uvmt_cv32e40s.flist` called `// CV32E40S test bench files` into a new file called `uvmt_cv32e40s_tb_files.flist`.

Now that new flist is used by formal too, so that it doesn't need to maintain its own copy of those lines, since it kept getting out of sync all the time and breaking the formal setup.

silabs-hfegran (approved):
Looks good to me

MikeOpenHWGroup (approved):

PR #1555, opened 2022-12-28
Questa Formal

silabs-robin:
This PR adds preliminary support for Questa Formal.

ci_check passes (except infamous "debug_test").
jg formal (internal) still runnable.

~~"Depends" on this open PR: https://github.com/openhwgroup/core-v-verif/pull/1554 (got merged)~~

@silabs-hfegran and @silabs-krdosvik I have changed your assertion modules. Please review and help me evaluate if they are still intact.


cv32e40s/fv/q.tcl
```diff
@@ -0,0 +1,40 @@
+# Copyright 2022 Silicon Labs, Inc.
```
MikeOpenHWGroup, line 1: `q.tcl` is a very short filename.  I assume this is a control file for Questa Formal.  If that is the case, then perhaps a more descriptive filename such as `questa_formal.tcl` would better convey this file's intent.

cv32e40s/fv/Makefile
```diff
@@ -35,12 +35,21 @@ DPI_DASM_SPIKE_REPO ?= TODO_dontwanttodefinethis
...
 
-fv:  $(CV_CORE_PKG)
+
+# Make Targets
+
+## JasperGold
+jg:  $(CV_CORE_PKG)
 	$(CV_SIM_PREFIX)  jaspergold  jg.tcl
 	# TODO:robin-silabs  Move to "mk/fv/jg.mk"?
+
```
MikeOpenHWGroup, line 51: `q` is a really short target name.  Maybe `questa_formal`?

silabs-robin: What about... using the same name as the executable?

That would be `jaspergold.tcl` and `qverify.tcl`.

The make targets could have "longhand" names correspondingly:
```
jaspergold: jg
jg: <etcetc>
  <etcetc>

qverify: q
q: <etcetc>
  <etcetc>
```

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -1405,39 +1405,41 @@ module uvmt_cv32e40s_tb;
...
-     end
-     else begin
-       if (vp_status_if.tests_passed) begin
-         tp <= 1'b1;
-         uvm_config_db#(bit)::set(.cntxt(null), .inst_name("*"), .field_name("tp"), .value(1'b1));
-       end
-       if (vp_status_if.tests_failed) begin
-         tf <= 1'b1;
-         uvm_config_db#(bit)::set(.cntxt(null), .inst_name("*"), .field_name("tf"), .value(1'b1));
+   `ifndef FORMAL  // uvm db not used in formal
```
MikeOpenHWGroup, line 1415: Can we do this with parameters and generate-blocks?  I worry about the over-use of macros in our environments and FORMAL appears all over the place and may or may not be used to mean the same thing in all instances.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -1355,6 +1357,7 @@ property p_parity_signal_is_not_invers_of_signal_set_major_alert(signal, parity_
   ////////// PC HARDENING BEHAVIOUR WHEN THERE ARE NO GLITCHES //////////
 
   sequence seq_dummy_if_id;
+    @(posedge xsecure_if.core_clk)
```
MikeOpenHWGroup, line 1360: This set of assertions uses multiple clocks.  Is that necessary?  If so, please add a comment.  If not, let's stick with one clock.

MikeOpenHWGroup (changes requested):
Thanks for this @silabs-robin!  I've made a few comments, but overall I _really_ like this.

Dumb question: is "Questa Formal" a new name for "OneSpin"?

cv32e40s/fv/Makefile
MikeOpenHWGroup: Works for me.

cv32e40s/fv/q.tcl
silabs-robin: I made new names for both the tcl files and the make targets.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
silabs-robin: There are some difficult cases, like this one:
```
   `ifndef FORMAL
   import rvviApiPkg::*;
   `endif
```

Looking at the file, I only think I know how to solve a few of them.

I think the "ignore this simulation thing in formal" ifndef might be one of the least bad ones.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-robin: I am not deeply familiar with the xsecure assertions, but I think "core_clk" is used to avoid false negatives when the core is sleeping, or something similar.

@silabs-krdosvik could you give a quick explanation?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
MikeOpenHWGroup: Sigh.  OK...

MikeOpenHWGroup:
Thanks for the update @silabs-robin.  Before approving/merging, I'd like to close the "clocking" discussion. @silabs-krdosvik can you comment?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-hfegran: I think these clock signals beg for a different naming scheme - I presume one is gated and one is free running?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
silabs-hfegran: While I don't like them, I think the alternative may be worse
Formal guards around initial/final-blocks can technically be removed, but that will (presumably) reduce readability, as one may be led to believe that these blocks are actually considered by the formal tools, while in reality it is ignored (and triggers yet another warning)

As for the example above.... these guarded imports would require dummy packages and manifests and I believe that would be less readable.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
silabs-krdosvik: You are right Henrik! I can comment it and change name.

silabs-krdosvik:
> Thanks for the update @silabs-robin. Before approving/merging, I'd like to close the "clocking" discussion. @silabs-krdosvik can you comment?

PR #1558 change the xsecure file to only use one clock signal

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Good-to-go from my end.

PR #1558, opened 2023-01-02
Xsecure clocking

silabs-krdosvik:
Only one clock signal. Add a gated clock signal. Made to fix the clock comment in #1555.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
@@ -139,14 +139,14 @@ module uvmt_cv32e40s_xsecure_assert
   assign rvfi_insn_opcode = rvfi_if.rvfi_insn[6:0];
   assign rvfi_insn_cmpr_opcode = rvfi_if.rvfi_insn[1:0];
 
-  //Signal used to check if the core clock is active or not.
-  logic core_clock_cycles;
+  //The alert signals used the gated clock, so we must therefore make sure the gated clock is enabled.
+  logic core_i_sleep_unit_i_core_clock_gate_i_clk_en_n;
 
   always @(posedge clk_i) begin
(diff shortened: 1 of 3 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert.sv
```diff
@@ -139,14 +139,14 @@ module uvmt_cv32e40s_xsecure_assert
   assign rvfi_insn_opcode = rvfi_if.rvfi_insn[6:0];
   assign rvfi_insn_cmpr_opcode = rvfi_if.rvfi_insn[1:0];
 
-  //Signal used to check if the core clock is active or not.
-  logic core_clock_cycles;
+  //The alert signals used the gated clock, so we must therefore make sure the gated clock is enabled.
+  logic core_i_sleep_unit_i_core_clock_gate_i_clk_en_n;
```
silabs-robin, line 143: This is `_n` for being clocked? Is there any precedent for it? I think RTL uses `_q`. And I've sometimes used `_d1` in cases where I've also had a `_d2` for double-flopped.

silabs-krdosvik: Good that you saw this. When I think about it, I might have used the _n wrongly... I mean to say the signal is a delayed signal, not the upcoming register value.

silabs-robin (approved):
Approving based on:
Looks good. Uses the one global clock, renamed the enable signal, the rest is utilizing the enable signal instead of at'ing on the old clock. The comments look nice too.

silabs-robin:
@silabs-hfegran or @MikeOpenHWGroup could you have a look?

MikeOpenHWGroup:
LGTM.  I will defer to @silabs-hfegran to merge as I am not sufficiently knowledgeable with this code.

PR #1566, opened 2023-01-04
basic VCS flow setup

abdullahyildiz:


MikeOpenHWGroup (approved):

pascalgouedo:
Ok for me, we are not using "core" test-bench.
It could be possible that we will use it later on for power consumption evaluation on back-annotated gate netlist but we mainly use questa.

PR #1571, opened 2023-01-09
Update CSR Gen Script

silabs-robin:
This PR updates the CSR access test generation script to be in sync with the latest update of the CSR Yaml definition.
Relates to https://github.com/openhwgroup/cv32e40s/pull/378


bin/gen_csr_access_test.py
```diff
@@ -17,34 +17,55 @@
...
 parser.add_argument("--pmp_num_regions",        help="Set number of Pmp regions",         type=str, default="0")
 parser.add_argument("--mhpmcounter_num",        help="Set number of mhpmcounters",        type=str, default="3")
 parser.add_argument("--num_triggers",           help="Set number of trigger registers",   type=str, default="1")
+parser.add_argument("--marchid",                help="Set number for marchid",            type=str, default="0")
+
 parser.add_argument("--clic_enable",            help="Enable clic support",               action="store_true", default=False)
-parser.add_argument("--zc_enable",              help="Enable Zc support",                 action="store_true", default=False)
 parser.add_argument("--clint_enable",           help="Enable Basic Interrupts",           action="store_true", default=False)
 parser.add_argument("--debug_enable",           help="Enable Debug Registers",            action="store_true", default=False)
+parser.add_argument("--umode_enable",           help="Enable Certain U-mode fields",      action="store_true", default=False)
```
silabs-robin, line 40: Mostly just new conditional names to support more features from the user manuals.

bin/gen_csr_access_test.py
```diff
@@ -215,8 +292,30 @@ def gen_riscv_dv_gen_csr_file(iteration = 1):
         file_handle.close()
         template_handle.close()
     except Exception as e:
+        print("error: exception in 'gen_riscv_dv_gen_csr_file'")
         print(e)
 
+def preprocess_yaml_m4(enabled_features, input_script_path, output_script_handle):
```
silabs-robin, line 298: Invocation of `m4`.

bin/gen_csr_access_test.py
```diff
@@ -148,6 +220,11 @@ def preprocess_yaml():
     if not args.dry_run:
         output_script_handle = open(output_script_path, "w")
 
+    # "m4" - Optionally used for all preprocessing
+    if (args.m4):
+        preprocess_yaml_m4(enabled_features, input_script_path, output_script_handle)
```
silabs-robin, line 225: Invocation of `m4`.

MikeOpenHWGroup: If we are going to use `m4` then we need to update [requirements.txt](https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/bin/requirements.txt#L1), and probably also the top-level README.

silabs-robin: All references I can find to `requirements.txt` indicate that this is used for Python packages: e.g. `pip install -r requirements.txt`. Is that correct?

But `m4` is not a Python package, it is a standard posix command line program along with `grep`, `ls`, `cd`, `chmod`, `cp`, `echo`, `make`, etc.
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/idx/utilities.html
_([main link](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/))_

MikeOpenHWGroup (approved):

silabs-hfegran:
Looks good to me, thanks!

silabs-hfegran (approved):

PR #1580, opened 2023-01-13
Cv32e40s/dev

eroom1966:
First cut of the 0.8.0 specification enhancement.
This will need an update to the reference simulator


eroom1966:
@silabs-hfegran

silabs-hfegran:
Thanks @eroom1966, Looks good to me, I will need to get the updated reference simulator onto our systems before merging this

PR #1583, opened 2023-01-16
Clic directed test bugfix

silabs-hfegran:


silabs-hfegran:
Self merging, fix removes code branch that is never taken in case of failure

PR #1584, opened 2023-01-16
Removed obsolete imperas_dv path

silabs-hfegran:


silabs-hfegran:
Self merging: Non-existing path from earlier imperas dv implementation, creates warning

PR #1586, opened 2023-01-17
Added Zc directed test to regressions and added zcb, zcmp to enabled …

silabs-hfegran:
Added Zc directed test to full regressions and added zcb, zcmp to enabled compiler extensions


cv32e40s/tests/cfg/default.yaml
```diff
@@ -176,4 +176,4 @@ plusargs: >
     +enable_zbb_extension=1
     +enable_zbc_extension=1
     +enable_zbs_extension=1
-cv_sw_march: rv32im_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zca_zifencei
+cv_sw_march: rv32im_zba1p00_zbb1p00_zbc1p00_zbs1p00_zicsr_zca_zcb_zcmp_zifencei
```
silabs-robin, line 179: Is zcmt missing because we await toolchain support?

cv32e40s/regress/cv32e40s_full.yaml
```diff
@@ -406,11 +406,10 @@ tests:
...
-#    dir: cv32e40s/sim/uvmt
-#    cmd: make test TEST=illegal
-#    num: 5
-#
+  zc_test:
+    description: Zc directed test
+    builds: [ uvmt_cv32e40s, uvmt_cv32e40s_pma_1]
+    dir: cv32e40s/sim/uvmt
+    cmd: make test TEST=zc_test
+    num: 5
```
silabs-robin, line 414: What does "5" mean here? Is it run 5 times for "uvmt_cv32e40s" and 5 times for "uvmt_cv32e40s_pma_1"? But it is a directed test, so what is the purpose of running it 5 times for each config?

silabs-robin (commented):
Looks good but I have 2 questions.

cv32e40s/regress/cv32e40s_full.yaml
MikeOpenHWGroup: My two cents:
1. Yes, "num" is the number of runs of this test.  (Maybe "runs" would have been better choice.)
2. Even if the test-program is static, the simulation _could_ be made random by enabling variable cycle delays and/or errors on the OBI, or even by enabling interrupt/debug events.

silabs-robin: > But it is a directed test, so what is the purpose of running it 5 times for each config?

> Even if the test-program is static, the simulation could be made random by enabling variable cycle delays and/or errors on the OBI, or even by enabling interrupt/debug events.

@silabs-mateilga could you comment on this?

--------

Quick link for convenience:
https://github.com/openhwgroup/core-v-verif/tree/cv32e40s/dev/cv32e40s/tests/programs/custom/zc_test

silabs-robin: Is it because of the `vp_random_num` and `glb_irq_delay` etc?

silabs-mateilga: The test has randomness built into it, so there is definitely a value in running it more than once. The timing of the interrupt hitting each multioperation instruction is randomized.

silabs-robin (approved):
I only have one outstanding question, but it is minor so I am hitting approve on the PR.

cv32e40s/tests/cfg/default.yaml
silabs-hfegran: yes

PR #1589, opened 2023-01-17
add fsdb waveform file generate code

guojiazhuxi:
Option `+UVM_VERBOSITY=$(VCS_UVM_VERBOSITY)`  is not recognized by vcs.

`$(VCS_UVM_ARGS)` option repeat

Signed-off-by: guojiazhuxi <378982405@qq.com>

### DIFF
--- a/cv32e40p/tb/uvmt/uvmt_cv32e40p_tb.sv
+++ b/cv32e40p/tb/uvmt/uvmt_cv32e40p_tb.sv
@@ -507,6 +507,12 @@ module uvmt_cv32e40p_tb;
     */
    initial begin : test_bench_entry_point
 
+     //Generate FSDB wave file
+     $fsdbDumpfile("wave.fsdb");  
+     $fsdbDumpvars("+all");       
+     $fsdbDumpSVA();              
+     $fsdbDumpMDA();              
+
      `ifdef PULP
        `ifdef NO_PULP
          `uvm_fatal("CV32E40P TB", "PULP and NO_PULP macros are mutually exclusive.")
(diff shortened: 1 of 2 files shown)


cv32e40p/tb/uvmt/uvmt_cv32e40p_tb.sv
```diff
@@ -507,6 +507,12 @@ module uvmt_cv32e40p_tb;
     */
    initial begin : test_bench_entry_point
 
+     //Generate FSDB wave file
```
MikeOpenHWGroup, line 510: This will dump a full FSDB database for every simulation.  We cannot do this in core-v-verif because not all simulators have the ability to dump FSDB.  Also, our usage model is that waves are not dumped by default.  Rather, the user must specify that waves should be dumped on the command-line.  Please see this [README](https://github.com/openhwgroup/core-v-verif/blob/master/mk/README.md#post-process-waveform-debug) for more information.

Rather than add system calls to the testbench, the way to handle this is with "do" files.  Have a look at the [Xrun Makefile](https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/mk/uvmt/xrun.mk#L91) for an example.

MikeOpenHWGroup (changes requested):

cv32e40p/tb/uvmt/uvmt_cv32e40p_tb.sv
guojiazhuxi: Thank you, I understand!

PR #1593, opened 2023-01-20
fencei test: fixed ISS error when ISS is explicitly disabled

silabs-hfegran:

### DIFF
--- a/cv32e40s/env/uvme/vseq/uvme_cv32e40s_vp_fencei_tamper_seq.sv
+++ b/cv32e40s/env/uvme/vseq/uvme_cv32e40s_vp_fencei_tamper_seq.sv
@@ -147,7 +147,9 @@ import "DPI-C" context function void rvviRefMemoryWrite(
 
 function void uvme_cv32e40s_vp_fencei_tamper_seq_c::write_iss_mem();
 
-  rvviRefMemoryWrite(0, addr, data, 4);
+  `ifdef USE_ISS
+    rvviRefMemoryWrite(0, addr, data, 4);
+  `endif
 
 endfunction : write_iss_mem
 


silabs-robin (approved):
Approving based on:
I am familiar with the issue that this stems from, and this seems like an appropriate fix.

But if there exists any way to avoid the ifdef I would prefer that.

eroom1966:
@silabs-hfegran @silabs-robin 
this appears to break the functionality
I am now seeing the fencei tests fail.
It appears that the code is no longer called to overwrite the memory in the REF memory model

the ifdef is not working as expected
to reproduce the failure run the following 
removing the ifdef will allow the test to pass

make comp_corev-dv comp gen_corev-dv TEST=corev_rand_fencei CV_CORE=cv32e40s CFG=default COREV=1 SIMULATOR=xrun USE_ISS=YES COV= SEED=308272962
make test TEST=corev_rand_fencei CV_CORE=cv32e40s CFG=default COREV=1 SIMULATOR=xrun USE_ISS=YES COV= SEED=849127855

I modified the code to do the following, is this acceptable ?

```
function void uvme_cv32e40s_vp_fencei_tamper_seq_c::write_iss_mem();

  if ($test$plusargs("USE_ISS")) begin
    rvviRefMemoryWrite(0, addr, data, 4);
  end

endfunction : write_iss_mem
```

silabs-hfegran:
> @silabs-hfegran @silabs-robin
> this appears to break the functionality
> I am now seeing the fencei tests fail.
> It appears that the code is no longer called to overwrite the memory in the REF memory model

Thanks for bringing this up @eroom1966, I will run a couple of tests later and incorporate the fix if it all checks out

PR #1600, opened 2023-01-26
support for 0.8.0 Zc

eroom1966:


silabs-hfegran (approved):

cv32e40s/env/uvme/uvma_cv32e40s_core_cntrl_agent.sv
silabs-robin, line 1: This PR was to cv32e40x/dev, but it touches the 40s directory. This makes merging harder so we should try to avoid it.

PR #1602, opened 2023-01-27
install-riscv-dv: Use RISCV_PREFIX

niwis:
Use the existing `RISCV_PREFIX` env variable to set up the gnu tool commands. Useful for installations that use the original RISC-V toolchain where the tools have different prefixes (e.g. `riscv64-unknown-elf-`)

### DIFF
--- a/cva6/regress/install-riscv-dv.sh
+++ b/cva6/regress/install-riscv-dv.sh
@@ -10,10 +10,10 @@
 # riscv-dv env variables
 export RISCV_TOOLCHAIN=$RISCV
 if ! [ -n "$RISCV_GCC" ]; then
-  export RISCV_GCC="$RISCV_TOOLCHAIN/bin/riscv-none-elf-gcc"
+  export RISCV_GCC="${RISCV_PREFIX}gcc"
 fi
 if ! [ -n "$RISCV_OBJCOPY" ]; then
-  export RISCV_OBJCOPY="$RISCV_TOOLCHAIN/bin/riscv-none-elf-objcopy"
+  export RISCV_OBJCOPY="${RISCV_PREFIX}objcopy"
 fi
 export SPIKE_PATH=$SPIKE_ROOT/bin
 export RTL_PATH=$ROOT_PROJECT/core-v-cores/cva6


JeanRochCoulon:
Hello @niwis,   I see you are submitted PR to core-v-verif/master. Be careful, PR must be submitted to core-v-verif/cva6/dev !

zchamski:
Hi @niwis , I agree that this implicit assumption about the "target triple" prefix used in the name of RISC-V tools needs addressing.  We faced this in the CVA6 bootcamp as well.

I propose to keep the path and the target triple separated, giving something like
```
export RISCV_GCC=$RISCV_TOOLCHAIN/bin/${RISCV_TARGET_TRIPLE}gcc
...
export RISCV_OBJCOPY=$RISCV_TOOLCHAIN/bin/${RISCV_TARGET_TRIPLE}objcopy
```

BTW, including the dash in the value of `RISCV_TARGET_TRIPLE` makes it possible to use a bare, unprefixed `gcc` / `objcopy` located in a target-specific directory.

niwis:
Hi @zchamski, I went with `RISCV_PREFIX` because this variable is already defined in `setup-ci`: https://github.com/openhwgroup/core-v-verif/blob/393e8f1be9b947b9417da2375dc254a89b986f51/.gitlab-ci/setup-ci-example/core-v-verif-cva6.yml#L3
It also includes the dash and can be left empty for an unprefixed version and doesn't constrain the path structure. But if there is a practical reason to re-compile the path to the riscv binaries, I can change this PR accordingly. In this case, where should I define `RISCV_TARGET_TRIPLE`?

niwis:
It makes sense to streamline the variables. A quick search for `RISCV` and `RISCV_PREFIX` suggests that this is a larger endeavour, though, and I'm not sure about all consequences. Feel free to close this PR and continue tracking this in #1607.

MikeOpenHWGroup:
> Feel free to close this PR and continue tracking this in #1607.

I agree with @niwis.  @JeanRochCoulon, what do you think?

JeanRochCoulon:
After a discussion with @zchamski, I think we all agree the fact that this point need to be adressed, but is more complex than expected. As the issue has been logged in GitHub issue by Mike, I close the PR.

PR #1606, opened 2023-02-01
Passive/Active agent for AXI4

AEzzejjari:
The axi agent can act as a passive or active agent.
For the active mode the memory system replaced by the memory class.
A extension not supported.


JeanRochCoulon:
Can you submit your PR to core-v-verif/cv32a6_v0.1.0 ?

PR #1608, opened 2023-02-01
cva6/sim/Makefile: Use VCS_HOME env variable

niwis:
The VCS path is currently hard coded. Use the `VCS_HOME` environment variable instead. `VCS_HOME` is required by VCS itself, so it should already be set in the environment calling this script.


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @niwis.

PR #1609, opened 2023-02-01
Update spike log parsing

yanicasa:
Spike output have been updated in latest releases
register index is now left align with space padding

image


yanicasa:
@niwis

niwis:
Is this also the case for the spike version specified here? https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/cva6/regress/install-spike.sh#L11

MikeOpenHWGroup:
OK... not now, but _soon_, we gotta replace this with step-and-compare.

PR #1611, opened 2023-02-02
ISACOV: FIX #1582

### LINKED ISSUE #1582: ISACOV Agent : wrong decode format
Hello,
So I notice something suspicion about assign some C_format in the [file](https://github.com/openhwgroup/core-v-verif/blob/master/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv) : 

1. line [336](https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L336) to [338](https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L338), a duplicate IF, so should be removed.
2. https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L395, I think it's a CL_TYPE for dasm_rvc_lw_imm(instr) not CS_TYPE because there is no function for CS_TYPE in dasm library.
3. https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L436 to https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L442, it's a rd function so why returning rs2 in [439](https://github.com/openhwgroup/core-v-verif/blob/6113f4f179645eb05c4178b6ec25be316a9a1aa8/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv#L439), beside that i thing there is some miss assign types.

I think it should be like :  `return (itype inside {CA_TYPE, CL_TYPE, CIW_TYPE}) ? rd[2:0] : rd;`

AyoubJalali:

### DIFF
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
@@ -165,8 +165,8 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
   if (mon_trn.instr.ext == C_EXT) begin
     mon_trn.instr.rs1     = dasm_rvc_rs1(instr);
     mon_trn.instr.rs2     = dasm_rvc_rs2(instr);
-    mon_trn.instr.rd      = dasm_rd(instr);
-    mon_trn.instr.c_rdrs1 = dasm_rd(instr);
+    mon_trn.instr.rd      = dasm_rvc_rd(instr);
+    mon_trn.instr.c_rdrs1 = dasm_rvc_rd(instr);
     mon_trn.instr.c_rs1s  = dasm_rvc_rs1s(instr);
     mon_trn.instr.c_rs2s  = dasm_rvc_rs2s(instr);
   end
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_instr.sv
@@ -333,11 +333,6 @@ function void uvma_isacov_instr_c::set_valid_flags();
     return;
   end
 
-  if (itype == CSRI_TYPE) begin
-    rd_valid = 1;
-    return;
-  end
-
 endfunction : set_valid_flags
 
 
@@ -392,7 +387,7 @@ function  int  uvma_isacov_instr_c::get_field_imm();
   if (this.itype == CIW_TYPE) begin
     return (dasm_rvc_addi4spn_imm(instr) >> 2);  // Shift 2 because [9:2] to [7:0]
   end
-  if (this.itype == CS_TYPE) begin
+  if (this.itype == CL_TYPE) begin
     return (dasm_rvc_lw_imm(instr) >> 2);  // Shift 2 because [6:2] to [4:0]
   end
   if (this.itype == CB_TYPE) begin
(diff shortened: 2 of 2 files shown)


MikeOpenHWGroup:
LGTM, thanks @AyoubJalali.  Before I merge, @silabs-robin can you have a look?

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -165,8 +165,8 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
   if (mon_trn.instr.ext == C_EXT) begin
     mon_trn.instr.rs1     = dasm_rvc_rs1(instr);
     mon_trn.instr.rs2     = dasm_rvc_rs2(instr);
-    mon_trn.instr.rd      = dasm_rd(instr);
-    mon_trn.instr.c_rdrs1 = dasm_rd(instr);
+    mon_trn.instr.rd      = dasm_rvc_rd(instr);
```
silabs-robin, line 168: Can this be trusted to always return the appropriate "rd" for any instruction format?

image

image

"7, 5" seems to mean, starting with bit 7 and growing leftwards to a size of 5 bits.

--------

It seems this change of dasm function might not by itself be enough.

silabs-robin (changes requested):
I replied in the original issue https://github.com/openhwgroup/core-v-verif/issues/1582 . (Except for 1 inline comment I left directly here on this PR.)

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
AyoubJalali: But It related to compressed instructions ?

silabs-robin: From its name, it looks like it should be related to compressed instructions. But I am not sure how it is used internally by Spike, and from looking at its definition it does not seem to take care of all of RVC's details.

_(I'm starting to suspect that a "core-v-decoder" might not be such a bad idea.)_

ASintzoff:
How can we proceed for the moment? Do we close this PR and wait for future core-v-decoder or do we merge it even we know this is not a perfect solution?

silabs-robin:
> Do we close this PR and wait for future core-v-decoder

I would not be opposed to making this PR work without waiting for core-v-decoder.

--------

> or do we merge it even we know this is not a perfect solution?

It could be merged in if there are no outstanding issues. IIRC, the dasm usage was not just "not a perfect solution" but it looked like it fetched an incorrect "rd".

silabs-robin:
To state it straight forward: I think the code is wrong and you have just merged bugs.

Until I have seen evidence that my previous concerns have either been corrected or were non-issues, I believe that the code is wrong.

AyoubJalali:
Hello @silabs-robin which part of the code you talking about, is it related to dasm library .

PR #1614, opened 2023-02-03
PMP Assert - Add Missing Check

silabs-robin:
This PR adds a PMP assert that was missing.

The updated asserts pass in formal.
ci_check passes (except infamous old known failure).
~~Only tested in `hello-world`.  (Hence the "draft" status.)~~

--------

_(Question to self: Why was this not here originally? What spec text would have directly led to its inclusion? What thinking-procedure would have led to it?)_


silabs-robin:
Hi @MikeOpenHWGroup could you have a look at this PR?

MikeOpenHWGroup (approved):

silabs-mateilga (approved):

MikeOpenHWGroup:
Merging based on @silabs-mateilga's approval.  Thanks!

PR #1616, opened 2023-02-03
Core Params - Add missing & Add external config

silabs-robin:
This PR adds missing parameters to `_constants.sv` and `dut_wrap`, as well as facilitating for defining these parameters in an external file not checked into this repo.

ci_check passes as much as it does for cv32e40s/dev.
Formal is still runnable.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
@@ -18,77 +18,160 @@
 `ifndef __UVMT_CV32E40S_CONSTANTS_SV__
 `define __UVMT_CV32E40S_CONSTANTS_SV__
 
-   // TODO: Expand as needed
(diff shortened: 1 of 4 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
```diff
@@ -18,77 +18,160 @@
 `ifndef __UVMT_CV32E40S_CONSTANTS_SV__
 `define __UVMT_CV32E40S_CONSTANTS_SV__
 
-   // TODO: Expand as needed
-   `ifdef DBG_NUM_TRIG_0
-      parameter CORE_PARAM_NUM_TRIGGERS = 0;
+
+   `ifdef PARAM_SET_0
+      `include  "cvverif_40s_params.svh"
```
silabs-robin, line 23: This refers to an external file that alternatively can be used to set all params.

silabs-hfegran (approved):
Looks good to me

PR #1619, opened 2023-02-06
Questa Formal - "Design Checks" Fix

silabs-robin:
This PR adds a few constraints to the Questa Formal setup. (Rationale below.)

I think I figured out why the counter-examples looked so nonsensical.
At first I thought it was a problem with the reset initialization sequence, but it was only partially that...

"Free nets" seems to be treated differently in Jasper and Questa. Jasper goes "it's a free net, it can do whatever, and Questa goes "this is not a valid control point, whoops there goes gravity".
Free nets needs to be turned into valid "control points", and one can do that by explicitly setting a cutpoint.
Before this change you could get e.g. rvfi_valid on cycle 0 and all kinds of weird behavior that seems to completely ignore what is written in the rtl.

### DIFF
--- a/cv32e40s/fv/qverify.tcl
+++ b/cv32e40s/fv/qverify.tcl
@@ -20,16 +20,17 @@
 proc cvfv_rerun {} {
   onerror  {exit 1}
 
-  # TODO:silabs-robin  Bring up the formal view before executing the rest
-
   puts "cvfv: compiling verilog"
   vlog  -mfcu  -f fv.flist
 
+  puts "cvfv: setting cutpoints"
+  netlist cutpoint {uvmt_cv32e40s_tb.clknrst_if.reset_n} -module uvmt_cv32e40s_tb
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Hi @silabs-robin.  This PR is failing its ECA check.  Did you create it from your Silabs email?

silabs-robin:
> Hi @silabs-robin. This PR is failing its ECA check. Did you create it from your Silabs email?

Hi. I wrote about this on Mattermost, in "Town Square".

silabs-robin:
> Hi @silabs-robin. This PR is failing its ECA check. Did you create it from your Silabs email?

Hi @MikeOpenHWGroup . The Eclipse ECA problem should be fixed now.

MikeOpenHWGroup:
Thanks @silabs-robin. Yes, the failing ECA check was a problem in EF's backend that has since been resolved.  This PR LGTM, but I will defer to @silabs-hfegran's approval.

PR #1620, opened 2023-02-06
Continued work on debug asserts

silabs-mateilga:
Another stepping stone towards getting the debug asserts up to speed again. Updates the pc_dpc-assert, and several ebreak-related asserts to RVFI timing.

Status of the assertion set is still "under development", but I'd like to PR now to accommodate efforts from other engineers coming in, and sharing expansions of the helper functions in rvfi_instr_if. 

A no-merge decision on this PR is fine, but I think the pros outweigh the cons.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
@@ -240,10 +240,10 @@ module uvmt_cv32e40s_debug_assert
         else `uvm_error(info_tag,$sformatf("Debug mode not restarted after ebreak"));
 
     cov_cebreak_dbg : cover property(
-        rvfi.is_cebreak() && rvfi.rvfi_dbg_mode
+        rvfi.is_ebreak_compr() && rvfi.rvfi_dbg_mode
     );
     cov_ebreak_dbg : cover property(
-        rvfi.is_ebreak_noncomp() && rvfi.rvfi_dbg_mode
+        rvfi.is_ebreak_noncompr() && rvfi.rvfi_dbg_mode
     );
 
     // Trigger match results in debug mode
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
@@ -25,6 +25,8 @@ module uvmt_cv32e40s_debug_assert
       uvma_rvfi_csr_if csr_dpc,
       uvma_rvfi_csr_if csr_mepc,
       uvma_rvfi_csr_if csr_mstatus,
+      uvma_rvfi_csr_if csr_mtvec,
+      uvma_rvfi_csr_if csr_tdata1,
       uvmt_cv32e40s_debug_cov_assert_if cov_assert_if
   );
 
@@ -40,6 +42,8 @@ module uvmt_cv32e40s_debug_assert
   localparam int DCSR_STEP_POS    = 2;
   localparam int DCSR_NMIP_POS    = 3;
   localparam int DCSR_STEPIE_POS  = 11;
+  localparam int DCSR_EBREAKM_POS  = 15;
+  localparam int DCSR_EBREAKU_POS  = 12;
 
   // ---------------------------------------------------------------------------
   // Local variables
@@ -47,14 +51,12 @@ module uvmt_cv32e40s_debug_assert
   string info_tag = "CV32E40S_DEBUG_ASSERT";
   logic [31:0] pc_at_dbg_req; // Capture PC when debug_req_i or ebreak is active
   logic [31:0] pc_at_ebreak; // Capture PC when ebreak
-  logic [31:0] halt_addr_at_entry;
-  logic halt_addr_at_entry_flag;
+  logic [31:0] halt_addr;
   logic [31:0] exception_addr_at_entry;
   logic exception_addr_at_entry_flag;
-  logic [31:0] tdata2_at_entry;
   // Locally track which debug cause should be used
   logic [2:0] debug_cause_pri;
-  logic [31:0] boot_addr_at_entry;
+  logic [31:0] boot_addr;
   logic [31:0] mtvec_addr;
   logic        is_trigger_match;
   logic        is_rvfi_nmi_handler;
@@ -66,6 +68,10 @@ module uvmt_cv32e40s_debug_assert
 
   logic first_fetch;
   logic fetch_enable_i_q;
+
+  logic         prev_was_ebreak_flag;
+  logic [31:0]  prev_was_ebreak_pc_wdata;
+  logic [31:0]  dpc_rdata_q;
 
   // ---------------------------------------------------------------------------
   // Clocking blocks
@@ -96,7 +102,7 @@ module uvmt_cv32e40s_debug_assert
 
   assign is_trigger_match = (cov_assert_if.trigger_match_in_wb || cov_assert_if.etrigger_in_wb) && cov_assert_if.wb_valid;
 
-  assign mtvec_addr = {cov_assert_if.mtvec[31:2], 2'b00};
+  assign mtvec_addr = {csr_mtvec.rvfi_csr_rdata[31:2], 2'b00};
 
   assign is_rvfi_nmi_handler =
     rvfi.rvfi_valid && (rvfi.rvfi_intr.cause & 11'h 400);
@@ -122,6 +128,7 @@ module uvmt_cv32e40s_debug_assert
 
 
     // Check that we enter debug mode when expected. CSR checks are done in other assertions
+    /*TODO:MT replace with specifics for each entry cause
     property p_enter_debug;
         $changed(debug_cause_pri) && (debug_cause_pri != 0) && !cov_assert_if.debug_mode_q
         |->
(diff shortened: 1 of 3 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -833,9 +883,9 @@ module uvmt_cv32e40s_debug_assert
     assign first_fetch = cov_assert_if.fetch_enable_i && !fetch_enable_i_q;
 
     always@ (posedge cov_assert_if.clk_i or negedge cov_assert_if.rst_ni) begin
-        if( !cov_assert_if.rst_ni || !cov_assert_if.fetch_enable_i) begin
+        if( !cov_assert_if.rst_ni) begin
             fetch_enable_i_q <= 0;
-        end else begin
+        end else if (cov_assert_if.fetch_enable_i) begin
             fetch_enable_i_q <= 1;
```
silabs-hfegran, line 889: This is not functionally equivalent, is this intended?

silabs-mateilga: Yes, the user manual has changed, the fetch_enable_i is now expected not to be deasserted after it has been asserted, except for a reset event.

Silabs-ArjanB: Where is that written in the user manual? I am not aware that we made a change for how fetch_enable_i behaves.

silabs-mateilga: image
Sorry @Silabs-ArjanB , that wasn't quite accurate, the input is "ignored". Maybe it's always been that way? I was under the impression it was a change, but it might have just been a weakness in the old assertion set.

Silabs-ArjanB: >> Maybe it's always been that way? 

Yes, this was already the case on the CV32E40P. I can't comment on the assertions themselves; I was just stating that the stated expectation above in the comments is not correct.

silabs-hfegran: I think in general the support logic should not make any assumptions on signals, but rather relay the actual signals themselves, or combinations thereof. If an input must be constrained in a specific way, that should be the responsibility of assumes on external interface ( or properly coded test bench in the case of simulation ). In cases like this one, where there may be wrong behavior on the input pins, it could get masked by support logic assuming correct input pin behavior, and harder to debug with formal tools, as it will not necessarily be reflected by assumes.

silabs-hfegran: Though it might not be relevant in this case, I will have to study the code in more detail; I mistakenly read this to be a part of the common support logic file, which is not the case.

silabs-mateilga: The logic here is a memory of the fact that fetch enable has gone high, so as to not get an issue with the signal being manipulated "out of spec" by the formal tool or any test bench, without limiting the test environment from doing exactly that. It is used exclusively in the debug_assert file, and I don't think the weaknesses of this causes any issues.

silabs-hfegran (approved):

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -131,42 +138,44 @@ module uvmt_cv32e40s_debug_assert
...
+        ) else `uvm_error(info_tag, $sformatf("Debug mode entered with wrong pc. pc==%08x", rvfi.rvfi_pc_rdata));
 
+    //TODO:MT rewrite to rvfi timing
     a_debug_mode_pc_dpc: assert property(
         $rose(first_debug_ins)
         |->
-        cov_assert_if.dpc_q == pc_at_dbg_req
-        ) else `uvm_error(info_tag, $sformatf("Debug mode entered with wrong dpc. dpc==%08x", cov_assert_if.dpc_q));
-
-    a_debug_mode_pc_dmode: assert property(
```
silabs-robin, line 150: Is this removed because it would be tautological?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -176,76 +185,70 @@ module uvmt_cv32e40s_debug_assert
         else `uvm_error(info_tag, $sformatf("Debug cause not correct for haltreq, cause = %d",cov_assert_if.dcsr_q[8:6]));
 
     // Check that debug with cause ebreak is correct
-    property p_cebreak_debug_mode;
-        $rose(cov_assert_if.debug_mode_q) && (cov_assert_if.dcsr_q[8:6] == cv32e40s_pkg::DBG_CAUSE_EBREAK)
-        |-> debug_cause_pri == cv32e40s_pkg::DBG_CAUSE_EBREAK;
+    property p_ebreak_debug_mode;
+        $rose(first_debug_ins) && (rvfi.rvfi_dbg == cv32e40s_pkg::DBG_CAUSE_EBREAK)
+        |-> prev_was_ebreak_flag;
```
silabs-robin, line 190: This previously checked both that an ebreak had occured and that it was the highest priority for entry, but now it only checks that an ebreak occured?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -176,76 +185,70 @@ module uvmt_cv32e40s_debug_assert
...
+    // TODO should check no change in dpc and dcsr
 
     property p_ebreak_during_debug_mode;
-        $rose(cov_assert_if.is_ebreak) && cov_assert_if.debug_mode_q
-        |->
-        s_conse_next_retire
-        ##0 cov_assert_if.debug_mode_q && (cov_assert_if.wb_stage_pc == halt_addr_at_entry);
-        // TODO should check no change in dpc and dcsr
+        rvfi.is_ebreak() &&
+        rvfi.rvfi_trap.debug_cause == cv32e40s_pkg::DBG_CAUSE_EBREAK && //The ebreak is actually taken
```
silabs-robin, line 232: With this line, the antecedent start with the core having decided to take a specific debug cause.

While the old assert checked "if we _should_ act upon the cause, then it must have the following effects",
the new assert checks "if the core _has decided_ ..."?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -256,13 +259,47 @@ function logic is_wfe();
...
 function logic is_ebreak();
-  return match_instr(INSTR_OPCODE_EBREAK, INSTR_MASK_FULL);
+  return match_instr(INSTR_OPCODE_EBREAK, INSTR_MASK_FULL) || match_instr(INSTR_OPCODE_C_EBREAK, INSTR_MASK_FULL);
 endfunction : is_ebreak
 
+function logic is_cebreak();
+  return match_instr(INSTR_OPCODE_C_EBREAK, INSTR_MASK_FULL);
+endfunction : is_cebreak
+
+function logic is_ebreak_noncomp();
```
silabs-robin, line 269: The ebreak naming is not symmetrical and obvious.
Could we do something like `is_ebreak_compr`, `is_ebreak_noncompr`, `is_ebreak`/`is_ebreak_any`?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -676,6 +680,79 @@ module uvmt_cv32e40s_debug_assert
     // Capture internal states for use in checking
     // -------------------------------------------
 
+    always @(posedge cov_assert_if.clk_i or negedge cov_assert_if.rst_ni) begin
+        if(!cov_assert_if.rst_ni) begin
+            pc_at_dbg_req <= 32'h0;
+            pc_at_ebreak <= 32'h0;
```
silabs-robin, line 686: Is this unused?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -676,6 +680,79 @@ module uvmt_cv32e40s_debug_assert
...
+    end
+    // remembering pc_wdata from an ecall, to deal with a req taking priority for debug entry
+    always @(posedge cov_assert_if.clk_i or negedge cov_assert_if.rst_ni) begin
+        if(!cov_assert_if.rst_ni) begin
+            prev_was_ebreak_flag <= 0;
+            prev_was_ebreak_pc_wdata <= 32'h0;
+        end else begin
+            if(rvfi.is_ebreak()) begin
+                prev_was_ebreak_flag <= 1;
+                prev_was_ebreak_pc_wdata <= rvfi.rvfi_pc_wdata;
```
silabs-robin, line 739: Why is ebreak PC modelled separate from the general PC modelling?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: The entire assert is being rewritten to rvfi timing, as it was to close to the core before. Removing this was more a "broader picture" thing. Also, this assert isn't proven yet, and is still being worked on.

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: Priority will be checked separately.

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: The line is there to separate out ebreaks that are deemed illegal, as the is_ebreak()function only checks that the instruction word is an ebreak. This check is not relevant for ebreaks that are not taken.

silabs-mateilga: The check you are looking for I think) is covered above, with the ebreak_Xmode_exception asserts

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: This bit of support logic is under reconstruction, not everything that has fallen out of use is removed yet.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: I have no strong feelings here, but the term "cebreak" is used quite a bit in old code. I don't mind changing it though, if we can agree about naming. These are not widely used yet, so the time is now... 

I guess is_ebreak, is_ebreak_compr, is_ebreak_noncompr is a good option, if we want to move away from the CEBREAK term.

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-robin: > The line is there to separate out ebreaks that are deemed illegal, as the is_ebreak()function only checks that the instruction word is an ebreak.

It makes sense to filter out "invalid cases", but this trusts what the core is claiming.
If there was a bug where the core _should_ have taken a DBG_CAUSE_EBREAK but didn't, then we would miss it.

What are the cases that make the ebreak invalid? Instruction access fault, instruction bus fault, and more? I think it could be better to exclude the invalid cases based on our expectations instead of trusting that the core has done the right thing.

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: It was easier to add the memory of pc from the ebreak event to the already existing ebreak memory logic, which was made first.

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-mateilga: I consider checking that an ebreak should be taken/not taken a separate issue, this assert checks what happens when an ebreak is taken. I intend to verify the "should be taken"-side of things elsewhere.

Also, as a note, the previous implementation checked that the core_fsm had decided there was an ebreak, so it wasn't exactly any better in that regard.

silabs-hfegran (approved):

PR #1622, opened 2023-02-06
Passive/Active agent for AXI4

AEzzejjari:
The axi agent can act as a passive or active agent.
For the active mode the memory system replaced by the memory class.
A extension not supported.


JeanRochCoulon:
Alae, I apologize but you need to (re)submit your PR to core-v-verif/cva6/dev (we have changed our mind Yesterday)

PR #1628, opened 2023-02-07
Hotfix + update to latest hash

silabs-hfegran:


silabs-robin (approved):
Approving based on:
Hash update is ok,
I remember that mclicbase should go iirc,
the param rename was recently changed,
and showoverrides should be off by default.

PR #1630, opened 2023-02-07
OBI Asserts - Make Formal-Friendly

silabs-robin:
This PR swaps out a dynamic array in the obi memory asserts with a packed array and rewrites the accompanying logic, so that these asserts can be used in formal.

Note: This is in "lib/" so it will affect other cores.
Note: The asserts are already bound in the tb but were not used in fv.

Passes ci_check (except for known debug test).
Runs in formal, and the assert in question passes.


silabs-robin:
Btw, what is this? https://github.com/openhwgroup/core-v-verif/blob/master/lib/uvm_agents/uvma_rvfi/uvma_rvfi_assert.sv

> rvfi agent

> obi signal names and asserts

silabs-mateilga (approved):

silabs-hfegran (approved):

PR #1631, opened 2023-02-08
RVFI Assert - Fix CEXes

silabs-robin:
This PR updates rvfi_assert to keep up with changes to the core and the environment.

Passes ci_check (except you know).
The asserts in question passes in formal.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
@@ -20,7 +20,9 @@ module uvmt_cv32e40s_rvfi_assert
   import cv32e40s_pkg::*;
   import cv32e40s_rvfi_pkg::*;
   import uvm_pkg::*;
-(
+#(
+  parameter logic  SMCLIC
+)(
   input  clk_i,
   input  rst_ni,
 
@@ -133,12 +135,15 @@ module uvmt_cv32e40s_rvfi_assert
     rvfi_trap.exception_cause
   ) else `uvm_error(info_tag, "rvfi_trap exceptions must have a cause");
 
-  a_interrupts_cause: assert property (
-    rvfi_valid  &&
-    rvfi_intr
-    |->
-    rvfi_intr.cause
-  ) else `uvm_error(info_tag, "rvfi_intr interrupts must have a cause");
+  if (!SMCLIC) begin: gen_clint_cause
+    a_interrupts_cause: assert property (
+      rvfi_valid  &&
+      rvfi_intr
+      |->
+      rvfi_intr.cause
+    ) else `uvm_error(info_tag, "rvfi_intr interrupts must have a cause");
+    // TODO:silabs-robin  Could also assert "cause < 2^clicwidth"
+  end : gen_clint_cause
 
   a_debugs_cause: assert property (
     rvfi_valid  &&
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
@@ -16,28 +16,32 @@
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 
 
+`default_nettype  none
+
+
 module uvmt_cv32e40s_rvfi_assert
   import cv32e40s_pkg::*;
   import cv32e40s_rvfi_pkg::*;
   import uvm_pkg::*;
 #(
-  parameter logic  SMCLIC
+  parameter logic  SMCLIC,
+  parameter int    SMCLIC_ID_WIDTH
 )(
-  input  clk_i,
-  input  rst_ni,
+  input wire  clk_i,
+  input wire  rst_ni,
 
-  input             rvfi_valid,
-  input [ 4:0]      rvfi_rs1_addr,
-  input [ 4:0]      rvfi_rs2_addr,
-  input [31:0]      rvfi_rs1_rdata,
-  input [31:0]      rvfi_rs2_rdata,
-  input [ 2:0]      rvfi_dbg,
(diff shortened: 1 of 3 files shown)


mk/uvmt/xrun.mk
```diff
@@ -244,6 +249,9 @@ XRUN_COMP_FLAGS += -nowarn CGPIDF
 # deselect_coverage -all warnings
 XRUN_COMP_FLAGS += -nowarn CGNSWA
 
+# Value Parameters without default values
+XRUN_COMP_FLAGS += -setenv CADENCE_ENABLE_AVSREQ_44905_PHASE_1=1
```
silabs-robin, line 253: For more info, look up `SVVMAP` on the Cadence support page.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -76,24 +78,51 @@ module uvmt_cv32e40s_rvfi_assert
...
+
+  a_dbg_cause_step: assert property (
+    p_dbg_cause_n(4)
+  ) else `uvm_error(info_tag, "'rvfi_dbg' did not match 'dcsr.cause'");
+
 
   // RVFI exception cause matches "mcause"
 
+  wire logic [10:0]  rvfi_mcause_exccode;
+  assign rvfi_mcause_exccode = (rvfi_csr_mcause_wdata & rvfi_csr_mcause_wmask);
```
silabs-robin, line 118: Had to limit this to `10:0` because CLIC uses `mcause` for wild new purposes.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -106,12 +135,15 @@ module uvmt_cv32e40s_rvfi_assert
     rvfi_trap.exception_cause
   ) else `uvm_error(info_tag, "rvfi_trap exceptions must have a cause");
 
-  a_interrupts_cause: assert property (
-    rvfi_valid  &&
-    rvfi_intr
-    |->
-    rvfi_intr.cause
-  ) else `uvm_error(info_tag, "rvfi_intr interrupts must have a cause");
+  if (!SMCLIC) begin: gen_clint_cause
```
silabs-robin, line 138: This property should not hold under CLIC, so it is now conditionally included.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -76,24 +78,51 @@ module uvmt_cv32e40s_rvfi_assert
...
+    p_dbg_cause_n(2)
+  ) else `uvm_error(info_tag, "'rvfi_dbg' did not match 'dcsr.cause'");
+
+  a_dbg_cause_haltreq: assert property (
+    p_dbg_cause_n(3)
+  ) else `uvm_error(info_tag, "'rvfi_dbg' did not match 'dcsr.cause'");
+
+  a_dbg_cause_step: assert property (
+    p_dbg_cause_n(4)
+  ) else `uvm_error(info_tag, "'rvfi_dbg' did not match 'dcsr.cause'");
```
silabs-robin, line 112: These case-splitting helper asserts are logically redundant, but they get proven faster than the general asserts, and when they are proven then the general assert gets proven quite quickly. (Note, I have not done comprehensive measurements of whether the time of helpers+general is significantly faster than just general, but the previous statements are demonstrably true.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -20,7 +20,9 @@ module uvmt_cv32e40s_rvfi_assert
   import cv32e40s_pkg::*;
   import cv32e40s_rvfi_pkg::*;
   import uvm_pkg::*;
-(
+#(
+  parameter logic  SMCLIC
```
silabs-robin, line 24: This deliberately has no default value, because it is much safer. Then it is impossible to forget to set it. If a new param is addet, no rogue instantiation will forget to add a deliberate specification for the value. Etc.

silabs-robin: I just (this week) saw in `tb` that we have not been keeping track with new core params at all. With default params, that will just go unnoticed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
Silabs-ArjanB: CLIC interrupts also have a 'cause'; the cause can however be 0. You could for example check that the cause is not above the maximum allowed value (based on id width)

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -76,27 +82,77 @@ module uvmt_cv32e40s_rvfi_assert
...
+    a_irq_cause_clint: assert property (
+      rvfi_valid  &&
+      rvfi_intr.interrupt
+      |->
+      (rvfi_intr.cause inside {3, 7, 11, [16:31], [1024:1027]})
+    ) else `uvm_error(info_tag, "unexpected interrupt cause");
+  end : gen_legal_cause_clint
+
+  if (SMCLIC) begin: gen_legal_cause_clic
+    localparam logic [31:0]  MAX_CLIC_ID = 2^SMCLIC_ID_WIDTH - 1;
```
silabs-robin, line 145: `^` is wrong... One moment, I'll fix it.

silabs-robin: Fixed.  Ready to continue review/merge.

silabs-mateilga (approved):

silabs-hfegran (approved):

PR #1634, opened 2023-02-09
Initial Commit for CV32E40P v2 Verification Plans

XavierAubert:
Here is the initial commit for CV32E40Pv2 verification plans. 

As discussed previously with Mike, he who would prefer to keep a history of v1 plan inside the excel sheets. So, for xpulp instructions, I moved the "legacy" v1 plans at the end of each vplan file. For other files, the features are simply added on top of the existing vplans.

Some plans in the Simulation directory are left untouched or contains small modifications: these are still relevant for v2 since the features they cover have not been updated since v1


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Hi @XavierAubert, this PR looks good.  However, this pull-request needs to be committed from the account (email address) that is registered with Eclipse.  Please rebase from that account.

PR #1643, opened 2023-02-13
Isacov - Link to "decoder.h" in Readme

silabs-robin:
> > Maybe we could link to https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/decode.h in the lib/dpi_dasm/ readme?
> 
> Good idea! Can you take care of that in a separate PR @silabs-robin?

https://github.com/openhwgroup/core-v-verif/issues/1582#issuecomment-1416380145


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks @silabs-robin!

PR #1644, opened 2023-02-13
Fencei Directed - Add to "full" Regression

silabs-robin:
This PR adds the FENCEI directed tests to the "full" regression list

Test status:
* full - The added tests pass in this regression.

### DIFF
--- a/cv32e40s/regress/cv32e40s_full.yaml
+++ b/cv32e40s/regress/cv32e40s_full.yaml
@@ -415,8 +415,7 @@ tests:
 #TODO:MT Comment back in when Zc is enabled in the core
 #  zc_test:
 #    description: Zc directed test
-#    builds: [ uvmt_cv32e40s, uvmt_cv32e40s_pma_1]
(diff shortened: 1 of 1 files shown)


cv32e40s/regress/cv32e40s_full.yaml
```diff
@@ -409,8 +415,7 @@ tests:
 #TODO:MT Comment back in when Zc is enabled in the core
 #  zc_test:
 #    description: Zc directed test
-#    builds: [ uvmt_cv32e40s, uvmt_cv32e40s_pma_1]
+#    builds: [uvmt_cv32e40s, uvmt_cv32e40s_pma_1]
```
silabs-robin, line 418: @silabs-mateilga do you know if there is a specific reason to use pma config "1" and not one of the others?

silabs-robin:
Sorry, I just noticed there is an issue with PMA CFG 3 and 4.

Will investigate and re-open/un-draft when done.

MikeOpenHWGroup (approved):

PR #1654, opened 2023-02-15
U-Mode Directeds - Add To "full" Regression

silabs-robin:
This PR adds the privilege directed tests to the "full" regression list

Test status:
* full - All but 1 ~~2~~ tests pass
    * ~~interrupt_priv_test disagrees with the ISS: Issue 1649~~
    * `custom_priv_gen_test` times out and is hence commented out from the "full" regression.


cv32e40s/tests/programs/custom/csr_priv_gen_test/README.md
```diff
@@ -1,5 +1,9 @@
-# Python generator files 
+Contains generated tests on U-mode csr instructions to insert illegall access
```
silabs-robin, line 1: I was looking for an explanation of what the test does, and when I found it I just copy-pasted it into the readme so it is easier to find.

silabs-mateilga (approved):

silabs-hfegran (approved):

Issue #1658, opened 2023-02-16
ISS – Off-by-One Haltreq Debug Entry


silabs-robin:
### Type

* Functionally incorrect behavior

### Steps to Reproduce

1. Branch: `pmadirectedfix`, https://github.com/silabs-robin/core-v-verif/tree/pmadirectedfix
1. Hash: `414984`
1. Compiler: `corev-openhw-gcc-centos7-20221213`
1. **Edit:** Remove everything `zc` from `cv32e40s/tests/cfg/pma_debug.yaml`  (except "zca")
1. Command: `make test TEST=pma_debug CFG=pma_debug`

The first mismatch is on PC, when the core has entered debug (per haltreq requested by the test program), but the ISS is not executing debug code yet.
On the subsequent PC mismatch errors, the ISS and core is off by one.

```
Error (IDV) PC mismatch (HartId:0, PC:0x000000ec main+32):
Info (IDV)  0>
Info (IDV)   . dut:0x1a110800 debugger_entry+0
Info (IDV)   . ref:0x000000ec main+32
Error (IDV) Insn. bit pattern mismatch (HartId:0, PC:0x000000ec main+32):
Info (IDV)  1>
Info (IDV)   . dut:7b211073 csrw    dscratch0,x2
Info (IDV)   . ref:49050513 addi    x10,x10,1168
Error (IDV) CSR register value mismatch (HartId:0, PC:0x000000ec main+32):
Info (IDV)  2> CSR 7b1 (dpc)
Info (IDV)   . dut:0x000000f0
Info (IDV)   . ref:0x000000ec
UVM_ERROR @ 15262.300 ns : rvvi-pkg.sv(94) reporter [] uvmt_cv32e40s_tb.imperas_dv.trace2api.state_compare @ 15262.000 ns: MIS
MATCH
Error (IDV) PC mismatch (HartId:0, PC:0x1a110800 debugger_entry+0):
Info (IDV)  3>
Info (IDV)   . dut:0x1a110804 debugger_entry+4
Info (IDV)   . ref:0x1a110800 debugger_entry+0
Error (IDV) Insn. bit pattern mismatch (HartId:0, PC:0x1a110800 debugger_entry+0):
Info (IDV)  4>
Info (IDV)   . dut:00001117 auipc   x2,0x1
Info (IDV)   . ref:7b211073 csrw    dscratch0,x2
[...]

PR #1669, opened 2023-02-20
PMA Directeds – Comment In & Add to Regress

silabs-robin:
This PR comments in parts of the "pma" test that were previously pending RTL updates, and it also adds the different "pma*" tests to the "full" regression list.

Test status:
* full - All "pma*" tests pass, but
    * There are ISS mismatches, ongoing investigation https://github.com/openhwgroup/core-v-verif/issues/1658
    * With zcmp enabled, the interrupt handler generates bad prolog/epilog (no ticket created for this yet)


cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -85,345 +104,220 @@ static void assert_or_die(uint32_t actual, uint32_t expect, char *msg) {
...
-  // Table jump failing second fetch should be the fault of the target
-  reset_volatiles();
-  /* TODO enable when RTL is implemented
-  make sure table is executable but target is not
-  __asm__ volatile("c.tbljal 1");
-  assert_or_die(mcause, EXCEPTION_INSN_ACCESS_FAULT, "error: tblj failing second should instruction access fault\n");
-  assert_or_die(mepc, TBLJ_TARGET_ADDR, "error: tblj second expected different mepc\n");
-  assert_or_die(mtval, MTVAL_READ, "error: tblj second expected different mtval\n");
-  */
+  // Zc features are in the vplan, but are much easier to check with assertions.
```
silabs-robin, line 320: I decided to not do the pma-zc checking in C/asm because it gets complicated by the "observer effect" (_"the act of observing will influence the phenomenon being observed"_).
With an assert you can get the given scenario and observe that the outcome is as expected, without the additional constraint that the system state maintains a healthy SW control flow.

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -19,62 +19,81 @@
...
+ * 0x1A11_1000
+ * 0x1A11_0800 + 16    <IO_ADDR>
+ *
+ * [r1, "main"]
+ * 0x1A11_0800 + 16
+ * 0x0000_0000         <MEM_ADDR_0>
+ */
+#define  MEM_ADDR_0        0
+#define  IO_ADDR           (0x1A110800 + 16)
+#define  MEM_ADDR_1        0x1A111000
```
silabs-hfegran, line 42: I believe this is the debug exception handler address - is this deliberate?

silabs-robin: Yes. This test does not use any debug features, so instead of doing anything difficult with the linker scripts it just re-uses the unused debug memory.

MikeOpenHWGroup (approved):

PR #1671, opened 2023-02-21
2 PULP tests corrections

pascalgouedo:
All 19 tests are then correctly running on CV32E40Pv2 and 20230217 corev Embecosm toolchain.


MikeOpenHWGroup (approved):
Thanks @pascalgouedo.  I am not in a position to review test-programs using the PULP ISA, but by inspection this looks fine.  In the worse case, the only impact will be to these specific test-programs.

PR #1673, opened 2023-02-23
Prevent runaway executions of Spike: Add patch, update cva6/sim Makefile.

zchamski:
This PR modifies Spike in order to prevent uncontrolled runaway executions if Spike encounters an unhandled exception while the trap vector is incorrectly configured.  This case can arise when the `trap_instruction_access_fault` entry contains all zeroes.

The PR adds a new Spike cmdline option `--steps=<n>` and the necessary internal modifications to stop the simulation when the estimated number of executed steps exceeds `n`.  It also adds the option `--steps=2000000` to Spike invocation in `cva6/sim/Makefile`.  This value is aligned with the default cycle limit in CVA6 RTL testbench.

Changed/added files:

* cva6/regress/install-spike.sh: Install 'max-steps' patch.
* cva6/regress/spike/patches/spike-max-steps.patch: New.
* cva6/sim/Makefile (spike): Add execution limit matching the default cycle limit in RTL testbench.

### DIFF
--- a/cva6/regress/install-spike.sh
+++ b/cva6/regress/install-spike.sh
@@ -26,6 +26,7 @@ if [ ! -f "$SPIKE_ROOT/bin/spike"  ]; then
     git apply $PATCH_DIR/spike/patches/spike-shared-fesvr-lib.patch
     git apply $PATCH_DIR/spike/patches/spike-cvxif-extension.patch
     git apply $PATCH_DIR/spike/patches/spike-dlopen-diagnostics.patch
+    git apply $PATCH_DIR/spike/patches/spike-max-steps.patch
     # Recursively copy Spike files related to CVXIF extension into current
     # directory.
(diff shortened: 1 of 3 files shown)


MikeOpenHWGroup:
This is a great idea @zchamski.  I will defer to others who are better positioned to review such changes (I am not very familiar with Spike's implementation).

Question: would this be a useful contribution to upstream to the [Spike repo](https://github.com/riscv-software-src/riscv-isa-sim)?

zchamski:
> Question: would this be a useful contribution to upstream to the [Spike repo](https://github.com/riscv-software-src/riscv-isa-sim)?
It might interest the Spike community indeed as it addresses a corner case of Spike operation that causes resource hogging.

zchamski:
Superseded by PR #1949, closing.

PR #1675, opened 2023-02-23
Changed SMCLIC to CLIC

silabs-oysteink:
Updated core hash to latest. Changed SMCLIC to CLIC to follow latest RTL.
Mostly search-and-replace, some indentation differences may occur. If you want the indentation fixed in this PR, please let me know.

### DIFF
--- a/cv32e40s/sim/ExternalRepos.mk
+++ b/cv32e40s/sim/ExternalRepos.mk
@@ -15,7 +15,7 @@ export SHELL = /bin/bash
 
 CV_CORE_REPO   ?= https://github.com/openhwgroup/cv32e40s
 CV_CORE_BRANCH ?= master
-CV_CORE_HASH   ?= df695cd46249ebdc93f141f76de8f45edfa57f79
+CV_CORE_HASH   ?= a9c6c716e4d1a219775b968bafec239f1190ac16
 CV_CORE_TAG    ?= none
 
 RISCVDV_REPO    ?= https://github.com/google/riscv-dv
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_clic_interrupt_assert.sv
@@ -20,8 +20,8 @@ module uvmt_cv32e40s_clic_interrupt_assert
   import cv32e40s_pkg::*;
   import uvma_rvfi_pkg::*;
   #(
-    parameter int   CLIC          = 0,
-    parameter int   CLIC_ID_WIDTH = 5,
+    parameter int   CLIC            = 0,
+    parameter int   CLIC_ID_WIDTH   = 5,
     parameter int   NUM_IRQ         = 32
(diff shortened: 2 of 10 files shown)


cv32e40s/sim/ExternalRepos.mk
```diff
@@ -15,7 +15,7 @@ export SHELL = /bin/bash
```
silabs-robin, line 15: Does it pass ci_check, and is formal still runnable?

silabs-robin (commented):
I am almost ready to approve but I have one post with 2 questions in it that I would like to resolve first.

Rationale:
Hash is updated (main purpose of the PR).
`*_constants.sv` updated (important, good).
Relevant yaml config updated (important, good).
Seems like the search-and-replace was comprehensive and correct in all other files too.

cv32e40s/sim/ExternalRepos.mk
silabs-robin: `ci_check` is suffering from a problem atm, but here is a reference of a run I did recently (pic below), so I would expect that the same tests pass/fail for this PR:

image

`Error: Zcm* is not compatible with the `c' extension`

`Error: illegal operands `cm.push {ra},-64'`

silabs-oysteink: I'll give ci_check and formal a go, and report when they're done.

Silabs-ArjanB: @silabs-robin Which tool is producing this 'Error: Zcm* is not compatible with the c' extension`' and why?

silabs-robin: > Which tool is producing this 'Error: Zcm* is not compatible with the c' extension`' and why?

`gcc` / `gas`, because we have used `.option norvc` and particularly `.option rvc`, and it is now in conflict with the `march` containing `*_zcb_zcmp`.

This is easy to fix (using `.option push` and `.option pop` around a section requiring `.option norvc`) for our directed tests, for our own random tests, and our bsp, but I think riscv-dv also uses this and that is more tricky. Henrik indicated he might have an idea for how to fix that last one.

silabs-robin (approved):
Approving based on:
Previously stated rationale (above comments).
Formal is still runnable and ci_check passes to an expected degree (as per DMs).

cv32e40s/sim/ExternalRepos.mk
silabs-oysteink: ci_check gives the same number of passed/failed, but with a few variations to which tests fail. The error seems consistent with the Zcm* mentioned above though.

silabs-oysteink: Formal seems ok

silabs-mateilga (approved):
Approving based on the discussion already had, and that I can't see any errors in the file changes.

MikeOpenHWGroup (approved):
Approving/Merged based on recommendation from @silabs-robin and @silabs-mateilga.

PR #1679, opened 2023-02-23
Re-use Core control Agent

AyoubJalali:
The following PR related to re-use of the existent core control agent.
Things to Know :
The CVA6 has only the **boot_addr** as an input of the signals in the core control agent, so the other signals have no effect.
The **fetch_toogle_seq** as @MikeOpenHWGroup mention was added to verify a requirement of the embedded cores.
We can change the **boot_addr** value from the **uvme_cva6_cfg**.sv file, or add a simulation option +boot_addr = (some value), with that you should give the same value to spike (add a [cva6.py](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/sim/cva6.py) option : **--isspostrun_opts**=some value ) and the [linker ](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/sim/link.ld) to guaranty that your test run very well, you should know also that there is some limitation in the **boot_addr** values (raise a INSTR_ACCESS_FAULT exception).

### DIFF
--- a/cva6/env/uvme/uvma_cva6_core_cntrl_agent.sv
+++ b/cva6/env/uvme/uvma_cva6_core_cntrl_agent.sv
@@ -1,0 +1,114 @@
+// Copyright 2023 OpenHW Group
+// Copyright 2023 Datum Technology Corporation
+// Copyright 2023 Silicon Labs, Inc.
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://solderpad.org/licenses/
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+
+`ifndef __UVMA_CVA6_CORE_CNTRL_AGENT_SV__
+`define __UVMA_CVA6_CORE_CNTRL_AGENT_SV__
+
+/**
+ * Core control agent defined for the CVA6
+ */
+class uvma_cva6_core_cntrl_agent_c extends uvma_core_cntrl_agent_c;
+
+
+   string log_tag = "CVA6CORECTRLAGT";
+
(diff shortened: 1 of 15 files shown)


cva6/env/uvme/uvme_cva6_cfg.sv
```diff
@@ -106,5 +165,33 @@ function uvme_cva6_cfg_c::new(string name="uvme_cva6_cfg");
...
+
+   uvma_cva6_core_cntrl_cntxt_c cva6_cntxt;
+
+   if (!$cast(cva6_cntxt, cntxt)) begin
+      `uvm_fatal("SAMPLECNTXT", "Could not cast cntxt to uvma_cva6_core_cntrl_cntxt_c");
+   end
+
+
+   num_mhpmcounters = cva6_cntxt.core_cntrl_vif.num_mhpmcounters;
+   //~ pma_regions      = new[cva6_cntxt.core_cntrl_vif.pma_cfg.size()];
```
MikeOpenHWGroup, line 178: Are these here for possible future use?  If so, please add a TODO comment header to say that.  If not, I recommend deleting it altogether.

cva6/env/uvme/uvma_cva6_core_cntrl_agent.sv
```diff
@@ -0,0 +1,118 @@
...
+
+   if (cfg.is_active) begin
+      fork
+         start_fetch_toggle_seq();//Do nothing
+      join_none
+   end
+
+endtask : run_phase
+
+task uvma_cva6_core_cntrl_agent_c::start_fetch_toggle_seq();
```
MikeOpenHWGroup, line 108: @AyoubJalali, as you point out, this is not needed by the CVA6. As such, this task is not expected to be called. I recommend adding a `uvm_error()` call.  Something like:
```
`uvm_error("CVA6 CORE CTRL", fetch toggle not supported in CVA6")
```

MikeOpenHWGroup (changes requested):
Good work @AyoubJalali.  I made a couple of minor comments, please have a look.

cva6/env/uvme/uvma_cva6_core_cntrl_agent.sv
MikeOpenHWGroup: Thanks!

MikeOpenHWGroup (approved):
LGTM @AyoubJalali.  I will leave it to either @JeanRochCoulon or @ASintzoff to execute the merge.

Issue #1681, opened 2023-02-24
Missing vcs simulation file in cv32e40p/sim/tools/vcs/vcs_batch.tcl


shariethernet:
Hello OpenHW Group Team,

I am trying to run workloads on CV32E40P using VCS. As a starting point I tried to compile the hello-world test. It seems like some tcl files are missing for vcs. 

Path: `cv32e40p/sim/core`
Command:  `make vcs-hello-world`
Error Message:  ```do: couldn't open ../tools/vcs/vcs_batch.tcl```

Logfile:
[logfile.txt](https://github.com/openhwgroup/core-v-verif/files/10820589/logfile.txt)

Could you also suggest any other working methods (besides verilator) to run workloads on the simulation?  I would need to use a commercial simulator, as I would then be performing gate-level simulations, for some research on power-energy optimizations of a certain commercial PDK. 

Thanks
Shrihari

MikeOpenHWGroup:
Hi @shariethernet.

VCS is rarely used for the "core" testbench so the Makefile rules for running it with VCS are out of date.  Thanks for pointing this out.  I believe I have a fix and will create a pull-request for it shortly.  However, my VCS run-time license has expired (!) and so I will need to get that restored before I can know for sure if the problem is completely resolve.

As for the other questions, please reach out to me at mike@openhwgroup.org and we can discuss.

MikeOpenHWGroup:
Hi @shariethernet, please have a look at pull-request #1682.  It _should_ resolve your issue.  I will not get this PR merged in until I can test it locally (once our VCS run-time license is restored).

shariethernet:
Hi @MikeOpenHWGroup. Thanks for the update. I got it working, but there were still a few bugs in the Makefile(s), after some tweaking, I got hello-world up. Other tests still require some work on the makefile as the programs are not getting compiled. I will resolve them and raise a PR once it works
Thanks

cheng0560011:
Hi @MikeOpenHWGroup. I am recently working on cv32e40p using vcs. Thank you for the #1682 update.
I also find that line 545 in Makefile may need som modification.
Whe I make vcs-hello-world, $(COSTOM) will return black and the elf will be put under /. May it be $(TEST_PROGRAM_PATH)/hello/hello.elf instead of $(COSTOM)/hello.elf ?

MikeOpenHWGroup:
I will have a look @cheng0560011.  Tomorrow is already busy for me, so it will be later this week.

[...]

PR #1682, opened 2023-02-24
Fix for VCS on core testbench

MikeOpenHWGroup:
Partial fix for #1681.


MikeOpenHWGroup:
This PR is now redundant after #2006 and #2014 have been merged in.   Closing.

PR #1686, opened 2023-02-28
PMA Assert - Add Missing Asserts

silabs-robin:
This PR adds assertions for the PMA based on what the link-to-coverage in the vplan revealed to be missing.

Test status:
* ci_check - Passes 7/16 (ci_check is currently this broken, unrelated from this PR.)
* Fv default - Passes (except 2 slow/non-converging)
* Fv pma 1/2/.../5 - Same as for Fv default


mk/uvmt/xrun.mk
```diff
@@ -252,6 +252,9 @@ XRUN_COMP_FLAGS += -nowarn CGNSWA
 # Value Parameters without default values
 XRUN_COMP_FLAGS += -setenv CADENCE_ENABLE_AVSREQ_44905_PHASE_1=1
 
+# Type Parameters without default values
+XRUN_COMP_FLAGS += -setenv CADENCE_ENABLE_AVSREQ_63188_PHASE_1=1
```
silabs-robin, line 256: Same as the above "Value Parameters" option, but for "Type Parameters".

This makes Xcelium behave LRM-compliant.
(Their explanation for why it is not default is that _"users may prefer that this issue should be reported during compilation stage and not later during elab"_, but it actually means that some checking will never happen at all.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -179,6 +186,48 @@ module uvmt_cv32e40s_rvfi_assert
   ) else `uvm_error(info_tag, "rvfi_trap debugs must have a cause");
 
 
+  // Mem accesses reflect actual bus
```
silabs-robin, line 189: Also, because some checking relies on RVFI, added a check that RVFI truthfully represent OBI bus transactions.

See https://github.com/openhwgroup/cv32e40s/issues/410

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -179,6 +186,48 @@ module uvmt_cv32e40s_rvfi_assert
...
+
+  var logic [31:0] writebuf_req_count_c;
+  var logic [31:0] rvfi_mem_count_c;
+  var logic [31:0] writebuf_req_count_n;
+  var logic [31:0] rvfi_mem_count_n;
+  var logic [31:0] rvfi_mem_new;
+
+  a_obi_vs_rvfi: assert property (
+    writebuf_req_count_c >= rvfi_mem_count_c
+  ) else `uvm_error(info_tag, "TODO");
```
silabs-robin, line 199: Woops! This lacks a uvm_error message. I will fix it. _(But I will wait until some more feedback accumulates so I can do it all in one go.)_

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
```diff
@@ -0,0 +1,60 @@
...
+`default_nettype none
+
+
+module  uvmt_cv32e40s_assumes (
+  input wire  clk_i,
+  input wire  rst_ni,
+
+  uvma_obi_memory_if  obi,
+  uvma_rvfi_instr_if  rvfi,
+  uvmt_cv32e40s_support_logic_for_assert_coverage_modules_if.slave_mp  sup
```
silabs-hfegran, line 34: Please use more descriptive name than "sup" - maybe support_if / support_if_mp etc.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -0,0 +1,207 @@
...
+  // Writebuffer Signals
+  input obi_data_req_t  writebuf_trans_i,
+  input obi_data_req_t  writebuf_trans_o,
+  input wire            writebuf_ready_o,
+
+  // PMA Verdict
+  input wire  pma_err,
+
+  // Support Interfaces
+  uvmt_cv32e40s_support_logic_for_assert_coverage_modules_if.slave_mp  sup,
```
silabs-hfegran, line 62: "sup", as comment above

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
```diff
@@ -0,0 +1,60 @@
...
+
+// Description:
+//   NB! Assumes are dangerous.
+//   Here is a collection of some assumes needed to have the environment usable.
+
+
+`default_nettype none
+
+
+module  uvmt_cv32e40s_assumes (
```
silabs-robin, line 28: Note: I am adding assumes. Please review thoroughly. (I think they are reasonable and safe.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -0,0 +1,207 @@
...
+    bus_trans_o.memtype[0] == bus_trans_bufferable
+    // Note: Depends on rest of checking to see that "bus_trans_bufferable is reliable
+  ) else `uvm_error(info_tag, "MPU bufferable flag must corespond to obi memtype[0]");
+
+
+  // DM region overrules PMA configs  (vplan:DebugRange)
+
+  a_dm_region: assert property (
+    core_trans_i.dbg  &&
+    (core_trans_i.addr inside {[DM_REGION_START:DM_REGION_END]})
```
silabs-hfegran, line 110: is DM_REGION_END inclusive?

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -0,0 +1,207 @@
...
+    p_partial_pma_allow (EXC_CAUSE_STORE_FAULT)
+  );
+
+
+  // MPU-accepted transactions must reach OBI  (vplan: not a vplan item)
+
+  property p_eventually_mpu2obi;
+    logic [31:0]  addr;
+    (bus_trans_valid_o && bus_trans_ready_i)  ##0
+    (1, addr = bus_trans_o.addr)
```
silabs-hfegran, line 177: This may lead to unnecessary sampling, as ##0 (1, ... ) might always be sampled (tool dependent), then discarded if not matching forked stream.  (And will most likely generate warning about false positives in formal verification tools)

Safer to use
```
(bus_trans_valid_o && bus_trans_ready_i,
 addr = bus_trans_o.addr)
 ```

File: cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv (hunk 3 above)
Comment (silabs-hfegran) [inline cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv:199]: Missing message?

File: cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv [hunk 9]
```diff
@@ -179,6 +186,48 @@ module uvmt_cv32e40s_rvfi_assert
...
+  end
+
+  always_comb begin
+    writebuf_req_count_n = writebuf_req_count_c;
+    if (writebuf_valid_i && writebuf_ready_o) begin
+      writebuf_req_count_n = writebuf_req_count_c + 1;
+    end
+
+    rvfi_mem_new = 0;
+    for (int i = 0; i < 16/*NMEM*/; i++) begin
```
silabs-hfegran, line 223: NMEM means? use constant/define instead?

silabs-hfegran (changes requested):
Mostly code-readability/maintainability issues and some questions, please se my other comments above

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
silabs-robin: I think that 1) the support interface is so famous now that it comes to mind easily, 2) it isn't even necessary to understand what it stands for because all the usage of it calls upon the more descriptive and understandable names within it, and 3) long prefixes makes the already-difficult-to-read properties even more difficult to read.

On point 3, this is similar to `import numpy as np` which is very common. Even a very simple expression like `a && !(b || c)` has a clearer logic than `my_interface_name.descriptive_name_a && !(my_interface_name.descriptive_name_b || my_interface_name.descriptive_name_c)`.

I can change the code if that is what the majority of people prefer.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
silabs-robin: It doesn't say so explicitly in the UM, AFAICT.
But the default values suggest that they are, because they are defined like this:

```
    .DM_REGION_START          (     32'hF0000000 ),
    .DM_REGION_END            (     32'hF0003FFF ),
```

@Silabs-ArjanB is DM_REGION_END inclusive?

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
silabs-hfegran: The problem is not in its instantiation where you can figure out what the port is based on its type. 
When you read the code where it is used, it would be much more readable for someone not intimately familiar with our source code and avoid having to cross reference to type then figure out its use. A lot of code is not read top to bottom, but rather start at the point of seeing some issue.
Reading the source code and seeing "sup" means I have to find instantiation, and see what kind of type this is, see that this is a modport in some interface that is called support logic and then deduce that this is the support logic interface.
If the signal was called something like support_if or support_logic_if it would be immediately apparent that this is the support logic interface instance. (Not to mention that our coding guidelines warrant an _if for interface instances and _if_t for interface types)

I had to stop and actually think for a bit to figure out what this actually is, even when just reading the port list.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -300,6 +300,14 @@ function logic is_umode();
           (rvfi_mode == cv32e40s_pkg::PRIV_LVL_U);
 endfunction : is_umode
 
+function logic  is_pma_fault();
+  return  rvfi_valid  &&
+          rvfi_trap.trap  &&
+          rvfi_trap.exception  &&
+          (rvfi_trap.exception_cause == 'h 1)  &&
+          (rvfi_trap.cause_type == 'h 0);
+endfunction : is_pma_fault
```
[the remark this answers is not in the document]
silabs-robin: Fixed.

mk/uvmt/xrun.mk
silabs-robin: Hiding, as reviewers probably have read it already.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
silabs-robin: Hiding, as reviewers probably have read it already.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -0,0 +1,207 @@
+// Copyright 2023 Silicon Labs, Inc.
```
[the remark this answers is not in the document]
silabs-robin: Hiding, as reviewers probably have read it already.

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -0,0 +1,207 @@
...
+    |->
+    !bus_trans_valid_o
+    //TODO:INFO:silas-robin Idea: rvfi-vs-obi sb refmodel prediction
+  ) else `uvm_error(info_tag, "pma must block obi reqs");
+
+
+  // memtype[0] matches bufferable flag  (vplan:InstructionFetches:0, vplan:DataFetches:0)
+
+  a_memtype_bufferable: assert property (
+    bus_trans_o.memtype[0] == bus_trans_bufferable
```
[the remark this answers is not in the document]
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
silabs-robin: These assumes are global and unconditionally included.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
silabs-robin: Got confirmation in chat. The answer is yes, start/end is inclusive.

cv32e40s/tb/uvmt/uvmt_cv32e40s_assumes.sv
silabs-robin: I know that these assumes were necessary before. But I cannot reproduce it (and I've tried a lot).
Either my code got tweaked during development, or something in the environment changed when I pulled the latest from upstream.

In any case, they are not necessary anymore so I have deleted them.

silabs-hfegran (approved):

PR #1688, opened 2023-03-01
Added split array aliases for the rvfi gpr- and mem-masks

silabs-mateilga:
These new signals from the rvfi_instr_if are split up and ordered versions of the `rvfi_gpr_*`- and `rvfi_mem_*` arrays, to help with verification. They are easier to deal with in waveform viewers, and function as better visualized alternatives to the helper functions available.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
@@ -165,7 +165,16 @@ module uvmt_cv32e40s_tb;
                                                                    .rvfi_mem_rdata(rvfi_i.rvfi_mem_rdata[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN*0    +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN]),
                                                                    .rvfi_mem_rmask(rvfi_i.rvfi_mem_rmask[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8*0  +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8]),
                                                                    .rvfi_mem_wdata(rvfi_i.rvfi_mem_wdata[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN*0    +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN]),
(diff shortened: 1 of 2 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -165,7 +165,16 @@ module uvmt_cv32e40s_tb;
                                                                    .rvfi_mem_rdata(rvfi_i.rvfi_mem_rdata[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN*0    +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN]),
                                                                    .rvfi_mem_rmask(rvfi_i.rvfi_mem_rmask[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8*0  +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8]),
                                                                    .rvfi_mem_wdata(rvfi_i.rvfi_mem_wdata[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN*0    +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN]),
-                                                                   .rvfi_mem_wmask(rvfi_i.rvfi_mem_wmask[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8*0  +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8])
+                                                                   .rvfi_mem_wmask(rvfi_i.rvfi_mem_wmask[uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8*0  +:uvma_rvfi_pkg::NMEM*uvme_cv32e40s_pkg::XLEN/8]),
+                                                                   .visualized_gpr_rdata(),
```
silabs-hfegran, line 169: Looks good, I only have one request/suggestion - maybe use something like rvfi_gpr_rdata_array instead of "visualized_" - visualized does not appear to give me any clues that this is an array of gprs, and the rvfi_* naming scheme makes it clear where the signal originates (after all it is technically just an alias for a 4096 bit vector represented in 32-bit blocks)

silabs-mateilga: I'm skeptical of giving it the rvfi_ postfix, as it is not an actual rvfi signal. Maybe @MikeOpenHWGroup can give his opinion?

silabs-mateilga:
Will merge in with #1693 to work around git branch issue

PR #1693, opened 2023-03-02
Debug assert fix

silabs-mateilga:
Fixed an issue in connection of obi interface that showed problems in simulation, and added checks for a feature not covered earlier, debug_pc_o. Updated both asserts and vplan.


silabs-mateilga:
Note that this PR now includes merging the changes from PR #1688, due to git branch issues.

silabs-hfegran (approved):

PR #1696, opened 2023-03-03
Quick start improvements and cv32e40p Makefile fix

MarioOpenHWGroup:
There are two main contributions to this MR:

- Make the user aware that the requirements can be installed using python virtual environments.
- Fix the `cv32e40p` Makefile, which had the RISCV_MARCH argument hardcoded.


cv32e40p/bsp/Makefile
```diff
@@ -3,19 +3,20 @@ RISCV            ?= $(CV_SW_TOOLCHAIN)
 RISCV_EXE_PREFIX ?= $(RISCV)/bin/riscv32-unknown-elf-
 RISCV_GCC = $(RISCV_EXE_PREFIX)gcc
 RISCV_AR = $(RISCV_EXE_PREFIX)ar
+RISCV_MARCH ?= rv32imc
 SRC = crt0.S handlers.S syscalls.c vectors.S
 OBJ = crt0.o handlers.o syscalls.o vectors.o
-LIBCV-VERIF = libcv-verif.a 
-CFLAGS ?= -Os -g -static -mabi=ilp32 -march=rv32imc -Wall -pedantic
+LIBCV-VERIF = libcv-verif.a
+CFLAGS ?= -Os -g -static -mabi=ilp32 -march=$(RISCV_MARCH) -Wall -pedantic
```
pascalgouedo, line 10: Already done a while ago on cv32e40p/dev

MarioOpenHWGroup: Oh, my bad I was using master as a base, I will discard this commit then :+1:

docs/VerifStrat/source/quick_start.rst
```diff
@@ -33,15 +33,32 @@ You will need:
...
+Python virtual environment and install them there::
+
+   $ git clone https://github.com/openhwgroup/core-v-verif.git $CORE_V_VERIF
+   $ cd $CORE_V_VERIF/bin
+   $ pip3 install virtualenv
+   $ python3 -m venv $VENV_NAME
+   $ source $(VENV_NAME)/bin/activate
+   $ pip3 install -r requirements.txt
+
+Before running the simulation the following environment variables
```
MikeOpenHWGroup, line 55: This is technically correct information but is already documented elsewhere.  It is good to follow the "DRY principle" of documentation: `Don't Repeat Yourself`.  This information is already documented in [mk/README](https://github.com/openhwgroup/core-v-verif/tree/master/mk#required-corev-environment-variables), so I would add a bullet item to the "You will need" list above.  Something like:

- Set the required shell environment variables ([link](https://github.com/openhwgroup/core-v-verif/tree/master/mk#required-corev-environment-variables)).

docs/VerifStrat/source/quick_start.rst
```diff
@@ -33,15 +33,32 @@ You will need:
...
 #. `Verilator <https://veripool.org/guide/latest/install.html>`_.
 
-An easy way to get the Python plug-ins installed on your machine is::
+There are two ways to get the Python plug-ins installed. The first one, which installs them in the host machine, is::
 
    $ git clone https://github.com/openhwgroup/core-v-verif.git $CORE_V_VERIF
    $ cd $CORE_V_VERIF/bin
    $ pip3 install -r requirements.txt
 
+The second one, which is more complex and more robust, is to create a
```
MikeOpenHWGroup, line 45: I caution about giving advice like this.  Everyone's compute environment will be different and we already get too many, "how do I setup my tools" questions.

MikeOpenHWGroup (changes requested):
Thanks for this PR @MarioOpenHWGroup!   It is good to have a fresh set of eyes on this.   I've made a couple of comments.

docs/VerifStrat/source/quick_start.rst
MarioOpenHWGroup: I totally agree with this. I like sentence you propose so I will change it👍

docs/VerifStrat/source/quick_start.rst
MarioOpenHWGroup: I agree with this. Although, the virtual python environments guarantee that we do not break other python projects by changing the versions of the python packages.

I will rephrase it, I will suggest it instead of looking that authoritative

MikeOpenHWGroup: Sounds good.

MikeOpenHWGroup (approved):

PR #1699, opened 2023-03-03
AXI agent: avoid read outside arrays

AEzzejjari:


ASintzoff:
Could you provide a better commit description?

Something like `AXI agent: avoid read outside arrays`

PR #1712, opened 2023-03-10
CSV Vplans

silabs-robin:
This PR adds a script to convert vplans from ".xlsx" to ".csv".

To be able to see a diff of vplan changes, it was proposed to save csv files next to the original spreadsheet files.
The reason for the script is that Excel has severe localization quirks that makes it practically impossible to do reliably from their gui.
For europeans defaulting to a ";" delimiter, a simple substitution of `s/,/;/g` is enough to make it workable.
There is also included an example on the UserMode Vplan.

Note: If you click "View file" on the csv, Github actually renders it correctly.

Example diff: https://github.com/silabs-robin/core-v-verif/compare/vplan_csv...silabs-robin:core-v-verif:vplan_csv_demo

### DIFF
--- a/bin/xlsx2csv.py
+++ b/bin/xlsx2csv.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+
+# Copyright 2023 Silicon Labs, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+#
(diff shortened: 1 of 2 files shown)


silabs-robin:
> Can we also add a README-like doc to explain the rationale and usage of this?

@MikeOpenHWGroup I have now added a docstring at the top of the python file.

MikeOpenHWGroup (approved):

PR #1713, opened 2023-03-10
Umode Assert - Add Missing Asserts

silabs-robin:
This PR adds U-mode assertions based on missing vplan link-to-coverage.

_(Will add updated vplan later, depending on this: https://github.com/openhwgroup/core-v-verif/pull/1712 .)_

Test status:
* ci_check - Passes  (except the 2 known bad tests.)
* Fv default - Passes (except slow-converging, and this: https://github.com/openhwgroup/cv32e40s/issues/414)


cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_assert.sv
```diff
@@ -51,9 +51,9 @@ module  uvmt_cv32e40s_pma_assert
   input wire           bus_trans_valid_o,
 
   // Writebuffer Signals
-  input obi_data_req_t  writebuf_trans_i,
-  input obi_data_req_t  writebuf_trans_o,
-  input wire            writebuf_ready_o,
+  input wire obi_data_req_t  writebuf_trans_i,
+  input wire obi_data_req_t  writebuf_trans_o,
```
silabs-robin, line 55: Note: Made one fix to PMA asserts because it gave warnings in sim.

(Will resolve this comment after the first review comment ticks in.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_assert.sv
```diff
@@ -179,14 +179,41 @@ module uvmt_cv32e40s_rvfi_assert
...
+    (rvfi_valid [->1])
+    |->
+    (rvfi_intr.exception == exception)  ||
+    rvfi_intr.interrupt
+    ;
+  endproperty : p_sync_cause
+
+  a_sync_cause: assert property (
+    p_sync_cause
+  ) else `uvm_error(info_tag, "rvfi_intr.exception can't happen unannounced");
```
silabs-robin, line 205: Note: Mostly added new asserts.

(Will resolve this comment after the first review comment ticks in.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -196,11 +203,7 @@ module  uvmt_cv32e40s_umode_assert
 
   reg [1:0]  effective_rvfi_privmode;
   always @(*) begin
```
silabs-robin, line 205: Note: The effective privilege mode was not modeled correctly, should be good now.

(Will resolve this comment after the first review comment ticks in.)

silabs-robin (commented):
Just adding some explanations.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -623,14 +653,23 @@ module  uvmt_cv32e40s_umode_assert
...
     |->
     (rvfi_csr_mstatus_rdata[MPRV_POS+:MPRV_LEN] == 1'b 0)
   ) else `uvm_error(info_tag, "exiting dmode towards umode should clear mprv");
 
+  a_dret_mprv_csr: assert property (
+    rvfi_if.is_dret()  &&
+    (rvfi_csr_dcsr_rdata[PRV_POS+:PRV_LEN] == MODE_U)  &&
+    rvfi_if.rvfi_dbg_mode
+    |->
+    (rvfi_csr_mstatus_wmask[MPRV_POS+:MPRV_LEN] == 1'b 1)  &&
```
silabs-hfegran, line 669: While I like the consistency of POS+:LEN, in this case this led me to believe that there might have been a bug here since there is a range specified on the left side and 1-bit signal rvalue.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -669,14 +708,14 @@ module  uvmt_cv32e40s_umode_assert
 
   property p_mret_from_mpp (int mode);
     is_rvfi_mret  &&
-    !rvfi_trap  &&
-    (rvfi_csr_mstatus_rdata[MPP_POS+:MPP_LEN] == mode)  &&
     (rvfi_mode == MODE_M)  &&
+    (rvfi_csr_mstatus_rdata[MPP_POS+:MPP_LEN] == mode)  &&
+    !rvfi_trap  &&
     !rvfi_dbg_mode
```
silabs-hfegran, line 714: is the behavior in debug mode covered somewhere else? (debug asserts?)

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -554,7 +584,7 @@ module  uvmt_cv32e40s_umode_assert
   a_ecall_umode_poststate: assert property (
     (is_rvfi_ecall && (rvfi_mode == MODE_U))
     |->
-    (mcause_poststate inside {EXC_CAUSE_ECALL_UMODE, 1, 2, 24, 25})  // ("It" or higher)
+    (mcause_poststate[11:0] inside {EXC_CAUSE_ECALL_UMODE, 1, 2, 24, 25})  // ("It" or higher)
```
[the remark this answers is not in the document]
silabs-hfegran: Perhaps add interrupt bit-check here as well to avoid interrupts with matching id falsely passing this assertion?

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -701,15 +740,17 @@ module  uvmt_cv32e40s_umode_assert
 
   a_mret_umode_nextmode: assert property (
```
silabs-hfegran, line 741: should we also check for trap here? This could pass if the instruction actually executes and changes mode to M-mode

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -718,8 +759,9 @@ module  uvmt_cv32e40s_umode_assert
 
   a_mret_umode_mprv: assert property (
```
silabs-hfegran, line 760: This sounds like an illegal instruction-exception, am I missing something here?

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -885,41 +927,96 @@ module  uvmt_cv32e40s_umode_assert
...
   a_prot_iside_legal: assert property (
     obi_iside_prot  inside  {3'b 000, 3'b 110}
   ) else `uvm_error(info_tag, "the prot on fetch must be legal");
 
+
+  // vplan:DataProt
+
+  a_data_prot: assert property (
+    rvfi_valid  &&
+    (|rvfi_if.rvfi_mem_rmask || |rvfi_if.rvfi_mem_wmask)
```
silabs-hfegran, line 957: would skipping the or-reduction here increase readability? nonzero variable == bool(true), or-reduce nonzero variable == bool(true)

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -885,41 +927,96 @@ module  uvmt_cv32e40s_umode_assert
...
+
   a_prot_dside_legal: assert property (
     obi_dside_prot  inside  {3'b 001, 3'b 111}
   ) else `uvm_error(info_tag, "the prot on loadstore must be legal");
 
+  wire logic [NMEM-1:0]  data_prot_equals;
+  wire logic [NMEM-1:0]  mem_act;
+  for (genvar i = 0; i < NMEM; i++) begin: gen_data_prot_equals
+    assign  data_prot_equals[i] = (rvfi_if.mem_prot[i*3+:3] == rvfi_if.mem_prot[2:0]);
+    assign  mem_act[i]          = |rvfi_if.check_mem_act(i);
```
silabs-robin, line 977: Use always_comb? Check for Xes in sim, ref tinfo problem.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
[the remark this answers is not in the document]
silabs-robin: The assertions above it check that it must be an exception.
But I've now also added `a_exc_mcause` to check that exception implies clearing mcause[31].

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: After thinking about it a bit, doesn't it serve as an additional check that the LEN is defined correctly? Because then the width of the literal must match and it would otherwise give a warning?

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: If the mret could ever go to U-mode then there would be a CEX, at least. But the assertion above `a_mret_umode_exception` also checks that there is an illegal instruction exception (or a higher priority exception).

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: It should be an illegal instruction exception.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: Fixed

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-robin: If there is a problem here, then it should be detectable if the two covers `cov_data_prot_equal_memact_*` are not be met in sim.

cv32e40s/docs/VerifPlans/Simulation/privileged_spec/CV32E40S_UserMode.csv
silabs-robin, line 1: I realized the CSR vplan PR got merged, so I uploaded the updated U-mode vplan as well.

cv32e40s/tb/uvmt/uvmt_cv32e40s_umode_assert.sv
silabs-hfegran: As I understand it, if the instruction traps, it should not have any side effects (e.g. not clear mprv)

silabs-hfegran (approved):

PR #1715, opened 2023-03-10
cv_regress script to generate regressions for Siemens verification manager tool

XavierAubert:
Added the needed templates to allow the generation of a RMDB File. 

Due to limitations of Questa Vrun, make sure to generate the rmdb file with coverage enabled, as the regression tool is not able to easily detect a failing make command as a test fail. It needs indeed to have access to an .ucdb file with the test status recorded inside. 

Also, Siemens tools expects one unique "testname" when merging ucdb, so tests that are executed with multiple configurations and seeds provoked an issue when merging. I was forced to update some coverage commands and script to solve this. 

I also updated one small thing to cv_regress script, to allow the same test to be executed in multiple configurations inside the same regression. That will be useful for handling all F/Zfinx tests that will be practically the same inside one unique regression.

### DIFF
--- a/bin/cv_regress
+++ b/bin/cv_regress
@@ -155,7 +155,7 @@ def read_file(args, file):
         # Determine if a test is valid, skip for compliance tests
         # Since it is not possible to determine apriori if a compliance test is valid
         if not 'compliance' in test.cmd:
-            check_valid_test(args.project, test.name)
+            check_valid_test(args.project, test.testname)
 
         # Determine if a test is indexed for setting test iterations
         if args.num:
@@ -200,6 +200,7 @@ parser.add_argument('--make', help='Substitute for make command (to use short-cu
 parser.add_argument('--makearg', action='append', help='Arguments to supply to each make command, can be specified multiple times')
 parser.add_argument('--sh', help='Select bash shell script output', action='store_true')
 parser.add_argument('--vsif', help='Select Vmanager VSIF output', action='store_true')
+parser.add_argument('--rmdb', help='Select Vrun RMDB output', action='store_true')
 parser.add_argument('--toolchain', help='Select toolchain to build with', choices=VALID_TOOLCHAINS, default=DEFAULT_TOOLCHAIN)
 args = parser.parse_args()
 
@@ -220,6 +221,8 @@ if not args.outfile:
         args.outfile = args.outfile + '.sh'
     elif args.vsif:
         args.outfile = args.outfile + '.vsif'
+    elif args.rmdb:
+        args.outfile = args.outfile + '.rmdb'
 
 regressions = []
 unique_builds = OrderedDict()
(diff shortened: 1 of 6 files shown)


bin/cv_regress
```diff
@@ -155,7 +155,7 @@ def read_file(args, file):
         # Determine if a test is valid, skip for compliance tests
         # Since it is not possible to determine apriori if a compliance test is valid
         if not 'compliance' in test.cmd:
-            check_valid_test(args.project, test.name)
+            check_valid_test(args.project, test.testname)
```
MikeOpenHWGroup, line 158: This _looks_ like it could break `cv_regress` for other simulators.

bin/templates/regress_rmdb.j2
```diff
@@ -0,0 +1,94 @@
+{% import 'regress_macros.j2' as regress_macros -%}
```
MikeOpenHWGroup, line 1: This looks like a tool-generated file, is that correct?  Can we place an SPDX header on this file?

XavierAubert: I don't totally understand your question. Are you talking about the whole *_rmdb.j2 file ? of only the _macros.j2 file and the import line ? 

If it's for the whole rmdb file, it's not, I wrote this by hand to manage regressions with Siemens tools, but yes I forgot to add a header on this one. 

If you're talking about the import line, it was my first time using Jinja2 templating, and I simply get inspired (a lot) by the already existing regress_vsif.j2 file, so I simply copied the imported line, and so the file, which is also missing a header as I can see.

bin/cv_regress
XavierAubert: Yes, but it only _looks_, don't be affraid. 


I also updated in consequence the constructor of the class test inside the file lib/cv_regression.py :

```
class Test:
        ...
        if not hasattr(self, 'testname'):
            self.testname = self.name
        ...
```

By doing so, the testname field is filled with the YAML top item, and backward compatibility is ensured 😄 

Now, a little explanation and more context on why it was necessary. I wanted to be able to run within the same regression a unique test with different core configuration. One practical example is for floating point tests, as we are going to have multiple values for latencies, but the same testcode underlying. With the already existing cv_regress YAML architecture, it was not possible because it would imply having twice the same entry with the same name, and would look like this : 

```
builds:
  uvmt_cv32e40p:
    cmd: make comp
    dir: cv32e40p/sim/uvmt

  uvmt_cv32e40p_pulp:
    ... (same as above) ... 
    cfg: pulp

tests:
  hello-world:    <- issue here
    ...
    build: uvmt_cv32e40p
    cmd: make test COREV=YES TEST=hello-world
    ...

  hello-world:    <- issue here
    ... 
    build: uvmt_cv32e40p_pulp
    cmd: make test COREV=YES TEST=hello-world
```

As you can see inside the above snippet, the already existing cv_regress function you mentionned **forces** to have the "real" test name as the YAML top item as it raises an error if not, and this leads to obvious issues while parsing the YAML file.  

So I added an optionnal field "testname" to the YAML parsing so that I am able to do what I wanted, for example like this : 


```
tests:
  hello-world:    
    ...
    build: uvmt_cv32e40p
    cmd: make test COREV=YES TEST=hello-world
    ...
    ^  ^  ^  
    |  |  | 
    same as before, untouched, without testname

  hello-world-pulp:            <- issue solved
    ... 
    build: uvmt_cv32e40p_pulp
    cmd: make test COREV=YES TEST=hello-world
    testname: hello-world      <- issue solved
```

bin/templates/regress_rmdb.j2
MikeOpenHWGroup: Thanks.  I was asking about whether the file was generated or not because we have had problems in the past with EDA tools generating files based on closed source artifacts.   If you have generated this manually, we should be fine.

Please do add a license header.  A two-liner will suffice:
```
// Copyright 2023 Dolphin Design
// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```

XavierAubert:
```suggestion
// Copyright 2023 Dolphin Design
// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1

{% import 'regress_macros.j2' as regress_macros -%}
```

MikeOpenHWGroup (approved):

PR #1721, opened 2023-03-14
CVXIF : Update CVXIF agent

AyoubJalali:


AyoubJalali:
Hello,
So i made some changes in the **cvxif** agent related to decode custom instructions, also I removed two other instructions, just to be aligned with @zchamski on one list of custom instructions, also I fixed some bugs in the agent like getting rs_valid info from the request, finally i add more options or we can say possibility to the agent if he got some invalid value of the **rs_valid**, sould raise an exception.

MikeOpenHWGroup:
I have a dumb question: where is the defintion of these custom instructions coming from?  It is compliant with the ISA? Do we need to update the CVA6 User Manual?

Silabs-ArjanB:
> I have a dumb question: where is the defintion of these custom instructions coming from? It is compliant with the ISA? Do we need to update the CVA6 User Manual?

Good question. Shouldn't these files be 'neutral' XIF agents that can be used by all projects using an extension interface as opposed to adding custom instructions in them?

AyoubJalali:
> Good question. Shouldn't these files be 'neutral' XIF agents that can be used by all projects using an extension interface as opposed to adding custom instructions in them?

For now there is a README in the cvxif agent repo, describes all cus instructions supported by the agent for now, but it should be add a section in the user guide to add these instructions.

AyoubJalali:
> I have a dumb question: where is the defintion of these custom instructions coming from? It is compliant with the ISA? Do we need to update the CVA6 User Manual?

I think it a good idea to add what the cvxif agent in the cvxif user guide section

AyoubJalali:
@MikeOpenHWGroup, @Silabs-ArjanB, I think this is the only common place, where the cus instructions was describe https://github.com/openhwgroup/core-v-verif/blob/master/lib/uvm_agents/uvma_cvxif/README.md

MikeOpenHWGroup:
@AyoubJalali, unless this Agent is going to be CVA6 specific, it needs to be architected such that:
1. Any behaviour in the Agent's components (driver, monitor, sequencer, sequences and coverage) must be based solely on the ISA.
2. Any behaviour related to a specific core (e.g. CVA6) must be implemented in either a core-specific extension or (preferred) a core-specific virtual sequence.

This agent lives in the `lib` directory so it is required to be generic across all cores that uses the CV-X-IF.  If it is to be CVA6-specific, then it needs to move to the `cva6/lib` directory.

My very strong perference is to implement this agent to be non-core-specific and put all core-specific extensions in a core-specific directory.  There are some good examples of this with the `OBI Memory Agent`.  The Agent itself is [here](https://github.com/openhwgroup/core-v-verif/tree/master/lib/uvm_agents/uvma_obi_memory) and you can find cv32e40s-specific extensions (for both components and sequences) [here](https://github.com/openhwgroup/core-v-verif/tree/master/cv32e40s/env/uvme).

MikeOpenHWGroup:
> I think this is the only common place, where the cus instructions was describe

Sorry @AyoubJalali, I have not been as alert about this situation as I should have.  The instruction decoding documented in the README should not be done in the Agent.  It needs to be moved to a virtual sequence.

AyoubJalali:
@MikeOpenHWGroup Yes I think you are right, because these instructions are here just for the agent to challenge the cvxif protocol, so I don't think this is a common for all cores

AyoubJalali:
Hello, I'm going to close this PR and creating a new one to avoid conflicts

PR #1722, opened 2023-03-14
Convert Vplans to CSV

silabs-robin:
This PR adds ".csv" variants of all 40s vplans (using `bin/xlsx2csv.py`), so that we in the future can see diffs when we update the vplans.

I did a spot check on the debug vplan, opening the csv files as a spreadsheet, and the cells are populated just like in the original xlsx file, except that "merged cells" are replaced with plain and simple indentation.


silabs-robin:
Example of what the CSV files look like:

Before:
image

After:
image

silabs-robin:
> as long as the csv variant is used only for reviews

That's the idea yeah. That the CSVs will be like a supplement - one works on the xlsx then auto-update the csv so the diff becomes available.

It will not only easy reviews, but also serve as a safety mechanism. I have myself mistakenly worked on a local copy of a vplan, then realized that I didn't start on the latest edition of the vplan, and had to manually port my updates into the latest. There is potential for accidentally deleting or otherwise disrupting things with a binary-only process.

silabs-hfegran (approved):
Looks good, thanks @silabs-robin

PR #1725, opened 2023-03-17
Fixed interrupt_test, removed from ci_check due to excessive run time

silabs-hfegran:
In the past this test most likely passed without running advertised functionality, and a change of compiler exposed the (many) latent problems within. Fixed by adding volatiles to variables to prevent the compiler from optimizing away central parts of the test. As the full test takes approximately 1.5 hours it has been removed from the ci_check regression list.

Note: There will be unknown exception handler entered messages printed during the last part of the test - this is due to common BSP code that is not overridden in the test.


silabs-robin (approved):
Approving based on:
Agree that the runtime is too long for ci_check.
The `volatile`/`__asm__`/`naked` fixes seem like a good way to rein in the compiler's zealous opportunism.

PR #1729, opened 2023-03-18
UVM test termination: Use end-of-test output of RVFI tracer iso. $finish.

zchamski:
This is a companion PR to cva6 PR https://github.com/openhwgroup/cva6/pull/1125.  It propagates the end-of-test information to the UVM testbench and adds a basic interpretation of the test exit code.

Note: this PR also reduces the severity of the absence of fetch toggle in CVA6 from ERROR to WARNING.  Using ERROR severity resulted in non-zero error count for the simulations, causing successful tests to be reported as failures.


cva6/tb/uvmt/uvmt_cva6_tb.sv
```diff
@@ -67,7 +67,8 @@ module uvmt_cva6_tb;
                                            );
    // DUT Wrapper Interfaces
    uvmt_rvfi_if                     rvfi_if(
-                                                 .rvfi_o()
+                                                 .rvfi_o(),
+                                                 .tb_exit_o()
```
MikeOpenHWGroup, line 71: Why is `tb_exit_o` added to the RVFI?  This is not a signal related to the RVFI Tracer.

cva6/tb/uvmt/uvmt_cva6_tb_ifs.sv
```diff
@@ -22,7 +22,8 @@
 
 
 interface uvmt_rvfi_if (
-                    output ariane_rvfi_pkg::rvfi_port_t  rvfi_o
+                    output ariane_rvfi_pkg::rvfi_port_t  rvfi_o,
+                    output logic[31:0] tb_exit_o
```
MikeOpenHWGroup, line 26: Same comment:  tb_exit_o is not a port of the RVFI.

cva6/tests/uvmt/compliance-tests/uvmt_cva6_firmware_test.sv
```diff
@@ -144,13 +144,20 @@ task uvmt_cva6_firmware_test_c::run_phase(uvm_phase phase);
    @(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
    repeat (33) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
    `uvm_info("TEST", "Started RUN", UVM_NONE)
-   // The firmware is expected to write exit status and pass/fail indication to the Virtual Peripheral
+   // The firmware is expected to write exit status and pass/fail indication to the Virtual Peripheral.
+   // The format of rvfi_vif.tb_exit_o is { wire[31:1] exit_code, wire test_finished }.
    wait (
-          (rvfi_vif.rvfi_o[0].halt       == 1'b1) ||
-          (rvfi_vif.rvfi_o[1].halt       == 1'b1)
+          (rvfi_vif.tb_exit_o[0]     == 1'b1)
```
MikeOpenHWGroup, line 150: Same comment,.

cva6/tests/uvmt/compliance-tests/uvmt_cva6_firmware_test.sv
```diff
@@ -144,13 +144,20 @@ task uvmt_cva6_firmware_test_c::run_phase(uvm_phase phase);
...
-   repeat (100) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
+   `uvm_info("TEST", "Test FINISHED", UVM_NONE)
+   // Set sim_finished (otherwise tb will flag that sim was aborted)
+   uvm_config_db#(bit)::set(null, "", "sim_finished", 1);
+   uvm_config_db#(int)::set(null, "", "test_exit_code", { 0'b0, rvfi_vif.tb_exit_o[31:1] });
+   // Let the termination-triggering instruction appear in the log.
+   @(posedge env_cntxt.clknrst_cntxt.vif.clk);
+   // Allow termination from now on.
    phase.drop_objection(this);
+   repeat (100) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
```
MikeOpenHWGroup, line 160: If you want this 100-cycle delay to gate the termination of the simulation, it should be placed _before_ dopping the objection.  By dropping its objection a UVM component is signalling to the UVM environment that the simulation can be terminated at any time.

MikeOpenHWGroup (changes requested):
Please see my comments related to the core_cntrl agent and the tb_exit signal.

cva6/tb/uvmt/uvmt_cva6_tb.sv
zchamski: This core-v-verif PR has a matching PR on the CVA6 side (https://github.com/openhwgroup/cva6/pull/1125) which replaces the `$finish` calls inside the RVFI tracer with an end-of-test signal forwarded to the TB.

cva6/tb/uvmt/uvmt_cva6_tb_ifs.sv
zchamski: See above: the corresponding signal was added to the RVFI in https://github.com/openhwgroup/cva6/pull/1125.

cva6/tests/uvmt/compliance-tests/uvmt_cva6_firmware_test.sv
zchamski: See above.

cva6/tests/uvmt/compliance-tests/uvmt_cva6_firmware_test.sv
MikeOpenHWGroup: If there needs to be a delay, then it should be before dropping the objection.  Once the objection has been dropped, this component is giving permission to the UVM run-flow to terminate the simulation immediately.  (Note that other components may still have their objections raised, so the simulation may not terminate here.)

cva6/tests/uvmt/compliance-tests/uvmt_cva6_firmware_test.sv
zchamski: @MikeOpenHWGroup , alternatively we could think of extending the definition of the `halt` signal in th RVFI interface, but the consequences of this change are unclear at this point.  What's your opinion in this?

MikeOpenHWGroup: The tracer should not signal end-of-test - that is not its purpose.  Rather, the `test-program` should indicate to the testbench that it is ready for the simulation to end.  Both the RISC-V compliance testsuite and the E40 testbenches do that by writing a specific code to a specific address (RISCV compliance calls it `tohost` and the E40 testbenches call it the [virtual peripheral status flags](https://docs.openhwgroup.org/projects/core-v-verif/en/latest/test_programs.html#virtual-peripherals)).

zchamski: Since the combined core-v-verif PR https://github.com/openhwgroup/core-v-verif/pull/1689  / CVA6 PR https://github.com/openhwgroup/cva6/pull/1075 the tracer tracks the writes to `tohost` performed by test programs, and interprets them in the RISC-V compliance / Spike / RISC-V HTIF way (bit 0 set == end of test, bits [31:1] == test exit code).  Before that, it terminated the simulation upon encountering an ECALL instruction.

Since the tracer is the only observer available in non-UVM settings, so it appears to be the best location to detect the "ready for the simulation to end" condition in a way that is common to all RTL simulations setups.

cva6/env/uvme/uvma_cva6_core_cntrl_agent.sv
```diff
@@ -107,7 +107,7 @@ endtask : run_phase
 
 task uvma_cva6_core_cntrl_agent_c::start_fetch_toggle_seq();
 
-      `uvm_error(log_tag, "fetch toggle not supported in CVA6");
+      `uvm_warning(log_tag, "fetch toggle not supported in CVA6");
```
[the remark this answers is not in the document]
AyoubJalali: @MikeOpenHWGroup, can't we just replace it with `uvm_info or we can let it empty, because this uvm_error cause problems in the end_of_simulation.

PR #1733, opened 2023-03-21
Added script to generate c header from yaml and generated header

silabs-hfegran:


bin/gen_c_header.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python3
+
+################################################################################
+#
+# Copyright 2020 OpenHW Group
+# Copyright 2020 Silicon Labs, Inc.
```
silabs-robin, line 6: Should be 2022 or 2023.

And I think @MikeOpenHWGroup said before that the openhw group copyright line isn't really necessary?

cv32e40s/bsp/csr.h
```diff
@@ -0,0 +1,955 @@
+/*
```
silabs-robin, line 1: Really nice! I like this new bsp header.

silabs-robin (approved):
There might be some license header issues, but I'm approving ahead of time since the files otherwise look good, assuming that the license issues will be resolved.

cv32e40s/bsp/csr.h
```diff
@@ -0,0 +1,955 @@
...
+#define MCAUSE           0x342
+#define MTVAL            0x343
+#define MIP              0x344
+#define MNXTI            0x345
+#define MINTSTATUS       0xFB1
+#define MINTTHRESH       0x347
+#define MSCRATCHCSW      0x348
+#define MSCRATCHCSWL     0x349
+#define TSELECT          0x7A0
+#define TDATA1           0x7A1
```
silabs-robin, line 207: What can we do about tdata1? It doesn't seem to have a struct.

PR #1736, opened 2023-03-23
Updated ci_check script to take yaml cfgs into account

silabs-hfegran:
ci_check only ran with the default configuration, and did not take yaml-specified builds into account. 
This should be working properly with this commit

- Caveat/TODO: This commit will build all the configurations specified in the metrics yaml, some of which will fail. Scripts should be updated to only build the necessary ones. The failing non-related builds have no functional impact on ci_check results.

### DIFF
--- a/cv32e40s/regress/cv32e40s_ci_check.yaml
+++ b/cv32e40s/regress/cv32e40s_ci_check.yaml
@@ -5,18 +5,22 @@ description: Commit sanity for the cv32e40s
 builds:
   clone_riscv-dv:
     cmd: make clone_riscv-dv
+    cfg: default
     dir: cv32e40s/sim/uvmt
 
   clone_svlib:
     cmd: make clone_svlib
+    cfg: default
     dir: cv32e40s/sim/uvmt
 
   clone_cv_core_rtl:
     cmd: make clone_cv_core_rtl
+    cfg: default
     dir: cv32e40s/sim/uvmt
 
   uvmt_cv32e40s:
     cmd: make comp_corev-dv comp
+    cfg: default
     dir: cv32e40s/sim/uvmt
 
   uvmt_cv32e40s_clic:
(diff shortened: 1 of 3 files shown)


silabs-hfegran:
@MikeOpenHWGroup, it would be good if you could have a look at this. 

Ideally we should deprecate the metrics-files as that simulator is no longer supported - I recon this script could be much cleaner if it actually parsed the relevant files directly, rather than depending on the metrics files.

silabs-hfegran:
This breaks cv_regress, will have to have another look before merging - I have an idea on how to fix

silabs-hfegran:
> > Ideally we should deprecate the metrics-files as that simulator is no longer supported
> 
> There is nothing "Metrics specific" in the `.metrics.json` file beyond the name. It is merely a JSON description of a regression test-list. We could translate this to another JSON schema, or a YAML schema or <_insert your favourite language-independent data format here_>. I hate them all equally and do not see that maintaining the JSON schema and the script to process it is any better or worse than a YAML schema and associated script.

Agree that these should be deprecated - this is a temporary fix to get Imperas' ci_check back up, we are not using this script at silabs (we only depend on the cv_regress script which works fine)

cv32e40s/regress/cv32e40s_ci_check.yaml
```diff
@@ -36,71 +40,64 @@ builds:
 
 tests:
   hello-world:
-    build: uvmt_cv32e40s
+    builds: [ uvmt_cv32e40s ]
```
silabs-robin, line 43: Does our scripts support both `build` and `builds`?

silabs-robin (approved):
Approving based on:
Yaml file changes make sense.
I presume that the metrics json and ci check python changes have been tested and are working.

cv32e40s/regress/cv32e40s_ci_check.yaml
silabs-hfegran: Internal scripts work with both, but my changes to cv_regress requires the use of builds.

PR #1690, opened 2023-03-02
Debug triggers assertions

silabs-krdosvik:
Trigger debug assertions.

Added an extra file to describe what the assertions covers.

The changes in the xsecure file is not important as this will be fixedin the upcomming xsecure PR

PR #1737, opened 2023-03-23
Add missing pmp-configuration to default run config

silabs-hfegran:
Default config has 0 enabled PMP regions, commits adds write-masks for the pmp-registers as these are present but WARL zero.

After this modification there should be no known errors in the ci_check tests.


cv32e40s/tests/cfg/default.yaml
```diff
@@ -9,6 +9,70 @@ ovpsim: >
     --override cpu/hpmcounter_undefined=T
     --override cpu/PMP_registers=64
     --override cpu/PMP_grain=0
+    --override cpu/mask_pmpaddr0=0x00000000
```
silabs-robin, line 12: Is there really no PMP_NUM_REGIONS equivalent that could simplify this commit?

silabs-robin (approved):
Approving based on:
I read the rationale and explanation in #1690.
While I think these yaml configs should be simpler and that the SV config should handle this, I agree that this is an ok quickfix.

cv32e40s/tests/cfg/default.yaml
silabs-hfegran: Negative, this has to be specified, either here or in sv - adding it here as a quickfix

PR #1750, opened 2023-03-28
Enable coverage collection flags

silabs-hfegran:
Enable coverage collection flags for
type parameterized modules and interfaces
coverage expressions with `inside` operator


silabs-robin (approved):
Approving based on:
Discussion in chat.

PR #1752, opened 2023-03-28
CVXIF : Moving the cvxif sequence from the lib to cva6/lib/

AyoubJalali:
Hello @MikeOpenHWGroup, this PR was created after our discuss in the PR #1721, I moved the cvxif sequence from the common code lib/


lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_pkg.flist
```diff
@@ -14,6 +14,7 @@
 +incdir+${DV_UVMA_CVXIF_PATH}/src/comps
 +incdir+${DV_UVMA_CVXIF_PATH}/src/obj
 +incdir+${DV_UVMA_CVXIF_PATH}/src/seq
++incdir+${CORE_V_VERIF}/${CV_CORE_LC}/lib/cvxif_seq
```
MikeOpenHWGroup, line 17: This creates a dependency on the directory structure of all environments in core-v-verif.  This line should be removed.

Access to core-environment-specific virtual sequences should be done in `<core>/env/uvme/uvme_<core>_pkg.flist`. For example, [cv32e40p/env/uvme/uvme_cv32e40p_pkg.flist](https://github.com/openhwgroup/core-v-verif/blob/760ecd3057d0de3998407ebd1e9cced9b3e7b074/cv32e40p/env/uvme/uvme_cv32e40p_pkg.flist#L1) has an incdir to a `vseq` subdirectory and this is where the virtual sequences for the e40p environment are.

A couple of things to point out:
1. The e40p environment puts _all_ virtual sequences for all its Agents into `cv32e40p/env/uvme/vseq`  You do not need to do it that way.  For example, you could have `cv32e40p/env/uvme/cvxif_vseq`, `cv32e40p/env/uvme/axi_vseq`, etc.
2. The sequences we are discussing here are _technically_ referred to as "virtual sequences".  (OK, now I'm just getting picky!).

MikeOpenHWGroup (changes requested):

lib/uvm_agents/uvma_cvxif/src/uvma_cvxif_pkg.flist
AyoubJalali: so if I created this cv32e40p/env/uvme/cvxif_vseq, should i add an incdir in the agent to locate this sequence i think

AyoubJalali: Ok this should be good, so should i add a "v" for virtual sequence

AyoubJalali: @MikeOpenHWGroup so i moved the sequence (vsequence I add the 'v') to the cva6 env, also I change the sequencer to vsequencer because it drive a vsequencer.

So could you have a look Please

PR #1753, opened 2023-03-28
PMA Coverage

silabs-robin:
This PR adds coverage definitions for PMA.

Test status:
* ci_check - Passes (except known "clic" test rvfi_intr.exception bug)
* Fv default - Converging pma covers/asserts all pass
* Fv altparams/pmp_pma* - Same as for fv default, but run for a shorter time


cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,401 @@
...
+    $past(rvfi_if.is_pma_fault,    , ,@(posedge clk_rvfi));
+  always_comb  was_rvfi_mem_wmask =
+    occured_rvfi_valid  &&
+    $past(|rvfi_if.rvfi_mem_wmask, , ,@(posedge clk_rvfi));
+  always_comb  was_rvfi_mem_act =
+    occured_rvfi_valid  &&
+    $past(rvfi_if.is_mem_act,      , ,@(posedge clk_rvfi));
+  always_comb  was_rvfi_pmamain_low =
+    occured_rvfi_valid  &&
+    $past(rvfi_pmamain_lowhigh[1], , ,@(posedge clk_rvfi));
```
silabs-robin, line 115: To illustrate the intention of these `$past` signals and the `clk_rvfi`:

image

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,401 @@
...
+  wire logic  is_mpu_activated;
+  assign  is_mpu_activated = (core_trans_ready_o && core_trans_valid_i);
+
+
+  // Helper Logic - RVFI Clock & Valid
+
+  var logic   was_rvfi_valid;
+  wire logic  clk_rvfi;
+  always_ff @(posedge clk_ungated) was_rvfi_valid <= rvfi_if.rvfi_valid;
+  assign  clk_rvfi = clk & was_rvfi_valid;
```
silabs-mateilga, line 77: As this relies on a clock not present in the rvfi_if, I don't think it belongs there.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
silabs-robin: From video call:

```
always_ff @(posedge clk_ungated) begin
  if (rvfi_valid) begin
    was_rvfi_mysignal <= rvfi_mysignal;
    ...
  end
end
```

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,401 @@
...
+
+    //TODO:INFO:silabs-robin more crosses are possible, but bordering on impractical/infeasible
+  endgroup
+
+  cg_mpu  mpu_cg = new;
+
+
+  // RVFI Coverage Definition
+
+  covergroup cg_rvfi @(posedge clk);
```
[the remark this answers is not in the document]
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
silabs-robin: Fixed.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -246,46 +274,46 @@ interface uvma_rvfi_instr_if
   //       valid values, indicated by rvfi_valid == 1
 
   // Check if instruction is of a certain type
-  // Usage: ref_mask sets the parts of the instruction you want to compare,
-  //        ref_instr is the reference to match
-  function logic match_instr( bit [ DEFAULT_XLEN-1:0] ref_instr,
-                              bit [ DEFAULT_XLEN-1:0] ref_mask
+  // Usage: instr_mask sets the parts of the instruction you want to compare,
+  //        instr_ref is the reference to match
+  function logic match_instr( bit [ DEFAULT_XLEN-1:0] instr_ref,
```
silabs-krdosvik, line 279: The change of name is ok for me

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -458,6 +490,22 @@ function logic is_ecall_f();
...
+
+function automatic logic is_tablejump_f();
+  return  match_instr_raw(INSTR_OPCODE_TABLEJUMP, INSTR_MASK_TABLEJUMP);
+endfunction : is_tablejump_f
+
+function automatic logic is_fencefencei_f();
+  return  match_instr(INSTR_OPCODE_FENCE,  INSTR_MASK_FENCE)  ||
+          match_instr(INSTR_OPCODE_FENCEI, INSTR_MASK_FENCEI);
+endfunction : is_fencefencei_f
+
```
silabs-krdosvik, line 508: I assume the functions is_pushpop_f, is_tablejump_f, and is_fencefencei_f work as they should, as I am unfamiliar with the match_instr and match_instr_raw functions. (I guess this is feedback work for Marton.)

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -501,15 +549,47 @@ function logic is_pma_instr_fault_f();
...
+      return  addr;
+  endcase
+endfunction : rvfi_mem_addr_word0highbyte_f
+
+function automatic logic is_split_datatrans_f();
+  logic [31:0]  low_addr  = rvfi_mem_addr;
+  logic [31:0]  high_addr = rvfi_mem_addr_word0highbyte;
+  return  is_mem_act && (low_addr[31:2] != high_addr[31:2]);
+endfunction : is_split_datatrans_f
+
```
silabs-krdosvik, line 590: Here I might be mistaken, but:

I believe you try check if the data fetched from OBI (bus) is unaligned, or not.
(f.eks. checking if we have fetched the aligned data word [b0,b1,b2,b3], or the unaligned data word [b1,b2,b3,b4], which will have a low_addr[31:2] != high_addr[31:2]) 

I know that this is not allowed on the OBI instruction bus (and that the alignment buffer fix the alignment after OBI fetches). I am unsure if the same goes for fetching data from the OBI data bus though. (I dont know anything about unaligned OBI data fetches, but for me it sounds impractical, so that is why I am sceptical)

If it is allowed: this code should definitively be included here.

If it is not allowed (and you only use this to check if it is not allowed in your assertion set) I think this function might be too specific for your pma asserts (but I guess it is okey to just include it anyways).

But in general: a name suggustion: change is_split_datatrans_f to is_unaligned_data_fetch, (reason: I was confused  about what was meant by is_split_datatrans_f. my first thought was that it was a memory transaction split into several parts like a push or pop instruction.

cv32e40s/tb/asserts/README.md
```diff
@@ -0,0 +1 @@
+Directory for assertion files and related models and support logic.
```
silabs-krdosvik, line 1: I like this change. As far as I understand the goal here is to add our  assert files in this directory instead of the uvmt directory, right?
In that case, comment about the directory name:
As we know asserts and coverage is often written in the same file (as it often seems fit), so I would suggest to name the new directory assert_and_cov or something that at state that we also include coverage in these files.

But I beleive I might be mistaken about my interpretation ^. Because I dont understand why you have included the assertion files in the uvmt folder, and the model files in the assert folder... (I would at least think that the assert file should go into the assert folder)

cv32e40s/docs/VerifPlans/Simulation/privileged_spec/CV32E40SX_PMA_VerifPlan.csv
```diff
@@ -290,19 +287,21 @@ COV: TODO",TODO missing cover.
...
+Note: Cover doesn't check that it goes through WB."
+,,Pop instructions,"pop, popret, pop.e, popret.e, c.pop, c.popret, c.pop.e, c.popret.e shall only occur for main memory regions, any such *pop* attempts from I/O shall cause a precise load access fault (exception code 7)","Attempt *pop* operations to main memory and IO, verify that all *pop* attempts to IO causes a precise loac access fault (code 7) and that this does not occur for main memory.",Self Checking Test,Directed Self-Checking,Testcase,DTC: cv32e40s/tests/programs/custom/pma/,TODO test is commented out.
 ,,,,,Check against RM,Constrained-Random,Functional Coverage,"RTC: TODO
 
-COV: TODO","TODO missing random test?
+COV: *.pma_cov_data_i.cover_item_covergroup_cg_inst_cg_inst_cross_x_loadstoreexec_main_pushpop_*","TODO missing random test?
 
-TODO missing cover."
+Note: Cover doesn't check that it goes through WB."
```
silabs-krdosvik, line 304: These changes looks fine. 
But is there any plans concerning the notes you have added? Or is the plan to just mention it, and dont do any thing about it?

cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
```diff
@@ -263,7 +265,7 @@
       const string pma_cfg_name = "pma_test_cfg_2";
       parameter int unsigned               CORE_PARAM_PMA_NUM_REGIONS = 7;
       parameter cv32e40s_pkg::pma_cfg_t CORE_PARAM_PMA_CFG[CORE_PARAM_PMA_NUM_REGIONS-1:0] = '{
-        '{word_addr_low : 32'hE010_0000>>2, word_addr_high : 32'hFFFF_FFFF>>2, main : 1'b0, bufferable : 1'b1, cacheable : 1'b0, integrity : 1'b0},
+        '{word_addr_low : 32'hE010_0000>>2, word_addr_high : 32'hFFFF_FFFF>>2, main : 1'b0, bufferable : 1'b1, cacheable : 1'b0, integrity : 1'b1},
```
silabs-krdosvik, line 268: That is a good idea for getting more coverage. I also needed coverage for cases where integrity was 1 for at least one region when working on the xsecure integrity interface assertion sets.

It is interessting to see this solution, which differs from mine solution. I walk on thin ice when it comes to pma/pmp configs, but my solution to this problem was to add a new config where all the regions had integrity. Do you have any comment on what is best?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tdefs.sv
```diff
@@ -104,4 +104,22 @@ typedef struct {
...
+  logic                        bufferable;
+  logic                        cacheable;
+  logic                        integrity;
+  logic                        override_dm;
+  logic [PMA_MAX_REGIONS-1:0]  match_list;
+  logic [31:0]                 match_idx;
+  logic                        have_match;
+  logic                        accesses_dmregion;
+  logic                        accesses_jvt;
+} pma_status_t;
```
silabs-krdosvik, line 122: This could also be an interface, but I think I like this better because it kind of is a packet, and it is clear in what direction it is sent (which differs a bit from how interfaces works).

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin: (I will comment on the first part of your remark too.)

> But in general: a name suggustion: change is_split_datatrans_f to is_unaligned_data_fetch

I think "split" is a necessary distinction.
If you have a halfword _misaligned_ by one byte, it would not get _split_, afaik.

"Split" is also the term used by the user manual and the archspec.

And it is for transactions in general, not just fetches/loads but also stores.

cv32e40s/docs/VerifPlans/Simulation/privileged_spec/CV32E40SX_PMA_VerifPlan.csv
silabs-robin: > is the plan to just mention it, and dont do any thing about it?

Short answer: Yes.

Long answer:
1)
For notes like the one above, I deem it to be too detailed and pedantic to be practical.
There is still a chance later, during signoff, that we decide that these should get a higher priority, but for now I don't think there is much ROI on it.
2)
For the one coverage vplan item that I did not link to (relating to isacov), there is a separate ticket for bringing isacov up to date.
But I should not trust that I remember this about PMA later, so I just added a reminder comment on that ticket, thanks.

cv32e40s/tb/asserts/README.md
silabs-mateilga: I don't think the number of files merits further division yet, although that might change in the future. Also the LRM argument makes sense, so i think the name "asserts" is the best option for the current structure.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+    cp_numregions: coverpoint  PMA_NUM_REGIONS {
+      bins zero = {0}      with (PMA_NUM_REGIONS == 0);
+      bins mid  = {[1:15]} with ((0 < PMA_NUM_REGIONS) && (PMA_NUM_REGIONS < 16));
+      bins max  = {16}     with (PMA_NUM_REGIONS == 16);
+    }
+
+    // vplan:"Overlapping PMA Regions"
+    cp_multimatch: coverpoint  num_matches  iff (is_mpu_activated) {
+      bins zero = {0}
+        with (!SIMPLIFY_FV);
```
silabs-mateilga, line 120: I think this is an adequate solution. An alternative that requires a bit more work but could be easier to read is to have two versions of each cover point within generate blocks, one for formal and one for sim. However, I'm not sure the benefit is worth the extra work, and it's only subjectively better(other people might not agree), so I will not push for a change.

cv32e40s/tb/asserts/README.md
silabs-krdosvik: Okei, I see. Sounds good. You mean the name "assertion" and not "assert", right @silabs-mateilga? Since assertion is the common-term?

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+    cp_loadstore: coverpoint  rvfi_if.rvfi_mem_wmask
+      iff (rvfi_if.is_mem_act)
+    {
+      bins  load  = {0};
+      bins  store = {[1:$]};
+      illegal_bins  undefined = default;  // Should be empty
+    }
+
+    cp_wasloadstore: coverpoint  rvfi_mem_wmask_q
+      iff (rvfi_mem_act_q && occured_rvfi_valid)
```
silabs-mateilga, line 307: Same question as above.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_cov.sv
```diff
@@ -0,0 +1,85 @@
...
+// limitations under the License.
+
+
+// Description:
+//   Coverage definitions for RVFI; mainly the support logic.
+//
+// Rationale:
+//   The RVFI support logic has many useful functions.
+//   They should be coverable in both formal and sim.
+//   They should be inspected in formal to see that they work as intended.
```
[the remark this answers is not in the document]
silabs-mateilga: Great addition

silabs-mateilga (changes requested):
Looks generally very good, left a few questions and a request for a name change.

cv32e40s/tb/asserts/README.md
silabs-mateilga: correct

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-krdosvik: I misunderstood. Never mind the questions and name suggestions. It looks good.

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_cov.sv
```diff
@@ -0,0 +1,85 @@
...
+
+
+  default clocking @(posedge clk_i); endclocking
+  default disable iff !rst_ni;
+
+
+  // Aliases
+
+  let  mem_rmask = rvfi_if.rvfi_mem_rmask;
+  let  mem_wmask = rvfi_if.rvfi_mem_wmask;
```
silabs-krdosvik, line 49: what is "let"? Will "let" indicate: logic mem_rmask; assign mem_rmask = rvfi_if.rvfi_mem_rmask?

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+      rvfi_mem_wmask_q   <= '0;
+      rvfi_mem_act_q     <= '0;
+      rvfi_pmamain_low_q <= '0;
+    end else if (rvfi_if.rvfi_valid) begin
+      occured_rvfi_valid <= 1;
+      rvfi_pma_fault_q   <= rvfi_if.is_pma_fault;
+      rvfi_mem_wmask_q   <= rvfi_if.rvfi_mem_wmask;
+      rvfi_mem_act_q     <= rvfi_if.is_mem_act;
+      rvfi_pmamain_low_q <= rvfi_pmamain_lowhigh[1];
+    end
```
silabs-krdosvik, line 100: I assume you make these "extra" signals, to make sure that they are set to 0 when undefined (rvfi_if.rvfi_valid==0)?
Or do you make them to avoid using the rvfi_if prefix in cover-points below? If neither, why do you add the extra signals?

cv32e40s/tb/uvmt/uvmt_cv32e40s_rvfi_cov.sv
silabs-robin: Here it is used as an alias, as in vhdl alias.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+      bins  misaligned = {1};
+      bins  aligned    = {0};
+    }
+
+    cp_pmafault: coverpoint  rvfi_if.is_pma_fault  iff (rvfi_if.rvfi_valid) {
+      bins  fault = {1};
+      bins  no    = {0};
+    }
+
+    cp_waspmafault: coverpoint  rvfi_pma_fault_q  iff (occured_rvfi_valid) {
```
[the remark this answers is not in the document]
silabs-robin: `rvfi_pma_fault_q` is the registered version of `rvfi_if.is_pma_fault`, so the coverpoint is not a duplicate.

The above coverpoint is checked exactly on `rvfi_if.rvfi_valid`, but this one will apply at any time where `occured_rvfi_valid` is true.

This is why I use the `is` and `was` naming.
You can check `is_fencei && was_load && was_io && !was_pmafault` to see a fencei that followed after a passing load from I/O.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
silabs-robin: They used to be called `was_rvfi_mem_act` etc, and they show what rvfi signals used to be on the previous rvfi_valid.

Their usage is e.g. `rvfi was XYZ on the previous rvfi_valid  &&  now rvfi is currently ABC`.

An alternative way to do this is to use e.g. a `let` construct, reintroduce `clk_rvfi`, and have a usage a la `rvfi_if.previous_value(rvfi_somesignal)`.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin: Fixed.

cv32e40s/tb/asserts/README.md
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+      illegal_bins  disallow_main =
+        binsof(cp_allow.disallow)  &&
+        binsof(cp_main.main);
+    }
+    x_loadstoreexec_main_pushpop: cross  cp_loadstoreexec, cp_main, cp_pushpop;
+    //x_loadstoreexec_bufferable:   cross  cp_loadstoreexec, cp_bufferable;
+      //Note: Filtering of this cross seems impossible.
+      //Each tool supports each their own disparate subset of the language, so
+      //you can seemingly make it work in one or the other but not both at once.
+    x_loadstoreexec_cacheable:    cross  cp_loadstoreexec, cp_cacheable;
```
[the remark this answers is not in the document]
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
silabs-krdosvik: Okei, understand. Both alternatives are fine by me

silabs-mateilga (approved):
Concerns are addressed, as soon as the suggested change of the TODO is comited, this is ready to go as far as I'm concerned. Approving.

silabs-krdosvik (approved):
Looks like you have addressed my comments.

cv32e40s/tb/uvmt/uvmt_cv32e40s_pma_cov.sv
```diff
@@ -0,0 +1,387 @@
...
+    }
+    x_allow_integrity:  cross  cp_allow, cp_integrity {
+      illegal_bins  disallow_integrity =
+        binsof(cp_allow.disallow) && binsof(cp_integrity.integrity);
+    }
+    x_allow_jvt:        cross  cp_allow, cp_jvt;
+
+    x_dmregion_dmode: cross  cp_dmregion, cp_dmode;
+
+    //TODO:INFO:silabs-robin more crosses are possible, but bordering on impractical/infeasible
```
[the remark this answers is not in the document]
silabs-robin: Fixed.

cv32e40s/tb/uvmt/uvmt_cv32e40s_constants.sv
silabs-hfegran: I think this needs to be tested with random simulations to be sure - I think it should not have an impact, but these are repurposed atomic bits so one needs to ensure that riscv-dv tests still function correctly with this change (with the given pma-configurations)

silabs-hfegran:
> @silabs-hfegran, @MikeOpenHWGroup, would one of you aid us by casting your committer eyes upon my plea for a merge.

Please test a few random tests that uses the PMA-configurations to ensure that nothing there breaks  + there are conflicts that needs to be resolved before merge.

silabs-robin:
I forgot that I had tweaked a todo note in the web gui... It turned ugly.

Merge conflict resolved now.

silabs-robin:
> Please test a few random tests that uses the PMA-configurations

I did this too.
About 80 of them.

5 failures.
1 timeout.
4 "vsof is deleted".

Screenshot of some of the tests run:
image

silabs-hfegran:
Thanks for the updates @silabs-robin, merging

PR #1755, opened 2023-03-29
Hotfix: Fixed missing zcmt for no_bitmanip cfg

silabs-hfegran:


silabs-hfegran:
Self merging to fix compilation errors

PR #1757, opened 2023-03-29
fix: optionally remove bad print after csrr mcycle

cathales:
The print happens after main() has finished if csrr was read via getStats() in main(), e.g. while running benchmarks.

The print is not possible at the moment and triggers a simulation fail.


cathales:
Issue created: https://github.com/openhwgroup/core-v-verif/issues/1775

Issue #1730, opened 2023-03-20
ISS: Mismatch on GPRs when clearing/setting dcsr bits in debug mode


silabs-hfegran:
## DCSR read/write bitfields GPR read value mismatch in debug mode

The following sequence of operations creates ISS mismatches, where the reference model appears to readback the value it wrote, not the value that was present prior to the csrrw instruction: enter debug mode, read dcsr and write all 0s, read dcsr and write all 1s, read dcsr and write back original value.

### Type
* Functionally incorrect behavior

### Steps to Reproduce
Please provide:
1. cv32e40s/dev 04a877e90098d126238caf32069cbfdf1b99acff
2. make test TEST=debug_test2 CFG=default USE_ISS=1  (This is at the end of a larger test, to only run this failing part, change the START_TEST_NUM define in debug_test2.c to 18)

![image](https://user-images.githubusercontent.com/75469886/226287999-4b8866d7-eeb2-41c0-87e6-546c6a30c958.png)

```
Info (IDV) Instruction executed prior to mismatch '0x3230(cover_known_iss_mismatches_dbg+12): 7b0717f3 csrrw   x15,dcsr,x14'
Error (IDV) GPR register value mismatch (HartId:0, PC:0x00003230 cover_known_iss_mismatches_dbg+12):
Info (IDV)  0> GPR x15
Info (IDV)   . dut:0x400004d3
Info (IDV)   . ref:0x400000d0
UVM_ERROR @ 40542.300 ns : rvvi-pkg.sv(93) reporter [] uvmt_cv32e40s_tb.imperas_dv.idv_trace2api.state_compare @ 40542.000 ns: MISMATCH
Info (IDV) Instruction executed prior to mismatch '0x3234(cover_known_iss_mismatches_dbg+16): 577d     c.li    x14,-1'
Error (IDV) GPR register value mismatch (HartId:0, PC:0x00003234 cover_known_iss_mismatches_dbg+16):
Info (IDV)  1> GPR x15
Info (IDV)   . dut:0x400004d3
Info (IDV)   . ref:0x400000d0
UVM_ERROR @ 40551.300 ns : rvvi-pkg.sv(93) reporter [] uvmt_cv32e40s_tb.imperas_dv.idv_trace2api.state_compare @ 40551.000 ns: MISMATCH
[...]

PR #1758, opened 2023-03-29
Removed no compare on DCSR ISS comparison

silabs-hfegran:
cf. issue #1730 comments


silabs-robin (approved):
Approving based on:
Agree that `dcsr` should be checked.
As I understand it, the checking was temporarily disabled before, and it is now brought back in again.

PR #1765, opened 2023-03-29
Switch Verilator testing to use v5.008 instead of v4.110

zchamski:
This PR introduces support for Verilator v5 (v5.008 at the time of PR creation).  The main modifications are:
*  A change in the use of shell variables to configure Verilator:
  * In Verilator v5, `VERILATOR_ROOT` is required to point to the root of the _source_ tree of Verilator
  * A separate variable (currently named `VERILATOR_INSTALL_DIR`) is needed to designate the installation directory of Verilator
  * If not defined explicitly by the user, `VERILATOR_INSTALL_DIR` is set to the parent directory of `$VERILATOR_ROOT`.
  * Path settings for Verilator-related executables and include files are expressed using `$VERILATOR_INSTALL_DIR`. 
* A change in the build flow of Verilator: instead of downloading and untarring a source package tarball, the source code is cloned from Verilator Github repository.
* A patch is applied to the source code of Verilator (it will be upstreamed shortly.)
* FORNOW, the `cva6/sim/Makefile` rule for Verilator invocation passes the `--no-timing` Verilator option as part of the value of Makefile variable `verilator`.

### DIFF
--- a/.gitlab-ci/cva6.yml
+++ b/.gitlab-ci/cva6.yml
@@ -102,6 +102,7 @@ pub_check_env:
     - echo $RISCV
     - echo $RISCV_PREFIX
     - echo $VERILATOR_ROOT
+    - echo $VERILATOR_INSTALL_DIR
     - echo $SPIKE_ROOT
     - echo $BBL_ROOT
     - echo $SYN_VCS_BASHRC
--- a/.gitlab-ci/cva6.yml
+++ b/.gitlab-ci/cva6.yml
@@ -149,15 +149,21 @@ pub_smoke:
     DASHBOARD_SORT_INDEX: 0
     DASHBOARD_JOB_CATEGORY: "Basic"
   script:
-    - mkdir -p artifacts/reports
+    - mkdir -p artifacts/reports artifacts/logs
     - python3 .gitlab-ci/scripts/report_fail.py
     - echo $SYN_VCS_BASHRC; source $SYN_VCS_BASHRC
-    - source cva6/regress/smoke-tests.sh
+    # In order to capture logs in case of test failure, the test script cannot fail.
+    - source cva6/regress/smoke-tests.sh || true
+    # The list of files must NOT fail on various DV_SIMULATORS values, so use 'v*_sim' to match
+    # 'veri-testharness_sim', 'vcs-testharness_sim' and 'vcs-uvm_sim' (one of them always applies,
+    # at least until new RTL simulator configurations are added.)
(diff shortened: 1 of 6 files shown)


zchamski:
For the matching CVA PR see https://github.com/openhwgroup/cva6/pull/1147.

MikeOpenHWGroup:
This all looks good to me @zchamski, but I will defer to @JeanRochCoulon and/or @ASintzoff to approve and merge this one.

umerimran-10xe:
Hi @zchamski, What changes do we need to have in our Infrastructure to enable `--timing` flag with Verilator ?

zchamski:
> Hi @zchamski, What changes do we need to have in our Infrastructure to enable `--timing` flag with Verilator ?

HI @umerimran-10xe , the only change needed is to remove the definition `verilator="verilator --no-timing"` in `cva6/sim/Makefile` line 119.  When this definition is skipped, the Makefile on the CVA6 side will pick up the default definition of `verilator` (which is simply `verilator` without any options, cf. CVA6 `Makefile` line 28).

umerimran-10xe:
Correct, but when we run with no option, it fails with the error as mentioned below. Either we need to run with `--timing` or `--no-timing` option. 

```
%Error-NEEDTIMINGOPT: /home/umerimran/Music/cva6/corev_apu/tb/common/SimDTM.sv:53:10: Use --timing or --no-timing to specify how timing controls should be handled
                                                                                    : ... In instance ariane_testharness.genblk1.i_SimDTM
   53 |   assign #0.1 debug_req_bits_data = __debug_req_bits_data[31:0];
      |          ^
```

Have you tried running with `--timing` option ?

zchamski:
> Have you tried running with `--timing` option ?

Not yet :-)

MikeOpenHWGroup:
```
%Error-NEEDTIMINGOPT: /home/umerimran/Music/cva6/corev_apu/tb/common/SimDTM.sv:53:10: Use --timing or --no-timing to specify how timing controls should be handled
                                                                                    : ... In instance ariane_testharness.genblk1.i_SimDTM
   53 |   assign #0.1 debug_req_bits_data = __debug_req_bits_data[31:0];
      |          ^
```
A better question would be, why is a "pound delay" assignment in the SimDTM module?  I would call that a bug.  Can we get rid of it?

JeanRochCoulon:
Great !! The PR passes the CI. COuld you squash the commits before I merge ?

PR #1768, opened 2023-03-30
CVXIF : Update the CVXIF agent

AyoubJalali:


AyoubJalali:
Hello, after Relocate the vseq in cva6 env, we're back to were we stopped.
So I made some updates in CVXIF agent, including more filtering in the base vseq, also in vseq, also i notice we had no information received related to **rs_valid**,  that because it was not implement in the monitor to drive this information, so I add it also.
I goona try to create small PR for updating the CVXIF agent, because I see a lot of things to be update. Thanks

cva6/env/uvme/cvxif_vseq/uvme_cvxif_base_vseq.sv
ASintzoff, line 1: Could you rework the way the tests are done to have only one `return ("ILLEGAL")`?
Mainly, use "positive" tests to identify non illegal instructions and if all the tests fail, return illegal.

ASintzoff (changes requested):

PR #1770, opened 2023-03-30
Adapt the environment and test-bench to execute the AXI agent in active/passive mode

AEzzejjari:
This commit allows to us switch the axi_agent mode without changing any part of code.
There is two way to change the agent mode: 

First way: From env_cfg we can change the variable axi_cfg.is_active to UVM_PASSIVE or UVM_ACTIVE.
Second:   From command line by using --axi_active option.

### DIFF
--- a/cva6/sim/cva6.py
+++ b/cva6/sim/cva6.py
@@ -816,6 +816,8 @@ def setup_parser():
                       help="custom configuration options, to be passed in config_pkg_generator.py in cva6")
   parser.add_argument("-l", "--linker", type=str, default="",
                       help="Path for the link.ld")
+  parser.add_argument("--axi_active", type=str, default="",
+                      help="Switch the AXI agent mode. yes for Active, no for Passive")
   return parser
 
 
@@ -931,7 +933,15 @@ def main():
     parser = setup_parser()
     args = parser.parse_args()
     global issrun_opts
-    issrun_opts = "\""+args.issrun_opts+"\""
+    if args.axi_active == "yes":
+      args.issrun_opts = args.issrun_opts + " +uvm_set_config_int=uvm_test_top.env,axi_is_passive,10"
+      issrun_opts = "\""+args.issrun_opts+"\""
+    elif args.axi_active == "no":
+      args.issrun_opts = args.issrun_opts + " +uvm_set_config_int=uvm_test_top.env,axi_is_passive,1"
+      issrun_opts = "\""+args.issrun_opts+"\""
+    else:
+      issrun_opts = "\""+args.issrun_opts+"\""
+
     global isspostrun_opts
     isspostrun_opts = "\""+args.isspostrun_opts+"\""
     global isscomp_opts
--- a/cva6/sim/cva6.py
+++ b/cva6/sim/cva6.py
@@ -817,7 +817,7 @@ def setup_parser():
   parser.add_argument("-l", "--linker", type=str, default="",
                       help="Path for the link.ld")
   parser.add_argument("--axi_active", type=str, default="",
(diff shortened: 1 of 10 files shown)


cva6/sim/cva6.py
```diff
@@ -931,7 +933,15 @@ def main():
     parser = setup_parser()
     args = parser.parse_args()
     global issrun_opts
-    issrun_opts = "\""+args.issrun_opts+"\""
+    if args.axi_active == "yes":
+      args.issrun_opts = args.issrun_opts + " +uvm_set_config_int=uvm_test_top.env,axi_is_passive,10"
+      issrun_opts = "\""+args.issrun_opts+"\""
```
ASintzoff, line 938: Lines 938, 941, 943 should be put outside if...else.

cva6/sim/cva6.py
```diff
@@ -816,6 +816,8 @@ def setup_parser():
                       help="custom configuration options, to be passed in config_pkg_generator.py in cva6")
   parser.add_argument("-l", "--linker", type=str, default="",
                       help="Path for the link.ld")
+  parser.add_argument("--axi_active", type=str, default="",
+                      help="Switch the AXI agent mode. yes for Active, no for Passive")
```
ASintzoff, line 820: help="switch AXI agent mode: yes for Active, no for Passive"

cva6/tb/uvmt/cva6_tb_wrapper.sv
```diff
@@ -109,22 +110,40 @@ module cva6_tb_wrapper
   logic [AXI_DATA_WIDTH-1:0]    rdata;
   logic [AXI_USER_WIDTH-1:0]    ruser;
 
+
```
ASintzoff, line 113: Unnecessary blank lines

ASintzoff (changes requested):
I would prefer to have `axi_is_active` instead of `axi_is_passive` (in `cva6.py`) as there is `cfg.axi_cfg.is_active` in System Verilog (principle of least surprise: don't confuse the poor reviewer).

Instead of using 1 and 10 in cva6.py, could you use 1 and 0?

MikeOpenHWGroup:
> First way: From env_cfg we can change the variable axi_cfg.is_active to UVM_PASSIVE or UVM_ACTIVE.
> Second: From command line by using --axi_active option.

Neither of these are the "UVM way" of controlling a configuration.  The UVM way is to declare `axi_cfg` as a random variable in the Agent's configuration object.  The UVM testcase can then add constraints to set the value as needed.

AEzzejjari:
The active vip agent should be the default mode, unless someone needs to revert to the old mode. For this, I implemented a constraint in the base test to set the mode directly. Also, I left the option to change the mode from the command line, as it is easier for someone who doesn't know the environment.

MikeOpenHWGroup:
> The active vip agent should be the default mode, unless someone needs to revert to the old mode.

Agreed.

> For this, I implemented a constraint in the base test to set the mode directly. Also, I left the option to change the mode from the command line, as it is easier for someone who doesn't know the environment.  

Are you talking about this:
```
 // Variable can be modified from command line, to change the AXI agent mode
   int force_axi_mode = -1;
```

It seems that the variable `force_axi_mode` is simply there to allow for command-line overrides.  This is not "the UVM way".  We can simply this a lot by creating a new testcase that extends the base test.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
OK, as we discussed in the project meeting, I am putting a little bit of water in my wine and approving this PR.

It cannot be merged until the conflict is resolved.

PR #1776, opened 2023-03-30
OBI Assert - Parity

silabs-robin:
This PR adds parity asserts (reqpar, etc) to the OBI assertion set, according to [OBI v1.2](https://raw.githubusercontent.com/openhwgroup/obi/main/OBI-v1.2.pdf).

Test status:
* ci_check - Passes
* Internal formal - N/A, not used (passes when enabled)
* Cvverif formal - Works as expected (core-side pass, env-side fail)

I did not expect to write this today, but I wanted to see if one of the latest RVFI bug fixes were reproducable with the core-v-verif formal setup and this is the only piece missing that I'm aware of. _(OBI parity assumes were already in cvverif fv, but lazily used xsecure asserts.)_

### DIFF
--- a/lib/uvm_agents/uvma_obi_memory/src/uvma_obi_memory_1p2_assert.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/uvma_obi_memory_1p2_assert.sv
@@ -262,4 +262,25 @@ module uvma_obi_memory_1p2_assert
   else
     `uvm_error(info_tag, "EXOKAY may only asserted in response to an LR or SC transaction (signaled via atop)")
 
+  // R-13: The reqpar signal shall be the inverse of req at each rising clk edge (when not in reset).
+  a_reqpar: assert property (
+    reqpar == !req
+  ) else `uvm_error(info_tag, "bad reqpar")
+
+  // R-14: The gntpar signal shall be the inverse of gnt at each rising clk edge (when not in reset).
+  a_gntpar: assert property (
+    gntpar == !gnt
+  ) else `uvm_error(info_tag, "bad gntpar")
+
+  // R-15: The rvalidpar signal shall be the inverse of rvalid at each rising clk edge (when not in reset).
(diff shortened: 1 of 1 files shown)


silabs-krdosvik:
By looking at the code, it seems like you have added 4 "easy" (and correct (according to my understanding of the spec)) assertions to that file, and I dont see any potential problems with that.

However, the description you have added make me more sceptical, as you say this is somehow connected to assumes (if im not mistaken?).
I beleive assumes have more consequences than asserts, so it would therefor be nice if I understood the connection here, and than hopefully also the actuall effect of this code, before approving.

If it is common knowledge (that I just simply dont know), you can explain to me internaly.

silabs-robin:
> you say this is somehow connected to assumes (if im not mistaken?). I beleive assumes have more consequences than asserts, so it would therefor be nice if I understood the connection here, and than hopefully also the actuall effect of this code, before approving.

Since formal needs assumes on a few things (bus protocol compliance, etc), I plan to use Tcl to switch some of these properties from being an assert into being an assume: https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/fv/jaspergold.tcl#L37
(This is about the core-v-verif formal setup, it does not change anything in our internal formal setup.)

The signals coming _from_ the core will be checked as asserts.
The signals going _to_ the core, from the environment, they are by default unconstrained in formal so that is were I'll flip the asserts into assumes.
(Glitch testing etc might not want this, but normal operation needs the assumes.)

Good that you point this out(!), because I see that the Tcl script is currently too generous in which properties it flips.

silabs-krdosvik (approved):

silabs-mateilga (approved):
Looks good

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Approving based on the above discussion and merging as I am aware that @silabs-hfegran is off today.

PR #1783, opened 2023-04-04
Updated/fixed pma.c

silabs-hfegran:


cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1150 @@
...
+    : [addr] "r"(mem_addr_1)
+    : "t0", "memory"
+  );
+
+  test_fail += *g_exp_fault;
+  test_fail += tmp           != 0xbbbbbbbb;
+  test_fail += g_mcause->raw != MCAUSE_LOAD_FAULT;
+  test_fail += *g_mepc       == 0;
+  test_fail += *g_mtval      != 0;
+  //cvprintf(V_LOW, "g_mcause: %08lx, g_mepc: %08lx, g_mtval: %08lx\n", g_mcause->raw, *g_mepc, *g_mtval);
```
silabs-hfegran, line 645: remove or change verbosity and comment in  (same for line 666)

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1150 @@
...
+    # Misaligned, boundary crossing store
+    li t0, 0x11223344
+    sw t0, 2(%[addr])
+  )"
+    :
+    : [addr] "r"(io_addr)
+    : "t0", "memory"
+  );
+
+  // Quirk: Split the assembly functions here as we
```
silabs-hfegran, line 789: TODO: check if this comment is still valid after fixes to interrupt handler

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
+    return index + 1;
+  }
+  cvprintf(V_LOW, "\nTest: \"%s\" OK!\n", name);
+  return 0;
+}
+
+// -----------------------------------------------------------------------------
+
+// TODO: implement the following for 40X:
+// Plan: Detect 40s/40x by checking csrs (avoid ifdefs)
```
silabs-hfegran, line 934: What the comment above mentions is possible, albeit tricky (have to use .word(machinecode) instead of asm instructions, otherwise the compiler will not compile the code for 40s

MikeOpenHWGroup: Maybe I am misunderstanding the intend here, but this should be easy since mvendor+march CSRs should be a globally unique way of determining which core the code is running on.

silabs-hfegran: That is corect Mike, the issue is compiler flags here - because we write explicit asm-instructions, the ones that are not enabled in the compiler flags for a given core will cause a compilation failure. Correct me if I am wrong, but I think we either need separate tests, separate includes or ifdefs to solve this for both cores.

MikeOpenHWGroup: Ahhh, yes. You are right.

MikeOpenHWGroup (approved):

cv32e40s/tests/programs/custom/pma/pma.c
silabs-hfegran: One option, albeit unreadable, is to use .word(machinecode) instead, as that will work with any assembler, but this will be a nuisance to debug (and not very feasible to do with jump/branch instructions)

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
+
+// ---------------------------------------------------------------
+// Helper functions
+// ---------------------------------------------------------------
+/*
+ * set_test_status
  *
- * [r0, "main"]
- * 0xFFFF_FFFF
- * 0x1A11_1000         <MEM_ADDR_1>
```
silabs-robin, line 30: The map of how the address space is removed? Is it because the tests no longer work that way or something else?

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
+volatile verbosity_t global_verbosity = V_LOW;
+
+// Global pointer variables
+volatile uint32_t * volatile g_has_clic;
+volatile uint32_t * volatile g_exp_fault;
+volatile mcause_t * volatile g_mcause;
+volatile uint32_t * volatile g_mepc;
+volatile uint32_t * volatile g_mtval;
+volatile uint32_t * volatile g_test_num;
+volatile uint32_t * volatile g_recovery_ptr;
```
silabs-robin, line 115: Why are these pointers instead of just variables?

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
 
-void load_misaligned_iomem(void)     {__asm__ volatile("lw t0, 0(%0)" : : "r"(MEM_ADDR_1 - 2) : "t0");}
-static void load_aligned_iomem(void) {__asm__ volatile("lh t0, 0(%0)" : : "r"(MEM_ADDR_1 - 2) : "t0");}
+int cvprintf(volatile verbosity_t verbosity, const char * volatile format, ...){
+  va_list args;
+  volatile int retval = 0;
 
-void load_misaligned_memio(void)     {__asm__ volatile("lw t0, 0(%0)" : : "r"(IO_ADDR - 1) : "t0");}
-static void load_aligned_memio(void) {__asm__ volatile("lb t0, 0(%0)" : : "r"(IO_ADDR - 1) : "t0");}
+  va_start(args, format);
```
silabs-robin, line 276: Would runs without prints be faster if va_start/end be inside the if-statement? (Not that it matters for this small test.)

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -1,5 +1,5 @@
-// Copyright 2021 OpenHW Group
-// Copyright 2021 Silicon Labs, Inc.
+//
+// Copyright 2023 Silicon Labs, Inc.
```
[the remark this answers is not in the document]
MikeOpenHWGroup: I am not 100% sure about this, but AFAIK, we should not remove previous copyright declarations, but we can add to existing ones.  So removing `Copyright 2021 OpenHW Group` is frowned upon, and updating the SiLabs copywrite is better done as `Copyright 2021, 2023 Silicon Labs, Inc.`

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
+#define NUM_TESTS 9
+// Start at 1 (ignore dummy test that is only used for env sanity checking during dev.)
+#define START_TEST_NUM 1
+// Abort test at first self-check fail, useful for debugging.
+#define ABORT_ON_ERROR_IMMEDIATE 0
+
+// Addresses of VP interrupt control registers
+#define TIMER_REG_ADDR         ((volatile uint32_t * volatile) (CV_VP_INTR_TIMER_BASE))
+#define TIMER_VAL_ADDR         ((volatile uint32_t * volatile) (CV_VP_INTR_TIMER_BASE + 4))
+#define DEBUG_REQ_CONTROL_REG *((volatile uint32_t * volatile) (CV_VP_DEBUG_CONTROL_BASE))
```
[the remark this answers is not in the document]
silabs-hfegran: No, not used - kept as a part of the template if one wants to access this functionality in the future if expanding the test. 
I do not mind removing it, but that will more easily lead to "template divergence" if someone realizes they need to reimplement the functionality in the future. Plan is to move this to some common test header file when the header generation script is completed.

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
+  MCAUSE_ACCESS_FAULT    = 1,
+  MCAUSE_ILLEGAL         = 2,
+  MCAUSE_BREAKPT         = 3,
+  MCAUSE_LOAD_FAULT      = 5,
+  MCAUSE_STORE_FAULT     = 7,
+  MCAUSE_UMODE_ECALL     = 8,
+  MCAUSE_MMODE_ECALL     = 11,
+  MCAUSE_INSTR_BUS_FAULT = 24,
+  MCAUSE_CHK_FAULT       = 25,
+} mcause_exception_status_t;
```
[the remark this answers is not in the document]
silabs-hfegran: Will be moved, but I deferred moving the common code until we have a suitable structure in place for that.

cv32e40s/tests/programs/custom/pma/pma.c
silabs-hfegran: Convenience, I find working with pointers easier in assembly - could also be variables.

cv32e40s/tests/programs/custom/pma/pma.c
```diff
@@ -13,319 +13,1147 @@
...
 
-#define  EXCEPTION_INSN_ACCESS_FAULT      1
-#define  EXCEPTION_LOAD_ACCESS_FAULT      5
-#define  EXCEPTION_STOREAMO_ACCESS_FAULT  7
+// External symbols
+extern volatile uint32_t mtvt_table;
 
-/* CFG Memory Layout
+// Message strings for use in assembly printf
+//const volatile char * const volatile asm_printf_msg_template = "Entered handler %0d\n";
```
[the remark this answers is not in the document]
silabs-hfegran: Kept as part of template example if one wants to print strings in assembly. E.g. 
the above would be printed by 
```
lw a0, asm_printf_msg_template # format
addi a1, zero, value_of_some_int # first %d in string
call printf
```

cv32e40s/tests/programs/custom/pma/pma.c
silabs-hfegran: technically the test is rewritten from scratch, so not removed per se - could add it back in as a comment in the test description.
Could also change my code to use this as defines/global constants, I will look into that.

cv32e40s/tests/programs/custom/pma/pma.c
silabs-hfegran: I have a feeling that this might be optimized by the compiler, but I will test and see if it makes a noticeable difference

cv32e40s/tests/programs/custom/pma/pma.c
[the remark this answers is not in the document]
silabs-hfegran: 1. Minor changes to file, Should year modified be added?
2. File cleared completely, all new content added, should old copyright be kept? 
3. File cleared completely, some/all new content is inspired by old content, should old copyright be kept?

I would expect 1. yes, 2. no, 3 yes - any inputs on this @MikeOpenHWGroup? (This file falls under category 3. as content is not (in most cases) copied from old test but rather rewritten with the same idea and methodology, to the extent possible as the original test)

silabs-robin (approved):
Approving based on:
The only outstanding question is about the copyright year, which I regard as minor because 1) it is our own original file, and 2) I assume the PR will not be merged before the matter is resolved.

cv32e40s/tests/programs/custom/pma/pma.c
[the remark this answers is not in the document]
MikeOpenHWGroup: My law degree got lost in the mail, so my opinion is probably useless, but FWIW I agree with you @silabs-hfegran.  I'm going to post this question to MM, but please do not gate a merge of this PR on the outcome of that discussion.

I'll let @silabs-robin decide if this conversation can be marked s resolved.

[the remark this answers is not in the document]
silabs-robin: The old file has its copyright in the git history. This is our own original file and not something we found and modified. I vote this is fine.

PR #1784, opened 2023-04-05
CVA6 CSR ACCESS TEST PR

spidugu444:
Hi @JeanRochCoulon, as per you suggestion, I simplify PR by addressing only MSCRATCH CSR.


spidugu444:
@zchamski could you please look on this PR.

ASintzoff:
Why are there old format (.pck) VPtool files?
The format to use is the YAML one.

I would change the name of `cva6/docs/VerifPlans/csr_verification` to `cva6/docs/VerifPlans/csr_access` or something like this to be more precise about the goal.

cva6/docs/VerifPlans/csr_verification/runme.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Set the destination directory of Markdown files for this project.
+# Since it will be used by VPTOOL, it shall NOT be a relative path.
+export MARKDOWN_OUTPUT_DIR=`readlink -f "$ROOTDIR/../source"`
+
+# Run VPTOOL overriding the default theme from Yaml config with 'winxpblue'.
+# FIXME: Introduce a suitably named shell variable that points to the root
+# directory of the tool set (TOOL_TOP etc.)
+# FORNOW use a hardcoded relative path.
+python3 $ROOTDIR/../../../../tools/vptool/vptool/vp.py -t winxpblue
```
zchamski, line 34: @spidugu444 , please replace the explicit option `-t winxpblue` with `$*` (as it is done for example in `cva6/docs/VerifPlans/ISA_RV32/runme.sh`).   This will allow you to pass any option to VPTOOL, including `-p` and `-t winxpblue`) interactively from the command line, e.g.
```
. cva6/docs/VerifPlans/csr_verification/runme.sh -p -t winxpblue
```

cva6/docs/VerifPlans/csr_verification/VP_IP004.pck
zchamski, line 1: @spidugu444 , it looks like you are using an old version of VPTOOL without support for Yaml database format (cf. the value of `vptool_gitrev` field, line 187 in the `.pck` file below).  This means your version of VPTOOL dates back to PR #1493.  Please make sure your VPTOOL is up to date: `git log -- tools/vptool/vptool/vp.py` should start with
```
commit 2768a433ac49ba93f3ecd9035508495effcbb5e4
Author: Zbigniew Chamski <zbigniew.chamski@thalesgroup.com>
Date:   Fri Feb 24 17:42:13 2023 +0100
```
Once this is done, update `runme.sh` as described in this review and then you will be able migrate the database from Pickle to Yaml format by invoking VPTOOL via `runme.sh` with the `-p` option:
```
. cva6/docs/VerifPlans/csr_verification/runme.sh -p
```
Save the database using VPTOOL and you will see a new `VP_IP004.yml` Yaml file created.  Remove the `.pck` file, run the `runme.sh` wrapper again **without** the `-p` option and you should see the same database content but this time read from the Yaml file.

zchamski (changes requested):
Please update your installation of VPTOOL, modify the `runme.sh` script and with these mods you will be able to update the Picke database file to Yaml format.

The explicit define of XLEN in the test specification file [cva6/tests/testlist_riscv-csr-access-test-cv32a60x.yaml](https://github.com/openhwgroup/core-v-verif/pull/1784/files#diff-244687b6f849b6ebe2f242f453bc1d45285ab2594aefbe0c310965d08c4283c8) should be dropped.

Also, please remove the two unused files.

cva6/tests/testlist_riscv-csr-access-test-cv32a60x.yaml
```diff
@@ -0,0 +1,7 @@
+- test: MSCRATCH_CSR
+  iterations: 1
+  path_var: TESTS_PATH
+  gcc_opts: "-DXLEN=32 -DTEST_CASE_1=True -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -I<path_var>/riscv-isa-sim//spike/"
```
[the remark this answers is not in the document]
spidugu444: Removed user-defined value -DXLEN=32

cva6/docs/VerifPlans/csr_verification/runme.sh
spidugu444: Updated to latest version VPTOOL

cva6/docs/VerifPlans/csr_verification/VP_IP004.pck
spidugu444: Generated yaml file by  using updated version VPTOOL .

JeanRochCoulon:
Hello @spidugu444,
Again, thank you for your PR, it makes us progress. 
You can find below some comments:
- Can you add the specification you are verifying to the CSR directory in pdf format?   It corresponds to an user manual version
- In DVplan, as the test is directive, can you specify the "random values" you will use. I suggest to choose values to allow the toggling of all the bits of the value (0x5555555555555555 and 0xAAAAAAAAAAAAAAAA for instance).
- In DVplan, as MSCRATCH CSR does not depend on priviliege mode (also if the name is **M**scratch), can you add to the DVplan a check that MSCRATCH CSR can be accessed also from User and Supervisor mode.
- An open discussion could be : how to implement functional coverage on different dvplan points ?  Do we reply on the fact that the test is executed or do we need to implement it in UVM format. Maybe @MikeOpenHWGroup would have an answer to it.
Thanks !!

spidugu444:
Hi @JeanRochCoulon, Thanks for your comments on this PR.
I modified DV plan as per your comments, please find the list of queries that i have

- For CSR Access Verification, I am using **CVA6 User manual and riscv-privileged-20211203 specification**, whether I need to add these documents?
- As per your suggestion, In DV plan I changed random values with test generated sequence i.e 0XFFFFFFFF, 0XA5A5A5A5, 0X5A5A5A5A ...
- I had a query on adding negative test scenarios, Accessing Machine mode CSRs on lower privilege modes will raise an exception. so to verify Machine Mode CSRs in lowest privilege mode we need to use two tests one is to access in supervisor mode and another one is to access in user mode whether i need to add these two tests also in PR?

MikeOpenHWGroup:
Hi @spidugu444, about your question:

> To verify Machine Mode CSRs in lowest privilege mode we need to use two tests one is to access in supervisor mode and another one is to access in user mode whether i need to add these two tests also in PR?

There is no right or wrong answer to this.  I recommend having a single test to verify all privileges and access types of a CSR.  For example, MSCRATCH has MRW privilege, and you have written an algorithm to verify that.  To the same test you can switch first to Supervisor mode and attempt to access MSCRATCH.  This should result in an exception.  If it does not the test should fail.  After returning from the exception, the same test can switch to User mode and perform the same checks.  It would also be useful to switch back to Machine mode and perform an access of MSCRATCH again, just to check that the privilege mode is restored after switching from Machine->Supervisor->User->Machine.

You could do this in multiple test-programs, but there are a lot of CSRs with Machine mode privilege and this would create a very large number of test-programs

JeanRochCoulon:
Hello
I suggest to copy all the spec you are refering into the directory where you have saved the DVplan, that is to say both specifications.
Another question: I am not clear abou the behaviour of MSCRATCH CSR when accessing in other mode than Machine because RISCV ISA specification does not tell that an exception is generated ?!? @MikeOpenHWGroup, @spidugu444 is it clear for you ?
Cheers

MikeOpenHWGroup:
It is my understanding that Machine CSRs (such as MSCRATCH) are not accessible in User or Supervisor mode. Accessing Machine CSRs in either User or Supervisor mode will cause an illegal instruction exception.

This is based on the second paragraph of Section 2.1, CSR Address Mapping Conventions, in the Privilege Specification which says:
> Attempts to access a non-existent CSR raise an illegal instruction exception. Attempts to access a CSR without appropriate privilege level or to write a read-only register also raise illegal instruction exceptions.

spidugu444:
> There is no right or wrong answer to this. I recommend having a single test to verify all privileges and access types of a CSR. For example, MSCRATCH has MRW privilege, and you have written an algorithm to verify that. To the same test you can switch first to Supervisor mode and attempt to access MSCRATCH. This should result in an exception. If it does not the test should fail. After returning from the exception, the same test can switch to User mode and perform the same checks. It would also be useful to switch back to Machine mode and perform an access of MSCRATCH again, just to check that the privilege mode is restored after switching from Machine->Supervisor->User->Machine.

As suggested this CSR test can be made into single test program considering all privilege modes ( Machine->Supervisor->User->Machine) However accessing CSR in lower privilege modes will raise exception which has to be handled to switch into higher privilege mode. I doubt how to control the exception report for control mechanism in test program.
if you can suggest on this point I would definitely try at my end.

spidugu444:
@RanjanThales

JeanRochCoulon:
Indeed this PR is more a strategy discussion... This should be out of GitHub PR. But how to read the files which are proposed by Sai Krishna ?  Doing like that is very useful to collaborate: we see the code and can discuss together.

MikeOpenHWGroup:
Hi @JeanRochCoulon, I suggest you join the meeting that @spidugu444, @ASintzoff and I have every Monday at 13h00 CEST.  I have sent you a meeting invitation.

JeanRochCoulon:
Sai Krishna,
In the DVPLAN I see one feature, one sub-feature and three items. I would suggest to define three sub-features: one dedicated to POR, one dedicated to RW accesses and one dedicated to privilege mode. The same sub-features should be defined for all the other CSRs. In that way, it will be more readable.
Cheers, Jean-Roch

JeanRochCoulon:
@spidugu444   Aziz has just commited the rst file of the CSR specification in https://github.com/openhwgroup/cva6/pull/1221. I suggest that you recopy this rst file in your DVPLAN directory and update VPTOOL to point on the right mscratch chapter.

cva6/regress/smoke-tests.sh
```diff
@@ -26,6 +26,7 @@ if ! [ -n "$DV_SIMULATORS" ]; then
 fi
 
 cd cva6/sim/
+python3 cva6.py --testlist=../tests/testlist_riscv-csr-access-test-cv32a60x.yaml --test MSCRATCH_CSR --iss_yaml cva6.yaml --target cv32a60x --iss=$DV_SIMULATORS $DV_OPTS  --linker=../sim/link.ld
```
JeanRochCoulon, line 29: The CSR tests will be executed from dv-riscv-csr-access-test.sh, you do not need to modify the smoke_tests.sh

JeanRochCoulon (changes requested):
On my side, I request the changes, but ready to merge as soon it is done. After the merge, I think I will submit a new PR to add the job to the CI.

cva6/regress/smoke-tests.sh
spidugu444: Removed CSR access test from smoke-tests.sh

cva6/regress/install-riscv-isa-sim.sh
```diff
@@ -11,7 +11,7 @@ if ! [ -n "$RISCV_ISA_SIM" ]; then
   RISCV_ISA_SIM="https://github.com/riscv-software-src/riscv-isa-sim.git"
   RISCV_ISA_SIM_BRANCH="master"
   RISCV_ISA_SIM_HASH="b9fc8e4e9087a6064dfcc627efabbe3fd4bdc309"
-  RISCV_ISA_SIM_PATCH="../../../cva6/regress/riscv-isa-sim.patch"
```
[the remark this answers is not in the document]
spidugu444: Removed patch related code.

spidugu444:
@JeanRochCoulon, I modified changes as per your suggestions, whether I need to raise new PR with updated changes?

PR #1786, opened 2023-04-06
Adding options to regressions and instruction generator

XavierAubert:
Added a way to customize the array `supported_isa` inside the file `riscv_core_settings.sv` using compile-time macros. This will allow one unique file to add RV32F, RV32ZFinx and RV32X extensions only when needed. 

Adding this feature implied also a way to specify which riscv-dv configuration is to be used, hence the addition of the makefile variable RISCVDV_CFG. As it is the case for the other CFG variable, the RISCVDV_CFG variable must be set to the name of the cfg .yaml file inside the test folder that will be load. To be backward-compatible and if not specified, the makefile will read the default "corev-dv.yaml" file. 

I added a support to this feature for shell, questa and vmanager regressions. 

I added the build.cfg support from commit 441deba1ef7584a0c0cf6e13839f9dde1f28f5c3 to questa and vmanager regressions


MikeOpenHWGroup (commented):
Hi @XavierAubert, this _looks_ good to me, but GitHub is saying there are some conflicts to resolve.

MikeOpenHWGroup:
Thank you sir!   Merging...

PR #1791, opened 2023-04-10
Optional compile of ImperasDV

MikeOpenHWGroup:
This PR restores the functionality of the `USE_ISS` Makefile variable.  Setting this variable to "NO" on any `make test` invocation will compile the environment excluding the ImperasDV model.  For example:
```
$ make test TEST=hello-world USE_ISS=NO
```

### DIFF
--- a/cv32e40p/tb/uvmt/uvmt_cv32e40p.flist
+++ b/cv32e40p/tb/uvmt/uvmt_cv32e40p.flist
@@ -47,9 +47,6 @@
 +incdir+${TBSRC_HOME}
 ${DV_UVMT_PATH}/uvmt_cv32e40p_pkg.sv
 
-// ImperasDV
--f ${TBSRC_HOME}/uvmt/imperas_dv.flist
-
 // CV32E40P test bench files
 ${DV_UVMT_PATH}/uvmt_cv32e40p_dut_wrap.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40p_tb.sv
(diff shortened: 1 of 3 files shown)


YoannPruvost:
Hi,
Thanks for the PR
Few comments:
 - This only works for xrun. It breaks all other simulators
 - Updating the vsim makefile the same way does not work because questa is sensitive to the file list order and Imperas_v_wrap references some uvme package.
   - on the e40p we can remove the reference to the uvme package and get it to work
   - for simulators other than xrun and vsim I haven't tried
   - For other cores I haven't tried but we will have the same issue with questa

MikeOpenHWGroup:
Thanks for the comments @YoannPruvost.

> This only works for xrun. It breaks all other simulators

Yes, this is known.  I wanted to get it working with Xcelium first because I know Dolphin uses that simulator, and then create follow on PRs for other simulators (they are all broken right now unless you've got ImperasDV).  Sorry I did not mention that before.

> Updating the vsim makefile the same way does not work because questa is sensitive to the file list order and 

Imperas_v_wrap references some uvme package. On the e40p we can remove the reference to the uvme package and get it to work.
Great idea.  Can you create a PR for that (assuming you accept this one)?

> For simulators other than xrun and vsim I haven't tried. For other cores I haven't tried but we will have the same issue with questa.

Let's get Xcelium and Questasim running on E40P and I'll move onto the other simulators and cores.

YoannPruvost:
>Let's get Xcelium and Questasim running on E40P and I'll move onto the other simulators and cores.
Ok let's do that !

YoannPruvost (approved):

MikeOpenHWGroup:
Perfect.  Thanks a lot @YoannPruvost.  Merging...

PR #355, opened 2020-11-10
Use state machine in step & compare instead of asynchronous always blocks

GTumbush:
Updated uvmt_cv32_step_compare.sv to use a state machine instead of asynchronous always blocks.  Passes ci_check in lee_moore branch that runs interrupt_test, corev_rand_interrupt_0, and corev_rand_interrupt_1. All pass.

Note, this file was cherry picked from https://github.com/GTumbush/core-v-verif/tree/lee_moore hash 12b6874 with log "Move stop/start_clk to state machine".  It does not contain Lee's latest modifications to this file.  It does not solve issues https://github.com/openhwgroup/cv32e40p/issues/549 and https://github.com/openhwgroup/cv32e40p/issues/548.  Since these issues do not gate RTL freeze I want to get this upgrade in master ASAP.

PR #1799, opened 2023-04-12
RVFI Assert - Check "rvfi_intr.interrupt" count

silabs-robin:
This PR adds an assertion relating to https://github.com/openhwgroup/cv32e40s/issues/355 .

Since the assert in the original issue is actually checking something unrelated, I wanted an assert that checks the actual problem that was discovered.
(ps. Includes a small update to the formal setup too.)

@silabs-oysteink  do you agree that this assert checks the scenario from issue 355?
It produces a similar case:
image


silabs-oysteink (approved):

silabs-oysteink:
Yes, this assertion looks like it is pinpointing issue #355 on cv32e40s.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_support_logic.sv
```diff
@@ -146,6 +155,37 @@ module uvmt_cv32e40s_support_logic
...
+        out_support_if.cnt_irq_ack = $past(out_support_if.cnt_irq_ack) + 1;
+      end
+    end
+  end
+
+
+  // Count rvfi reported interrupts
+
+  logic  do_count_rvfi_irq;
+  always_comb  do_count_rvfi_irq =
```
silabs-hfegran, line 174: Please use `begin/end` for large expressions like this

silabs-hfegran (commented):
Looks good in general, there is a conflict that needs to be resolved (so might as well fix the minor thing I commented on at the same time)

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_support_logic.sv
silabs-robin: What is the benefit?

image

silabs-hfegran: > What is the benefit?
> 
> image

Readability, less error-prone, clearly defined region where always_comb statement begins/ends 

From our coding guidelines:
https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/docs/CodingStyleGuidelines.md :
image

silabs-robin: That is automatically true.
A block is defined as the construct that uses begin/end.

silabs-hfegran: My personal opinion is that things that can (with good readability) be placed on one line can use the non-begin/end forms, other should use begin end. I will not complain about missing begin/end in those cases - even though I personally will use them.

That said, thank you for fixing this and the merge conflicts, approving

silabs-hfegran (approved):

PR #1812, opened 2023-04-17
update xsecure ECC in register file

silabs-krdosvik:


silabs-mateilga (approved):
I've left some suggestions for simplifications, but they are not strictly necessary, so I'm approving.

silabs-hfegran (approved):
Looks good to me

PR #1813, opened 2023-04-17
update xsecure security alert assertions

silabs-krdosvik:


silabs-mateilga (approved):
Looks good to me, nothing to comment on. Approving.

PR #1818, opened 2023-04-18
update xsecure dummy and  assertions

silabs-krdosvik:


cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
```diff
@@ -0,0 +1,827 @@
...
+  //Verify that dummy and hint instructions are either add, and, mul, or bltu instructions
+
+  property p_dummy_hint_instr_is_either_add_and_mul_or_bltu(id_dummy_or_hint);
+
+    if_valid
+    && id_ready
+    ##1 id_dummy_or_hint
+
+    |->
+    (id_instr[6:0] == OPCODE_OP
```
silabs-mateilga, line 244: General comment: We should consider making a general function set based on the "match_instr.." and "is_.." functions in the rvfi_instr_if, for use with any 32 bit instruction. Code like this would be faster to write, easier to get correct, and easier to read if we did that.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
```diff
@@ -0,0 +1,827 @@
...
+
+  a_xsecure_hint_instr_destination_is_r0: assert property (
+    p_dummy_hint_destination_is_r0(id_hint)
+  ) else `uvm_error(info_tag, "The result of a hint instruction is not stored in the x0 GPR.\n");
+
+
+
+  //Verify that the LFSR registers are updated if a dummy or hint instruction is executed
+
+  property p_dummy_hint_update_lfsr0(if_dummy_or_hint, lfsr, lfsr_n, write_lfsr_en, write_lfsr_value);
```
silabs-mateilga, line 445: Two unused variables are taken in to this property.

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
```diff
@@ -0,0 +1,827 @@
...
+    p_dummy_hint_update_lfsr0(
+      if_hint,
+      lfsr0,
+      lfsr0_n,
+      lfsr0_seed_we,
+      lfsr0_seed)
+  ) else `uvm_error(info_tag, "A hint instruction does not update the LFSR0 register.\n");
+
+
+  property p_dummy_hint_update_lfsr1_lfsr2(if_dummy_or_hint, lfsr, lfsr_n, default_value, lfsr_we, lfsr_w_value);
```
silabs-mateilga, line 479: As LFSR changing in normal operation is very important to the functionality, this assert look a bit to "open ended" to build confidence. Please look at stricter alternatives. Maybe check that if an LFSR is enabled and not written, it has to change?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
```diff
@@ -0,0 +1,827 @@
...
+
+    //Make sure we detect a dummy instruction
+    if_dummy_or_hint
+    && id_valid
+    && ex_ready
+    && id_last_op
+
+    |=>
```
silabs-mateilga, line 487: I think this assert needs to be turned around a bit. As of now it verifies that the lfsr is set to a set of three possibilities, without checking that those possibilities actually change. I would start by changing consequent to only  `lsfr != $past(lfsr)`, and see if you get any legal counterexamples to that assert.

silabs-mateilga (changes requested):

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
silabs-krdosvik: Thanks! Will change it in the same manner as for the property below

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
silabs-krdosvik: Thanks! I guess this is about the same as the comment below? I will use lsfr != $past(lfsr) :-)

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_dummy_and_hint_assert.sv
silabs-krdosvik: Agreed!

silabs-krdosvik:
I have added everyone for review because I have made a new pkg file with functions as Marton suggested. Since this can become commonly used file I think you should state your opinions, so I can adjust it before it is overly used. 
I am also open to do/structure it in another way, or undo it if that what that seems to be best. @silabs-mateilga , @silabs-hfegran , @silabs-robin , @MikeOpenHWGroup

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
```diff
@@ -0,0 +1,347 @@
...
+                match_instr_isb_f(instr, INSTR_OPCODE_CSRRC)  ||
+                match_instr_isb_f(instr, INSTR_OPCODE_CSRRWI) ||
+                match_instr_isb_f(instr, INSTR_OPCODE_CSRRSI) ||
+                match_instr_isb_f(instr, INSTR_OPCODE_CSRRCI));
+    end
+  endfunction : is_csr_instr_f
+
+  // NOTE!  This instruction differs from the strict definition of "reading a CSR"
+  //        in the RISCV-spec, as it returns true only if the read value is actually
+  //        stored somewhere.
```
silabs-mateilga, line 213: After a few discussions I have reconsidered on this, and I think these should be changed to follow the spec definition of "reading a CSR". This would also apply to the original in rvfi_instr_if. 

I think these changes should be made before committing this new file, I can chip in of you don't have the time.

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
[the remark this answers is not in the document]
silabs-mateilga: I agree with Robin. Also, these are not cv32e40s specific, so the file can be renamed to reflect that. I'd suggest "support_instr_functions", "isa_instr_functions" or a short version of either where we drop the "_instr_".

silabs-mateilga (changes requested):
Over all good improvements on the previous feedback, and thanks for doing the support functions addition.

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
silabs-krdosvik: Or maybe have two functions, one for each case? I can fix it. I think I have put a lot of work on you already, considering all the PRs that needed reviews, hehe :-)

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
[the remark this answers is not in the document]
silabs-krdosvik: Aaa, yess!

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
silabs-krdosvik: Inside discussion: Marton fixes this (since he must do something similar anyways)

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_functions.sv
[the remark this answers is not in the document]
silabs-krdosvik: I have tried to copy the structure other pkg files uses, but you should probably look into how I have done it in case you preferer something else.

lib/support/support_instr_functions.sv
```diff
@@ -0,0 +1,347 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+`ifndef __UVMT_CV32E40S_INSTR_FUNCTIONS__
+`define __UVMT_CV32E40S_INSTR_FUNCTIONS__
```
silabs-robin, line 17: This does not match the filename, and it says "40S" while being in "lib/".

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_pkg.sv
```diff
@@ -0,0 +1,31 @@
...
+`ifndef __UVMT_CV32E40S_INSTR_PKG_SV__
+`define __UVMT_CV32E40S_INSTR_PKG_SV__
+
+
+/**
+ * Encapsulates all the functions for instruction words
+ */
+package uvmt_cv32e40s_instr_pkg;
+
+   `include "uvmt_cv32e40s_instr_functions.sv"
```
silabs-robin, line 26: Where is this file? Should it be `support_instr_functions.sv`?

lib/support/support_pkgs.sv
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+
+`ifndef __SUPPORT_PKG_SV__
+`define __SUPPORT_PKG_SV__
+
+
+package support_instr_pkg;
+  `include "support_instr_functions.sv"
```
silabs-robin, line 22: Is this file included from multiple places?
I think our compilation might be using a single compilation unit? Not sure. If it does, then one of the includes will probably not include anything.

cv32e40s/fv/dummy_pkg.sv
```diff
@@ -49,6 +49,10 @@ package uvmt_cv32e40s_pkg;
   import cv32e40s_pkg::*;
 endpackage
 
+package uvmt_cv32e40s_instr_pkg;
+  `include "uvmt_cv32e40s_instr_functions.sv"
+endpackage
```
silabs-robin, line 54: Just for everyones information:

Ideally, `dummy_pkg.sv` shouldn't exist.
So, ideally, none of these pkgs should be redefined here.
The only reason they exist is to avoid the non-formal-friendly aspects of the UVM world.

So if this new pkg is 100% formal friendly, it could rather be included via the flists instead of via this dummy pkg.
But if you want, just leave it as it is above and I can take care of fixing it, as I sporadically maintain the core-v-verif formal setup, and nobody is really actively using it anyway.

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_pkg.sv
silabs-krdosvik: I have changed it to what we agreed on in our meeting earlier today :-) But to answer: support_instr_functions.sv was a file containing only function, while support_pkgs.sv was a file containing packages (but currentlig only one). Instead of writing the functions into the package, the packet included the file. This structure makes it possible for future packages to also include the same file, therby also containing the functions in the function files. The reason I did this was because it seems like this structure is used for the other packages. 
image
I hope this answers your question.

cv32e40s/fv/dummy_pkg.sv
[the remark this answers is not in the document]
silabs-krdosvik: Yess!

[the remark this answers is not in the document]
silabs-krdosvik: Should be good now

silabs-krdosvik: okei, hem, i think i understand. But than I will just remove it for now.

lib/support/support_pkgs.sv
silabs-krdosvik: I see. I think this is only included in the place that corresponds to that shown in the comment (given the new structure and name of course).

cv32e40s/fv/dummy_pkg.sv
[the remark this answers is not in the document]
MikeOpenHWGroup: `support_pkg.sv` in `lib/support` is a good as any name I could come up with.   Years from now someone will hate us for this decision, but they can fix it then.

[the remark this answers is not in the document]
silabs-krdosvik: Haha, hem, yeah... well well for now then.

silabs-hfegran:
Thanks for this @silabs-krdosvik, looks good to me

silabs-hfegran (approved):

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_pkg.sv
silabs-robin: > `include "uvmt_cv32e40s_instr_functions.sv"

Can you post a link to the file being included here?

cv32e40s/tb/uvmt/uvmt_cv32e40s_instr_pkg.sv
```diff
@@ -0,0 +1,31 @@
+// Copyright 2023 Silicon Labs, Inc.
```
silabs-robin, line 1: Where is this file used?

silabs-krdosvik: Yess. I have forgot to delete this file, hehe...

PR #1823, opened 2023-04-18
UVMA RVFI - Remove 40s-dependency

silabs-robin:
The current `uvma_rvfi` is locked to only work with the 40s.
This PR removes that dependency.

Test status:
* Hello world passes


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thank you sir!

PR #1824, opened 2023-04-18
Allow compilation without ImperasDV (USE_ISS=0)

MikeOpenHWGroup:
Reinstates capability to compile and run the environment with the ImperasDV reference model excluded.  For example,
```
$ make test TEST=hello-world USE_ISS=0
```
Note that only the `xrun` Makefile has been updated.  I will create another PR for other simulators at a later date.


MikeOpenHWGroup:
Hi @silabs-hfegran, please review this PR when you have time.  I believe it is correct, but I cannot be sure until I get the proper toolchain for the E40S installed.  Until then I have added the `DO NOT MERGE` label.

mk/uvmt/xrun.mk
```diff
@@ -179,8 +179,13 @@ XRUN_FILE_LIST ?= -f $(DV_UVMT_PATH)/uvmt_$(CV_CORE_LC).flist
 XRUN_USER_COMPILE_ARGS += +define+$(CV_CORE_UC)_TRACE_EXECUTION
 XRUN_USER_COMPILE_ARGS += +define+UVM
 ifeq ($(call IS_YES,$(USE_ISS)),YES)
-	XRUN_PLUSARGS += +USE_ISS
+	XRUN_PLUSARGS          += +USE_ISS
+	XRUN_FILE_LIST_IDV     ?= $(DV_UVMT_PATH)/imperas_dv.flat
```
MarioOpenHWGroup, line 183:
```suggestion
	XRUN_FILE_LIST_IDV     ?= $(DV_UVMT_PATH)/imperas_dv.flist
```
Is this a typo?

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
```diff
@@ -51,7 +51,7 @@
 ${DV_UVMT_PATH}/uvmt_cv32e40s_pkg.sv
 
 // ImperasDV
--f ${TBSRC_HOME}/uvmt/imperas_dv.flist
+//-f ${TBSRC_HOME}/uvmt/imperas_dv.flist
```
MarioOpenHWGroup, line 54: Probably you wanted to delete these lines as the filelist is splitted?

mk/uvmt/xrun.mk
MikeOpenHWGroup: A big one!  Thanks.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
MikeOpenHWGroup: Yes, you are right.  I had intended to remove this just prior to removing the DO NOT MERGE label.  Thinking about it now, it makes sense to just deleted the line.  Good catch.

MikeOpenHWGroup:
Good catches @MarioOpenHWGroup.  I've pushed in fixes to this PR.

MikeOpenHWGroup:
@silabs-hfegran, this PR is now ready for your review.  @MarioOpenHWGroup, I believe I have resolved your review comments, please check and let me know if you agree by marking the discussions as resolved.

This _works_ but I am not sure I like it because it relies on conditional compilation (ifdefs).  Unfortunately, we may be forced into this as users who lack an ImperasDV installation need a way to exclude any code that invokes the ImperasDV DPI.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -33,9 +33,11 @@ module uvmt_cv32e40s_tb;
    import cv32e40s_pkg::*;
    import uvmt_cv32e40s_pkg::*;
    import uvme_cv32e40s_pkg::*;
+   `ifdef USE_ISS
```
silabs-hfegran, line 36: This needs to be inside the ifndef FORMAL block, otherwise the following code will break formal runs

silabs-hfegran: edit: on second thought, not strictly necessary, but I think it looks odd this way

silabs-hfegran (commented):
In general I think this looks good, just a bit odd as mentioned above with giving USE_ISS priority over !FORMAL

silabs-hfegran:
I merged this one a bit too fast - This breaks compilation with imperas-dv, @silabs-robin will have a fix up shortly (no need to revert)

MikeOpenHWGroup:
Thanks @silabs-hfegran and @silabs-robin.

PR #1827, opened 2023-04-19
adding CV_TOOL_PREFIX to XRUN executable to account for separate LSF …

LeeHoff:
…and Tool strings

I am using CV_SIM_PREFIX for my LSF command, but I also need to load the module file for xcelium before the xrun command. I am using CV_TOOL_PREFIX = "module load xcelium;"


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Great idea @LeeHoff.

PR #1828, opened 2023-04-20
Adding RISCOF riscv-arch-test suite simulation support for cv32e40p and related flow updates

dd-vaibhavjain:
Purpose of this PR is to add support for running RISCOF riscv-arch-test suite within core-v-verif environment.

The files added for the flow follow the riscof documentation at : https://riscof.readthedocs.io/en/stable/

### **New Files Added:**
1. **RISCOF setup related files are added under a new dir: core-v-verif/cv32e40p/sim/riscof/** 
- config.ini -> config file for riscof
- cv32e40p/ -> directory with required cv32e40p DUT intput files for ricof
- sail_cSim/ -> directory with required sail reference model input files for riscof
- README.md -> readme for this riscof setup explaining riscof run steps and requirements
- Makefile -> this is same as the makefile in core-v-verif/cv32e40p/sim/uvmt
 
2. **new uvm test for riscof:**
core-v-verif/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_riscof_firmware_test.sv -> _contains riscof related test configs and signature writer function_

3. **new make file for riscof flow support:**
core-v-verif/mk/riscof.mk -> _contains all new variables and targets for riscof in one place_

4. Other major file updates:
core-v-verif/mk/uvmt/vsim.mk , xrun.mk, vcs.mk, uvmt.mk -> _to add new riscof dut sim targets for different simulators_ . Updates are strictly added as new block of targets for riscof while keeping all existing code as such.

- The directory structure chosen in this way to avoid disrupting existing setup and have good distinction for riscof but still be able to make use if needed the exisiting bench's debugging and other uvm infrastructure
- the riscof setup files are created using initial templates provided by the riscof package and added with cv32e40p related commands
- signature writer is added as a function inside the uvmt_cv32e40p_riscof_firmware_test.sv rather than creating a new class of signature writer, as it seemed only relevant in this context of riscof and not something which has relevance for reusability
- signature writer has a FIXME at the moment. As this has some hardcoded addresses in line with linker file and some dependency on current data section of memory. But may need a way to parse the hex file and find the start and end of signature addresses to keep it generic and independent of linker file but this may have a performance penality to consider.
- more riscof config files and dut targets with different isa/platform values can be added but at the moment only 1 config is added to run with all possible standand extentions for cv32e40p and available in the test suite RV32IMFCZicsr_Zifencei.

 The setup and flow works with vsim and xrun and a riscof report is able to generate at the end with some failures.


MikeOpenHWGroup (approved):
Good job @dd-vaibhavjain!   I am not sure I agree with your decision to handle the signature writer in the test_case class.  My thought about this is that it could be done as a virtual sequence running on the OBI UVM Agent. Having said that, your approach _does_ work, and  I can see no reason _not_ to accept it.

Thanks for your contribution.

dd-vaibhavjain:
Hi @MikeOpenHWGroup,

Thanks a lot for accepting this PR and your kind words and valuable feedback!
 
On the signature writer here is my reasoning. 
My initial thinking was same as yours to create a new core specific signature writer class with a virtual peripheral like the one which is currently existing in the testbench. But I dropped this approach due to some issues and lack of specific benefits which I could think of with this.
(1) Main issue I encountered was that the signature start/end address varies across different riscof tests and is not fixed or pre determined and as such the virtual peripheral for this purpose if say rely on end-of-test writes to trigger this signature writer, had no way to figure out the signature region on its own through obi monitor. 
(2) Also monitoring obi data to check/determine a write of signature header or footer unique key words to figure out the start/end address is not possible, as in riscof tests this keyword does not have an associated store instruction or as such an obi transaction and is just pre-loaded in the memory as part of data section.

Thus I felt that as such monitoring of OBI by adding a new virtual peripheral in this case doesnt help much in our case. And could be just simply handled in the test itself as uvm test already monitors for end of test and we can just do signature dump task at that point .
Doing same thing in vp is also possible but I didn't see any specific benefit assuming the obi monitoring is not helpful, rather some minor drawbacks of this would be that we will have to touch some exisitng TB env files to add a new virtual peripheral and as such the TB will have more riscof code embedded in it at various places. 
My thinking at the time was that by adding a new uvmt_<core>_riscof_test.sv we can probably as much as possible limit the code for riscof at one place through this test.

Let me know your thoughts on this.

Thanks,
Vaibhav

MikeOpenHWGroup:
Good points all @dd-vaibhavjain.  In verification there is always more than one path to success and your path is prefectly valid.  Good job!

PR #1829, opened 2023-04-20
Xsecure c test

silabs-krdosvik:


cv32e40s/tests/programs/custom/xsecure_test/xsecure_test.c
```diff
@@ -96,97 +182,194 @@ void u_sw_irq_handler(void) {
...
 
 }
 
 
+
 int main(void){
 
   unexpected_irq_beh = 0;
 
+  // seed the rand() function, which is used in some of the test functions
```
silabs-hfegran, line 358: Please use a source of random from the systemverilog environment (have a look at interrupt_test.c, `random_num(...)` on how to do that. 

As you seed `rand()` with `srand(1)`, this test will always use the same fixed (but random) value. Furthermore, rand() is not guaranteed to be stable across different simulators, versions and systems, so this can cause issues when trying to reproduce a specific failure

silabs-hfegran (changes requested):
Thanks for this @silabs-krdosvik, In general it looks good, but there are some things I would like to have changed before this is merged, have a look at my comments above.

silabs-hfegran (approved):

PR #1830, opened 2023-04-20
Xsecure vplan

silabs-krdosvik:


cv32e40s/docs/VerifPlans/Simulation/xsecure/CV32e40s_xsecure_vplan.csv
silabs-robin, line 1: Here are just my immediate thoughts:
The deleting/adding could have been done in just one commit.
The rename introduces a new scheme, but that seems to be for the better.
Since the vplan update and the rename happens at the same time, the csv is not diffable.

cv32e40s/docs/VerifPlans/Simulation/xsecure/xsecure.xlsx
silabs-robin, line 1: The requirement locations says "CV32E40S User manual ++VERSION++" and should maybe be updated so we at a later time don't get confused about what the latest available version was at the time of the last xsecure vplan update..

silabs-robin (changes requested):
Looks good. I only have the comments below. Mainly that I would prefer if the document reference uses a version number. If there is a spec change in the future then we can use the version number to get a narrower diff of what needs reconsidering.

cv32e40s/docs/VerifPlans/Simulation/xsecure/xsecure.xlsx
silabs-krdosvik: Yess!

silabs-krdosvik:
New PR with update: https://github.com/openhwgroup/core-v-verif/pull/1832

PR #1836, opened 2023-04-23
Vplan try3

silabs-krdosvik:
try 1: https://github.com/openhwgroup/core-v-verif/pull/1830 (comments, but made next PR to address the comments)
try 2: https://github.com/openhwgroup/core-v-verif/pull/1832 (was merged before review)
try 3: the current one


silabs-mateilga (approved):
Hard to review the excel vplans, but the plan looks good. Only question is if we should await merging until the remaining links to coverage are added? I'm approving anyway, but we should consider delaying to save work effort.

silabs-krdosvik:
> Hard to review the excel vplans, but the plan looks good. Only question is if we should await merging until the remaining links to coverage are added? I'm approving anyway, but we should consider delaying to save work effort.

Conclusion from offline discussion:
No time for doing statistical distribution check. But we add check that verify that all possible situations is reached.

PR #1838, opened 2023-04-24
Enabling floating point register and CSR check

YoannPruvost:


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
LGTM @YoannPruvost.  Also passes a local `ci_check` regression (w/o ImperasDV).

PR #1841, opened 2023-04-25
Stand-up the ci_check regression for CV32E20

MikeOpenHWGroup:
This PR sets up a preliminary CI regression for CV32E20.  The command-line to run it:
```
$ cd bin
$ ./ci_check --core cv32e20 --simulator xrun --iss 0
```

If you have an ImperasDV license, omit the `--iss 0` switch.

### DIFF
--- a/.metrics.json
+++ b/.metrics.json
@@ -9,6 +9,12 @@
   "builds": {
     "list": [
       {
+        "name":     "uvmt_cv32e20",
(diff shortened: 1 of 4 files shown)


szbieg:
My comment does not refer to the changes, but I hope it would be relevant and worth noting that package distutils is [deprecated](https://peps.python.org/pep-0632/).

szbieg (approved):

PR #1842, opened 2023-04-26
Bug correction for xpulp instruction generation and basic xpulp test.…

XavierAubert:
… This commit redo modifications that were inside PR#1819. This PR has been cancelled as it contained modifications to files that has been renamed in PR#1840, and it was faster and simpler to just redo the modifications from a clean state.


MikeOpenHWGroup (approved):
Thanks @XavierAubert, LGTM.   Also passes local CI.

PR #1844, opened 2023-04-27
ci: check that at least one test is run

cathales:


JeanRochCoulon:
ok @yanicasa ?

yanicasa:
👍this will fix when simulation have no test but this not prevent the case when pipeline has no job.

Anyway, for me we can merge

PR #1845, opened 2023-04-27
Corrections and PULP instruction generation updated

XavierAubert:
Typo and code correction here and there. 

Added basic SIMD/MAC streams and improved base pulp instruction generation to be more flexible.


MikeOpenHWGroup (approved):
LGTM.  Passes local CI under both xrun and vsim (sans ImperasDV).

PR #1849, opened 2023-04-28
Updates to asserts for formal regression

silabs-mateilga:
Some fixes triggered by cleaning up formal regression:

- Updates to debug_asserts to start using common support logic for exception triggers
- Fixed csr helper functions in rvfi_instr_if to match spec definition, added csr_write_intended, which gives a intended value for the update of a csr
- cleaned up and expanded scope for a_rlblifts_lockedexec in pmprvfi_asserts


lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -365,39 +367,56 @@ function automatic logic is_csr_instr(logic [11:0] csr_addr = 0);
...
+    return is_csr_instr(csr_addr);
+  end
+endfunction
+
+// returns intended write data for any CSR write instruction,
+// without regard for what the legal values are in the CSR
+// input current value of the csr, and the csr address
+// NOTE: that this will work for CSRRW with unspecified csr address,
+// but CSRRS/CSRRC will give incorrect return values
+function automatic logic [DEFAULT_XLEN-1:0] csr_write_intended( logic [DEFAULT_XLEN-1:0] csr_rdata,
```
silabs-robin, line 402: Can we call it `csr_wdata_intended`?

Because I was just reading the assert that uses this function and even if I know what the function does, my brain reads "csr_write_intended" as "this csr instructions intends to write" and not "this is the _data_ that the instruction tries to write.

And because it is consistent with `rvfi_csr_<csrname>_wdata`.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -365,39 +367,56 @@ function automatic logic is_csr_instr(logic [11:0] csr_addr = 0);
   end
```
silabs-robin, line 367: Btw, I think the `32'h0 |` above might be unnecessary.

silabs-robin (approved):
Approving based on:
1. Debug asserts switched to no longer duplicate the support logic, nice.
2. Pmprvfi change seems to be both correct and an improvement.
3. Tb just connects signals for the pmprvfi updates.
4. The instr if logic seems correct to me as well.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: Probably, yes. Might fix later.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -365,39 +367,56 @@ function automatic logic is_csr_instr(logic [11:0] csr_addr = 0);
...
+    return is_csr_instr(csr_addr);
+  end
+endfunction
+
+// returns intended write data for any CSR write instruction,
+// without regard for what the legal values are in the CSR
+// input current value of the csr, and the csr address
+// NOTE: that this will work for CSRRW with unspecified csr address,
+// but CSRRS/CSRRC will give incorrect return values
+function automatic logic [DEFAULT_XLEN-1:0] csr_intended_wdata( logic [DEFAULT_XLEN-1:0] csr_rdata,
```
silabs-krdosvik, line 402: Just to make sure I follow: In other assertion files is_csr_write should be changed to csr_intended_wdata, because csr_intended_wdata is now what used to be is_csr_write, or?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -365,39 +367,56 @@ function automatic logic is_csr_instr(logic [11:0] csr_addr = 0);
...
-// NOTE!  This instruction differs from the strict definition of "reading a CSR"
-//        in the RISCV-spec, as it returns true only if the read value is actually
-//        stored somewhere. If you need the spec definition, please use the
-//        rvfi_csr_if-signals directly.
+// This function follows the spec definition of a csr read
 function automatic logic is_csr_read(logic [11:0] csr_addr = 0);
-  if (rvfi_rd1_addr != 0) begin
-    return is_csr_instr(csr_addr);
-  end else begin // rd is X0, not a read instruction
+  if (  (rvfi_rd1_addr == 0) &&
```
silabs-krdosvik, line 372: When making this change, you should also make sure that the assertions files that used the old version of the is_csr_rdata either dont fail, or make an alternative to what was there before and exchange the assertions to use that signal instead. (Unless I have misunderstood of course.)

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -365,39 +367,56 @@ function automatic logic is_csr_instr(logic [11:0] csr_addr = 0);
...
 function automatic logic is_csr_write(logic [11:0] csr_addr = 0);
-  if (csr_addr == 0) begin //not specified
-    return  ( (rvfi_rs1_addr != 0) && match_instr_isb(INSTR_OPCODE_CSRRW))  ||
-            ( (rvfi_rs1_addr != 0) && match_instr_isb(INSTR_OPCODE_CSRRS))  ||
-            ( (rvfi_rs1_addr != 0) && match_instr_isb(INSTR_OPCODE_CSRRC))  ||
-            match_instr_isb(INSTR_OPCODE_CSRRWI) ||
-            //TODO:MT add set and clear with immediate nonzero
-            match_instr_isb(INSTR_OPCODE_CSRRSI) ||
-            match_instr_isb(INSTR_OPCODE_CSRRCI);
+  if (  ( (rvfi_rs1_addr == 0) &&
```
silabs-krdosvik, line 383: Same comment as for is_csr_read

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin: Excellent point.

silabs-krdosvik:
Apart from my previous comments it looks good :-)

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-mateilga: This has been checked, found no problems.

silabs-hfegran (approved):

PR #1852, opened 2023-04-28
fix idv flist inclusion

silabs-robin:
Ref: https://github.com/openhwgroup/core-v-verif/pull/1824


silabs-hfegran:
Thanks for the fix @silabs-robin!

silabs-hfegran (approved):

PR #1855, opened 2023-05-02
Timeout fixes

silabs-hfegran:
- Added watchdog to timeout excessive sequences of irq and single step
- ~Added watchdog to timeout after NMIs~ edit: This happened in a previous PR
- Turned off linker relaxation for tests written in assembly only
- Removed mepc increment in case of (random) bus errors on instr. interface in random tests.


cv32e40s/env/uvme/vseq/uvme_cv32e40s_irq_ss_timeout_vseq.sv
```diff
@@ -0,0 +1,99 @@
...
+
+   super.new(name);
+
+endfunction : new
+
+
+task uvme_cv32e40s_irq_ss_timeout_vseq_c::body();
+
+   int unsigned irq_cnt_min  = 10;
+   int unsigned ss_cnt_min   = 10;
```
silabs-robin, line 63: Are these two unused?

silabs-robin (approved):
Approving based on:
1. New vseq is a smart solution.
2. Relaxation doesn't seem appropriate in pure-assembly tests.
3. The problem with sticky irqs has been explained offline.
4. As long as we don't have "permanent bus error addressess", leaving mepc untouched in the handler makes sense.
5. My one remark is not reason to block the PR.

silabs-robin (approved):

cv32e40s/env/uvme/vseq/uvme_cv32e40s_irq_ss_timeout_vseq.sv
```diff
@@ -0,0 +1,93 @@
...
+       forever begin
+         @(posedge clk) begin
+           if (rvfi_valid) begin
+             // Threshold exceeded
+             if (irq_cnt > cfg.irq_min_limit &&
+                 single_step_cnt > cfg.single_step_min_limit &&
+                (single_step_cnt + irq_cnt) >= cfg.irq_single_step_threshold)
+             begin
+               `uvm_info("IRQ_SS_TIMEOUT_WATCHDOG", $sformatf("IRQ/SINGLE STEP timeout: single step count: %0d, irq count: %0d, threshold: %0d", single_step_cnt, irq_cnt, cfg.irq_single_step_threshold), UVM_LOW);
+               cntxt.vp_status_vif.exit_valid = 1;
```
MikeOpenHWGroup, line 81: Here we are replicating some of the end-of-sim logic in the "firmware_test" and by-passing the normal UVM shutdown flow.  Could this vseq signal the firmware_test to end?

MikeOpenHWGroup (changes requested):

cv32e40s/env/uvme/vseq/uvme_cv32e40s_irq_ss_timeout_vseq.sv
silabs-hfegran: > Here we are replicating some of the end-of-sim logic in the "firmware_test" and by-passing the normal UVM shutdown flow. Could this vseq signal the firmware_test to end?

How do you propose we do that? (I thought that this is exactly what I am doing here, the only real difference is that I do not explicitly set test_passed or test_failed, for the simple reason that if the test fails, it will write the fail-status to the end test register and test will terminate. If the test is either long running or in an unrecoverable state it will not be able to write its status prior to exiting, and the pass/fail condition should be dependent on ISS checking only - thus the self-test part will say passed.

The test-termination mechanism already in place, uses a vsequence to check for OBI writes to the status flags memory location, and the vseq in this PR should terminate the test in the same manner as that sequence, but without checking for memory accesses. 

This is how we terminate regular tests upon writing to the status register:
https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/env/uvme/vseq/uvme_cv32e40s_vp_status_flags_seq.sv

The firmware test in use is this one (I have no clue why our only firmware test resides in the compliance-tests directory) :
https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/tests/uvmt/compliance-tests/uvmt_cv32e40s_firmware_test.sv
and this section of code is what actually terminates the test:
```
   phase.raise_objection(this);
   @(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
   repeat (33) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
   `uvm_info("TEST", "Started RUN", UVM_NONE)
   // The firmware is expected to write exit status and pass/fail indication to the Virtual Peripheral
   wait (
          (vp_status_vif.exit_valid    == 1'b1) ||
          (vp_status_vif.tests_failed  == 1'b1) ||
          (vp_status_vif.tests_passed  == 1'b1)
        );
   repeat (100) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
   //TODO: exit_value will not be valid - need to add a latch in the vp_status_vif
   `uvm_info("TEST", $sformatf("Finished RUN: exit status is %0h", vp_status_vif.exit_value), UVM_NONE)
   phase.drop_objection(this);
```
^ looking at this code raises a few questions that you may be able to answer - why do we require 100 cycles of valid exit status to actually exit a test? This is why the vseq i made gives quite spammy output when terminating - though I find that a good thing, as that makes the warning more visible. 

I could overload the vp_status_flags sequence to do a whole lot of other things than just checking for those memory writes; but I think it is cleaner to have this runaway test-detection in a separate sequence. e.g. one sequence for each abnormal terminate condition.

MikeOpenHWGroup: > How do you propose we do that?

It would require the vseq to terminate and return a value.  Not sure if that is worth it.

> I have no clue why our only firmware test resides in the compliance-tests directory

This is a hang-over from the days when I thought we would need up to four distinct types of [test-programs](https://docs.openhwgroup.org/projects/core-v-verif/en/latest/uvm_tests.html#test-programs-in-the-core-v-verif-uvm-environment) in core-v-verif.  This should have been refactored long ago, but as we generally have only one UVM testcase it has never been a priority.

> why do we require 100 cycles of valid exit status to actually exit a test?

Actually, we do not.  The above code is not sensitive to the number of cycles that `exit_valid` is asserted.  It merely waits 100 clock cycles after it is asserted before dropping its objection.  An old (and probably unnecessary) habit of mine intended to allow for the collection of additional cycles so that the termination condition is more easily viewable on waves.

> This is why the vseq i made gives quite spammy output when terminating - though I find that a good thing, as that makes the warning more visible.

Should it be a warning or error?

> I could overload the vp_status_flags sequence to do a whole lot of other things than just checking for those memory writes; but I think it is cleaner to have this runaway test-detection in a separate sequence. e.g. one sequence for each abnormal terminate condition.

Agreed - I like the "runaway test-detection".  But as you are not generating errors or fatals it seems odd that you want to terminate the sim.

MikeOpenHWGroup:
Approving/Merging.  I am still not 100% comfortable with the new timeout, but I do see the utility of it and a "proper" fix would involve a lot of re-work of long-standing and stable code.

PR #1857, opened 2023-05-02
add benchmarks

cathales:
Make it possible to easily run CoreMark and Dhrystone. Run them in CI.


MikeOpenHWGroup:
> we should discuss about the criteria we want to evaluate to get a pass/fail.

If there are no pass/fail criteria, these tests should not be part of the CI regression. The purpose of a CI regression is to have a reasonable (not complete) sanity check of the DUT and TB in the shortest possible time.  So, only those tests that contribute unique coverage and have an unambiguous pass/fail indication should be part of a CI regression.  Typically, performance tests are part of a full regression, not a CI regression.

> I can open another issue do discuss the pass/fail criteria for benchmark. Is it okay for you?

Yes, that would be a good idea.  Thanks!

JeanRochCoulon:
Indeed, we will add a criteria based on the number of execution cycles. Of course, a modification of compiler or RTL could impact this and could contrain to update the number, but it is what we want !

cathales:
See #1884

PR #1866, opened 2023-05-04
Clone Verif Environment - Script

silabs-robin:
This PR adds a script to clone the external 40x verif environment subtree repo.

It is clean, simple, and extensible to other cores.

I can rewrite this in Python or make if we think that would make things better. But I honestly think this is an uncomplicated problem and that this simple shell script is a fitting solution.

@silabs-hfegran, @MikeOpenHWGroup, this continues #1867

### DIFF
--- a/bin/clonetb
+++ b/bin/clonetb
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+
+# Copyright 2023 Silicon Laboratories Inc.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+#
+# Licensed under the Solderpad Hardware License v 2.1 (the "License"); you may
+# not use this file except in compliance with the License, or, at your option,
+# the Apache License version 2.0.
+#
+# You may obtain a copy of the License at
+# https://solderpad.org/licenses/SHL-2.1/
+#
+# Unless required by applicable law or agreed to in writing, any work
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Replace "cv32e40x/" dir
+rm -rf cv32e40x  &&
+git clone  -b cv32e40xdev_subtree_gitmerge  https://github.com/silabs-robin/core-v-verif.git  cv32e40x  &&
+
+# Ignore the cloned directory
+echo cv32e40x >> .git/info/exclude
+git ls-files cv32e40x/ | xargs git update-index --skip-worktree
--- a/bin/clonetb
+++ b/bin/clonetb
@@ -20,10 +20,97 @@
 # limitations under the License.
 
 
-# Replace "cv32e40x/" dir
-rm -rf cv32e40x  &&
-git clone  -b cv32e40xdev_subtree_gitmerge  https://github.com/silabs-robin/core-v-verif.git  cv32e40x  &&
+# Description:
+#   This script is used to clone core-specific verification environments.
+#   (E.g. "core-v-verif/cv32e40x/".)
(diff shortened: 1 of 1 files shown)


bin/clonetb
```diff
@@ -0,0 +1,116 @@
...
+
+clone() {
+  check_env_vars
+
+  rm -rf ${CV_CORE}
+  git clone  -b ${VERIF_ENV_BRANCH}  ${VERIF_ENV_REPO}  ${CV_CORE}
+}
+
+clone_cv32e40x() {
+  CV_CORE=cv32e40x
```
MikeOpenHWGroup, line 73: Two comments about this:
1. It would ease usability and maintenance if these were clearly visible as global vars at the top of the file.
2. Would it be possible to set things up so that if the user control these via shell env vars or command-line options?

bin/clonetb
```diff
@@ -0,0 +1,116 @@
...
+ignore_cloned_directory() {
+  check_env_vars_cvcore
+
+  echo ${CV_CORE} >> .git/info/exclude
+  git ls-files ${CV_CORE} | xargs git update-index --skip-worktree
+}
+
+unignore_cloned_directory() {
+  check_env_vars_cvcore
+
```
MikeOpenHWGroup, line 91: Anytime you use `sed` or `awk` there should be some comments to explain extactly what is happening.  :stuck_out_tongue_closed_eyes:

MikeOpenHWGroup (changes requested):

bin/clonetb
silabs-robin: Comments are now added, and each pipe is split to its own line.

bin/clonetb
silabs-robin: > It would ease usability and maintenance if these were clearly visible as global vars at the top of the file.

It would be the opposite of "progressive disclosure".
For normal usage there is no need to know, and for advanced usage it is trivial to figure it out.

Nevertheless, I've added a comment at the top of the file which explains the env vars.
It can be useful when modifying the script at a later time.

--------

> Would it be possible to set things up so that if the user control these via shell env vars or command-line options?

Having two ways of doing it?

I also played with that idea (and know a way to do it), but ultimately I think it adds unnecessary clutter.
Env vars like `CV_CORE` are already in use in core-v-verif.
It's nice and quick and clean and gets things done.

bin/clonetb
MikeOpenHWGroup: Good comments, thanks.

The resolution of a conversation is the responsibility of the individual who opened the conversation (in this case, me), or a Committer.

bin/clonetb
MikeOpenHWGroup: >  Having two ways of doing one thing?

We already have that.  The Makefiles are written such that you can either set vars on the command-line or as shell environment variables:
```
$ make test TEST=hello-world SIMULATOR=xrun
```
...is the same as...
```
$ export SIMULATOR=xrun
$ make test TEST=hello-world
```

silabs-robin: > We already have that. The Makefiles are written such that you can either set vars on the command-line or as shell environment variables
> [...]
> ...is the same as...
> [...]

Oh, I thought you were asking for command-line options, like `$ myprogram --myarg`.
Those two examples you gave are both using environment variables, not command-line options.
Our makefiles are not written to facilitate that, the syntax of make allows macro definitions to be given as arguments.
By default, posix (and gnu) shell allows you to pass environment variables by prepending them to a command, like `$ MYVAR=somevalue mycommand`, and the "Examples:" section of my comment header shows an example of this.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Approved.  I shall defer to @silabs-hfegran for final approval and merge.

silabs-hfegran (approved):

silabs-hfegran (approved):

PR #1877, opened 2023-05-05
ISACOV : Add ifdef to avoid VCS compilation error due to unsupported 'with' key word

AyoubJalali:
As we discuss, for the CVA6 env we have a vcs compilation error related to the vcs version, the 'with' isn't supported for now, the fix is programmed for July, so to integrate the ISACOV in the CVA6 env, I add ifdef (a final solution after a lot of discuss) as a temporary fix.

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -15,6 +15,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 
+`ifdef UNSUPPORTED_WITH //TODO - Remove ifdef when the issue in VCS simulator is fixed
+  `define WITH iff
+`else
+   `define WITH with
+`endif
+
 covergroup cg_executed_type(
     string name,
     instr_name_t instr_name
@@ -49,24 +55,24 @@ covergroup cg_zb_rstype(
   cp_rd: coverpoint instr.rd;
 
   cp_rd_rs_hazard: coverpoint instr.rd {
-    ignore_bins IGN_RS_HAZARD_OFF = {[0:$]} with (!reg_hazards_enabled);
+    ignore_bins IGN_RS_HAZARD_OFF = {[0:$]} `WITH (!reg_hazards_enabled);
     bins RD[] = {[0:31]} iff (instr.rd == instr.rs1);
   }
 
   cross_rd_rs: cross cp_rd, cp_rs {
-    ignore_bins IGN_OFF = cross_rd_rs with (!reg_crosses_enabled);
+    ignore_bins IGN_OFF = cross_rd_rs `WITH (!reg_crosses_enabled);
   }
 
   cp_rs_value: coverpoint instr.rs1_value_type {
-    ignore_bins POS_OFF = {POSITIVE} with (!rs_is_signed);
-    ignore_bins NEG_OFF = {NEGATIVE} with (!rs_is_signed);
-    ignore_bins NON_ZERO_OFF = {NON_ZERO} with (rs_is_signed);
+    ignore_bins POS_OFF = {POSITIVE} `WITH (!rs_is_signed);
+    ignore_bins NEG_OFF = {NEGATIVE} `WITH (!rs_is_signed);
+    ignore_bins NON_ZERO_OFF = {NON_ZERO} `WITH (rs_is_signed);
   }
 
   cp_rd_value: coverpoint instr.rd_value_type {
-    ignore_bins POS_OFF = {POSITIVE} with (!rd_is_signed);
-    ignore_bins NEG_OFF = {NEGATIVE} with (!rd_is_signed);
-    ignore_bins NON_ZERO_OFF = {NON_ZERO} with (rd_is_signed);
+    ignore_bins POS_OFF = {POSITIVE} `WITH (!rd_is_signed);
+    ignore_bins NEG_OFF = {NEGATIVE} `WITH (!rd_is_signed);
+    ignore_bins NON_ZERO_OFF = {NON_ZERO} `WITH (rd_is_signed);
   }
 
   `ISACOV_CP_BITWISE(cp_rs_toggle, instr.rs1_value, 1)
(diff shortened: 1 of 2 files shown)


lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -15,6 +15,1448 @@
 //
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 
+`ifdef UNSOPPRTED_WITH //ifdef to replace 'with' with 'iff' VCS version don't support 'with' key word
```
silabs-robin, line 18: `UNSOPPRTED` seems to be a typo.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -15,6 +15,1448 @@
...
+    ignore_bins IGN_RS_HAZARD_OFF = {[0:$]} iff (!reg_hazards_enabled);
+    bins RD[] = {[0:31]} iff (instr.rd == instr.rs1);
+  }
+
+  cross_rd_rs: cross cp_rd, cp_rs {
+    ignore_bins IGN_OFF = cross_rd_rs iff (!reg_crosses_enabled);
+  }
+
+  cp_rs_value: coverpoint instr.rs1_value_type {
+    ignore_bins POS_OFF = {POSITIVE} iff (!rs_is_signed);
```
silabs-robin, line 62: I am not sure what the `iff` adds here.

Using `with`, the bin would not exist.
Using `iff`, the bin does still exist, and for those instances were it was supposed to not exist, we disallow sampling of it?
Is that the correct interpretation?
If so, then we disallow sampling of something that nevertheless should not exist, so I can't see what the benefit is.

silabs-robin (changes requested):

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: I talking about VCS, because i think the iff work so fine, and effectively works as with

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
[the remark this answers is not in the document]
AyoubJalali: I made ifdef <TAG>
I duplicate all the cover group but with iff instead of with (with cause a compilation error for us)

AyoubJalali: I think I should modify the TAG

MikeOpenHWGroup (changes requested):
We need to come up with a solution that does not involve duplicating >1000 lines of code.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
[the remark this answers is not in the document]
silabs-robin: > We need to come up with a solution that does not involve duplicating >1000 lines of code.

@MikeOpenHWGroup 

> Could the "with/iff"-differentiation be made local to the place where it is used instead

Here is one example:
```
`ifdef WITH_CONSTRUCT_UNSUPPORTED
  `define  WITHIFF  iff
`else
  `define  WITHIFF  with
`endif
```

Her is another:
```
`ifndef  WITH_CONSTRUCT_UNSUPPORTED
  cp_rs_value: coverpoint instr.rs1_value_type {
    ignore_bins POS_OFF = {POSITIVE} iff (!rs_is_signed);
`endif
```

_(NB! I did not check the correctness of the logic nor syntax in those two examples, I just jotted something down.)_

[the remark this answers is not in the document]
silabs-robin: NB! Before going with the first option, read my other comments to see what I think of the usage of `iff` here.

AyoubJalali:
> We need to come up with a solution that does not involve duplicating >1000 lines of code.

I think l'm lazy boy, because i don't want a solution that add **ifdef** in every line **with** is in there, so let's be it, i'll change it

MikeOpenHWGroup:
I am ready to approve this, but before I do, please have a look at #1881 to see if `iff` gets you what you expect.

silabs-robin (changes requested):
I must agree that 500 lines of duplicated code is better than 1000 lines of duplicated code.
But preferably there would be 0 lines of duplicated code.
So I would like to know why Mike's suggestion of defining a `WITH` macro can't be used.

--------

In addition, these two comments have not been sufficiently addressed:
https://github.com/openhwgroup/core-v-verif/pull/1877#discussion_r1186105098
https://github.com/openhwgroup/core-v-verif/pull/1877#issuecomment-1540565027

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: > Using `iff`, the bin does still exist, and for those instances were it was supposed to not exist, we disallow sampling of it?

No for the ignore_bins will ignore them & also for illegal, and actually it's exclude from coverage

AyoubJalali: image

AyoubJalali:
@MikeOpenHWGroup @silabs-robin, what mike suggest works after a little change so now there less than 500 lines duplicated

silabs-robin (approved):
> what mike suggest works after a little change

Excellent! I'm glad it worked. Looks like a nice temporary fix for the VCS problem, and hopefully everyone got what they wanted in the end.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -15,6 +15,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 
+`ifdef UNSUPPORTED_WITH
```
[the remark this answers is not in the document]
JeanRochCoulon: We can add a TODO or create a github issue to remind it. To me, both can be done, comment can include the github issue reference.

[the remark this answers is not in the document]
silabs-robin: Either one is fine by me.

_(I have a slight preference for doing it in code, but both options are good.)_

I have given my approval of this PR nonetheless.

PR #1880, opened 2023-05-08
add to and update trigger assert file

silabs-krdosvik:
This PR updates the trigger asserts, and adds more asserts.
It uses the support logic in #1876, so it should not be merged before that PR.


cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-robin, line 1: Is the compilation time still big? (3 min instead of 30 sec.)

If yes, then this might be the best time to try and find a solution.

(Some of those generate blocks in `_triggers_assert_cov` are the cause. And now that they get split into different files it might become difficult to pinpoint the cause at a later time.)

silabs-krdosvik: Do you mean split into the support logic and trigger module? I cant time it untill the the #1876 PR is in (since I get errors without it). I think the compilation time is reduces a lot, but I am not sure if it is as small as 30 seconds. I will com back to you with a time estimate when #1876 is in :-)

silabs-robin: > Do you mean split into the support logic and trigger module?

I don't have a full overview of what the new changes include, but here is my point:

Today, cv32e40s/dev formal compiles in 3 minutes.
If one comments out everything within the triggers_assert_cov module, then it takes only 30 seconds. (In particular, the second generate block from the bottom, IIRC.)
So I think we should look for a way to not 6x the compilation time (if possible), and I think it is better to do it now than later.

silabs-mateilga (commented):
I've left a few comments, but I'll postpone a full review until the generate/compile time issues are resolved.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-krdosvik: The fv compilation time is now 1 min and 30 sec. So this addition will increase the time by 1 minute. I will look more into reducing the compilation time.

silabs-krdosvik: Measurements [compile time]:
- No trigger or support module: ish 1 min
- 128 memory access: ish 1 min and 30 sec with the new file structure
- 13 memory access: ish 1 min

It seems that the measurements vary a little for every time I run the same experiment, therefore "ish"-measurements.
It also seems like 1 memory access take the same time as 13.
It didn't seem that modifying the generate blocks had any effect.

silabs-mateilga (changes requested):
Generally very solid code! It's a bit hard to review such large changes, but there isn't much I could find to point out, so I think we are very close to merging here. Left a few questions and comments that should be addressed pre merge.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_csr_if.sv
```diff
@@ -67,14 +67,14 @@ interface uvma_rvfi_csr_if_t
   // -------------------------------------------------------------------
 
 
-  function automatic logic [XLEN-1:0] read();
-    read = (rvfi_csr_rdata & rvfi_csr_rmask);
-  endfunction : read
+  function automatic logic [XLEN-1:0] pre_state();
```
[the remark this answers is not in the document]
silabs-krdosvik: It is checked. :-)

silabs-mateilga (approved):
All issues resolved, approving

PR #1882, opened 2023-05-08
Preliminary RV32I decoder & support

silabs-hfegran:
Initial work on a decoder for RV32I

### DIFF
--- a/lib/support/isa_support.sv
+++ b/lib/support/isa_support.sv
@@ -164,10 +164,10 @@
     // Compressed
     CEBREAK,
     // Pseudo name, class of instructions
-    STORE_INSN,
-    LOAD_INSN,
+    STORE_INSTR,
+    LOAD_INSTR,
     // Unknown for instructions that cannot be decoded
-    UNKNOWN
+    UNKNOWN_INSTR
   } instr_names_e;
 
   // GPR Registers
@@ -347,19 +347,6 @@
     FUNCT3_AND     = 3'b111
   } op_minor_opcode_e;
 
-
-  // Specific function subtypes
-  typedef struct packed {
-    csr_name_e           csr;
-    union packed {
-      gpr_t         rs1;
-      logic [19:15] uimm;
-    }n;
-    csr_minor_opcode_e funct3;
-    gpr_t          rd;
-    major_opcode_e opcode;
-  } csr_instr_t;
-
   // U type
   typedef struct packed {
     logic [31:12]  imm;
@@ -426,18 +413,12 @@
 
   // B type
   typedef struct packed {
-    logic [31:25]  imm;
+    logic [31:25]  imm_h;
     gpr_t          rs2;
     gpr_t          rs1;
     logic [14:12]  funct3;
-    gpr_t          rd;
+    logic [11:7]   imm_l;
   } b_type_t;
-
-  function logic[11:0] read_b_imm(logic[31:0] instr);
-    automatic logic [11:0] imm;
-    imm = {instr[31], instr[7], instr[30:25], instr[11:8]};
-    return imm;
-  endfunction : read_b_imm
 
   // S type
   typedef struct packed {
@@ -553,6 +534,58 @@
     uncompressed_instr_t uncompressed;
   } instr_t;
 
+  typedef struct packed {
+    gpr_t gpr;
+    bit   valid;
+  } reg_operand_t;
+
+  typedef struct packed {
+    logic[31:0] imm;
+    // TODO
+    //imm_e       imm_type;
+    //int         width;
+    //bit         sign_ext;
+    bit         valid;
+  } imm_operand_t;
+
+  typedef struct packed {
+    int   offset;
+    gpr_t gpr;
+    bit   valid;
+  } mem_operand_t;
+
+  // TODO Zc
+  // typedef struct packed {
+  //   rlist_t rlist;
+  //   bit     valid;
+  // } rlist_operand_t;
+
+  typedef enum logic[7:0] {
+    I_TYPE,
+    I_TYPE_LOAD,
+    J_TYPE,
+    S_TYPE,
+    R_TYPE,
+    R4_TYPE,
+    B_TYPE,
+    U_TYPE,
+    // Compressed formats
+    CR_TYPE,
+    // Others
+    UNKNOWN_FORMAT
+  } instr_format_e;
+
+  typedef struct packed {
+    instr_names_e  instr;
+    instr_format_e format;
+    reg_operand_t rd;
+    reg_operand_t rs1;
+    reg_operand_t rs2;
+    reg_operand_t rs3;
+    imm_operand_t imm;
+    // rlist_operand_t rlist;
+  } asm_t;
+
   function logic [11:0] get_cj_offset(compressed_instr_t instr);
     get_cj_offset = {
       instr.format.cj.imm[11+1],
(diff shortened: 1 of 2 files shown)


lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+  } op_minor_opcode_e;
+
+
+  // Specific function subtypes
+  typedef struct packed {
+    csr_name_e           csr;
+    union packed {
+      gpr_t         rs1;
+      logic [19:15] uimm;
+    }n;
```
silabs-krdosvik, line 357: n is not that describing.

silabs-krdosvik: Just to come with a proposal: "addressing", since it seems like we divide between register and immediate _addressing_ :
image

Obs: based on random google search result.

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+      logic [31:7]         raw;
+      i_type_t             i;
+      i_type_load_t        i_load;
+      j_type_t             j;
+      s_type_t             s;
+      r_type_t             r;
+      r4_type_t            r4;
+      b_type_t             b;
+      u_type_t             u;
+    } format; // Would like to use type, but type is reserved keyword in sv
```
silabs-krdosvik, line 469: I think this is a good alternative :-)

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+       && (instr.compressed.format.cr.rs2.gpr     == X0)
+       && (instr.compressed.format.cr.funct4      == 4'b1001)) :
+        decode_instr = CEBREAK;
+
+      default: decode_instr = UNKNOWN;
+    endcase
+
+  endfunction : decode_instr
+
+  function is_instr(instr_t instr, instr_names_e instr_type);
```
silabs-krdosvik, line 805: just to make sure I understand the code. We use this function to add decoded signals? and then we use the instr_name in the rvfi instr interface to check instr_name ==<desired_instr>?

lib/support/isa_support.sv
silabs-hfegran: > n is not that describing.

Yup, this one slipped through - Just a copy-paste from the previous code, where i used "n" to symbolize an anonymous union/struct, which is not possible in sv - i believe it can be better solved with the approach below.

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+      (   (instr.uncompressed.opcode              == BRANCH)
+       && (instr.uncompressed.format.b.funct3     == FUNCT3_BGEU)) :
+        decode_instr = BGEU;
+
+      // Compressed
+      (   (instr.compressed.opcode                == 2'b10)
+       && (instr.compressed.format.cr.rd_rs1.gpr  == X0)
+       && (instr.compressed.format.cr.rs2.gpr     == X0)
+       && (instr.compressed.format.cr.funct4      == 4'b1001)) :
+        decode_instr = CEBREAK;
```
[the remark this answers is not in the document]
silabs-hfegran: just a regular case statement, but instead of using case (something) I used case (1), where the first condition that is true in the list below will match.

lib/support/isa_support.sv
silabs-hfegran: this function `is_instr( instruction_binary, expected instruction) ` returns true or false, `decode_instr(instruction_binary)` returns the name of the instruction as an enum-value from the list of instructions.

silabs-hfegran: currently decode_instr should cover most (if not all) of rv32i - can be expanded as needed.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -212,6 +215,8 @@ interface uvma_rvfi_instr_if_t
 
   // assigning signal aliases to helper functions
   always_comb begin
+    instr_name          <= decode_instr(rvfi_insn);
```
silabs-hfegran, line 218: This basically creates a signal that can be used in waveforms, that presents the name of the last executed instruction

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+      instr.format.cj.imm[10+1],
+      instr.format.cj.imm[6+1],
+      instr.format.cj.imm[7+1],
+      instr.format.cj.imm[3+1:1+1],
+      instr.format.cj.imm[5+1],
+      1'b0
+    };
+  endfunction : get_cj_offset
+
+  function instr_names_e decode_instr(instr_t instr);
```
silabs-hfegran, line 570: One idea I have for this function, is to return a struct that contains `instr_names_e` + operands; and operands set as INVALID if they are not used for that specific function. This will basically enable us to build a valid asm-string from the contents of the struct.

lib/support/isa_support.sv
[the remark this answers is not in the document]
silabs-robin: LRM "12.5.2 Constant expression in case statement" has the full explanation.

image

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+    SRA,
+    OR,
+    AND,
+    // Compressed
+    CEBREAK,
+    // Pseudo name, class of instructions
+    STORE_INSN,
+    LOAD_INSN,
+    // Unknown for instructions that cannot be decoded
+    UNKNOWN
```
silabs-mateilga, line 170: Maybe invalid or illegal are good name alternatives, as they fall more in line with what the core would do.

silabs-mateilga (commented):
Looks very good, it will be really interesting to start using this in the rvfi_instr_if and other files. I'm holding off on a full review until the PR is more complete.

lib/support/isa_support.sv
silabs-krdosvik: Sounds good. Thanks! :-)

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+  typedef struct packed {
+    logic [31:12] imm;
+    gpr_t         rd;
+  } j_type_t;
+
+  function logic[20:0] read_j_imm(logic[31:0] instr);
+    automatic logic [20:0] imm;
+    imm = instr >> 11;
+    return { imm[20], imm[10:1], imm[11], imm[18:12], 1'b0 };
+  endfunction : read_j_imm
```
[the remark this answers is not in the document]
silabs-robin: Or you could do something like you have done below: `imm_12_10` etc.

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+  } compressed_instr_t;
+
+  typedef union packed {
+    compressed_instr_t   compressed;
+    uncompressed_instr_t uncompressed;
+  } instr_t;
+
+  function logic [11:0] get_cj_offset(compressed_instr_t instr);
+    get_cj_offset = {
+      instr.format.cj.imm[11+1],
```
silabs-robin, line 558: Why is there +1 on everything?

lib/support/isa_support.sv
silabs-robin: Can that be a separate function?

silabs-robin, line 570: Since this function finds an instruction's name, can the function be called something with "name"?

`decode_instr_name()`?
`get_instr_name()`?

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+    OR,
+    AND,
+    // Compressed
+    CEBREAK,
+    // Pseudo name, class of instructions
+    STORE_INSN,
+    LOAD_INSN,
+    // Unknown for instructions that cannot be decoded
+    UNKNOWN
+  } instr_names_e;
```
silabs-robin, line 171: The other enums are not in plural.

lib/support/isa_support.sv
silabs-robin:
```
function is_instr(instr_t instr, instr_names_e instr_type);
  return  instr_type == decode_instr(instr);
endfunction
```

silabs-robin, line 805: The name `is_instr` initially made me think this functions checks whether a binary is a legal instruction or just random non-instruction data.

silabs-mateilga: An alternative name could be `match_instr(..)`

lib/support/isa_support.sv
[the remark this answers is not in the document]
silabs-hfegran: This is on my TODO-list

lib/support/isa_support.sv
silabs-hfegran: functionality changed, so the new functionality is more in line with the intended usage

lib/support/isa_support.sv
silabs-hfegran: fixed

lib/support/isa_support.sv
silabs-robin: Ok.

lib/support/isa_support.sv
silabs-hfegran: fixed

lib/support/isa_support.sv
```diff
@@ -68,6 +70,887 @@
...
+  typedef enum logic [6:0] {
+    LOAD   = 7'b000_0011, LOAD_FP  = 7'b000_0111, CUS_0 = 7'b000_1011, MISC_MEM = 7'b000_1111, OP_IMM = 7'b001_0011, AUIPC_OP = 7'b001_0111,OP_IMM_32 = 7'b001_1011,
+    STORE  = 7'b010_0011, STORE_FP = 7'b010_0111, CUS_1 = 7'b010_1011, AMO      = 7'b010_1111, OP     = 7'b011_0011, LUI_OP   = 7'b011_0111,OP_32     = 7'b011_1011,
+    MADD   = 7'b100_0011, MSUB     = 7'b100_0111, NMSUB = 7'b100_1011, NMADD    = 7'b100_1111, OP_FP  = 7'b101_0011, RES_1    = 7'b101_0111,CUS_2     = 7'b101_1011,
+    BRANCH = 7'b110_0011, JALR_OP  = 7'b110_0111, RES_0 = 7'b110_1011, JAL_OP   = 7'b110_1111, SYSTEM = 7'b111_0011, RES_2    = 7'b111_0111,CUS_3     = 7'b111_1011
+  } major_opcode_e;
+
+  typedef enum logic [1:0] {
+    C0 = 2'b00, C1 = 2'b01, C2 = 2'b10, C3 = 2'b11
+  } compressed_opcode_e;
```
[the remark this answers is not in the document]
silabs-hfegran: fixed

silabs-oysteink (approved):

PR #1883, opened 2023-05-09
fixed a_csrs_mmode_only

silabs-mateilga:
This PR fixes an assert that started failing after rvfi started reporting exception triggers.


silabs-robin (approved):
Approving based on:
Offline discussion.
This change is closer to the original intent, and is the simplest fix.
(Note for the potential future: There is a hole inthat  dbg_trigger can mask a real bug, but it is deemed extremely unlikely that something like that would get introduced, etc.)

MikeOpenHWGroup (approved):
Local CI (w/o ImperasDV) passes.

PR #1884, opened 2023-05-09
ci: add benchmarks pass/fail

cathales:


MikeOpenHWGroup:
I still have concerns about adding a performance benchmark tests to a CI regression. It is very costly in terms of time and licenses.  Remember that the goal of a CI regression is to ensure that a pull request doesn't break anything obvious - it is not intended to check all functions.  Performance tests typically take a long time to run and consume valuable licenses so they are not a good fit for CI.  I strongly recommend moving the performance benchmark tests to a weekend regression.

Secondly, I worry about the pass/fail criteria:
```
valid_cycles = {
    'dhrystone': 221425,
    'coremark': 700447,
}
```
That is an extremely precise criteria.  Are we going to reject a pull-request if the coremark benchmark requires 700448 cycles?

MikeOpenHWGroup:
> The sum of all job durations is more than 8 hours

Are you saying that the CI regression requires 8 hours to run?

> So the total [time for benchmarks to run] is 4 minutes

That is what I expected.  Given that I would expect a CI regression to run 15 to 20 minutes, this is a significant increase.

cathales:
It is 8 hours of runner work but as we have several runners handling jobs in parallel, the duration of the whole CI run is about one hour. To put it in a nutshell, the global impact of 4 minutes (for the benchmarks to run) split on 8 runners is 30 seconds.

I would say that the cost is:

- 4 minutes of work for the runners (0.8%)
- 30 seconds of latency for the user, but as the jobs are run only at a specific time in the day, if I submit a PR today I have to wait for tomorrow before I can check the results so I won’t mind the 30 second increase in latency.

MikeOpenHWGroup:
OK, you win @cathales!

Having said that, a one hour CI regression is very long.

PR #1890, opened 2023-05-10
Fixed debug_pc outputs not connected warning in sim

silabs-hfegran:
New signals added to the core caused simulation warnings, hooked these up to respective signals in the testbench to avoid unconnected ports-warning


silabs-mateilga (approved):
approving.

silabs-oysteink (approved):

PR #1891, opened 2023-05-10
Test-bench and scripts updates to align to CV32E40P PR ...

pascalgouedo:
[#801](https://github.com/openhwgroup/cv32e40p/pull/801)


cv32e40p/sim/uvmt/Makefile
MikeOpenHWGroup, line 1: Why remove the `CV_SW` variables?

mk/uvmt/uvmt.mk
```diff
@@ -63,6 +63,7 @@ MKDIR_P = mkdir -p
 # Compile compile flags for all simulators (careful!)
 WAVES        ?= 0
 SV_CMP_FLAGS ?= "+define+$(CV_CORE_UC)_ASSERT_ON"
+SV_CMP_FLAGS = ""
```
MikeOpenHWGroup, line 66: Does this mean the assertions are not compiled or used in simulation?

cv32e40p/sim/uvmt/Makefile
pascalgouedo: Because if they are not set in env it puts empty string and nothing works after that.
And the check in Common.mk traps nothing because it is set to empty string.
Removing them here allows the check to work correctly.

cv32e40p/tb/uvmt/uvmt_cv32e40p_imperas_dv_wrap.sv
```diff
@@ -479,8 +479,8 @@ module uvmt_cv32e40p_imperas_dv_wrap
     // rvviRefCsrCompareEnable(hart_id, `CSR_DCSR_ADDR,   RVVI_FALSE);
 
     // rvviRefCsrCompareEnable(hart_id, `CSR_MSTATUS_ADDR,   RVVI_FALSE);
-    void'(rvviRefCsrSetVolatile(hart_id, `CSR_MSTATUS_ADDR          ));
-    rvviRefCsrCompareEnable(hart_id, `CSR_MISA_ADDR,   RVVI_FALSE);
+//  void'(rvviRefCsrSetVolatile(hart_id, `CSR_MSTATUS_ADDR          ));
```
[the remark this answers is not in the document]
pascalgouedo: Done

MikeOpenHWGroup (approved):
(Mostly) looks good to me @pascalgouedo.  It also passes a local CI (sans ImperasDV).

I have made a few comments, but these are not worth gating the merge.  When you have time, please respond to these comments within this PR.  Your answers are mostly just for my information.

mk/uvmt/uvmt.mk
MikeOpenHWGroup: > What is the rule in core-v-verif about those asserts which are inside the some design RTL files?

There is no rule about this, but in general we try to use assertions in both formal and simulation.

PR #1894, opened 2023-05-11
Update cv32e40s csr access test

silabs-krdosvik:
What this PR does:
- update gen_csr_access_test script:
1) make it possible to specify input path for the yaml file
2) make sure we use the m4 yaml file if the m4 flag is on
3) update help/args description 

- cv32e40s_csr_access_test:
1) update the file by using an updated m4 yaml file (https://github.com/openhwgroup/cv32e40s/blob/master/yaml/csr.yaml.m4, will make a PR for this in the cv32e40s repo)

- test.yaml
1) make sure we dont let the compiler optimize the code. The optimization changes the bne instructions to a cm.jt, and consequently generates an infinit loop.

Extra:
Tried rebasing, which almost work (but eventually ended up with a mess). So unless anyone complains about the commits, I will let it be.

This PR is related to another PR in the cv32e40x repo: https://github.com/openhwgroup/cv32e40x/pull/853 (which initially was in the *s repo (https://github.com/openhwgroup/cv32e40s/pull/453))


bin/gen_csr_access_test.py
```diff
@@ -71,11 +74,19 @@
...
     print('{}'.format(args))
 
+
 script_path      = os.path.join(topdir, '{}/vendor_lib/google/riscv-dv/scripts/gen_csr_test.py'.format(args.core.lower()))
-yaml_file_path   = os.path.join(topdir, '{}'.format(args.core.lower()) + '/env/corev-dv/{}'.format(args.core.lower()) + '_csr_template.yaml')
 template_path    = os.path.join(topdir, './bin/templates/csr_access_test_template.S')
 output_yaml_path = ""
 
+if (args.input_yaml_path != None):
+    yaml_file_path   = '{}'.format(args.input_yaml_path.lower())
```
silabs-robin, line 83: Why is there `.lower()` on this?

cv32e40s/sim/uvmt/make
silabs-robin, line 1: Doesn't seem quite right.

silabs-robin (changes requested):

cv32e40s/sim/uvmt/make
silabs-krdosvik: Agreed!

bin/gen_csr_access_test.py
silabs-krdosvik: Original though: because the other path uses lower when handling the core argument. But this is a good point. paths don't necessarily use lower case.

silabs-robin (approved):
Approving based on:
The points in the description makes sense.
The code seems to match those points.
All my feedback has been fixed.

silabs-hfegran (approved):
Approving based on Robins review

PR #1904, opened 2023-05-15
Cv32e40p/dev baoshan add fp tests

dd-baoshan:
Updated fp stream lib and add example tests for extended f and zfinx


dd-baoshan:
branch wrongly selected

PR #1907, opened 2023-05-15
CLIC Vplan - Update vs User Manual v0.9.0

silabs-robin:
This PR updates the CLIC vplan according to the changes highlighted in our internal documentation.

If we don't trust our internal documentation then we could have a thorough look at the CLIC spec diff compared to the last version mentioned in the vplan.
https://github.com/riscv/riscv-fast-interrupt/compare/8675ec9090c2d1cd50fb9960cb7ce9aedb757ebf...master


silabs-hfegran (approved):
Looks good to me, thanks @silabs-robin

PR #1908, opened 2023-05-15
add test of the shadow copy to mcause register

silabs-krdosvik:
Add assertion for the mcause register, and reverse-shift the assertions in the clic assertion files.


silabs-mateilga (approved):
This PR is evidence of a well structured assertion set, as all the work that was necessary is to reuse existing properties. It also makes it very easy to review, and to get right. Approving.

silabs-hfegran (approved):

PR #1909, opened 2023-05-15
Small modifications to regression generation

XavierAubert:
Added an option for cv_regress script to enable check_sim_results for questa CI

I also noticed that the yesorno jinja macro was somehow weird, if the variable is set to False, the macro will convert it to "False" instead of "NO" as I *think* it should be. 

By default, a test is supposed to inherit its configuration from its build, but I added a way to override this behavior if someone wants to, only for questa CI tool first.


MikeOpenHWGroup:
Passes local CI (sans ImperasDV) under Xcelium.

PR #1915, opened 2023-05-17
ISACOV Agent : Integrate the ISACOV Agent in the CVA6 env

AyoubJalali:


cva6/env/uvme/uvme_cva6_cfg.sv
```diff
@@ -129,18 +135,34 @@ class uvme_cva6_cfg_c extends uvma_core_cntrl_cfg_c;
...
-         clknrst_cfg.trn_log_enabled   == 1;
+         clknrst_cfg.trn_log_enabled   == 0;
          axi_cfg.trn_log_enabled       == 1;
+         rvfi_cfg.trn_log_enabled      == 1;
+         isacov_cfg.trn_log_enabled    == 1;
       }
 
       if (cov_model_enabled) {
-         cvxif_cfg.cov_model_enabled == 1;
+         cvxif_cfg.cov_model_enabled  == 0;
```
MikeOpenHWGroup, line 164: Why are we disabling the XIF fcov?

cva6/env/uvme/uvme_cva6_constants.sv
```diff
@@ -26,5 +26,6 @@ parameter uvme_cva6_debug_default_clk_period = 10_000; // 10ns
 
 parameter XLEN = 32;
 parameter ILEN = 32;
+parameter RVFI_NRET = ariane_pkg::NR_COMMIT_PORTS;
```
MikeOpenHWGroup, line 29: We should not be using a package from the RTL in the verification environment.  We should have a package included in `cva6/env/uvme/uvme_cva6_pkg.sv` and use that here.

MikeOpenHWGroup (changes requested):
Looks pretty good to me @AyoubJalali.  I have a couple of comments.

cva6/env/uvme/uvme_cva6_cfg.sv
AyoubJalali: Hello,
It's just a decision that i made for now because we don't have a final coverage model for CVXIF (need update), also we wanna see the isacov coverage alone

cva6/env/uvme/uvme_cva6_constants.sv
MikeOpenHWGroup: Of course, you are afraid of this - you _should_ be!

The whole point here is to ensure that the _testbench_, not the _DUT_ is in control of the configuration.  If this is not true, then the DUT is exposed to potential bugs in-which a top-level chip cannot properly configure multiple, heterogeneous CVA6 cores.

So if someone accidentally misconfigured NR_COMMIT_PORTS in the RTL, hopefully we will catch it (when the CI fails).

cva6/env/uvme/uvme_cva6_cfg.sv
MikeOpenHWGroup: The coverage viewer is hierarchical, so we will always be able to see ISACOV independently.

cva6/env/uvme/uvme_cva6_constants.sv
AyoubJalali: Great, thanks for convincing me

cva6/env/uvme/uvme_cva6_cfg.sv
AyoubJalali: For this I'll make the change

PR #1916, opened 2023-05-19
[cv32e40p] tb update for debug and debug coverage with f and xpulp instructions

dd-vaibhavjain:
This PR is doing following things:

1.) reconnect debug_sequencer -> was preventing debug tests to run
2.) update debug rom for FPR push and pop for floating point random stream tests
3.) create all xpulp instruction encodings as params in uvme tb for use. At the moment use case is for coverage but can be used with assertions or scoreboards if needed.
4.) add new debug cover groups for floating point and xpulp instructions according to v2 test plans

### DIFF
--- a/cv32e40p/env/corev-dv/cv32e40p_debug_rom_gen.sv
+++ b/cv32e40p/env/corev-dv/cv32e40p_debug_rom_gen.sv
@@ -68,6 +68,11 @@ class cv32e40p_debug_rom_gen extends riscv_debug_rom_gen;
             end
             // Need to save off GPRs to avoid modifying program flow
             push_gpr_to_debugger_stack(cfg_corev, debug_main);
+
+            // Need to save off FPRs incase f-extension instructions are used to avoid modifying program flow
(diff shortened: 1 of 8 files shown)


MikeOpenHWGroup (approved):
Hi @dd-vaibhavjain, I am going to approve and merge this because it is a clean update and it passes a local CI (without ImperasDV).

However, I have question about the "debug_rom" code.  It looks a lot like what I would expect to see in a real debug rom.  Is this required, and if so, why?  Our pseudo-random test-programs do not do anything purposeful, so why should the debug rom code?  Why not simply have some random code and return?

MikeOpenHWGroup (approved):
Hi @dd-vaibhavjain, I am going to approve and merge this because it is a clean update and it passes a local CI (without ImperasDV).

However, I have question about the "debug_rom" code.  It looks a lot like what I would expect to see in a real debug rom.  Is this required, and if so, why?  Our pseudo-random test-programs do not do anything purposeful, so why should the debug rom code?  Why not simply have some random code and return?

PR #1918, opened 2023-05-22
[cv32e40p] enable Imperas functional coverage and fix some tb flow issues

dd-vaibhavjain:
PR is for changes related to following things:

1.) Add imperas func coverage enable switches for cv32e40p tb and update simulator make scripts for this.
2.) Add new defines in uvme corev-dv env for setting default ISA config for each test without need of adding a specific define in test's corev_dv.yaml. This was missing for new F, ZFINX, XPULP tests and without this would need explicit defines for all new FPU, XPULP tests.
3.) Minor issue in vsim.mk for waves dump after a recent change for waves variable in mk files, due which waves.tcl seems to not being able to load based on order of vsim arguments.


cv32e40p/tb/uvmt/uvmt_cv32e40p_imperas_riscv_coverage_config.svh
```diff
@@ -0,0 +1,25 @@
+/*
```
MikeOpenHWGroup, line 1: I have two concerns here:
1. Using compiler macros (defines) to control the inclusion of coverage models is a bad approach - this is better done with run-time conditional checks.
2. If the Imperas function coverage model is external to CORE-V-VERIF, its control should be as well.

MikeOpenHWGroup (changes requested):

cv32e40p/tb/uvmt/uvmt_cv32e40p_imperas_riscv_coverage_config.svh
dd-vaibhavjain: @MikeOpenHWGroup My reasoning for doing like this:
> I have two concerns here:
> 
> 1. Using compiler macros (defines) to control the inclusion of coverage models is a bad approach - this is better done with run-time conditional checks.
Vaibhav:
1.) For Imperas riscvISACOV model, this is as per Imperas docs, they use various ifdefs for entire coverage model and all these different defines somewhere needs to be added (either directly in TB compile command or somewhere in TB) to enable various parts of this cover model. 
Thats why I define all  these numerous defines at one place in .svh file and include this .svh within imperas filelist only (this is like including all other imperas stuff).  So this is used only with "USE_ISS=yes" from make files.

> 2. If the Imperas function coverage model is external to CORE-V-VERIF, its control should be as well.
Vaibhav:
(2) So you are suggesting in core-v-verif/cv32e40p/tb/uvmt/imperas_dv.flist -> Instead of adding this new .svh file in tb/uvmt/ path , we shd add a path for a corresponding .svh file somewhere from local ${IMPERAS_HOME}. ?
I can do this if this is the the case for Imperas. 
My understanding while making adding this file was that it is aligned with the way Imperas is built/added in make files/uvm tb with its intantiation, interfaces etc as in uvmt_cv32e40p_imperas_dv_wrap. Only thing is this being not part of imperas docs, any other entity using Imperas with cv32e40p in core-v-verif will not see the defines and the coverage will not be enabled for them.

MikeOpenHWGroup: 1. If I were an Imperas customer, I would complain (a lot!) about their inappropriate use of compiler macros.  CORE-V-VERIF already has too much of this and over time it creates very hard to maintain code.
2. Yes, I like the idea of separating out the manifests.  It makes things much cleaner for those users of CORE-V-VERIF who are not ImperasDV customers.  Thanks for the good suggestion.

dd-vaibhavjain: @MikeOpenHWGroup then i need to discuss internally with @pascalgouedo to check if we can use some paths in Imperas installation as by default I dont have permissions to write files to Impera's installation( $IMPERAS_HOME) path. Since we want to add a new file in this IMPERAS_HOME path and provide it in imperas.flist to still keep it generic for users of Imperas.

But for me this whole requirement of having only this file for Imperas related defines placed outside of core-v-verif TB is very odd and inconsistent with how Imperas is used at the moment. As you can see now core-v-verif has allowed entire Imperas connections, instantiation within the core-v-verif TB along with Make files. So after this now not allowing some part of Imperas, seems not consistent.
 
Unless we put in a requirement and move all Imperas TB related files out of core-v-verif or atleast move to a new dir outside of tb/uvmt directory, like a new dir for ref_model/imperas for example. Already these defines are well masked by USE_ISS and within Imperas I see no issue/errors for anyone without Imperas using core-v-verif.

MikeOpenHWGroup:
It was my understanding that _anything_ related to ImperasDV had to be external  to CORE-V-VERIF.  (For the record, I do not like this situation, but Imperas sets the terms).  If that is not the case then let's keep things as they are.

MikeOpenHWGroup (approved):

pascalgouedo:
> It was my understanding that anything related to ImperasDV had to be external to CORE-V-VERIF. (For the record, I do not like this situation, but Imperas sets the terms). If that is not the case then let's keep things as they are.

Don't totally agree.
Imperas installation area contains models for a lot of Cores, not only CV32E40P or RISC-V ones.
Variables to use correct model and configuration files should be local to your verification environment because it totally depends of the core you are verifying.
This is like ovpsim.ic model configuration file which depends of your RTL parameters which are driven by the test/RTL configuration.

We could think together with Imperas about a different way to manage all that but it is quite late to make a revolution now.

Cheers,
Pascal.

PR #1921, opened 2023-05-22
Added Support of riscv-arch-test suites and infrastructure in core-v-verif

### LINKED ISSUE #1919: Update core-v-verif for CVA6 to support the latest riscv-arch-tests and its infrastructure.
### Task Title
Update core-v-verif to support latest [riscv-arch-tests](https://github.com/riscv-non-isa/riscv-arch-test) infrastructure.

### Why we need an update
Currently core-v-verif supports a bit old version riscv-arch-test suites. We need to update it because:
- Generally, there are some fixes added to the existing tests, and test suites for more extensions are added (Bit Mnip (`B`), Double precision floating point (`D`), Single Precision floating point (`F`), some basic privilege). These tests suites are not supported by core-v-verif or at least not available these yaml files ([link](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/tests/testlist_riscv-arch-test-cv64a6_imafdc_sv39.yaml))
- The main reason for this update is to leverage the latest infrastructure of arch-tests which allows us to develop and run the privilege architecture tests. arch-tests now has a mechanism to support traps from S/U with and without Virtualization enabled (Related PRs: [[1]](https://github.com/riscv-non-isa/riscv-arch-test/pull/277), [[2]](https://github.com/riscv-non-isa/riscv-arch-test/pull/318)). This feature can be useful and leveraged in executing the virtual memory test plan (PR [here](https://github.com/openhwgroup/core-v-verif/pull/1789)) for the architectural testing and verification of the MMU of CVA6.

### LINKED ISSUE #1920: Hardcoded ISA string to compile riscv-arch-tests
Each test suite from [riscv-arch-tests](https://github.com/riscv-non-isa/riscv-arch-test) has ISA string that is only required by that specific extension and that string is scanned by RISCOF and passed to the `march` flag of toolchain while compiling the test.

For example, c.ebreak test has the ISA string `RV64IC_Zicsr` available [here](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/riscv-test-suite/rv64i_m/C/src/cebreak-01.S#L13), and when it is compiled using RISCOF, ISA string will pass to the march flag as: `-march=rv64ic_zicsr`.

As core-v-verif does not use RISCOF to run arch-tests, it has a hardcoded ISA string (for example [here](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/sim/cva6.py#L859)) containing all the extension supported by the target configuration
I may explicitly mention that this is not a bug and it is working fine this way. 
The only little problem is that when it compiles the tests that do not intend to include any `C` instructions, this hardcoded ISA string (rv64gc) gets compiled such that it may compress some I-type instructions.
Another thing is that when we want to run tests of some non-general extensions such as BitManip, the toolchain requires to have corresponding zba, zbb, zbc, zbs as part of ISA string.
So, it is just a better approach that `march` comes from a specific test suite that needs to get compiled. Since core-v-verif uses yaml based framework instead of RISCOF, this can be adapted into that yaml-based framework.

Abdulwadoodd:
This PR adds the support of an updated arch-test suite and infrastructure in core-v-verif framework.

### Related issues
- Resolves #1919 
- Resolves #1920 

### What is added/changed?
- Hash of riscv-arch-test is updated to latest (3.6.5) version.
- Hash of CVA6 is updated that points to the head of the master branch
- Yaml for riscv-arch-test suites is updated: 
-  This removes the redundant flags for compilation, which are already part of the compilation command in [cva6.py](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/sim/cva6.py#L414). Added `march` and `mabi` variables for each test suite in yaml file.
- Added `--en_arch_tests` boolean command line argument to pick the march ISA string and mabi, if we are running arch-tests.
- Added BitManip, Single and Double Precision Floating point, and privilege arch-test suites.

I have used the latest release of the toolchain that takes into account the isa string passed to the `-march` flag of the toolchain. Compilation may fail with old versions of the toolchain.

:warning: One of the Double Precision Floating point tests (`fcvt.l.d_b28-01.S`) fails (probably) due to a potential bug.

**Note:** Test compilation may throw warning messages and we don't need to worry about it. Tests work fine with the warning. It will disappear after [this](https://github.com/riscv-non-isa/riscv-arch-test/pull/352) PR gets merged.


JeanRochCoulon:
Thanks for this PR. So far, the CI is failed: some fails with vcs-uvm target and with arch-test regression.
please refer to the https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_core-v-verif_0.html dashboard

Abdulwadoodd:
Hi @JeanRochCoulon, 
By inspection, the latest CI results tells me that CI is probably using older version of toolchain. CI fails on `arch_test cv64a6_imafdc_sv39` and `csr_test cv32a60x` tests which uses CSR instructions and it is passing the updated `arch_test cv32a60x` tests because these tests does not have any CSR instruction. 

#### Reason
The latest release of toolchain require `zicsr` included as part of ISA string of -march flag (**and older toolchain does not recognize `zicsr`, hence throws compilation error**) unless we are using rv32/64`g` which is not the case for the updated tests. Updated arch-tests follows right convention of ISA string which is: to compile tests with ISA string having only those extensions/features which are supported by the test itself and not to add any extra extension/feature in ISA string.

This is already addressed in the description of this PR. So, in conclusion, this might require a toolchain update in CI flow. I have tested it locally with latest release of toolchain and all the arch-tests are passing.

zchamski:
Hi @Abdulwadoodd , by looking at the Thales CI logs I confirm that the failures in `csr_test cv32a60x` and `arch_test cv64a6_imafdc_sv39` jobs come from the GCC 10.2 (current CI compiler) not knowing about the `_z*` way of specifying ISA extensions.

Upgrading the CI toolchain to GCC 12.x or even 13.x is indeed necessary, but it implies a major review+modification of test suite setups.  I propose to sync up on this with @MikeOpenHWGroup and @JeanRochCoulon, and to schedule the GCC 10.2 -> GCC 12+ transition in an orderly way.

MikeOpenHWGroup:
> Upgrading the CI toolchain to GCC 12.x or even 13.x is indeed necessary, but it implies a major review+modification of test suite setups.

It is not clear to me why this would be a **_major_** modification. Can you explain a bit more about this?

zchamski:
"thorough" would be a more appropriate term...  The compiler options are handled in two places (in the driver script `cva6.py` and in the test lists).  This approach needs a critical review and proper streamlining to have the options in one place and to prevent redundant settings.

MikeOpenHWGroup:
> The compiler options are handled in two places (in the driver script `cva6.py` and in the test lists). This approach needs a critical review and proper streamlining to have the options in one place and to prevent redundant settings.

Arg! That is a separate issue so I have opened #1939 to discuss.

PR #1923, opened 2023-05-23
Enable dummy instruction testing, remove no_bitmanip configuration

silabs-hfegran:
- Made dummy instruction generation available for riscv-dv
- Replaced no_bitmanip configuration with dummy_instr
- Increased per-test timeout to two hours (expect that dummy instr. tests may cause some tests that otherwise would not time out to fail)
- Removed dummy_instr cfg from running in regression with directed tests, as these needs modifications to enable dummy instructions.

ci_check OK


silabs-robin (approved):

MikeOpenHWGroup:
LGTM

PR #1931, opened 2023-05-25
Update trigger asserts and general test environment to new spec

silabs-krdosvik:
- Updates trigger assert and enviroment to new spec (important)
- fix bug in csh and csb opcode
- Add TODOs for stuff that is not yet 100% verified (I will continue looking into this, but since this PR holds environment changes, it is a good idea to merge it)
- ci check passes when not using the ISS

### DIFF
--- a/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
+++ b/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
@@ -532,7 +532,7 @@ function void uvme_cv32e40s_cfg_c::set_unsupported_csr_mask();
    unsupported_csr_mask[uvma_core_cntrl_pkg::SCOUNTEREN] = 1;
 
    // TODO:ropeders re-evaluate this when 40s is more stable
-   //unsupported_csr_mask[uvma_core_cntrl_pkg::TDATA3] = 1;
+   unsupported_csr_mask[uvma_core_cntrl_pkg::TDATA3] = 1;
    unsupported_csr_mask[uvma_core_cntrl_pkg::TCONTROL] = 1;
 
    unsupported_csr_mask[uvma_core_cntrl_pkg::MCONTEXT] = 1;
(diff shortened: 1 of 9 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -190,30 +140,83 @@ module uvmt_cv32e40s_triggers_assert_cov
...
-  localparam HW_ZERO_4 = 4;
-  localparam HW_ZERO_3 = 3;
-  localparam HW_ZERO_2 = 2;
-  localparam HW_ZERO_1 = 1;
-  localparam HW_ZERO_0 = 0;
-
+  //Initial settings
   localparam TDATA1_DISABLED = 32'hF800_0000;
   localparam MAX_NUM_TRIGGERS = 5;
+  localparam MAX_MEM_ACCESS = 13; //Push and pop can do 13 memory access. XIF can potentially do more (TODO xif: check this when merging to cv32e40x)
```
silabs-robin, line 146: > XIF can potentially do more (TODO xif: check this when merging to cv32e40x

One could maybe add an assert to check that there are never more reads/writes than what is assumed here?
Then we don't have to remember to manually update anything, it will simply fail if it stops being true.

silabs-robin (approved):
Approving based on:
The diff seems to match what I know has changed in the related specs.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-krdosvik: Good idea!

cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
```diff
@@ -532,6 +532,7 @@ function void uvme_cv32e40s_cfg_c::set_unsupported_csr_mask();
    unsupported_csr_mask[uvma_core_cntrl_pkg::SCOUNTEREN] = 1;
 
    // TODO:ropeders re-evaluate this when 40s is more stable
```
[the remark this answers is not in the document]
silabs-krdosvik: Resolved (from my side)

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-krdosvik: Now, wmask is also added to the assertion

PR #1943, opened 2023-05-30
Add two arguments in cva6.py to controlle systemverilog seed

AEzzejjari:
The first argument --gen_sv_seed run a test X times with a random seed and generates X logs. (0<X)

The second argument --sv_seed executes a test with a specific seed.

We cannot use both options at the same time.

### DIFF
--- a/cva6/sim/cva6.py
+++ b/cva6/sim/cva6.py
@@ -426,7 +426,10 @@ def run_assembly(asm_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, outp
   # ISS simulation
   for iss in iss_list:
     run_cmd("mkdir -p %s/%s_sim" % (output_dir, iss))
-    log = ("%s/%s_sim/%s.log" % (output_dir, iss, asm))
+    if log_format == 1:
+      log = ("%s/%s_sim/%s_%d.log" % (output_dir, iss, asm, test_iteration))
+    else:
+      log = ("%s/%s_sim/%s.log" % (output_dir, iss, asm))
     log_list.append(log)
     base_cmd = parse_iss_yaml(iss, iss_yaml, isa, target, setting_dir, debug_cmd)
     cmd = get_iss_cmd(base_cmd, elf, target, log)
@@ -561,7 +564,10 @@ def run_c(c_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, output_dir,
   # ISS simulation
   for iss in iss_list:
     run_cmd("mkdir -p %s/%s_sim" % (output_dir, iss))
-    log = ("%s/%s_sim/%s.log" % (output_dir, iss, c))
+    if log_format == 1:
+      log = ("%s/%s_sim/%s_%d.log" % (output_dir, iss, c, test_iteration))
+    else:
+      log = ("%s/%s_sim/%s.log" % (output_dir, iss, c))
     log_list.append(log)
     base_cmd = parse_iss_yaml(iss, iss_yaml, isa, target, setting_dir, debug_cmd)
     cmd = get_iss_cmd(base_cmd, elf, target, log)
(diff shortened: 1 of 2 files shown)


cva6/sim/cva6.py
```diff
@@ -555,13 +555,13 @@ def run_c(c_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, output_dir,
                       linker, gcc_opts, elf))
   cmd += (" -march=%s" % isa)
   cmd += (" -mabi=%s" % mabi)
-  run_cmd(cmd, debug_cmd = debug_cmd)
+  run_cmd_output(cmd.split(), debug_cmd = debug_cmd)
```
cathales, line 558: I think this is why the dashboard is failing. Is it an issue if you restore it?

```suggestion
  run_cmd(cmd, debug_cmd = debug_cmd)
```

cva6/sim/cva6.py
```diff
@@ -818,6 +819,10 @@ def setup_parser():
                       help="Path for the link.ld")
   parser.add_argument("--axi_active", type=str, default="",
                       help="switch AXI agent mode: yes for Active, no for Passive")
+  parser.add_argument("--gen_sv_seed", type=int, default=0,
+                      help="Run test X tims with random seed")
```
cathales, line 823:
```suggestion
                      help="Run test N times with random seed")
```

cva6/sim/cva6.py
```diff
@@ -933,10 +938,25 @@ def main():
...
+    if args.gen_sv_seed > 0:
+      args.issrun_opts = args.issrun_opts + " +ntb_random_seed_automatic"
+    elif args.gen_sv_seed == 0:
+      args.issrun_opts = args.issrun_opts + " +ntb_random_seed=" + args.sv_seed
+      args.gen_sv_seed = 1
+    else:
+      logging.error('gen_sv_seed can not take a negative value')
+
+
+
```
cathales, line 959: Jumping one line here is enough.

cathales (commented):
Thanks for the update! I think that there is still one typo left, see below.

The CI should pass this time.

Can you mark the previous conversations as solved, please?

cva6/sim/cva6.py
```diff
@@ -426,7 +426,7 @@ def run_assembly(asm_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, outp
   # ISS simulation
   for iss in iss_list:
     run_cmd("mkdir -p %s/%s_sim" % (output_dir, iss))
-    log = ("%s/%s_sim/%s.log" % (output_dir, iss, asm))
+    log = ("%s/%s_sim/%s_%s.log" % (output_dir, iss, asm, test_iteration))
```
cathales, line 429: Isn’t `test_iteration` an integer? So maybe `%d` is better there?

```suggestion
    log = ("%s/%s_sim/%s_%d.log" % (output_dir, iss, asm, test_iteration))
```

cva6/sim/cva6.py
```diff
@@ -561,7 +561,7 @@ def run_c(c_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, output_dir,
   # ISS simulation
   for iss in iss_list:
     run_cmd("mkdir -p %s/%s_sim" % (output_dir, iss))
-    log = ("%s/%s_sim/%s.log" % (output_dir, iss, c))
+    log = ("%s/%s_sim/%s_%s.log" % (output_dir, iss, c, test_iteration))
```
cathales, line 564:
```suggestion
    log = ("%s/%s_sim/%s_%d.log" % (output_dir, iss, c, test_iteration))
```

cathales (commented):
Now the issue seems to be that the `.log` name has changed as you have added a suffix.

Either update the lines containing `report_benchmark` in `.gitlab-ci/cva6.yml` or make it so that names are not changed if the arguments provided are not changed.

The second option might be better as it does not change the default behavior. With the current version of your changes, people would have to update their scripts to take this new suffix into account even if they haven’t changed the way they are calling `cva6.py`.

cva6/sim/cva6.py
AEzzejjari: Thanks for the review. I will updated the file accordingly.

PR #1947, opened 2023-05-31
Formal - Sporadic Maintenance

silabs-robin:
This PR updates the formal scripts so it can run with the latest core-v-verif code.

### DIFF
--- a/cv32e40s/tests/uvmt/base-tests/uvmt_cv32e40s_base_test_pkg.sv
+++ b/cv32e40s/tests/uvmt/base-tests/uvmt_cv32e40s_base_test_pkg.sv
@@ -18,7 +18,6 @@
 `define __UVMT_CV32E40S_BASE_TEST_PKG_SV__
 
 // Pre-processor macros
-`include "uvm_macros.svh"
 `include "uvmt_cv32e40s_macros.sv"
 
 package uvmt_cv32e40s_base_test_pkg;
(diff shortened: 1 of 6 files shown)


cv32e40s/tests/uvmt/base-tests/uvmt_cv32e40s_base_test_pkg.sv
```diff
@@ -18,7 +18,6 @@
 `define __UVMT_CV32E40S_BASE_TEST_PKG_SV__
 
 // Pre-processor macros
-`include "uvm_macros.svh"
```
silabs-robin, line 21: Note: I don't think this was necessary. It is already defined by simulators that support UVM? Removed it. ci_check still passes.

silabs-robin: @silabs-hfegran do you agree that it is safe to remove this?

silabs-robin:
I pushed one more feature: Support for running external config files (AKA `PARAM_SET_0`).

MikeOpenHWGroup:
Hi @silabs-robin, I will not be in a position to run a local CI on this until I return to Ottawa (2023-06-12).  If that is gating you, please work with @silabs-hfegran to review/test this PR.

silabs-hfegran (approved):
Looks good to me

PR #1952, opened 2023-06-01
Set Spike execution limit to 2M steps across all verification runs.

zchamski:
In order to prevent runaway executions of Spike, this PR sets an execution limit of 2 million cycles on all Spike executions triggered by the CVA6 DV scripts.  The mechanism used is the new Spike cmdline option introduced in PR #1949.

Sullary of file changes:

* cva/sim/Makefile (spike_stepout): Set up Spike stepout option if caller supplied a step count limit.
  (spike): Invoke Spike with a 'stepout' limit option (possibly empty).
* cva6/sim/cva6.yaml (iss: spike): Set step limit of Spike runs to 2M.  Add explanatory comments.

### DIFF
--- a/cva6/sim/Makefile
+++ b/cva6/sim/Makefile
@@ -38,6 +38,10 @@ issrun_opts     ?=
 isspostrun_opts ?=
 log             ?=
 variant         ?=
+# Set Spike step count limit if the caller provided a step count value in variable 'steps'.
+ifneq ($(steps),)
+  spike_stepout = --steps=$(steps)
+endif
 
 # TRACE_FAST, TRACE_COMPACT and VERDI are mutually exclusive and imply non-empty DEBUG.
 ifneq ($(TRACE_FAST),)
@@ -101,7 +105,7 @@ endif
 # Spike specific commands, variables
 ###############################################################################
 spike:
-	$(tool_path)/spike --log-commits --isa=$(variant) -l $(elf)
+	$(tool_path)/spike $(spike_stepout) --log-commits --isa=$(variant) -l $(elf)
 	cp $(log).iss $(log)
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


zchamski:
The 2M steps limit is safe as the RTL side uses a timeout of 2M cycles defined in the RVFI tracer.  At the time of writing this limit is sufficient for all CI jobs including benchmarks.

MikeOpenHWGroup:
I am fundamentally opposed to this idea.  A reference model is an individual component of a verification environment and components should not have the ability to unilaterally stop a simulation.

Individual components _may_ participate in the decision to stop a simulation (in UVM this is called "dropping your objection"), or not.

If we think we need this, we can add a "watchdog timer" in the verification environment that forces a shutdown after a (large) number of instructions are retired.

zchamski:
Hi @MikeOpenHWGroup , this PR is intended for the "pre-tandem" approach to verification that is based on full, post-mortem  trace comparison between simulators.  In this setting Spike does not interact with UVM and the Spike simulation can be terminated in four possible ways:

1. By executing a valid test termination sequence (writing an odd value to memory at the address of variable `tohost`.)  This is the expected way and it shoud be the only one ever needed.
1. In response to an interactive Ctrl-C from the user.
1. By reaching/exceeding a specified number of retired instructions (including those executed in indefinitely-looping exception handlers).
1. **By exhausting disk space on the CI server while tracing an indefinitely-looping exception handler, effectively killing the server.**

The PR aims at preventing the case 4 which still occurs way too often.

MikeOpenHWGroup:
OK..., let's do this for now.  We will need a long term solution.

MikeOpenHWGroup (approved):

zchamski:
Reworked the PR to simplify the change and remove irrelevant history.  Some of the removed changes will land in a separate PR.

MikeOpenHWGroup (approved):
Approving.  Will defer to @ASintzoff or @JeanRochCoulon for merge.

PR #1953, opened 2023-06-01
ISACOV : Add ISACOV jobs for CI & test description & script 

AyoubJalali:


MikeOpenHWGroup (commented):
LGTM @AyoubJalali.  Aa far as I can tell, this PR only affects CVA6 and it is passing CI so there is no reason not to merge.  Nevertheless, I will defer to @ASintzoff for the formal approval.

PR #1955, opened 2023-06-06
40s fixes for 0.9.0

silabs-hfegran:
- Commented out assertion/support logic that creates an unintentional loop (needs a rewrite)
- Removed tdata3 from debug_test2
- Updated core hash


cv32e40s/sim/ExternalRepos.mk
silabs-robin, line 1: ci_check still passes?
Formal still runs?

silabs-robin: (yes)

PR #1958, opened 2023-06-06
Fix #1948

### LINKED ISSUE #1948: RVFI agent : compilation error when disable debug mode
Hello,
As i know the debug mode isn't supported in the STEP1 configuration of the CVA6, so i try to disable it from the [uvme_cva6_cfg.sv](https://github.com/openhwgroup/core-v-verif/blob/cva6/dev/cva6/env/uvme/uvme_cva6_cfg.sv), but i got a compilation error says : 
###   Please make sure that the object is allocated before using it.

so i think the problem is we should create the debug csrs, if only if we're in debug mode (after the condition in [line](https://github.com/openhwgroup/core-v-verif/blob/e3ec071f66542c16a75c3ecee61e0efa62b76ae4/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv#L210) )

AyoubJalali:


lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_mon.sv
```diff
@@ -205,13 +205,13 @@ task uvma_rvfi_instr_mon_c::monitor_rvfi_instr();
...
          end
 
-         dcsr_ret_data = mon_trn.csrs["dcsr"].get_csr_retirement_data();
          // In debug mode, detect NMIP event for a data bus error
          if (mon_trn.dbg_mode &&
              !last_dcsr_nmip &&
              mon_trn.nmip[0] &&
              dcsr_ret_data[3])
          begin
+            dcsr_ret_data = mon_trn.csrs["dcsr"].get_csr_retirement_data();
```
silabs-robin, line 214: The if-condition uses `dcsr_ret_data`.
If we move its assignment down, then it will not be assigned in time for the if-statement's evaluation?

The LRM also has this:
image

AyoubJalali: you're right, i didn't see that

AyoubJalali: I'll close this, I created anothor one for the fix, (sorry lot of git issues)

PR #1959, opened 2023-06-06
convert array, since the corresponding array is convertes in the rtl

silabs-krdosvik:
1) Fixed todo, by inverting the alb_resp_q array (which is connected to the resp_q array in the alignment buffer), since the corresponding array is inverted in the rtl:
image


silabs-mateilga (approved):
Looks good!

silabs-robin (approved):

PR #1964, opened 2023-06-08
Disabled and added todo for 3 coverage crosses that needs rework

silabs-hfegran:


silabs-robin (approved):
Approving based on:
Known issue.

PR #1965, opened 2023-06-09
Updated 40S core hash

silabs-hfegran:


cv32e40s/sim/ExternalRepos.mk
silabs-robin, line 1: Both sim and formal still works?

silabs-robin: (It has been used in local regressions for the past few days.)

silabs-robin (approved):

PR #1966, opened 2023-06-09
Removed iss workaround

silabs-hfegran:
- Fixed: removed override. No longer needed, and also had a conflict with the 0-triggers configuration


silabs-robin (approved):
Approving based on:
The TODO-comment prophesized this change.

PR #1968, opened 2023-06-12
Updated default u_sw_irq_handler to support clic

silabs-hfegran:
- Fixed: u_sw_irq_handler compares mcause incorrectly when clic is used
- Fixed: replaced tabs with spaces


silabs-robin (approved):
Approving based on:
The previous reading didn't heed CLIC-specific mcause fields, but now that is fixed.
Logic seems correct to me.

PR #1970, opened 2023-06-12
Fixed marchid in cv32e40s_readonly_csr_access_test

silabs-hfegran:
- Fixed: marchid 0x14->0x15 for cv32e40s in cv32e40s_readonly_csr_access_test


silabs-robin (approved):
Approving based on:
0x15 is the correct value indeed.

PR #1971, opened 2023-06-12
Fix #1969 - use suggested module shutil instead of deprecated distutils

### LINKED ISSUE #1969: Deprecate distutils module
https://github.com/openhwgroup/core-v-verif/blob/9c8ac94d156dfc95c7e7e58be0ef1555b1280b39/bin/ci_check#L50

After [PEP 632](https://peps.python.org/pep-0632/):

For these modules or functions, use the standard library module shown:

- distutils.spawn.find_executable — use the shutil.which function

szbieg:
As suggested.

### DIFF
--- a/bin/ci_check
+++ b/bin/ci_check
@@ -47,7 +47,7 @@ import subprocess
 import pprint
 import yaml
 import re
-import distutils.spawn
+import shutil
 
 if (sys.version_info < (3,0,0)):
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (approved):
Thanks @szbieg!  Looks (and works) good to me.

PR #1972, opened 2023-06-13
Corrected an issue with corev-dv compilation using vsim & regress improvements 

XavierAubert:
I had an issue where vsim crashes while running multiple corev-dv instruction generation in parallel. After some code digging, I found that for non-corev-dv simulations, vlog/vopt combo is not called in regression environment thanks to COMP=0 parameter. This behavior was not implemented for corev-dv comp/opt/simulation, so here it is (commit 2cc69ed), and it solves my issue.

While I was looking after this, I also saw two things I considered issues for the corev-dv commands, that includes: 

- Compilation of the whole core just for generating the assembly files
- vsim flags used for corev-dv compilation/simulations are the same for the "real" simulation run

So, inside the same commit 2cc69ed, I removed core compilaton & added 2 distinct variables. Let me know if there is any specific reason to compile the core files for a testbench that doesn't stimulate it, I may have missed something... 

The second commit of the PR allows to use the `precmd` field of cv_regress script to generate in one unique simulation all assembly files for a same test, it allows reducing compilation time/CPU usage.

### DIFF
--- a/bin/templates/regress_rmdb.j2
+++ b/bin/templates/regress_rmdb.j2
@@ -15,9 +15,7 @@
 {% for r in regressions %}
             <member>{{r.name}}</member>
 {% endfor %}
-{% if coverage != false %}
-            <member>cov_report</member>
-{% endif %}
+            <member>report</member>
         </members>
     </runnable>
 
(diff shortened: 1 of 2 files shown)


XavierAubert:
I will issue a new pull requests for this, things has changed sinced then

PR #1973, opened 2023-06-13
bug correction for xpulp instruction generation

XavierAubert:
Two bug corrected: 

- assembly string not correctly generated if a comment is added to the instruction 
- bad format specifier for post_inc_store immediate


MikeOpenHWGroup (approved):
LGTM.  Passes local CI under Questasim.

PR #1975, opened 2023-06-13
ci: add synthesis checks

cathales:
The CI will flag big variations in gates count (±100 gates).


MikeOpenHWGroup:
A difference of 100 gates is less than 0.2%.  That seems very restrictive.

JeanRochCoulon:
Yes, 100 gates is restrictive, but 500 gates should be too much.
100 gates allows us to track the gate regression. I would suggest to comit 100, and update the value appears to be too restrictive.

cathales:
> BTW the alue is not weel configured because the 3 synthesis CI are RED.

Err yes I didn’t take the correct values 😁 but let’s say it was to test the condition 😉

PR #1976, opened 2023-06-13
ISACOV : Filter hazard cross

AyoubJalali:
Hello @MikeOpenHWGroup @silabs-robin ,
This PR related to filter hazard cross (RAW : Read After Write -> instr_prev.rd == instr.rs1 || instr.rs2) so all instruction do not have rd or one of the register source should not appear (**UNKNOWN_GROUP, FENCE_GROUP, FENCE_I_GROUP, RET_GROUP, ENV_GROUP**) in, the cross also :  
for current group of instruction should not have : **JUMP_GROUP** -> jump instruction do not have source register (only have rd)
for Previous group of instruction should not have : **STORE_GROUP & BRANCH_GROUP** -> Store & branch instruction do not have destination register (only have rs)

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -1459,10 +1459,23 @@ covergroup cg_sequential(string name,
   cross_seq_group_x3: cross cp_group, cp_group_pipe_x2, cp_group_pipe_x3;
   cross_seq_group_x4: cross cp_group, cp_group_pipe_x2, cp_group_pipe_x3, cp_group_pipe_x4;
 
-  // FIXME: This will need more filtering
   cross_seq_gpr_raw_hazard: cross cp_group, cp_group_pipe_x2, cp_gpr_raw_hazard {
     // Ignore non-hazard bins
     ignore_bins IGN_HAZ = binsof(cp_gpr_raw_hazard) intersect {0};
+    ignore_bins IGN_GROUP = binsof(cp_group) intersect {UNKNOWN_GROUP,
+                                                        FENCE_GROUP,
+                                                        FENCE_I_GROUP,
+                                                        RET_GROUP,
+                                                        WFI_GROUP,
+                                                        ENV_GROUP};
+    ignore_bins IGN_PREV_GROUP = binsof(cp_group_pipe_x2) intersect {UNKNOWN_GROUP,
+                                                                     FENCE_GROUP,
+                                                                     FENCE_I_GROUP,
+                                                                     RET_GROUP,
+                                                                     WFI_GROUP,
+                                                                     ENV_GROUP,
+                                                                     STORE_GROUP,
+                                                                     BRANCH_GROUP};
   }
 
   cross_seq_csr_hazard_x2: cross cp_csr, cp_instr, cp_csr_hazard {


lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
silabs-robin, line 1: Hi. I have read through and tried to understand.

Why is the JUMP_GROUP filtered out from cp_group?
The PR description says _"jump instruction do not have source register (only have rd)"_, but [JUMP_GROUP](https://github.com/openhwgroup/core-v-verif/blob/ed36379808945b6ed01115cf561b8a12f450df15/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv#L774) includes JALR, and JALR uses rs1.

Can these be illegal bins?
Adding ignores is always risky, because it is difficult to know if one accidentally have ignored something one didn't mean to (like possible the JUMP_GROUP above, or future instructions being added to a group).
(Note, this is not a checker of the DUT, it is a checker of the coverage model's own integrity, just like how the [get_instr_group](https://github.com/openhwgroup/core-v-verif/blob/ed36379808945b6ed01115cf561b8a12f450df15/lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv#L815) function ends with a fatal error if it failed to map incomming instructions to a group.)

AyoubJalali: I think i confuse it with J_type, sorry for that.


> Can these be illegal bins?

If you wanna do that, you should set all the bins in the cg as illegal.

silabs-robin: > illegal means that's should not have for example (BRANCH -> JUMP), and that's not true, because we ignore it because we don't see the goal to cover it for HAZARD

This is within `cross_seq_gpr_raw_hazard`, so wouldn't it mean that `(BRANCH -> JUMP) && hazard` is illegal?
If we are ignoring this case, we must be ignoring it because it truly cannot happen, and if we claim that it truly cannot happen we might as well ask SystemVerilog to check if our claim is correct, IMO.

silabs-robin: What exactly is not possible?

Why do you propose to ignore a coverage if it CAN be hit?

AyoubJalali: we can't set all bins illegal, and specially **cp_gpr_raw_hazard** , that's why it's ignored

AyoubJalali: > Can these be illegal bins?

for this, I get an error because we hit this sequence BRANCH -> ANYTHING, here a group instruction of branch isn't illegal but, isn't have any goal to cover for hazard cg, because BRANCH INSTRUCTIONS does not have destination register, so we should ignore.

silabs-robin: Even when using `&& binsof etc etc (hazard)`?

Coverage filtering can be very tricky sometimes, so it might reach a point where the effort is not worth it.

I think it would be safer to figure out the correct way to mark these impossible bins as illegal, so that we don't accidentally declare something as ignored when we in reality would like to see covers for it, but if it is too cumbersome then it might be more practical and expedient to just keep these as ignores.

AyoubJalali: it didn't work because simply, it's not an illegal bins, it just bins that we don't need on the HAZARD context.

silabs-robin: > it's not an illegal bins, it just bins that we don't need

Then why ignore it?

--------

We have 6 cases:
1. Bin can be hit + we cover it: This is the ideal case.
2. Bin can be hit + we ignore it: Why would anyone want to do this?
3. Bin can be hit + we illegalize it: Why would anyone want to do this?
4. Bin can not be hit + we cover it: This cannot happen so you get unmet coverage.
5. Bin can not be hit + we ignore it: Decent way to handle unmet coverage, but you must _hope_ that you write the filter correct.
6. Bin can not be hit + we illegalize it: Decent way to handle unmet coverage, and you can _know_ that you write the filter correct.

AyoubJalali: > Then why ignore it?

take this a example FENCE -> ALU GROUP , this sequence is a normal one, that's means it's not illegal, we can have a FENCE then ALU GROUP instruction, but we're ignore it because this cover group is related to hazard bins, and having a prev instruction without a destination register or a current instruction without rs1 && rs2 is not a bins that i'm interesting covering it because it have not sense in our context (hazard).

AyoubJalali: > because then the "hazard" function is defined incorrectly).

the hazard function uses rx_valid signals, which set related to instruction type, but in the cross_seq_gpr_raw_hazard, we use instruction group, so that's different, for example J_type -> has only rd, but JUMP_GROUP has rs1 or rs2 because it have also jalr, so that's why there a problem in some cases

silabs-robin: > the hazard function uses rx_valid signals, which set related to instruction type, but in the cross_seq_gpr_raw_hazard, we use instruction group, so that's different

Yes they are different, but I have not seen any reason to think that it matters.

--------

> that's why there a problem in some cases

Please show me one case and I will change my mind immediately.

Please show a case of any two instructions, matching the groups in the new ignore bins, which could come with a hazard, such that it would be a violation if it was an illegal bin.

MikeOpenHWGroup: @silabs-robin said:
> We have 6 cases:
> ...snip...
> 5. Bin can not be hit + we ignore it: Decent way to handle unmet coverage, but you must hope that you write the filter correct.
> 6. Bin can not be hit + we illegalize it: Decent way to handle unmet coverage, and you can know that you write the filter correct.

If I am following this discussion thread correctly, these are the two cases @AyoubJalali is talking about.  Furthermore, I _think_ he has implemented case six.  If that is the case, can we resolve this conversation and merge this in now?

silabs-robin: I think case 5 is implemented, which is an acceptable solution. Let's mark it as resolved.

silabs-robin (approved):

PR #1979, opened 2023-06-14
AXI/assertion modification

AEzzejjari:
CVA6 can perform a read request with a length equal to 0 or 1, so I've modified this assertion to ensure that ARLEN can only be equal to 0 or 1.


MikeOpenHWGroup:
@AEzzejjar, please provide an english language description of the intention of the modification in the Description field.

PR #1985, opened 2023-06-19
change where integrity signal is fetched from

silabs-krdosvik:
Make sure we fetch a valid obi req integrity signal.
(And place the mcause_shadow signal with the other shadow signals)

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
@@ -1369,11 +1369,10 @@ module uvmt_cv32e40s_tb;
         .lrfodi_bus_req (core_i.load_store_unit_i.buffer_trans_valid),
         .lrfodi_bus_gnt (core_i.load_store_unit_i.buffer_trans_ready),
 
-        .req_is_store (core_i.load_store_unit_i.bus_trans.we),
-        .req_instr_integrity (core_i.if_stage_i.mpu_i.bus_trans_integrity),
-        .req_data_integrity (core_i.load_store_unit_i.mpu_i.bus_trans_integrity),
+        .req_is_store (core_i.m_c_obi_data_if.req_payload.we),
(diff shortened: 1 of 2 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -1369,11 +1369,10 @@ module uvmt_cv32e40s_tb;
         .lrfodi_bus_req (core_i.load_store_unit_i.buffer_trans_valid),
         .lrfodi_bus_gnt (core_i.load_store_unit_i.buffer_trans_ready),
 
-        .req_is_store (core_i.load_store_unit_i.bus_trans.we),
-        .req_instr_integrity (core_i.if_stage_i.mpu_i.bus_trans_integrity),
-        .req_data_integrity (core_i.load_store_unit_i.mpu_i.bus_trans_integrity),
+        .req_is_store (core_i.m_c_obi_data_if.req_payload.we),
+        .req_instr_integrity (core_i.m_c_obi_instr_if.req_payload.integrity),
+        .req_data_integrity (core_i.m_c_obi_data_if.req_payload.integrity),
```
silabs-robin, line 1374: I am not familiar with this interface.

Was the previous probing done too early in the signal chain?

Reading this signals from this interface is better because it has the same timing as `.data_bus_req (core_i.m_c_obi_data_if.s_req.req),` etc above, right?

silabs-krdosvik: Exactly,  the probing was done too early.

The interface is connected to the obi signals in the core. But it seems like there is not an obi signal for integrity, so the m_c_obi_data/instr_if is the closes we get to obi. So yess, the timing should be the same as for the .data_bus_req signal.

silabs-robin: Ok. Thanks.

silabs-robin (approved):
Approving based on:
Explanation given in comments.

PR #1989, opened 2023-06-20
cv32e40s: Testbench updates

silabs-hfegran:
- Updated cv32e40s hash to latest
- Added $assertkill() to avoid liveness assertion errors upon test termination
- Added advanced debugging option flags to xrun makefile (SRC_DEBUG)
- Qualified CSR logic in imperas_dv_wrap with rvfi_valid


silabs-hfegran:
ci_check OK, changes should have no impact on formal verification.

silabs-robin (approved):
Approving based on:
All of this was discussed IRL before.
Looks like expected.

PR #1990, opened 2023-06-20
[cv32e40p] adding rv32f/zfinx functional coverage based on v2 testplans

dd-vaibhavjain:
Adding new functional coverage for RV32F and Zfinx extensions for cv32e40p. This is based on testplans for cv32e40p v2.
We have testplan reviews ongoing and this first version of the functional coverage and can be modified based on review outcomes and coverage analysis.
Functional coverage is for F-multicycle scenarios, operand register corner cases and different cases for various cv32e40p FPU configurations.

### DIFF
--- a/cv32e40p/env/uvme/cov/uvme_rv32f_covg.sv
+++ b/cv32e40p/env/uvme/cov/uvme_rv32f_covg.sv
@@ -1,0 +1,612 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright 2023 OpenHW Group
+// Copyright 2023 Dolphin Design
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://solderpad.org/licenses/
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+//
+///////////////////////////////////////////////////////////////////////////////
+
+class uvme_rv32f_zfinx_covg extends uvm_component;
+    /*
+    * Class members
+    */
+    uvme_cv32e40p_cntxt_c  cntxt;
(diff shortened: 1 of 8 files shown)


cv32e40p/env/uvme/cov/uvme_rv32f_covg.sv
```diff
@@ -0,0 +1,1127 @@
...
+            option.weight = 5;
+        }
+
+        //cross coverage for F-inst in ID-stage with preceeding F-multicycle instr
+        cr_f_inst_at_id_stage_inp_with_fpu_multicycle_req : cross cp_id_stage_inst_valid, cp_id_stage_f_inst, cp_apu_req_valid, cp_apu_grant_valid, cp_curr_fpu_apu_op {option.weight = 50;}
+
+        //cross coverage for F-inst in ID-stage with preceeding F-multicycle - case with apu_busy or APU needing more than 1 clock cycle 
+        cr_f_inst_at_id_stage_inp_while_fpu_busy : cross cp_id_stage_inst_valid, cp_id_stage_f_inst, cp_apu_busy, cp_curr_fpu_apu_op {
+            option.weight = 50;
+`ifdef FPU_LAT_0_CYC
```
MikeOpenHWGroup, line 207: Please do _not_ use compiler macros (ifdefs) in the functional coverage model.  We can use parameterization for this.

MikeOpenHWGroup (changes requested):
This is a great contribution @dd-vaibhavjain!  A couple of comments:
1. We need to find a way to implement conditional coverage bins without using compiler macros (ifdefs). 
2. It would be **great** to integrate this into `ISACOV`.

cv32e40p/env/uvme/cov/uvme_rv32f_covg.sv
dd-vaibhavjain: Thanks @MikeOpenHWGroup for reviewing this. Agreed with your suggestion, that should be the right way to do. Will work to make this covergroup parameterized.

dd-vaibhavjain: Regarding integrating it in ISACOV, I think at the moment this covergroup will not be directly able to integrate there. And would need more work and possibe re-code from scratch. 
Maybe I can take this as an action item for this at the moment. And do this from scratch again for ISACOV ?

dd-vaibhavjain:
This is currently being re-worked to use ISACOV and incorporate other comments

dd-vaibhavjain:
Closing this and all comments to be incorporated in a new PR

PR #1995, opened 2023-06-20
Updated README file for CV32E40P vplans and first vplans update 

XavierAubert:
I updated the readme file to be easier and cleaner to read, with direct link and status to the vplans relevant for v2 verifications

I also added a first batch of updates to the vplans, containing small updates from the first reviews, work is still in progress


cv32e40p/docs/VerifPlans/README.md
```diff
@@ -1,15 +1,22 @@
...
+
 ## Verification Plan Status
 
-The tables below capture the current status of the Verification Plan for the CV32E40P by high-level feature, as long with status update with respect to CV32E40Pv1 verification plans.  Under the heading `Capture`, the test plan can be **Incomplete**, **draft**, or one of the following: 
-* **Captured _(v1)_**: Vplan has been captured during v1 verification and has no modifications since then. 
-* **Captured _(new)_**: Vplan captures either new features or features not tested in v1
-* **Captured _(formal)_**: Vplan captures a different methodology from v1, using formal tool
-* **Captured _(updated)_**: Vplan has been captured in v1, but contains modifications for v2
+The tables below capture the current status of the Verification Plan for the CV32E40P by high-level feature, as long with status update with respect to CV32E40Pv1 verification plans.  Under the heading `Capture`, the test plan can be **Incomplete**, **draft**. If the verification is **captured**, it has one of the following status:
```
MikeOpenHWGroup, line 15: The heading `Capture` is used in the V1 tables, but in the V2 table the heading is `VPlan Status`.  Maybe we can use VPlan Status for both.

cv32e40p/docs/VerifPlans/README.md
```diff
@@ -18,66 +25,88 @@ Under the heading `Review` is one of following:
...
-| FNMADD/FMADD | Captured (**new**) | Ready for Review | Formal coverage issue: some bits of operands have been tied low | 
-| FNMSUB/FMSUB | Captured (**new**) | Ready for Review | Formal coverage issue: some bits of operands have been tied low |
-| Other F instructions | Captured (**new**) | Ready for Review | Formal pipeline issue: Add preceeding F multicycle instruction |
+| OBI     | Complete | Reviewed | |
+| Sleep Unit | Complete | Reviewed | Updates pending based on review feedback |
+| Pipelines | Complete | Reviewed | Updates pending based on review feedback|
 
 ### Xpulp instruction extensions
-**Note**: Xpulp instructions, as other instructions, has been mainly verified by formal tools. Some aspects needed a complementary simulation verification
+**Note**: Xpulp instructions are "exercised, but not fully verified" in CV32E40P.
```
MikeOpenHWGroup, line 100: Is that true?  It was my understanding that as far as possible, Xpulp instructions were fully verified in formal and any Xpulp instructions not fully verified in formal would be fully verified in simulation.

MikeOpenHWGroup (commented):
This is great @XavierAubert.  I will start reviewing the Xpulp spreadsheets now (but we can merge this in even before that).  I have a couple of comments about the README, please have a look.

cv32e40p/docs/VerifPlans/README.md
pascalgouedo: When talking about partly verified Pulp instructions, it is mainly about multiplication operands for either scalar or SIMD instructions.
Same case for Floating-Point multiplications.

cv32e40p/docs/VerifPlans/README.md
XavierAubert: There is a typo in the part of the file you are showing, I meant to use `VPlan Status` as a header and not `Capture`, I forgot to change that. For the V1 table, the idea was to left the v1 work as it was before starting working on the v2, but I can change little things like this if you prefer

cv32e40p/docs/VerifPlans/README.md
XavierAubert: I restored the v1 file as it was, and it was saying "Xpulp instructions are "exercised, but not fully verified" in CV32E40P."

The part you mention is about v1 verification **only**.

For v2, is it exactly as Pascal said

MikeOpenHWGroup: Interesting...  I have used similar strategies in the past and consider that to be "fully verified".  There is no need to verify that a specific operation is correct for all possible input values.  Your strategy to constrain bits [29:02] to either 28'b0 or 28'b1 is perfectly valid and sufficient to call the operation "fully verified".

pascalgouedo: Hi @MikeOpenHWGroup 
I agree with you for the general case of 32 bits scalar integer instructions.
It is not fully verified for SIMD operations. I would like to have this free bits strategy applied depending on data size.
So when cutting the 32 bits in 2 or 4 independent pieces, having free bits on each part would allow full verification.

MikeOpenHWGroup:
Thanks for your prompt comments @XavierAubert and @pascalgouedo.  All conversations in this PR are now resolved so I will merge it in.

PR #2001, opened 2023-06-22
Adding hwloop support for cv32e40p

YoannPruvost:


MikeOpenHWGroup (approved):
LGTM @YoannPruvost, it passes a local regression, but I cannot properly test this without ImperasDV.  In any case, I doubt it will break anything, so I am approving and merging it.

Issue #1569, opened 2023-01-09
OBI coverage model does not compile with VCS


guojiazhuxi:
<br>**Environment**: OS：VMware Ubuntu 20.04.5 LTS  VCS:O-2018.09-SP2-3

## Bug 1：

**core-v-verif/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv** Line 51-68 of the file, use VCS to compile and report an error. The error is shown in the figure below.
![image](https://user-images.githubusercontent.com/49388283/211302357-3162768d-7d17-4f09-ad5e-deb672cedaf0.png)
In order to run smoothly, I first commented out lines 51-68

## Bug 2：

In **core-v-verif/mk/Common.mk** file. The **CFLAGS** variable in line 316 of the file should be **export**, Because this variable is transmitted to the Makefile file behind.

## Bug 3：

1.Line 31 in file **core-v-verif/cv32e40p/tb/uvmt/uvmt_cv32e40p.flist**, It should not be **$(DV_SVLIB_PATH)**，It should be **${DV_SVLIB_PATH}**

## Bug 4：In file **core-v-verif/mk/vcs.mk**.

1.Line 33, `SIMV = $(CV_TOOL_PREFIX) simv -licwait 20` is error , It should be `SIMV = $(CV_TOOL_PREFIX)simv -licwait 20` ,No space before simv

2.Line 189, `cd $(SIM_CFG_RESULTS)/$(CFG) && $(VCS) $(VCS_COMP) -top uvmt_$(CV_CORE_LC)_tb` is error, It should be `cd $(SIM_CFG_RESULTS) && $(VCS) $(VCS_COMP) -top uvmt_$(CV_CORE_LC)_tb`

3.Line 223, **VCS_RESULTS** is no define var, It should be use **SIM_RESULTS**

## problem:
After modifying the above problems and verification software, I took the following steps. May I ask, how can I join the License of OVPSIM?
![image](https://user-images.githubusercontent.com/49388283/211313286-ca566280-7e40-42cd-9209-808843178715.png)

MikeOpenHWGroup:
Hi @guojiazhuxi, thanks for your issue.  AFAIK, currently there are no OpenHW members using VCS for CV32E40P in CORE-V-VERIF, so the Makefiles have suffered from "bit rot".

We try to avoid discussing multiple topics in a signal issue, so we will use this issue to resolve **Bug 1** only.
[...]

PR #2006, opened 2023-06-23
Reinstate VCS support for CV32E40Pv2 on master branch

MikeOpenHWGroup:
This branch was originally developed for Prabha Krishnaswami of Synopsys to reinstate VCS support of CV32E40P v2 on the master branch.   It addresses two problems:
1. VCS support in the Makefiles and related scriptware was out-of-date.
2. The "Functional Coverage Invalid With Clause Specification" (FCIWCS) VCS error in the OBI Memory functional coverage model ([link](https://github.com/openhwgroup/core-v-verif/blob/4bb9858cd7c58f8856ff544f53fe102c76ea9683/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv#L52)).


MikeOpenHWGroup:
Hi @XavierAubert, I have asked for your review for obvious reasons: this will affect the CV32E40P environment on the master branch of CORE-V-VERIF.

Hi @silabs-robin and @AyoubJalali, I have asked for your review because it contains a fix suggested by SNPS for the "Functional Coverage Invalid With Clause Specification" (FCIWCS) error emitted by VCS.  The fix here is applied to the OBI coverage model (`lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv`), but it if works there it should also work for ISACOV.

MikeOpenHWGroup:
Note.  This PR should also resolve #1569.

pascalgouedo:
> Hi @XavierAubert, I have asked for your review for obvious reasons: this will affect the CV32E40P environment on the master branch of CORE-V-VERIF.

Hi @mike
We never used master because of that:

image

MikeOpenHWGroup:
Yes, I am aware that the master branch is not used by the CV32E40P v2 team.  But others who use the v1 want to use VCS and the purpose of this PR is to allow them to do that on the master branch.  Once this PR is "in" I will create another PR to allow VCS support for v2 on the cv32e40p/dev branch.

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv
```diff
@@ -49,23 +49,23 @@ covergroup cg_obi(string name,
...
    }
 
    memtype: coverpoint (trn.memtype) {
-      ignore_bins IGN_MEMTYPE = {[0:$]} with (!is_1p2);
+      ignore_bins IGN_MEMTYPE = {[0:$]} with ((item>1) && (!is_1p2));
    }
 
    prot: coverpoint (trn.prot) {
-      ignore_bins IGN_MEMTYPE = {[0:$]} with (!is_1p2);
+      ignore_bins IGN_MEMTYPE = {[0:$]} with ((item>1) && (!is_1p2));
```
silabs-robin, line 63: What about `trn.prot == 0`?

MikeOpenHWGroup: Are you suggesting:
```
ignore_bins IGN_MEMTYPE = {[0:$]} with ((trn.prot==0) && (!is_1p2));
```

silabs-robin: Will that work?
There is no _"item"_ in that, so VCS will probably still throw an error/warning?

I think maybe `(item >= 1)` should work though.
My point was only that the coverpoint starts out with `[0:$]`, but the `item>1` would inadvertently remove the `0`.

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv
```diff
@@ -49,23 +49,23 @@ covergroup cg_obi(string name,
    option.name         = name;
 
    we: coverpoint (trn.access_type) {
-      ignore_bins IGN_WRITE = {UVMA_OBI_MEMORY_ACCESS_WRITE} with (!write_enabled);
-      ignore_bins IGN_READ =  {UVMA_OBI_MEMORY_ACCESS_READ} with (!read_enabled);
+      ignore_bins IGN_WRITE = {UVMA_OBI_MEMORY_ACCESS_WRITE} with ((item>1) && (!write_enabled));
+      ignore_bins IGN_READ  = {UVMA_OBI_MEMORY_ACCESS_READ}  with ((item>1) && (!read_enabled));
```
silabs-robin, line 53: Isn't `UVMA_OBI_MEMORY_ACCESS_READ` defined to be `0`?

image

So `(item>1)` would leave this bin empty?

MikeOpenHWGroup: I think you're right.  So it should be `item>0`, no?

silabs-robin: Woops, sorry I think I wrote it wrong in my other comment as well.

Actually I would hope that `(item || 1)` would be enough to silence VCS.

But `(item >= 0)` should be correct when dealing with unsigned signals.

MikeOpenHWGroup: Done.

lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_cov_model.sv
MikeOpenHWGroup: Changed to `(item >= 0)`.

silabs-robin (approved):
Approving based on:
The changes in `uvma_obi_memory_cov_model.sv` and `mk/Common.mk` look ok to me.
I don't foresee any big problems with `libdpi_dasm.so`.
And the rest is 40p or vcs related changes which I am not concerned about.

MikeOpenHWGroup:
Thanks @silabs-robin.  Merging.

PR #2008, opened 2023-06-26
ISACOV : Update test integration

AyoubJalali:


cva6/tests/custom/isacov/branch_to_zero.S
```diff
@@ -34,12 +34,21 @@ branch:
   c.bnez     s1, branch
 branch1:
   c.beqz     s0, branch1
+branch2:
```
MikeOpenHWGroup, line 37: How does the code break out of this loop?  Is there a timer-based trap in the BSP that changes `t5`?

MikeOpenHWGroup (changes requested):
Hi @AyoubJalali, I have a question about the `branch_to_zero` test-program and also a request. My request is that you add a proper description to all of your pull-requests.  You should _always_ explain the purpose of the pul-request in the description field to answer questions such as:
- **Why** is this PR needed? (does it fix a bug?  does it add a new capability?)
- **How** does this PR do that?

image

cva6/tests/custom/isacov/branch_to_zero.S
AyoubJalali: To cover immb = 0x0 I give a false condition (isn't like jump); so if we give a s0 != zero we gonna cover immb = 0x0 because it's been decoded also we don't get a infinite loop.

MikeOpenHWGroup: So where does `s0` get set?

cva6/tests/custom/isacov/branch_to_zero.S
```diff
@@ -34,12 +34,21 @@ branch:
...
+  beq        a1, zero, branch4
+branch5:
+  bne        t5, t5, branch5
   addi       a0, a0, 22
   srl        a4, a5, a0
   csrrsi     t4, 0x340, 0
   sltiu      t4, s9, 236
   remu       t4, tp, a1
   c.andi     a5, -1
+  lui        zero, 0
```
silabs-hfegran, line 51: Loading 0 to x0 is not really meaningful, is this just an intentional nop? if so.. why?

cva6/tests/custom/isacov/branch_to_zero.S
silabs-hfegran: The code makes sense to me, as long as you initialize all the registers that are used for comparison (so I am also missing an `s0` initialization here)

PR #2012, opened 2023-06-27
Cv32e40p/dev baoshan xpulp mac

dd-baoshan:
PR for following updates
1)  Update pulp sanity test; Add random tests for MAC and SIMD
2) Fix MAC assignment for has_rd
3) Add IMM_BRANCH support in post_processing function
4) Update pulp base sanity, MAC and SIMD classes


cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
```diff
@@ -16,118 +17,334 @@
...
+      gen_directed_instr(.idx(i));
+    end
+    super.post_randomize();
 
-    all_xpulp_instr = placeholder_ins.instr_group[RV32X];
+  endfunction : post_randomize
 
-    if (all_xpulp_instr.size() == 0) begin
-      `uvm_fatal(this.get_type_name(), "No instruction inside RV32X. Check that RV_DV_ISA is correctly set")
+  virtual function void print_stream_setting();
```
dd-vaibhavjain, line 94: Should use `uvm_object_utils_begin and uvm's print() method?

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
```diff
@@ -16,118 +17,334 @@
...
+    use_pulp_only_for_directed_instr == 1;
+    use_no_repetitive_instr_per_stream == 1;
+    if (use_no_repetitive_instr_per_stream) {
+      num_of_instr_per_stream == riscv_instr::instr_group[RV32X].size() - bypass_xpulp_instr_list.size() - 1;
+    }
+  }
+
+  virtual function void update_current_instr_arg_list(int idx=0);
+    super.update_current_instr_arg_list(idx);
+    if (!(CV_BEQIMM inside {exclude_instr})) exclude_instr = new[exclude_instr.size() + 1] ({exclude_instr, CV_BEQIMM});
```
dd-vaibhavjain, line 255: new[] is unnecessary here.  this should be enough -> exclude_instr = {exclude_instr, CV_BEQIMM};

and also since CV_BEQIMM/CV_BNEIMM are already supported as i am guessing, these should be controlled as part of any other global config for branch instructions something like isnt it?
if !(cfg.no_branch)

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
```diff
@@ -16,118 +17,334 @@
...
+    solve use_pulp_only_for_directed_instr   before use_no_repetitive_instr_per_stream;
+    solve use_no_repetitive_instr_per_stream before num_of_instr_per_stream;
+  }
 
-  function void post_randomize();
-    super.post_randomize();
-  endfunction : post_randomize
+  function void pre_randomize();
+    super.pre_randomize();
+    num_of_avail_regs_gpr = 4; // limit to 4 only registers: for rd, rs1, rs2 and +1 extra
```
dd-vaibhavjain, line 335: Why such a tight constraint in base test? It should be random and maybe have some constraints with distribution to limit to a small number like this in some cases.
even then there should be a range rather than such a hard constraint.

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
```diff
@@ -16,118 +17,334 @@
...
+  int unsigned            num_of_avail_regs_gpr, num_of_avail_regs_rv32c;
+  bit                     more_xpulp_instr;
+  bit                     is_rv32f = riscv_instr_pkg::RV32F     inside {riscv_instr_pkg::supported_isa};
+  bit                     is_zfinx = riscv_instr_pkg::RV32ZFINX inside {riscv_instr_pkg::supported_isa};
+
+  rand int unsigned       num_of_instr_per_stream;
+  rand bit                use_same_reglist_per_instr;         // use same reglist for directed instrs per stream
+  rand bit                use_rand_fields_per_instr;          // workaround on vsim issue: same instr per stream has same fields value
+  rand bit                use_pulp_only_for_directed_instr;   // use pulp instr only as directed instrs per stream
+  rand bit                use_no_repetitive_instr_per_stream; // directed instr is not allow to repeat in a stream
```
dd-vaibhavjain, line 45: I dont see any need to have such constraint in random tests to not allow same instruction to be repeated. 
ISA has limited instructions, and if the stream enforces such things then it is contrary to randomization objective and also it will limit the max number of instructions that can be even generated in this stream.

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
```diff
@@ -16,118 +17,334 @@
...
-      gen_xpulp_instr({all_xpulp_instr[i]});
+  virtual function void update_next_instr_arg_list(riscv_instr instr=null, int idx=0);
+    assert(instr != null);
+    if (use_no_repetitive_instr_per_stream) begin
+      exclude_instr = new[exclude_instr.size() + 1] ({exclude_instr, instr.instr_name});
     end
-  endfunction : post_randomize
+  endfunction : update_next_instr_arg_list
+
+  virtual function void modify_instr_fields(riscv_instr instr=null, int idx=0);
```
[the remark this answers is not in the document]
dd-baoshan: -	This is a workaround function in conjunction with vsim issue when same instruction within a stream have same fields values (rd, rs1, rs2)
-	  rand bit                use_rand_fields_per_instr;          // workaround on vsim issue: same instr per stream has same fields value
-	E.g without workaround running with vsm (highlighted instr have same rd, rs1, rs2)
image
-	E.g with workaround function
image

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
dd-baoshan: -	In this case, the array is expanding to accept new element. It need to be resized
-	This line is under sanity test. This is to retain the legacy sanity test to cycle through all the RV32X instructions. Previously it exclude cv_<branch> and in this case I Include it back into sanity test

cv32e40p/env/corev-dv/instr_lib/cv32e40p_pulp_instr_lib.sv
dd-baoshan: all the use_* are knobs in manipulating the constraints. the no_repeat is mainly use for legacy sanity tests and also has less weightage in rand distribution for other extended streams.

dd-baoshan:
Need further revise internally

PR #2014, opened 2023-06-27
Reinstate VCS support for CV32E40Pv2 on cv32e40p/dev branch

MikeOpenHWGroup:
For longer than I can remember, no active member has been using VCS to run simulations of CV32E40P in CORE-V-VERIF.  This pull-request restores the ability to use VCS (and also fixes a latent bug in `ci_check` that I did not realise was there).

Note that this PR does (almost) the same thing as #2006 on the master branch.

### DIFF
--- a/mk/uvmt/vsim.mk
+++ b/mk/uvmt/vsim.mk
@@ -682,7 +682,7 @@ cov: $(COV_MERGE_TARGET)
 # Clean up your mess!
 
 clean:
-	rm -rf $(SIM_RUN_RESULTS)
+	rm -rf $(SIM_RESULTS)
 
 clean_rtl:
 	rm -rf $(CV_CORE_PKG)
--- a/bin/ci_check
+++ b/bin/ci_check
@@ -47,7 +47,7 @@ import subprocess
 import pprint
 import yaml
 import re
-import distutils.spawn
+import shutil
 
 if (sys.version_info < (3,0,0)):
     print ('Requires python 3')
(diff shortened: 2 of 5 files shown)


mk/uvmt/vsim.mk
```diff
@@ -682,7 +682,7 @@ cov: $(COV_MERGE_TARGET)
 # Clean up your mess!
 
 clean:
-	rm -rf $(SIM_RUN_RESULTS)
+	rm -rf $(SIM_RESULTS)
```
dd-vaibhavjain, line 685: Here if this ->SIM_RESULTS if used, it would mean the entire directory vsim_results/xrun_result/vcs_result will be removed. 
So need to be careful using clean after this for older users of cv32e40p.
Previously SIM_RUN_RESULTS would only clean the current test's run directory.

Maybe we will add a new target like "clean_test_dir" to only test run dir.

dd-vaibhavjain (approved):

mk/uvmt/vsim.mk
XavierAubert: As there is already the `clean_all` target, maybe add that `rm -rf $(SIM_RESULTS)` under that ?

dd-vaibhavjain:
Simulation results with Xcelium - **Pass**
Simulation result with questa - **Pass** (Checked that modified coverage bins are still correctly created)
Simulation result with VCS - **Pass** for (default config only) , fail for other configs.

I think we can go ahead with this PR merge and will fix the failure with other(non-defult) configs for vcs after this PR.

There are also few things in this vcs.mk change which are removed but i thought were useful and I would like to ask if it would be fine to add those things again after this PR is merged (via a new PR from my side), things like:
FSDB=yes, riscof setup, POST_TEST etc, -full64 (vcs option) , old "clean" target equivalent.

MikeOpenHWGroup:
Thanks @dd-vaibhavjain.  Your point is well taken and I have created #2018 to track it.

PR #2021, opened 2023-07-04
ci: move expected results to cva6 repository

cathales:
It makes it possible to perform changes which have an impact on CVA6
area without having to open a second MR on core-v-verif.

It is still needed to "declare" the change of area in a diff which is
reviewed, but now the diff is in the same repository as the
modification.

### DIFF
--- a/.gitlab-ci/scripts/report_synth.py
+++ b/.gitlab-ci/scripts/report_synth.py
@@ -10,6 +10,7 @@
 import re
 import sys
 import os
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (approved):

MikeOpenHWGroup:
LGTM @cathales.  I will defer to @JeanRochCoulon and/or @ASintzoff to merge.

PR #2023, opened 2023-07-04
ci: report environment failures in dashboard

cathales:
Infrastructure issues sometimes occur, preventing some jobs requiring specific things from starting. So these jobs do not produce reports at all. As a result, jobs in the dashboard are "successful", which is misleading.

To fix that, this PR adds a job which is run only if one of the previous jobs failed in GitLab words (including the case when the job did not start). There are no specific requirements for this new job so there is no reason it doesn’t start.

### DIFF
--- a/.gitlab-ci/cva6.yml
+++ b/.gitlab-ci/cva6.yml
@@ -71,6 +71,7 @@ stages:
   - smoke tests
   - verif tests
   - backend tests
+  - find failures
   - report
 
 .verif_test:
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
Côme, To be compliant with Eclipse rules, a description need to be provided. Please.

cathales:
I set draft status until #2039 is merged, then I’ll rebase

MikeOpenHWGroup (approved):
CI is passing, so I will approve/merge

PR #2027, opened 2023-07-05
Removed b_ext_abs cfg from b-ext test in full regression 

silabs-hfegran:
With the addition of zbc-instructions, the b_ext_test is incompatible with the b_ext_abs-cfg, and this configuration has been deprecated from full regression runs


silabs-robin (approved):
Approving based on:
While I am not intimately familiar with the relationship between zbc and b_ext_abs, the diff seems consistent with the PR description and I trust the intention behind the PR description.

Issue #2018, opened 2023-06-30
(Re) add features to VCS Makefile for CV32E40P


MikeOpenHWGroup:
As pointed out by @dd-vaibhavjain in #2014, there are few things in [vcs.mk](https://github.com/openhwgroup/core-v-verif/blob/309ac0c9ea81b0a2956e8b26d2cca754666b7403/mk/uvmt/vcs.mk#L1) which are removed we would like to add those things again.  Things like:
- FSDB=yes
- riscof setup,
- POST_TEST
- -full64 (vcs option)

dd-vaibhavjain:
This issue can be closed now as all these changes are added back to vcs.mk for cv32e40p through the PR #2029 and merged successfully to cv32e40p/dev branch.

PR #2029, opened 2023-07-05
[cv32e40p] make file updates and minor updates for riscof setup

dd-vaibhavjain:
This PR includes:
- fix for open action item for vcs.mk #2018 
- add new clean_test target in make files for just removing given test(left clean target untouched), improve clean_all
- update riscof plugin and related make targets for minor improvements, enable iss and coverage.
- update for fix from riscv_arch_test suite for mtvec align
- add 2 regress lists for these riscv_arch_test

- Add support for additional runtime plusargs passed from a separate yaml file in test_cfg/ dir
- Use the test_cfg yaml as additional "test configuration" to build the test simulation run dir paths, log names, coverage db name.
- add 2 test_cfg yaml

- Fix for Imperas coverage defines name change


bin/cfgyaml2make
```diff
@@ -48,6 +48,7 @@ TOPDIR = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)
 REQUIRED_KEYS = ('name', 'description',)
 CFG_PATHS = (
              '<CV_CORE>/tests/cfg',
+             '<CV_CORE>/tests/test_cfg',
```
XavierAubert, line 51: After a second reading, I am wondering if this can cause issues, because the main CFG= option will also look inside tests/test_cfg

XavierAubert (commented):
Just a small comment for cfgyaml2make modifications

bin/cfgyaml2make
dd-vaibhavjain: The cfg/test_cfg file names are expected to be unique, and incase multiple cfg files are found with same names, the script already flags a fatal error. So any one creating a new config and running a check for those new configs should be able to hit this issue and create a unique name for new cfg file.
And ensuring the regress list is created with correct cfg/test_cfg is always the responsibilty of the dv which ensures we are creating correct scenario. 
@XavierAubert  do you still see need for 2 separate yaml2make scripts for these 2 ?

XavierAubert: yes, true I did not see it was also checking for duplicates.

XavierAubert (approved):

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for this @dd-vaibhavjain.  I have a few minor questions about a couple of things:
1. I am still not clear on the need for `<cv-core>/test/test-cfg`.  Yes, the names must be unique, but is that a problem?
2. Running `riscof` enables ImperasDV by default.   Is there a way to disable this?
Neither of these gate merging this PR.  We can discuss in our weekly meeting.

dd-vaibhavjain:
Big thanks to you @MikeOpenHWGroup for taking time to go through this PR!
Sure I will be able to present these things better in our weekly and will have better opportunity to discuss, and we can see if we need to redo certain things.
Your point on Imperas enable by default is noted. Currently there is no way to disable from command line so will wait for the meeting and discuss what we can do here.

PR #2030, opened 2023-07-06
use hash=709bec4 instead of main in clonetb

silabs-krdosvik:
Clonetb uses a hash instead of main, as a hash will be more stable than main.

Formal comiles, ci_check has these results:
image


silabs-robin (approved):
Approving based on:
Changes look like expected, based on our previous experimentation and discussion.
I also tried to run it (and also run it wrong to see what happens), and the results are as expected.

PR #2037, opened 2023-07-10
Added Zc* instructions to disassembler

silabs-inhaahje:


lib/support/isa_support.sv
```diff
@@ -773,38 +967,294 @@
...
+    CONST_GEN_CI_H,
+    // Others
+    UNKNOWN_HINT
+  } hint_name_e;
+
+  // Get the correspopnding name of the hint instruction
+  function hint_name_e get_hint_name(instr_name_e name);
+    hint_name_e hint_name;
+
+    if (name inside { SLTI, SLTIU, ANDI, ORI, XORI, SLLI, SRLI, SRAI })
```
silabs-hfegran, line 1160: Perhaps this structure would be more readable with a case-statement? 
``` 
casex(name)
  SLTI, SLTIU, ANDI, ORI, (...): hint_name = REG_IMM_I_H;
  ADD, SUB, (...): hint_name = REG_REG_R_H;
  (etc...)
```

lib/support/isa_support.sv
```diff
@@ -773,38 +967,294 @@
...
+    else if (name == C_LI)
+      hint_name = C_LI_H;
+    else
+      hint_name = UNKNOWN_HINT;
+
+    return hint_name;
+  endfunction
+
+  // Find out if the instruction is a HINT.
+  function logic check_if_hint(instr_name_e name, instr_format_e format, instr_t instr);
```
silabs-hfegran, line 1187: Two comments on this section:
I would prefer if/else-blocks to have begin/end-statements. That said, I think this section could be simplified a bit; perhaps by initializing hint to 0 and using ternary operators or just assign the output of the logic check:

e.g. 
```
casex (get_hint_name(name))
  ADDI_H:      hint = (instr.uncompressed.format.i.rd == X0 && (instr.uncompressed.format.i.rs1 != X0 || instr.uncompressed.format.i.imm != 12'b0));
  FENCE_H:     hint = ((instr.uncompressed.format.i.imm.funct7[27:25] == 3'b0 && instr.uncompressed.format.i.imm.shamt[24] == 1'b0) || instr.uncompressed.format.i.imm.shamt[23:20] == 4'b0);
  REG_IMM_I_H: hint = (instr.uncompressed.format.i.rd == X0);
  (etc....)

File: lib/support/isa_support.sv [hunk 3]
```diff
@@ -773,38 +967,294 @@
...
+    stack_adj_operand_t stack_adj; // structure to handle stack_adj fields for Zcmp-instructions
   } asm_t;
 
 
   // ---------------------------------------------------------------------------
   // Non-trivial immediate decoder
   // ---------------------------------------------------------------------------
-  function logic [20:0] get_j_imm(instr_t instr);
-    get_j_imm = {
+  function logic [20:1] sort_j_imm(instr_t instr);
```
silabs-krdosvik, line 979: Why did you change the name? Both works ofc,, but I must admit I liked the get functions better. And I dont know what is best of [20:1] or [19:0], but yeah, same same anyways.

silabs-hfegran: @silabs-krdosvik: I guess the purpose here is to have one set of functions for sorted immediates (but not interpreted), and one for the human readable format

lib/support/isa_support.sv
```diff
@@ -773,38 +967,294 @@
...
+    // Others
+    UNKNOWN_HINT
+  } hint_name_e;
+
+  // Get the correspopnding name of the hint instruction
+  function hint_name_e get_hint_name(instr_name_e name);
+    hint_name_e hint_name;
+
+    if (name inside { SLTI, SLTIU, ANDI, ORI, XORI, SLLI, SRLI, SRAI })
+      hint_name = REG_IMM_I_H;
```
silabs-krdosvik, line 1161: Niice function. Not sure if I understand how this names are constructed (I am not up to date when it comes to HINT instruction), so could you briefly explain :-) (Dont need to change the code for now)

lib/support/isa_support.sv
```diff
@@ -892,55 +1389,294 @@
...
-        asm.rs2.valid   = 1;
-        asm.imm.valid   = 1;
+        asm.rs1.gpr             = instr.uncompressed.format.b.rs1.gpr;
+        asm.rs2.gpr             = instr.uncompressed.format.b.rs2.gpr;
+        asm.imm.imm_raw         = {instr.uncompressed.format.b.imm_h, instr.uncompressed.format.b.imm_l};
+        asm.imm.imm_raw_sorted  = sort_b_imm(instr);
+        asm.imm.imm_type        = IMM;
+        asm.imm.width           = 12;
+        asm.imm.sign_ext        = 1;
+        asm.imm.imm_value       = get_imm_value_b(sort_b_imm(instr)); //TODO: FIX function names
```
silabs-krdosvik, line 1399: Is this still something you want to fix?

lib/support/isa_support.sv
```diff
@@ -892,55 +1389,294 @@
...
+          asm.rd.gpr              = instr.compressed.format.ci.rd_rs1.gpr;
+          asm.imm.imm_raw         = { instr.compressed.format.ci.imm_12, instr.compressed.format.ci.imm_6_2 };
+          asm.imm.imm_raw_sorted  = { instr.compressed.format.ci.imm_12, instr.compressed.format.ci.imm_6_2 };
+          asm.imm.imm_type        = IMM;
+          asm.imm.width           = 6;
+          asm.imm.sign_ext        = 1;
+          asm.imm.imm_value       = get_imm_value_ci({ instr.compressed.format.ci.imm_12, instr.compressed.format.ci.imm_6_2 });
+          asm.rd.valid            = 1;
+          asm.imm.valid           = 1;
+          asm.is_hint             = 0;
```
silabs-krdosvik, line 1453: I see you have specified that is_hint is 0 here. It seems like this is not specified everywhere, so I assume is_hint is default 0. My point is, if is_hint is set to 0 you dont need to include it. But if this is a interesting remark for these particular instructions, it can be nice to explicit state that is_hint=0.

lib/support/isa_support.sv
```diff
@@ -892,55 +1389,294 @@
...
+        asm.imm.imm_value       = instr.compressed.format.cmjt.index;
+        asm.imm.valid           = 1;
+      end
+      CMPP_TYPE: begin
+        asm.imm.imm_raw         = instr.compressed.format.cmpp.spimm;
+        asm.imm.imm_raw_sorted  = instr.compressed.format.cmpp.spimm;
+        asm.imm.imm_type        = SPIMM;
+        asm.imm.width           = 1;
+        asm.imm.sign_ext        = 0;
+        asm.rlist.rlist         = instr.compressed.format.cmpp.urlist;
```
silabs-krdosvik, line 1670: Quick question: why do you refer to rlist and urlist, and not only use rlist or urlist?

lib/support/isa_support.sv
silabs-hfegran: @silabs-robin, I believe that 3 is covered by the imm_sorted, and 4 is covered by imm_value; is there anything else we would want to include apart from potential naming discussions?

silabs-robin: > I believe that 3 is covered by the imm_sorted, and 4 is covered by imm_value;

Ok.

> is there anything else we would want to include apart from potential naming discussions?

Not from my part.

lib/support/isa_support.sv
silabs-inhaahje: Thank you! I had forgotten to remove it

lib/support/isa_support.sv
```diff
@@ -1286,112 +2030,337 @@
...
-       && (instr.compressed.format.cr.funct4      == 4'b1001)) :
+      (   (instr.compressed.opcode                        == C2)
+       && (instr.compressed.format.cr.rd_rs1.gpr          == X0)
+       && (instr.compressed.format.cr.rs2.gpr             == X0)
+       && (instr.compressed.format.cr.funct4              == 4'b1001)) :
         asm = build_asm(C_EBREAK, CR_TYPE, instr);
 
+      (   (instr.compressed.opcode                        == C2)
+       && (instr.compressed.format.cr.rs2.gpr             != X0)
+       && (instr.compressed.format.cr.funct4              == 4'b1000)) :
```
silabs-krdosvik, line 2145: I see you have made enums for funct7, which is very good. Is it possible to make enums for funct4 and funct3 as well, or would it be complicated?

lib/support/isa_support.sv
silabs-krdosvik: Aaa, I understand. Sounds good then. I think 3 and 4 is good enough.

silabs-krdosvik: It would be nice if you could comment the [20:1] Vs. [19:0].

silabs-inhaahje: > It would be nice if you could comment the [20:1] Vs. [19:0].

The reason why I use [20:1] instead of [20:0]:
In the original function get_j_imm the immediate was left shifted by one. The sort_j_imm only sorts the bits in correct order, but does not shift them. 

The reason why I used [20:1] instead of [19:0] was because [20:1] correlates with where the bits are located in the immediate. I understand if this is confusing here since the bits are not shifted into that position. Should I change these all to start at bit 0?

silabs-robin: > Should I change these all to start at bit 0?

I would prefer if everything matches the spec as far as possible.

If this signal is about a bit field, and the bit field does not have bit 0, then I would prefer that the signal also doesn't have bit 0.

silabs-robin: Otherwise we might have to make explicit mappings later, to say that bit 0 really means bit 1 etc, and that is exactly were many of the past bugs in ISACOV has been introduced.

The more direct the connection, and the less steps involved, the less chance for human error, is what I think.

lib/support/isa_support.sv
```diff
@@ -357,6 +385,65 @@
...
+    rlist_abi_name_e rlist_abi;
+  } rlist_t;
+
+  // ---------------------------------------------------------------------------
+  // Stack_adj for zcmp instructions
+  // ---------------------------------------------------------------------------
+  function int get_stack_adj( rlist_t rlist, logic[5:4] spimm);
+    int stack_adj_base;
+    int stack_adj;
+
```
silabs-hfegran, line 433: I would suggest using `case (rlist) inside` for the nested ifs below

lib/support/isa_support.sv
silabs-krdosvik: Thanks for the explanation. Sounds good!

silabs-krdosvik: RESOLVED

lib/support/isa_support.sv
```diff
@@ -1286,112 +2030,337 @@
...
+       && (instr.uncompressed.format.r.funct7     == FUNCT7_M)) :
         asm = build_asm(REMU, R_TYPE, instr);
 
       // Compressed
+      (   (instr.compressed.opcode                        == C0)
+       && (instr.compressed.format.ci.rd_rs1.gpr          == X0)
+       && (instr.compressed.format.ci.imm_12              == 1'b0)
+       && (instr.compressed.format.ci.imm_6_2             == 5'b0)
+       && (instr.compressed.format.ci.funct3              == 3'b0)) :
+        asm = build_asm(ILLEGAL_INSTR, CI_TYPE, instr);
```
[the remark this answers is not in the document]
silabs-krdosvik: This is not really that important comment, and dont need change. Just wanted to mention what I was thinking :-)

RESOLVED

lib/support/isa_support.sv
silabs-krdosvik: Okei, I see, looks good then. Thanks for the explanation.

RESOLVED

lib/support/isa_support.sv
silabs-krdosvik: I didnt see any function naming that was questionable that I didnt comment, so in general I think it is good. Are there any names in particular you want to discuss?

RESOLVED

lib/support/isa_support.sv
silabs-krdosvik: RESOLVED

lib/support/isa_support.sv
silabs-inhaahje: I agree.  I forgot about making enums for them but will fix

silabs-inhaahje: Made enums for funct3 and funct5. I was not able to find a good way of naming enums for  funct6 and funct4, but can still make if this is something we want:)

silabs-krdosvik: Looks good! In that case I think it is fine letting the funct6 and funct4 be bit sequences.

RESOLVED

lib/support/isa_support.sv
silabs-krdosvik: Looks good!

RESOLVED

silabs-krdosvik (approved):
In general very good! 

I approve as my comments have been addressed.

silabs-hfegran (approved):
Looks very good @silabs-inhaahje, thanks for these updates!

PR #2038, opened 2023-07-10
Merge x into s

silabs-krdosvik:
Merge changes from X into S.


silabs-robin (approved):
Approving based on:
Matches latest development of updating the script, changing the readme, and deleting all other X files.

PR #2046, opened 2023-07-12
Added clic tests to cover coverage holes, fixed clic test issues

silabs-hfegran:
- Added mnxti, mscratchcsw(l), mintthresh, minhv-tests to clic tests
- Fixed test issues
- Made sticky/non-sticky clic irq requests configurable


silabs-krdosvik:
Interesting!

PR #2051, opened 2023-07-13
Added support for 40X in debug_test2, fixed typo

silabs-hfegran:
* Added support for 40x in debug_test2 (deliberately on 40S to ease merging efforts
* Fixed typo in hello_world comment


silabs-robin (approved):
Approving based on:
This test is known to fail on the X, and this looks like a nice fix to it.

PR #2057, opened 2023-07-17
Update clonetb hash

silabs-krdosvik:
image

### DIFF
--- a/bin/clonetb
+++ b/bin/clonetb
@@ -80,7 +80,7 @@ clone() {
 clone_cv32e40x() {
   CV_CORE=cv32e40x
   VERIF_ENV_REPO=https://github.com/openhwgroup/cv32e40x-dv.git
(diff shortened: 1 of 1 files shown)


silabs-robin (commented):
Does ci_check reasonably pass and is formal still able to run?

bin/clonetb
silabs-robin, line 1: Does ci_check reasonably pass and is formal still able to run?

I guess it would, because the latest changes to x-dv were minor.

silabs-krdosvik: Formal compiles. Sim dont compile on cv32e40x/mater. Need to wait for this PR to be merged: https://github.com/openhwgroup/cv32e40x/pull/890

silabs-robin: Ok. Let's wait then. And when it is in, it seems we must update the core hash (ExternalRepos) at the same time.

PR #2059, opened 2023-07-18
Made it possible to use the Risc-v disassembler with isacov

silabs-inhaahje:
To verify the implementation, the test hello-world was ran to see if the coverage when using the SPIKE decoder is close to the coverage when using isa_support. 

The results showed that the two decoders had approximately the same coverage, indicating that the implementation works. 

Screenshot of coverage when using Spike:
image

Screenshot of coverage when using isa_support:
image

### DIFF
--- a/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
+++ b/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
@@ -243,6 +243,8 @@ class uvme_cv32e40s_cfg_c extends uvma_core_cntrl_cfg_c;
       isacov_cfg.seq_instr_x2_enabled       == 1;
       isacov_cfg.reg_crosses_enabled        == 0;
       isacov_cfg.reg_hazards_enabled        == 1;
+      isacov_cfg.decoder                    == RISCV_DISASSEMBLER;
+
 
       rvfi_cfg.nret                    == RVFI_NRET;
       rvfi_cfg.nmi_load_fault_enabled  == 1;
@@ -417,6 +419,7 @@ function uvme_cv32e40s_cfg_c::new(string name="uvme_cv32e40s_cfg");
      clic_irq_clear_on_ack_plusarg_valid = 1;
      clic_irq_clear_on_ack.rand_mode(0);
    end
+
 
    isacov_cfg           = uvma_isacov_cfg_c::type_id::create("isacov_cfg");
    clknrst_cfg          = uvma_clknrst_cfg_c::type_id::create("clknrst_cfg");
--- a/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
+++ b/cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
@@ -243,7 +243,7 @@ class uvme_cv32e40s_cfg_c extends uvma_core_cntrl_cfg_c;
       isacov_cfg.seq_instr_x2_enabled       == 1;
       isacov_cfg.reg_crosses_enabled        == 0;
       isacov_cfg.reg_hazards_enabled        == 1;
-      isacov_cfg.decoder                    == RISCV_DISASSEMBLER;
+      isacov_cfg.decoder                    == ISA_SUPPORT;
 
 
       rvfi_cfg.nret                    == RVFI_NRET;
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
@@ -54,6 +54,9 @@ class uvma_isacov_cfg_c extends uvm_object;
     `uvm_field_enum(decoder_e, decoder, UVM_DEFAULT);
   `uvm_object_utils_end;
 
+   constraint defaults_cons {
+      soft decoder  == SPIKE;
+   }
   extern function new(string name = "uvma_isacov_cfg");
 
 endclass : uvma_isacov_cfg_c
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
@@ -33,7 +33,7 @@ class uvma_isacov_cfg_c extends uvm_object;
   rand bit                     reg_crosses_enabled;
   rand bit                     reg_hazards_enabled;
 
-  rand decoder_e            decoder;
+  rand decoder_e               decoder;
 
   // Core configuration object to filter extensions, csrs, modes, supported
   uvma_core_cntrl_cfg_c        core_cfg;
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
@@ -33,6 +33,8 @@ class uvma_isacov_cfg_c extends uvm_object;
   rand bit                     reg_crosses_enabled;
   rand bit                     reg_hazards_enabled;
 
+  rand decoder_e            decoder;
+
   // Core configuration object to filter extensions, csrs, modes, supported
   uvma_core_cntrl_cfg_c        core_cfg;
 
@@ -48,6 +50,8 @@ class uvma_isacov_cfg_c extends uvm_object;
     `uvm_field_int(seq_instr_x2_enabled, UVM_DEFAULT);
     `uvm_field_int(reg_crosses_enabled, UVM_DEFAULT);
     `uvm_field_int(reg_hazards_enabled, UVM_DEFAULT);
+
+    `uvm_field_enum(decoder_e, decoder, UVM_DEFAULT);
   `uvm_object_utils_end;
 
   extern function new(string name = "uvma_isacov_cfg");
--- a/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
+++ b/lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
@@ -154,21 +154,27 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
   if (cfg.decoder == SPIKE) begin
     // Attempt to decode instruction with Spike DASM
     instr_name = dasm_name(rvfi_instr.insn);
+  end else if (cfg.decoder == ISA_SUPPORT) begin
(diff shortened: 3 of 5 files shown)


lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -138,62 +145,117 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
...
+      mon_trn.instr.c_rdrs1 = instr_asm.rd.gpr;
+      mon_trn.instr.c_rdp   = instr_asm.rd.gpr;
+      mon_trn.instr.c_rs1s  = instr_asm.rs1.gpr;
+      mon_trn.instr.c_rs2s  = instr_asm.rs2.gpr;
+    end
+    else begin
+      mon_trn.instr.rs1     = instr_asm.rs1.gpr;
+      mon_trn.instr.rs2     = instr_asm.rs2.gpr;
+      mon_trn.instr.rd      = instr_asm.rd.gpr;
+      mon_trn.instr.immi    = instr_asm.imm.imm_raw_sorted;
```
silabs-krdosvik, line 237: Is the mon_trn.instr.imm* set to sorted immediates or actual value of the immediates? I see it says this in the above : 
 code: mon_trn.instr.immb = dasm_sb_imm(instr) >> 1; // Because dasm gives [12:0], not [12: 1]

silabs-inhaahje: instr_asm.imm.imm_raw_sorted is sorted but not shifted. It is my understanding that instr_asm.imm.imm_raw_sorted will give the wanted value without the need to shift like dasm.

cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
```diff
@@ -243,6 +243,8 @@ class uvme_cv32e40s_cfg_c extends uvma_core_cntrl_cfg_c;
       isacov_cfg.seq_instr_x2_enabled       == 1;
       isacov_cfg.reg_crosses_enabled        == 0;
       isacov_cfg.reg_hazards_enabled        == 1;
+      isacov_cfg.decoder                    == RISCV_DISASSEMBLER;
```
silabs-krdosvik, line 246: Did we agree on using SPIKE or RISCV_DISASSAMBLER as default. Considering that we add this change to the sx/dev branches I think it is nice to have RISC_DISASSAMBLER as default, and other people will not be affected unless we do a merge (if I am not mistaken), aaand, as long as the RISC_DISASSAMBLER produce the same results as SPIKE it should be fine.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-inhaahje: I believe it is set to the sorted immediates, but not the value.

lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
```diff
@@ -33,6 +33,8 @@ class uvma_isacov_cfg_c extends uvm_object;
   rand bit                     reg_crosses_enabled;
   rand bit                     reg_hazards_enabled;
 
+  rand decoder_e               decoder;
```
silabs-robin, line 36: Is there anything that specifies a default for this?

I wonder if maybe it should fall back to "SPIKE" for anyone that doesn't explicitly configure this variable, for seamless backwards compatibility.
(Or maybe there should be some way to require that it gets explicitly configured?)

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-robin, line 1: Could you include some descriptions and/or screenshots of how this integration was tested? To demonstrate that it works.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -138,62 +145,117 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
...
+    end else begin
+      // Undecodable instruction
+      mon_trn.instr.name = UNKNOWN;
+    end
+
+    mon_trn.instr.itype = get_instr_type(mon_trn.instr.name);
+    mon_trn.instr.ext   = get_instr_ext(mon_trn.instr.name);
+    mon_trn.instr.group = get_instr_group(mon_trn.instr.name, rvfi_instr.mem_addr);
+
+    if (mon_trn.instr.ext == C_EXT) begin
```
silabs-robin, line 224: Do we need this if-statement?

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
```diff
@@ -138,62 +145,117 @@ function void uvma_isacov_mon_c::write_rvfi_instr(uvma_rvfi_instr_seq_item_c#(IL
...
+    if (instr_name_lookup.exists(instr_name)) begin
+      mon_trn.instr.name = instr_name_lookup[instr_name];
+    end else begin
+      // Undecodable instruction
+      mon_trn.instr.name = UNKNOWN;
+    end
+
+    mon_trn.instr.itype = get_instr_type(mon_trn.instr.name);
+    mon_trn.instr.ext   = get_instr_ext(mon_trn.instr.name);
+    mon_trn.instr.group = get_instr_group(mon_trn.instr.name, rvfi_instr.mem_addr);
```
silabs-robin, line 222: Is this code exactly the same regardless of using spike or the new decoder?

```
    if (instr_name_lookup.exists(instr_name)) begin
      mon_trn.instr.name = instr_name_lookup[instr_name];
    end else begin
      // Undecodable instruction
      mon_trn.instr.name = UNKNOWN;
    end

    mon_trn.instr.itype = get_instr_type(mon_trn.instr.name);
    mon_trn.instr.ext   = get_instr_ext(mon_trn.instr.name);
    mon_trn.instr.group = get_instr_group(mon_trn.instr.name, rvfi_instr.mem_addr);
```

I wonder if maybe we should only if-else the parts that are actually different, so it gets easier to understand what the point of the monitor is.
Already from before this PR, I have thought this function is so long that it is difficult to understand, and I've wanted to refactor it a bit. If we keep code that does the same thing together then it should be easier to refactor out.
(The refactoring itself doesn't have to be done now.)

I am not sure.
Leave the code as it is unless more people agree to change it.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-krdosvik: Didnt look that much into it, so wanted to mentioned it. But I agree, so looks good how it is :-) 

RESOLVED

cv32e40s/env/uvme/uvme_cv32e40s_cfg.sv
silabs-krdosvik: RESOLVED

lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
```diff
@@ -18,6 +18,11 @@
 `ifndef __UVMA_ISACOV_TDEFS_SV__
 `define __UVMA_ISACOV_TDEFS_SV__
 
+typedef enum {
+  SPIKE,
+  RISCV_DISASSEMBLER
```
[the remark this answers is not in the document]
silabs-inhaahje: It is no problem to change it! Should I change it to isa_support or core-v-decoder?

silabs-krdosvik (approved):
Looks good.
Dont see anything that might be out of place, so I approve.

lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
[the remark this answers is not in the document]
silabs-robin: Well, I think the problem is that the _decoder_ is called "isa_support" and not something with "decoder".
If we don't change it now, it will get merged to master later and we're stuck with it forever.
@silabs-hfegran, thoughts?

The easiest thing to do is probably just to keep the "isa_support" name and change the enum name to reflect it (ISA_SUPPORT).
I think I vote for that, because it is minimal effort.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-inhaahje: > Could you include some descriptions and/or screenshots of how this integration was tested? To demonstrate that it works.

Done! Let me know if you think something is missing.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-inhaahje: No, we do not:) Thank you!

lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
silabs-robin: See also: https://github.com/openhwgroup/core-v-verif/pull/2059#discussion_r1266513139

lib/uvm_agents/uvma_isacov/uvma_isacov_tdefs.sv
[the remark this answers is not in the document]
silabs-robin: Thanks.

lib/uvm_agents/uvma_isacov/uvma_isacov_mon.sv
silabs-robin: That looks good. At least I got convinced that it is a good substitute for the spike decoder.

lib/uvm_agents/uvma_isacov/uvma_isacov_cfg.sv
silabs-inhaahje: Added a soft constraint in uvma_isacov_cfg setting Spike as decoder. Hope this could be what you had in mind.

silabs-robin: Looks very good. Thanks.

silabs-robin (approved):
Approving based on:
Useful addition.

PR #2061, opened 2023-07-18
Add a short script to challenge generated tests using CVA6-DV

AyoubJalali:
I add a small script named smoke-gen_tests.sh, to challenge the CVA6-DV (extend RISCV-DV), because some changes on the repo effect test generation, and we don't have something like smoke-test to assure that any change is not gonna break the generation test.


JeanRochCoulon:
@AyoubJalali  I cannot merge because the CI fails

AyoubJalali:
> @AyoubJalali I cannot merge because the CI fails

We can't debug ?

PR #2069, opened 2023-07-21
Corrected matmul_32b_float test for v1 and removed 3 cycles FPU latencies configs

pascalgouedo:


MikeOpenHWGroup (approved):
This PR will only affect a single test-program (`matmul_32b_float`) which is not (yet?) part of CI, so I am approving and merging based on visual inspection only.

PR #2070, opened 2023-07-21
Expanded disassembler with A extension. Not tested

silabs-inhaahje:


lib/support/isa_support.sv
```diff
@@ -169,6 +169,18 @@
     SRA,
     OR,
     AND,
+    //A
```
silabs-krdosvik, line 172: I cant see any immediate flaws, so looks good.

RESOLVED

lib/support/isa_support.sv
```diff
@@ -614,12 +626,28 @@
     FUNCT3_REMU   = 3'b111
   } m_minor_opcode_e;
 
+  // Minor opcode for atomic instructions, "A".
+  typedef enum logic [2:0] {
+    FUNCT3_A    = 3'b010
```
silabs-hfegran, line 631: I suggest we name this FUNCT3_A_W, as this field refers to the .w (word) in the A-instructions.(i guess it would be easy to add 011 as FUNCT3_A_D as well (double) even though it is not used by us).

lib/support/isa_support.sv
```diff
@@ -614,12 +626,28 @@
...
+    FUNCT5_SC_W             = 5'b00011,
+    FUNCT5_AMOSWAP_W        = 5'b00001,
+    FUNCT5_AMOADD_W         = 5'b00000,
+    FUNCT5_AMOXOR_W         = 5'b00100,
+    FUNCT5_AMOAND_W         = 5'b01100,
+    FUNCT5_AMOOR_W          = 5'b01000,
+    FUNCT5_AMOMIN_W         = 5'b10000,
+    FUNCT5_AMOMAX_W         = 5'b10100,
+    FUNCT5_CZEXTB_AMOMINUW  = 5'b11000,
+    FUNCT5_AMOMAXU_W        = 5'b11100
```
silabs-hfegran, line 650: I think we need to split FUNCT5 here, as there are overlapping use-cases between C-extension and A-extension (e.g. funct5_a_something)

lib/support/isa_support.sv
```diff
@@ -959,6 +987,12 @@
     bit valid;
   } stack_adj_operand_t;
 
+  // Atomic operand to specify additional memory ordering constraints for atomic instructions
+  typedef struct packed {
+    bit aq;
+    bit rl;
+    bit valid;
+  } atomic_operand_t;
```
silabs-hfegran, line 995: I think this is fine for now, as the spec is a bit confusing on the syntax of these wrt. assembly.
we could consider making separate instruction names in the future.

silabs-hfegran (changes requested):
Thank you for these updates @silabs-inhaahje, Looks good, I only have a few minor things I think we should consider changing

silabs-hfegran (approved):

silabs-hfegran:
Thanks, looks good now

PR #2074, opened 2023-07-24
Added illegal access tests for mscratchcsw* to clic test

silabs-hfegran:
* Added tests to check for illegal csr-instruction usage with mscratchcsw/mscratchcswl-registers


silabs-robin (approved):
Looks like a good addition.

PR #2075, opened 2023-07-24
Switch to GCC 13

corentinmaraisT:
A parameter called `--isa_extension` has been added in `cva6.py`. Depending on the set of instruction the user want to use while compiling/executing a code (assembly/c/elf) the user can choose the extension to enable via this argument. i,m,a,f,d,c,g,v are not supposed to be specified here only z\_, s\_, or x\_ extensions can be added via this argument. This argument is set to zicsr by default.

Ex:

> python3 cva6.py --testlist=../tests/testlist_riscv-compliance-cv32a60x.yaml --test rv32i-I-ADD-01 --iss_yaml cva6.yaml --target cv32a60x --iss=  **--isa_extension=zicsr,zbb**

In this example, the extension zicsr and zbb has been added to the isa initially associated to the target cv32a60x.


MikeOpenHWGroup (changes requested):
I do not see anything in this pull-request that implements a **_Switch to GCC 13_**.  There are several changes to run-scripts.  Does the user _need_ to use GCC 13 for these to work?

If we require users to use GCC 13, then this PR should also include:
1. An update to the README to explain to the User that GCC 13 is required.
2. A run-time check that fatals-out if the User's GCC version is not supported.

fatimasaleem:
This PR will resolve the https://github.com/openhwgroup/core-v-verif/issues/1920

JeanRochCoulon:
Mike, you are right, support need to be provided to user. But as the current PR blocks the execution of other pipelines ( pipeline options are gcc10 but the gcc used by the CI is GCC13 => CRASH!! ), I will merge this PR and @corentinmaraisT to add README information and checkers.

PR #2086, opened 2023-07-27
Fixed issue with always_comb block using signals written within

silabs-hfegran:
Signals may be have dependencies on each other, thus to
avoid having to micromanage order of execution, they are split
into separate always_comb blocks. Assign will not work due to
sensitivity list issues with functions, and a single always_comb
block will cause issues due to clause b in section 9.2.2.2.1 in
ieee 1800-2017


silabs-robin (approved):
Approving based on:
IRL discussions.

PR #2089, opened 2023-07-27
Fixed issue where clic tests would fail due to (unintended) too high mintstatus.mil level

silabs-hfegran:
Added a function to reset dut mintstatus.mil state in tests that may end up failing due to too high mintstatus.mil level.


silabs-robin (approved):
Approving based on:
It fixes a problem, has self-checking and was tested repeatedly in simulation.

PR #2090, opened 2023-07-28
Adapt for new parameterization

zarubaf:
Companion commit to https://github.com/openhwgroup/cva6/pull/1321.

It deprecates the top-level `ArianeCfg` config struct in favor of the CVA6 one. Furthermore, the testbench now defaults to the Default config struct in the `ariane_pkg`. While that isn't super useful for reproducibility, this should ideally limit the changes to `coer-v-verif` at the moment. Once the config struct is stable a "known-good" configuration can be fixed here.


zarubaf:
> @zarubaf Could you rebase your PR on latest cva6/dev commit? It is needed as the path of config_pkg_generator.py was missing in cva6.py.

Ooops, I guess that was me when cleaning up the root dir. Re-based on `cva6/dev`!

zarubaf:
No longer applicable.

PR #2091, opened 2023-07-28
ci: rework and document

cathales:
1. Rely on stages more than needs relationships: "needs" relationships are tricky as they bypass stage ordering. For instance, "initjob" was useless, which was a bug.
2. Refactor job descriptions: each job needs less code in cva6.yml. It also fixes an issue in CVV pipelines, where the CVA6 hash in the dashboard was not the one used in some jobs of the pipeline.
3. Document the CI structure


cathales:
The only failing test is `tests/isa/rv64ua/lrsc.S`. Can you merge this PR, please?

PR #2093, opened 2023-07-31
add atomic to rvfi_instr_if, expand/divide the support pkg

silabs-krdosvik:
Added is_amo and is_atomic instr to rvfi_instr_if. Also restructured the isa_support. Will continue working on isa_support files when the merges from cv32e40s/dev to cv32e40x/dev, and cv32e40s to cv32e40x-dv is done.

Must wait to merge this PR untill https://github.com/openhwgroup/cv32e40x-dv/pull/13 is merged, and we can update the clonetb hash.

Same status as https://github.com/openhwgroup/cv32e40x-dv/pull/13 (ci check passes partially and formal compiles)

### DIFF
--- a/lib/support/isa_constants.sv
+++ b/lib/support/isa_constants.sv
@@ -11,21 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//
-// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 
-`ifndef __SUPPORT_PKG_SV__
-`define __SUPPORT_PKG_SV__
-
-
-package support_pkg;
-  `include "isa_constants.sv"
-  `include "isa_constants_csr.sv"
-  `include "isa_tdefs.sv"
-  `include "isa_tdefs_csr.sv"
-  `include "isa_support.sv" //TODO: krdosvik, change name to isa_disassembler when no outstanding PR
-  `include "isa_functions.sv"
-endpackage
-
-`endif // __SUPPORT_PKG_SV__
-
+// -------------------------------------------------------------------
+// This file holds constants related to the ISA
+// -------------------------------------------------------------------
--- a/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
+++ b/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
@@ -87,6 +87,7 @@ interface uvma_rvfi_instr_if_t
   localparam INSTR_MASK_DIV_REM     = 32'h FE00_607F;
   localparam INSTR_MASK_FULL          = 32'h FFFF_FFFF;
   localparam INSTR_MASK_R_TYPE        = 32'h FE00_707F;
+  localparam INSTR_MASK_AMO_TYPE      = 32'h F800_707F;
   localparam INSTR_MASK_I_S_B_TYPE    = 32'h 0000_707F;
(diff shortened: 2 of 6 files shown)


lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -167,6 +168,18 @@ interface uvma_rvfi_instr_if_t
   localparam INSTR_OPCODE_CMPOPRET   = 32'b 00000000_00000000_101_11110_0000_00_10;
   localparam INSTR_OPCODE_CMPOPRETZ  = 32'b 00000000_00000000_101_11100_0000_00_10;
 
+  localparam INSTR_OPCODE_LRW       = 32'b 00010_0_0_00000_00000_010_00000_0101111;
```
silabs-hfegran, line 171: do we still need these masks with the new disassembler?

lib/support/support_pkg.sv
```diff
@@ -19,7 +19,12 @@
 
 
 package support_pkg;
-  `include "isa_support.sv"
+  `include "isa_constants.sv"
+  `include "isa_constants_csr.sv"
+  `include "isa_tdefs.sv"
+  `include "isa_tdefs_csr.sv"
+  `include "isa_support.sv" //TODO: krdosvik, change name to isa_disassembler when no outstanding PR
+  `include "isa_functions.sv"
```
silabs-robin, line 27: The includes in `isa_functions.sv` are also already included above this line.

lib/support/isa_constants.sv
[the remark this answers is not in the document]
silabs-krdosvik: Yess, I wasnt actually going for a PR now, but I guess it doesnt matter that if these files are out or not. Each file have a descriptive text: F.eks.: 
// This file holds constants related to the ISA

This means that for future expansion of the ISA support we add the constants in this file, type defs in another, and so on. This is to reduce the amount of code that will gather in what was the the isa_support file, and make it more overviewable.

I must wait for the merges mentioned in the introduction untill I can continue developing these files.

lib/support/support_pkg.sv
silabs-krdosvik: You mean isa_function is indirectly included by including the isa_support.sv? If that is what you mean: I didnt think about that, is that a problem? If not I think it is more readable to explicitly define what is meant to be included in the package, even though files might already include them.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-hfegran: I think I can do that conversion quite quickly, but it would be nice to have things a little more stable before we start changing these things.

lib/support/support_pkg.sv
silabs-hfegran: Regardless of the above, all these files needs compile-guards, otherwise there will be issues down the line.

silabs-robin: Beware of dangers like this:
If 1) an include file has guards, and 2) compilation of all source files is done as _one_ "compilation unit", and 3) two other files include that include file, then C) only one of those other files will get any actual content from the include file.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-krdosvik: sounds good!

silabs-robin:
I think this PR is a stale duplicate of another PR that got merged?
Therefore I'm closing it.

If I'm wrong then please just re-open it.

PR #2095, opened 2023-07-31
Add (too many) warning suppressions for VSIM

MikeOpenHWGroup:
This PR does two things:
1. Updates the hash for the RTL.
2. Updates `vsim.mk` to allow the CV32E20 environment to compile/run under Questasim (vsim).

The updates to `vsim.mk` add a set of warning suppressions - these need to be investigated further, but for now this gets simulations running with Questa.


LeeHoff (approved):
These changes look good to me.  I don't have a vsim license to make sure it all works, but I saw the effect of the changes when we worked together.

PR #2100, opened 2023-08-03
Merge cv32e40s/dev --> cv32e40x/dev

silabs-krdosvik:
Merge from cv32e40s/dev to cv32e40x/dev.
This is related to the PR: https://github.com/openhwgroup/cv32e40x-dv/pull/16 in cv32e40x-dv
Formal compiles and ci_check:
image

### DIFF
--- a/bin/cv_regress
+++ b/bin/cv_regress
@@ -151,6 +151,8 @@ def read_file(args, file):
             test.sub_make(args.make)
         if args.iss != None:
             test.iss = args.iss
+        else:
+            args.iss = 'YES'
         if args.cfg:
             test.cfg = args.cfg
         if not test.builds and test.build:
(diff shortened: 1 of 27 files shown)


bin/cv_regress
silabs-robin, line 1: image

Two questions:
1. Did the first merge 100% auto-merge without conflicts or any need for manual intervention?
2. How could the second merge have anything to add?

silabs-krdosvik: > image
> 
> Two questions:
> 
> 1. Did the first merge 100% auto-merge without conflicts or any need for manual intervention?
> 2. How could the second merge have anything to add?

1. The first merge had some conflicts yess.
2.  I forgot to run a core-v-verif fetch, so in the first merge I merged the changes I had locally, which was not the updated version of cv32e40s/dev. As I didn't discover this, I had to do some fixing to make the code run formal and simulation, hence the commit in between the merges. Then I discovered that cv32e40s/dev had more commits than what I had locally and needed to do a new merge. The last merge was mostly auto-merge, but conflicts where I had done fixes in the "add missing decoder"-commit (if I remember correctly).

silabs-krdosvik: > According to this (link and screenshot), it should be possible to automatically merge s/dev into x/dev. And since it is much easier to do this merge compared to the x-dv merge, should we maybe re-do the merge so we have just one commit instead of those three above?
> 
> [cv32e40x/dev...cv32e40s/dev](https://github.com/openhwgroup/core-v-verif/compare/cv32e40x/dev...cv32e40s/dev)
> 
> image

Sure! :-)

silabs-krdosvik:
Will redo this merge. See comment above!

PR #2106, opened 2023-08-07
ISACOV : Add csr directed tests

AyoubJalali:


MikeOpenHWGroup:
Can we roll all these into a single test?

AyoubJalali:
> Can we roll all these into a single test?

I think we can, but I prefer it like that, every test related to a CSR instruction (CSRRW, CSRRS,C ...)

MikeOpenHWGroup:
> > Can we roll all these into a single test?

> I think we can, but I prefer it like that, every test related to a CSR instruction (CSRRW, CSRRS,C ...)

In general I do as well.  However, we are moving towards a time when we will have very large regressions of many hundreds of tests.  The overhead of multiple trivial tests will start to be significant.  At this point the stability of the CVA6 is such that we expect these CSR instruction tests to always pass, so the value of many trivial tests is now less than the cost of running may trivial tests in regression.

I strongly recommend merging these into a single test-program.

AyoubJalali:
@MikeOpenHWGroup could you check this please, I merge all the tests in one.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Good work @AyoubJalali, thanks!

PR #2110, opened 2023-08-07
ISACOV : Ignore OFLOW BIN when check_overflow = 0

AyoubJalali:
Ignore OFLOW bins when it's disable

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -549,9 +549,9 @@ covergroup cg_div_special_results(
   }
 
   cp_div_arithmetic_overflow : coverpoint instr.rs1_value {
-    //ignore_bins IGN_OVERFLOW = {[0:$]} with (!check_overflow);
     `ifdef UNSUPPORTED_WITH
-     bins OFLOW = {32'h8000_0000} iff (check_overflow && instr.rs2_value == 32'hffff_ffff); //TODO
+     ignore_bins IGN_OVERFLOW = cp_div_arithmetic_overflow iff (!check_overflow);
+     bins OFLOW = {32'h8000_0000} iff (instr.rs2_value == 32'hffff_ffff); //TODO
     `else
      bins OFLOW = {32'h8000_0000} with (check_overflow) iff (instr.rs2_value == 32'hffff_ffff);
     `endif


lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -549,9 +549,9 @@ covergroup cg_div_special_results(
...
 
   cp_div_arithmetic_overflow : coverpoint instr.rs1_value {
-    //ignore_bins IGN_OVERFLOW = {[0:$]} with (!check_overflow);
     `ifdef UNSUPPORTED_WITH
-     bins OFLOW = {32'h8000_0000} iff (check_overflow && instr.rs2_value == 32'hffff_ffff); //TODO
+     ignore_bins IGN_OVERFLOW = {[0:$]} iff (!check_overflow);
+     bins OFLOW = {32'h8000_0000} iff (instr.rs2_value == 32'hffff_ffff); //TODO
     `else
      bins OFLOW = {32'h8000_0000} with (check_overflow) iff (instr.rs2_value == 32'hffff_ffff);
     `endif
```
silabs-robin, line 557: To me it seems like the `else` clause using `with` would never need the ignore bins.

If that is correct, then this change only pertains to the `iff` clause.
We are not using that part of the code, and when time comes to resolve the VCS-related TODO then this `iff`-using clause should no longer be of relevance.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -549,9 +549,9 @@ covergroup cg_div_special_results(
   }
 
   cp_div_arithmetic_overflow : coverpoint instr.rs1_value {
-    //ignore_bins IGN_OVERFLOW = {[0:$]} with (!check_overflow);
     `ifdef UNSUPPORTED_WITH
-     bins OFLOW = {32'h8000_0000} iff (check_overflow && instr.rs2_value == 32'hffff_ffff); //TODO
+     ignore_bins IGN_OVERFLOW = {[0:$]} iff (!check_overflow);
+     bins OFLOW = {32'h8000_0000} iff (instr.rs2_value == 32'hffff_ffff); //TODO
```
silabs-robin, line 554: Btw, instead of `{[0:$]}` I think it should be possible to use `cp_div_arithmetic_overflow`.
It might not make a big difference here, but it is much easier to say "for every value" than to say "from this specific start value to this specific end value", particularly if unsigned numbers are involved.

silabs-robin (approved):

PR #2115, opened 2023-08-08
Merge script

silabs-krdosvik:
The plan is to do merging between S and X-dv more frequent. Ideally, every PR is merged into x-dv (in that case the person that made the code can determent whether the code is suitable for X, instead of a random merge that don't have control of all code)

Script for the merge processes: 
- cv32e40s to cv32e40x-dv
- cv32e40s/dev to cv32e40x/dev
- cv32e40x/dev to cv32e40s/dev

Could not fin a good method for doing merging for cv32e40x-dv to cv32e40s.

The script works for me, so I hope it will work in general. If we just start using it we will figure out if there are needed changes.

I have made a lot of comments, let me know if you think it is too much.

I remember there was some suggestions for doing moving and substituting of 40x to 40s in a different manner, but it works as it is so I have let it be (as I cant remember the exact suggestions).


bin/merge_script
```diff
@@ -0,0 +1,166 @@
+#!/bin/bash
+
```
silabs-robin, line 2: I think it should have a license header.
Even if it is "just a shell script", I think one would be better safe than sorry by adding it.

bin/merge_script
```diff
@@ -0,0 +1,166 @@
...
+#Merge of cv32e40s into cv32e40x-dv:
+merge_cv32e40s_into_cv32e40x-dv () {
+
+  echo "Merge of cv32e40s into cv32e40x-dv:"
+
+  #Enter the cv32e40x-dv repo in cv32e40x subdirectory
+  cd cv32e40x
+
+  #Make a branch in cv32e40x-dv that contain core-v-verif's cv32e40s folder from the cv32e40s/dev branch
```
silabs-robin, line 29: Maybe `#` could be substituted with `echo `, in most places?

(Or maybe `echo "==== $0:"` or something even more easily identifiable.)

bin/merge_script
```diff
@@ -0,0 +1,166 @@
+#!/bin/bash
+
+#TODO:
+#List up the PRs these changes are from (so it is easy to see what to include in the cv32e40s/dev to cv32e40x/dev merge)
+
+#Variables:
+date_time=$(date +%Y.%m.%d-%k.%M)
```
silabs-robin, line 7: Really nice.

silabs-robin (commented):
This is a very nice script. I'm looking forward to doing a merge now.

I only have a few comments.

bin/merge_script
silabs-krdosvik: True!

bin/merge_script
silabs-krdosvik: Good idea

bin/merge_script
silabs-robin, line 1: Looks like there is now 2 files, instead of the old file being renamed.

silabs-krdosvik: Hem, wonder how that happened. But anyways, thanks! Should be fixed now.

silabs-robin (approved):
Approving based on:
I tried the script locally (before the feedback-fixing additions) and it worked.
The scrit is useful and clean, and all feedback conversations are resolved.

PR #2081, opened 2023-07-26
TB updates and fixes 

XavierAubert:
This PR updates includes those changes from @dd-vaibhavjain :

- Core hash to latest at the time of PR
- add rvviRefMetricGet to solve core-v-verif issue https://github.com/openhwgroup/core-v-verif/issues/2015
- Add param for trace2api ON_MISMATCH_RECONVERGE
- Update mk/README to add desc for TEST_CFG_FILE

FIX for simd and hwloop test issues:
- fix SIMD UIMM case for update_imm_str
- update imm randomization for cv32e40p_inst type
- Fix constraints issues in hwloop tests for setupi instructions:
- This is done by simplifying constraints and and fix issues with cv.setupi, resereved_rd.
- Add checks to make sure number random generated instructions never underflow due to constraints

Issue #2007, opened 2023-06-23
Regression script in corev-dv/corev_fp_zfinx_instr_test_for_regr?


MikeOpenHWGroup:
As of hash [25c4199](https://github.com/openhwgroup/core-v-verif/commit/25c4199488abcd99afcbd8a2b86d2266dc2a23e0) (on branch cv32e40p/dev) there is a regression script in [cv32e40p/tests/programs/corev-dv/corev_fp_zfinx_instr_test_for_regr/regr.sh](https://github.com/openhwgroup/core-v-verif/blame/318a4df4e3b7055dc5af5a9ebb4c4eeff2f95c9c/cv32e40p/tests/programs/corev-dv/corev_fp_zfinx_instr_test_for_regr/regr.sh#L1).  I know that I reviewed and approved the pull-request that added this file, but now I have some concerns.  Adding a script to run test-programs in the test-program directory is a significant change to the way tests are compiled and run today, so I have a few questions:

1. Is the use if this script documented anywhere (such as a README)?
2. Why was this changed introduced?
3. Do we need a new script to run the `corev_fp_zfinx_instr_test_for_regr` test-program, or can we update an existing script to support it?
4. Can we put the run script in the `bin` with the others?

dd-baoshan:
Hi @MikeOpenHWGroup 
Please refer to my inline response. 

1) Is the use if this script documented anywhere (such as a README)?
No, Currently I only put some notes in this script to describe how this script works.

2) Why was this changed introduced?
3) Do we need a new script to run the corev_fp_zfinx_instr_test_for_regr test-program, or can we update an existing script to support it?
During the initial stage when I tried to construct the yaml file that needed for cv_regress, I noticed that it has some limitations in current flow whereby it could not meet my test intention. 
e.g the testcase corev_fp_zfinx_instr_test_for_regr can use to run with different random stream by using runargs for test overriding, I choose in this way is to reduce the number of tests folder into one test folder only.
Current cv_regress requires a yaml file and ths yaml file takes in one testname for a testcase (in this case corev_fp_zfinx_instr_test_for_regr). This also mean if there are many different runargs, all of them are still using the same testname (users can put multiple same testname with different runargs but it will takes one out of them)
BUT if let said  multiple testname is allow and this will lead to another issue whereby same testname using same result path regardless the runargs input. Currently the yaml file does not have a way to define different run folder.

[...]

PR #2117, opened 2023-08-09
Corrections and updates to CV32E40Pv2 simulation environment

XavierAubert:
This PR adds many corrections and updates added by @dd-baoshan, @dd-vaibhavjain and I. 
Sorry for piling up so many commits, there was a lot of back-and-forth between the three of us as we progressed in implemeting TB features and solving bugs as we encountered them
Below is a summary for each main topic

**1. rmdb template (for Questa Vrun) & cv_regress**
- Cleaned unused parameters for make command line and bug fixing (3cd8a15ff2d49a19bb0ce1616b6bdfc389186bc0, eb9a77fd1e82dcf9f39d1e656603d9326f638158)
- corev-dv compilation is done one time for each configuration, and never compiled again to save performance (9a9b5075fbaa54b59ec3b0af98a9af937f5ca91d & cc0f2fb7f74b49be42e37ebdc1d3aeb0aa88c939)

**2. Regression files** 
- typo fix (aafbd639e2be4da77922b9ed29802a39428ea338 & fcddc8eac0f0687dd459e04f17693d8e0b1efbeb) 
- simplified ci_check (e30325448bb74b5fc5e70d8b9ee6d1b56011028c)
The regression was too long for a simple ci_check, only hello-world test is run with all configurations. Non-relevant tests have been removed to keep the essentials. 
- Various additions (9fe4d9da596eaf9688ee92c59404def30db6a2df, 

**3. Issue #2007 solved** (60dcb276a3a7ed843a7590c216f737cd6645c203,  03fbb5f51ce0b092c16a82cb527ed2bf5b297d46, 523ac6f5e6f45c1cf14e67946956c685903d0e1f)

**4. HWLoop generation improved and implemented some fixes** (Commits 201479dd5c07e26f1dc1a73e6769b3165c14a69b, 1d474e8b2cda782d75b989d67b31257cb28ebb08, 4566b76203e2c139c8e60fa2d474b6b1afb9a4de)

**5. FP stream coverage improved** (2138c06ae384184c1620448b5b7293f61403e1c3 & 523ac6f5e6f45c1cf14e67946956c685903d0e1f)

**6. Updated configurations for riscof arch test to add FPU latency** 91d6d75f7fc5033c1732e059040006e8e0839778 & 9434db41129455eebf1ffd29ca42c4955348e9b6

**7. Default ISS usage has been set to no in various contexts as it requires a licence** 
- inside cv_regress (49ad67d2b647952e68eed12522869214822b555b)
- inside riscof arch test (6048bff3b451a031fb0f1ad42d0b299efa8ce837)

**8. Make usage of feature introduced in PR #2081** (FYI, implemented to solve #2007)
Commits 9e6cc3450c92cecd659fca11a1ae740a49c7d0fb


cv32e40p/sim/riscof/config_cv32e40p_v2.ini
```diff
@@ -12,7 +12,7 @@ target_run=1
 jobs=6
 PATH=riscof
 corevverifPATH=../../../
-imperas_iss=yes
+imperas_iss=no
```
MikeOpenHWGroup, line 15: One of the benefits of running `riscof` within the CORE-V-VERIF environment is the ability to run the RISC-V architectural test suite against the ImperasDV reference model (ISS).  We should therefore enable it by default.

MikeOpenHWGroup (changes requested):
Whew!  Big PR.  :sweat_smile:

For the most part, this all looks good and it passes a local CI (under Questasim).  I have a couple of comments/questions about the exclusion of the reference model (ISS) by default.  I would expect that the reference would be enabled by default.

cv32e40p/sim/riscof/config_cv32e40p_v2.ini
dd-vaibhavjain: Hi @MikeOpenHWGroup , this change is actually based on feedback from an older PR where this was actually first set to "yes" to make is default on. 
And the decision after that discussion was to set it "no" by default as you mentioned it may be a problem for users who want to run without ISS as there is no other way to disable it from command line. And in such cases this will cause compilation issues and would need to know what config file changes are required to fix this.

And thats why this change to revert back to "no" for imperas_iss here.

Also in this PR we have added separate config files to run with ISS=yes (new files suffixed with "_ForIssDebugOnly").
So we plan to use these new config files with ISs=yes to run riscv_arch_test suite.

MikeOpenHWGroup (approved):
OK, I'll approve since there was clearly some forethought about whether or not to include the reference model by default.  Let's discuss offline.

PR #2121, opened 2023-08-14
Merge Script - "x-dv" Detection

silabs-robin:
This PR changes how the merge script checks whether `cv32e40x/` is an `x-dv` clone or not. (Plus some other consistency tweaks.)

Rationale: None of my remotes were called `origin`, so the script tried to continue even if its prerequisites were not fulfilled.

### DIFF
--- a/bin/merge.sh
+++ b/bin/merge.sh
@@ -127,26 +127,22 @@ merge_xdev_into_sdev () {
 check_cv32e40x_repo() {
 
   echo "=== Check if cv32e40x exist ==="
-  if [ ! -d "cv32e40x" ]
-  then
+  if [ ! -d "./cv32e40x/" ]; then
     echo "Directory cv32e40x does not exists."
     echo "Run: ./bin/clonetb -x"
     echo "before running: ./bin/merge_script --s_into_x-dv"
     exit 1
   fi
+  printf "OK\n\n"
 
-  echo "=== check if cv32e40x is the core-v-verif repo ==="
-  cd cv32e40x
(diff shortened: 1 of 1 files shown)


bin/merge.sh
```diff
@@ -127,26 +127,22 @@ merge_xdev_into_sdev () {
...
+  printf "OK\n\n"
 
-  echo "=== check if cv32e40x is the core-v-verif repo ==="
-  cd cv32e40x
-
-  if [ $(git config -l | grep remote.origin.url | grep core-v-verif) ]
-  then
-    echo "Directory cv32e40x is a core-v-verif repo and not a cv32e40x-dv repo."
+  echo "=== Check if cv32e40x is the core-v-verif repo ==="
+  if [ ! -d "./cv32e40x/.git/" ]; then
```
silabs-robin, line 139: Note:
I'm checking for the existence of the `.git/` directory instead of checking the URL of the remotes.
I'm not convinced this method is 100% bulletproof, but the previous method weren't either and this one works for me.
Please let me know if anyone can think of ways this new method can fail.

kristinedosvik: Make sense! Sounds like an improvement change.

kristinedosvik (approved):
The changes looks good! Agree that checking ./git is better than checking for origin remote.

MikeOpenHWGroup (approved):
Good idea @silabs-robin.  Making this check truly bullet proof may well be impossible, but this is a good attempt.

PR #2136, opened 2023-08-18
Added covers for nmi in debug followed by single stepping with stepie = 0/1

silabs-hfegran:
* Adds covers for the case where we encounter an nmi in debug mode, drets to single step with both options for step interrupt enable

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
@@ -1075,8 +1075,8 @@ module uvmt_cv32e40s_debug_assert
       ##1
           rvfi.is_dret
        && rvfi.rvfi_nmip
-       && csr_dcsr.rvfi_csr_rdata[DCSR_STEPIE_POS]
-       && csr_dcsr.rvfi_csr_rdata[DCSR_STEP_POS] == stepie_value
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEPIE_POS] == stepie_value
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEP_POS]
        && rvfi.rvfi_valid
       ##1
           csr_mtvec.rvfi_csr_rdata > 3 // ignore lower bits
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
@@ -1070,4 +1070,41 @@ module uvmt_cv32e40s_debug_assert
         end
     end
 
+    sequence dbg_with_nmi_dret_stepie;
+          rvfi.rvfi_dbg_mode
+      ##1
+          rvfi.is_dret
+       && rvfi.rvfi_nmip
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEPIE_POS]
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEP_POS]
+       && rvfi.rvfi_valid
+      ##1
+          csr_mtvec.rvfi_csr_rdata > 3 // ignore lower bits
+       && csr_dpc.rvfi_csr_rdata != 0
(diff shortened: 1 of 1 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -1070,4 +1070,24 @@ module uvmt_cv32e40s_debug_assert
         end
     end
 
+    sequence dbg_with_nmi_dret_stepie(bit stepie_value);
+          rvfi.rvfi_dbg_mode
+      ##1
+          rvfi.is_dret
+       && rvfi.rvfi_nmip
```
silabs-robin, line 1077: Since `rvfi_nmip` is 2-bit:
The implementation probably doesn't allow `rvfi_nmip[1] && !rvfi_nmip[0]`, but if it does then the cover doesn't necessarily match NMIs?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -1070,4 +1070,24 @@ module uvmt_cv32e40s_debug_assert
...
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEPIE_POS] == stepie_value
+       && csr_dcsr.rvfi_csr_rdata[DCSR_STEP_POS]
+       && rvfi.rvfi_valid
+      ##1
+          csr_mtvec.rvfi_csr_rdata > 3 // ignore lower bits
+       && csr_dpc.rvfi_csr_rdata != 0
+       && csr_mtvec.rvfi_csr_rdata != csr_dpc.rvfi_csr_rdata
+       && !rvfi.rvfi_trap.exception
+       && rvfi.rvfi_valid[->1]
```
silabs-robin, line 1086: Does this mean `(A && B && C && valid)[->1]`?

If yes, then it doesn't have to match the first upcoming rvfi_valid?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
```diff
@@ -1070,4 +1070,24 @@ module uvmt_cv32e40s_debug_assert
         end
     end
 
+    sequence dbg_with_nmi_dret_stepie(bit stepie_value);
+          rvfi.rvfi_dbg_mode
+      ##1
```
silabs-robin, line 1075: Would removing the `##1` preserve the intention of the cover?

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-hfegran: Good catch, index 0 should be added here

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-hfegran: Agreed, the ##1 is not needed here (And is potentially in violation of rvfi semantics)

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-hfegran: Yes,

This might be better to do with a reject_on-clause, the intention here is to actually not cover the cases where the csrs match, as that specific case makes it impossible to discern if the rtl actually performed the operation intended. I will have to think about this one to see if there is a better way (without creating an overly complex cover)

silabs-hfegran: Fixed

cv32e40s/tb/uvmt/uvmt_cv32e40s_debug_assert.sv
silabs-hfegran: Fixed

silabs-robin (approved):

PR #2139, opened 2023-08-21
CVA6 CSR Access Verification PR

spidugu444:
please find the PR for CVA6 32-bit CSR Access Verification.

Highlights:
**1.Yaml files have been added for CVA6 32-bit CSRs, which are divided into supervisor mode read-write, and machine mode read-write and read-only CSRs.
2. Added test programs for CVA6 32-bit CSRs, which can be classified as supervisor mode read-write, machine mode read-write, read-only, and machine counter CSRs.
3. The behavior of the counter CSR in CVA6 Machine mode and the scenario when the counter CSR value reaches its maximum value were verified.
4.  Added CVA6 32-bit CSR access failing test cases.**

Thanks


spidugu444:
Closing this PR as I raised new PR on CVA6 repo.
please find the PR link here https://github.com/openhwgroup/cva6/pull/1380

PR #2140, opened 2023-08-21
Update x-dv hash

silabs-hfegran:


silabs-robin (approved):
Approving based on:
Ref https://github.com/openhwgroup/cv32e40x-dv/pull/24 it looks good.

PR #2143, opened 2023-08-22
Updated 40x-dv hash to 653a93fc

silabs-hfegran:
* Fix outdated/broken tests


silabs-robin (approved):
Approving based on:
This PR should point to the recent merges on x-dv, and the recent merges on x-dv are assumed to have been demonstrated to not break ci_check and formal.

PR #2144, opened 2023-08-23
Updated 40x-dv hash to 4a0a4d53

silabs-hfegran:


silabs-robin (approved):
Approving based on:
Occasional hash updates are nice.
Assuming that ci_check and formal was demonstrated to still work.

PR #2146, opened 2023-08-23
Latency options for OBI VIP configuration (plus other additions) 

XavierAubert:
Most important point of this PR is the addition on OBI VIP configuration done by @dd-vaibhavjain on commits 03de62ae005432aaf7da53a1f5e6a464898073d2 & fdd80c58ca4299fb37cee6f14ce49c9b3c063f5b. 

@MikeOpenHWGroup, you should specifically have a look at these modifications, and tell us if something's missing. 

The other commits from @dd-baoshan are mainly fixes. One in particular (commit 0c12d6d9789675bd8738dde344cef86a70915845) addresses an issue we talked a couple of weeks ago, regarding assertions errors not being catched by the flow. We will be using this while all assertions are not converted to a more standard way using uvm_errors.


cv32e40p/env/uvme/uvme_cv32e40p_cfg.sv
```diff
@@ -36,6 +36,37 @@ class uvme_cv32e40p_cfg_c extends uvma_core_cntrl_cfg_c;
...
+   bit                        random_instr_stall; // variable set by plusarg +random_instr_stall
+
+   bit                        long_random_data_stall; // variable set by plusarg +long_random_data_stall
+   bit                        long_random_instr_stall; // variable set by plusarg +long_random_instr_stall
+
+   bit                        max_rand_data_latency; // variable set by plusarg +max_rand_data_latency=<>
+   int                        rand_data_latency;
+
+   bit                        max_rand_instr_latency; // variable set by plusarg +max_rand_instr_latency=<>
+   int                        rand_instr_latency;
```
MikeOpenHWGroup, line 69: The name of this variable (`rand_instr_latency`) hints that this is a random value, yet it is not.  It is used in a constraint to set the upper bound of a random variable, so maybe it could be called `latency_limit`.

cv32e40p/env/uvme/uvme_cv32e40p_cfg.sv
```diff
@@ -109,6 +140,116 @@ class uvme_cv32e40p_cfg_c extends uvma_core_cntrl_cfg_c;
...
+      if (max_rand_instr_latency) {
+         obi_memory_instr_cfg.drv_slv_gnt_random_latency_max == rand_instr_latency;
+         obi_memory_instr_cfg.drv_slv_rvalid_random_latency_max == rand_instr_latency;
+      }
+   }
+
+   constraint long_random_data_stall_sim_cons {
+      if (long_random_data_stall) {
+         obi_memory_data_cfg.drv_slv_gnt_mode    == UVMA_OBI_MEMORY_DRV_SLV_GNT_MODE_RANDOM_LATENCY;
+         obi_memory_data_cfg.drv_slv_gnt_random_latency_min == 0;
```
MikeOpenHWGroup, line 204: The name of this constraint block is `long_random_data_stall_sim_cons` which implies we want long delays.  However the minimum delay is 0, which is the same as the minimum delay for `zero_data_max_instr_stall_sim_cons`, so you will see a lot of small delays even when `long_random_instr_stall` is set.

Is there a reason the min latency is 0 for both constraint blocks?

MikeOpenHWGroup (commented):
Look good @XavierAubert, although I have a couple of questions.

cv32e40p/env/uvme/uvme_cv32e40p_cfg.sv
dd-vaibhavjain: Yes @MikeOpenHWGroup agree with you the variable name rand_instr{data}_latency can be more meaningful with _limit suffix as the name actually specifies the maximum random latency limit. Will update this.

cv32e40p/env/uvme/uvme_cv32e40p_cfg.sv
dd-vaibhavjain: Yes the thought behind keeping min latency to 0 was basically to use this "random" latency mode to mainly cover a longer range here upto 16 cycles and not only an exclusive long range (like 8 to16). (it was mainly intended to extend the other zero_data_max_instr_stall cases).

The reason is to just exercise random delay across the longer range especially to trigger cases such as for example  misaligned accesses which causes a single load instruction to trigger 2 separate OBI transfers and with having minimum at still 0 we can see cases 1 transfer finishes immediately and 1 may take upto 16 cycles.
And reason to extend the range to 16 is also to hit more corner cases with all type of FPU instructions.

Do you think this case maybe redundant and covered elsewhere or we dont need this explicit config?

MikeOpenHWGroup: Whether we need this explicit constraint is best determined by running a full regression and reviewing the coverage.  In v1 a full coverage regrerssion required ~2000 test runs, so depending on the size of your compute farm and number of simulator licenses this could be done over a weekend or even overnight.

I like the idea of creating a scenario for long OBI cycle delays.  After a coverage regression we'll know if we need to tune the long constraints.

pascalgouedo: v1 was 114 instructions, v2 has 434 instructions and some/most of them are much more complex than standard RISC-V ones. So much more than 2000 test runs would be necessary to cover everything.
v1 was only 1 configuration, v2 has 9 configs.
We are limited by Imperas RM licenses, only 8 of them.

Long OBI delays are needed to check Load returning data with respect to long multicycle FPU instructions like FDIV/FSQRT.
Except if you have an explicit functional cross coverage between Loads and FDIV/FSQRT coverage analysis and RTL code coverage will not give you any information.

cv32e40p/env/uvme/uvme_cv32e40p_cfg.sv
dd-vaibhavjain: This is updated now.

XavierAubert: Done with commit cb7a414cd7b7502ea3f16113c18005a98c6d068e

MikeOpenHWGroup: Thanks!

MikeOpenHWGroup (approved):
Thanks for the changes.  Local CI runs and passes.  Merging...

PR #2150, opened 2023-08-24
Coverage Analysis - 40s Non-DUT Refines

silabs-robin:
This PR adds refinement files for non-DUT code coverage.


cv32e40s/sim/tools/xrun/README.md
silabs-robin, line 1: Note: Added spacing, reduced wordiness, because I struggled to read it with non-rendered text-only formatting.

cv32e40s/sim/tools/xrun/cv32e40s.auto.vRefine
silabs-robin, line 1: Note: The "auto" file is the UNR results from the latest (at the time of re-generation) "full" regression results.

cv32e40s/sim/tools/xrun/cv32e40s.hierarchy.vRefine
silabs-robin, line 1: Note:
This "hierarchy" file was about excluding non-DUT files (e.g. tb, asserts, etc) code coverage.
It was from 2021 40p so I re-did it for the 40s, and split it into multiple files with the "non_dut_code_cov_" prefix.

cv32e40s/sim/tools/xrun/cv32e40s.auto.vRefine
silabs-krdosvik: Okei, I will assume the auto file is good then, since it is automatically generated

silabs-krdosvik: Since this is not reachable for formal, we must make sure design agrees that this make sense considering the parameter set, right?

cv32e40s/sim/tools/xrun/cv32e40s.hierarchy.vRefine
silabs-krdosvik: So we should find these excluded lines somewhere in the other files?

silabs-hfegran (approved):
Looks good to me

cv32e40s/sim/tools/xrun/cv32e40s.hierarchy.vRefine
silabs-krdosvik: I see it is not 100% match, but that the difference is refinement of instances Vs. modules, and it seems ok.

silabs-robin: > So we should find these excluded lines somewhere in the other files?

Approximately, yes.

It's this about interfaces and assertion files etc. Excluding code cov on them.

bilde

cv32e40s/sim/tools/xrun/cv32e40s.auto.vRefine
silabs-robin: > we must make sure design agrees that this [the auto-generated UNR refines] make sense considering the parameter set, right?

I've added a note in our internal task tracker now.

PR #2151, opened 2023-08-24
adding initial support for imperas ref model CSR checking.

LeeHoff:
added cv32e20 to cv_regress script to further that effort.
added MPHMCounterNum =10 to dut_wrap file
Turned on CSR checking in the Impersas model. Added several CSRs that are needed before asynch. signals can be integrated.
7 of the 10 ci_check tests pass with USE_ISS=YES.

### DIFF
--- a/cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
+++ b/cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
@@ -161,6 +161,9 @@ module uvmt_cv32e20_tb;
                                                       .irq_i     (dut_wrap.irq),
                                                       .irq_ack_o (dut_wrap.irq_ack),
                                                       .irq_id_o  (dut_wrap.irq_id),
+    .fetch_enable_i          (),
+    .debug_req_i             (),
+    .core_sleep_o            (),
                                                       .*);
 
    // Debug assertion and coverage interface
--- a/cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
+++ b/cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
@@ -161,9 +161,9 @@ module uvmt_cv32e20_tb;
                                                       .irq_i     (dut_wrap.irq),
                                                       .irq_ack_o (dut_wrap.irq_ack),
                                                       .irq_id_o  (dut_wrap.irq_id),
-    .fetch_enable_i          (),
-    .debug_req_i             (),
-    .core_sleep_o            (),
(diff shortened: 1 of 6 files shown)


cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
```diff
@@ -161,6 +161,9 @@ module uvmt_cv32e20_tb;
                                                       .irq_i     (dut_wrap.irq),
                                                       .irq_ack_o (dut_wrap.irq_ack),
                                                       .irq_id_o  (dut_wrap.irq_id),
+    .fetch_enable_i          (),
```
MikeOpenHWGroup, line 164: I hate to complain about such things, but it really would be better if these lines added spaces (not tabs) to align with the preceeding lines.

cv32e20/tests/cfg/default.yaml
```diff
@@ -1,9 +1,13 @@
...
 description: Default configuration for CV32E20 simulations
 compile_flags: 
+# Alternate reference model configuration using parameter overrides and tool control
+#  --override cpu/priv_version=20190608
 ovpsim: >
-# Alternate reference model configuration using parameter overrides
-#    --override cpu/misa_Extensions=0x001104
-#    --override cpu/marchid=4
-#    --override cpu/noinhibit_mask=0xFFFFFFF0
+  --override cpu/mvendorid=0
```
MikeOpenHWGroup, line 7: The value of `mvendorid` for OpenHW cores should be 0x0000_0602.  See the [CV32E40P User Manual](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/control_status_registers.html#machine-vendor-id-mvendorid) for an example.

MikeOpenHWGroup (changes requested):
This looks good @LeeHoff.  A couple of minor comments.

cv32e20/tb/uvmt/uvmt_cv32e20_tb.sv
LeeHoff: I updated the pull request to fix this formatting.

cv32e20/tests/cfg/default.yaml
LeeHoff: The RTL currently sets the mvendorid register to 0x0.  This is an override so the test will pass (until the RTL is changed).  What the correct value for mvendorid, and marchid is currently an unresloved discussion on MM / github.
The csr specification file lists mvendorid=0x00000602, and that is what the reference model uses.

PR #2153, opened 2023-08-28
Updated 40x-dv hash to 0a6e8865

silabs-hfegran:


bin/clonetb
silabs-robin, line 1: Does ci_check pass and formal still work?

PR #2156, opened 2023-08-29
Enable cvxif extension for Spike when cvxif variable is set

AyoubJalali:
Hello, this PR related to enable/disable when we desire, using an environment variable.
Cvxif extension is enabled when **cvxif ** variable is set, and disable otherwise


JeanRochCoulon:
Thanks @AyoubJalali @cathales to propose this fix, it allow to have a functional CI (also if a long temr solution need to be found). Great !

PR #2159, opened 2023-08-29
New directed test for debug single-step on different HW loops scenarii.

pascalgouedo:


MikeOpenHWGroup:
In order to compile this test-program I needed to do two things:

1. Download and install the latest COREV toolchain ([corev-openhw-gcc-centos7-20230815](https://www.embecosm.com/resources/tool-chain-downloads/#corev)).
2. Set CV_SW_MARCH to `rv32imc_zicsr_zifencei_xcvhwlp` (adding `xcvhwlp`).

Please update the [TOOLCHAIN](https://github.com/openhwgroup/core-v-verif/blob/cv32e40p/dev/mk/TOOLCHAIN.md) markdown file to let users know of this requirement.

The testcase compiles and run, but fails:
```
UVM_INFO @ 16.800 ns : uvmt_cv32e40p_base_test.sv(302) uvm_test_top [BASE TEST] set load_instr_mem
UVM_INFO @ 114.300 ns : uvmt_cv32e40p_firmware_test.sv(146) uvm_test_top [TEST] Started RUN
CV32E40P BSP: illegal instruction exception handler entered

CV32E40P BSP: illegal instruction exception handler entered

CV32E40P BSP: illegal instruction exception handler entered

... lots more of these BSP messages...

UVM_INFO @ 192591.300 ns : uvmt_cv32e40p_firmware_test.sv(155) uvm_test_top [TEST] Finished RUN: exit status is 1
UVM_INFO @ 192591.300 ns : uvm_objection.svh(1276) reporter [TEST_DONE] 'run' phase is ready to proceed to the 'extract' phase
UVM_ERROR @ 192591.300 ns : uvmt_cv32e40p_base_test.sv(345) uvm_test_top [END_OF_TEST] DUT WRAPPER virtual peripheral flagged test failure.
UVM_ERROR @ 192591.300 ns : uvmt_cv32e40p_base_test.sv(350) uvm_test_top [END_OF_TEST] DUT WRAPPER virtual peripheral signaled exit_value=1.
UVM_INFO @ 192591.300 ns : uvm_report_server.svh(904) reporter [UVM/REPORT/SERVER] 
```
Is this expected?  Perhaps I am using the wrong toolchain.

pascalgouedo:
Hi @MikeOpenHWGroup 
In fact this test can only be run using pulp_fpu_zfinx* configuration.
I guess the errors are related to fdiv assembly instruction which generates an illegal instruction exception.

I don't know how to express that configuration restriction in test.yaml file of the test.
For instance there is nothing in any pulp_* test yaml file saying those tests should only by used with any pulp_* config...

MikeOpenHWGroup:
> In fact this test can only be run using pulp_fpu_zfinx* configuration.

Ah.  that helped.  I ran this command:
```
$ make test SIMULATOR=vcs USE_ISS=NO CFG=pulp_cluster_fpu_zfinx TEST=pulp_hardware_loop
```
With the above, the compilation of the test-program used the MARCH value of `rv32imc_zfinx_zicsr_zifencei_xcvhwlp1p0_xcvmem1p0_xcvmac1p0_xcvbi1p0_xcvalu1p0_xcvsimd1p0_xcvbitmanip1p0_xcvelw1p0` as expected.   However, the test runs, but does not exit.

> I don't know how to express that configuration restriction in test.yaml file of the test.

Yes, this is a known problem.  We _could_ add a "default_cfg" option to the `test.yaml` file, but this was never done.  I created #2171 to discuss an idea I have about it.

pascalgouedo:
> Ah. that helped. I ran this command:
> 
> $ make test SIMULATOR=vcs USE_ISS=NO CFG=pulp_cluster_fpu_zfinx TEST=pulp_hardware_loop
> 
> With the above, the compilation of the test-program used the MARCH value of rv32imc_zfinx_zicsr_zifencei_xcvhwlp1p0_xcvmem1p0_xcvmac1p0_xcvbi1p0_xcvalu1p0_xcvsimd1p0_xcvbitmanip1p0_xcvelw1p0 as expected. However, the test runs, but does not exit.

I didn't say to use any *_cluster_* configuration 😏.

In fact cluster configuration is making wfi act as a nop. The core is going to sleep mode using another instruction (cv.elw) together with pulp_clock_en_i input.
This will need specific handling in both the test-bench and the ISS to verify that particular mechanism.

MikeOpenHWGroup:
> By the way this new test is called pulp_hardware_loop_debug_test.
So you should launch:

> make test SIMULATOR=vcs USE_ISS=NO CFG=pulp_fpu_zfinx TEST=pulp_hardware_loop_debug_test

This also compiles and runs, but does not terminate.  Of course, that does not affect any CI test is from my end it is OK to merge if you would like.  So I have approved this PR, let me know if you want me to keep working on it.

MikeOpenHWGroup (approved):

pascalgouedo:
Strange. It stops correctly with questasim.
I will try with cvs and xcelium.

pascalgouedo:
For whatever reason, test is ok with vsim and xrun but not with vcs.
Investigating.

pascalgouedo:
Found the root cause!
Pb comes from VCS make file which does not use CFG_COMPILE_FLAG so only runs in v1 mode.
After having updated it, the test is passing.
Just pushed the update.
(using this PR to push coremark test slight update!)

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Wow, great catch @pascalgouedo.  _I_ should have caught the problem with `vcs.mk`, thanks for your efforts.

PR #2158, opened 2023-08-29
Debug Vplan - Updates & Link-to-cov

silabs-robin:
This PR updates the debug vplan.

It is the one we all edited in parallel.
It contains both added link-to-coverage, and it was also updated along with updates to the debug spec some time ago.

PR #2164, opened 2023-08-30
New Script - (Auxiliary) Vplans in JSON

silabs-robin:
This PR adds a script that converts the CSV vplans into JSON.

I've not yet convinced myself whether its a good/bad idea, so please give very honest feedback.

The background for this script is the [recent PR to update the debug vplans](https://github.com/openhwgroup/core-v-verif/pull/2158/files#diff-72df94a4887d68d6b16e53ed8373367fe4aa9b4e27cc3ff748ce9d90e00708d6), where the diff is very difficult to review.
As a comparison, [here is the same diff in json](https://github.com/silabs-robin/core-v-verif/compare/json_vplans_debugtest...silabs-robin:core-v-verif:json_vplans_debugtest_mergediff#diff-e8338da9547f3dd6b8e0c2479f7359b5a160b5fd783d4db6fe0322f9b5de8197).
_(Both diffs suffer from the original xlsx files having weird quirks/glitches.)_

--------

Does the json diff help in reviewing the diff?
Enough that it worth having the json files checked in?
Is it even worth reviewing the diffs when the vplans (hopefully?) have seen some kind of review beforehand?
How perfectly must the script handle variations in how vplans are written?
NB! If the original xlsx has quirks, the json will have quirks too (the script is simple and does a compromise between adapting the original csv and just letting data in be data out).

Here are some selected samples:

bilde

bilde

bilde

bilde

vs

bilde

bilde

bilde


silabs-robin:
Here's a nice comparison:

XLSX:
bilde

CSV:
bilde

JSON:
bilde

bin/csv2json
silabs-robin, line 1: Does anyone remember why we need XLSX vplans?

silabs-hfegran:
I like the idea here! 

As for xlsx -> There must be a better way... (Though I do not know the answer to that). There is the vplan tool that was brought up a while back, it has been a while since we last had a look at that.

bin/csv2json
silabs-robin: One (easily surmountable) reason for the XLSX vplans:

The "DONOTDELETE" sheet.
It holds legal values for "data validation", constraining the values certain columns may use:
bilde

That's easy to replace with 3 lists and a nested for-loop in Python (applied to the ".json" files).

silabs-robin: A second (easily surmountable) reason for the XLSX vplans:

The default column names.
They contain what info the vplan is required to have.
bilde

That's easy to replace with (approximately) 1 list, a for-loop, and the `in` keyword, in Python.

silabs-robin: A third (more difficult-to-solve?) reason for the XLSX vplans:

The "References" sheet.
It ties an ID to a full name and a URL.
bilde

It could be solved by just having all of those 3 pieces of information within the "Requirement Location" field.
_(But that isn't super-readable...)_
Alternatively, one could have some special object (json object) to hold that look-up information, but that absolutely tarnishes the super-simple system that is proposed, so I think that's worse than the option above.

silabs-mateilga:
The JSON diff is definitely easier to review than the CSV diff, so there is at least some merit to this. I still think that the solution to this is to find a better option than XLSX, and this should be looked in to, but perhaps that is a consideration for the next project?

bin/csv2json
silabs-robin: A fourth (maybe medium-difficulty to solve?) reason for the XLSX vplans:

Working with a spreadsheet might(?) be easier to edit and view than the json list of textual objects.
bilde

This might be solved by having a 1-1 equivalency between the ".csv" and the ".json".
You can edit/review in your prefered format, then a 50-line (30-sloc) Python script converts to the other.

silabs-robin, line 1: > perhaps that is a consideration for the next project?
> -- @silabs-mateilga 

Agreed.
Let's at least keep all XLSX files, and let them continue to be the main source.

Does it make sense to merge the script still? Even if it is not difficult to write from scratch in the future, it could save a couple of hours to just save this one for later.

Btw, we already have a CSV variant of all vplans, they just lie there to ease review, and all editing still happens on the XLSX.

silabs-robin: @MikeOpenHWGroup is there a 5th reason for the XLSX vplans?

silabs-mateilga: `Does it make sense to merge the script still? Even if it is not difficult to write from scratch in the future, it could save a couple of hours to just save this one for later.`
Yes. I agree that the script should be merged, and I also think there is value in committing JSON files, as they simplify review compared to CSV. 

My comment about "next project" was regarding finding an alternative to the XLSX.

silabs-robin, line 1: > There is the vplan tool that was brought up a while back, it has been a while since we last had a look at that.
> -- @silabs-hfegran 

That one is interesting indeed.

IIRC, they didn't have a human-readable textual format yet?
And I wonder if it might be a bit complicated the way they are doing it.

This json format is dead-simple, and it should have a near 1-1 equivalency with the xlsx.

It can be summarized as:
`A vplan is 1 json list, containing vplan objects, each holding key-value pairs.`
`A vplan is validated by 1) certain mandatory keys exists in each object, and 2) certain key values are constrained instead of free-form.`

silabs-robin: Excellent! Thanks for the feedback. Then I'll un-draft this PR.

MikeOpenHWGroup: Opinions on this vary, but the 5th reason for using a spreadsheet to capture DVplans is that everybody is familiar with them, and they are easy to use to capture simple tables.

silabs-robin: > Opinions on this vary, but the 5th reason for using a spreadsheet to capture DVplans is that everybody is familiar with them, and they are easy to use to capture simple tables.

That's a good point.

I can't see a big difference from this and point 4.
Luckily, CSV can be directly opened, edited, and saved in Excel and Libreoffice.
And, luckily, a flat list of JSON objects is trivially convertible to CSV.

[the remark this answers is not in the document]
silabs-robin: Result from OpenHW meeting:
1) We merge in this script now.
2) I generate the json files in a new PR.
3) We pick up the discussion about VPTOOL etc after 40s/x is done.

PR #2166, opened 2023-08-31
Resolved clic assert fixmes for mscratchcsw and mscratchcswl

silabs-hfegran:
Fixme was in place to account for an rtl error that has since been resolved, the invalid csr instructions for mscratchcsw and mscratchcswl should now trap as expected, and the assertions have been updated accordingly


silabs-mateilga (approved):
Looks good to me

silabs-robin (approved):
Approving based on:
Marton's CLIC-knowledge approval.
And, the assert looks stricter now.

PR #2169, opened 2023-08-31
fix for propagating reset signal 

eroom1966:
- requires IDV : commit 89b5aa66cccdb98b116568181f39c999f1ffa651 (HEAD -> main, origin/main, origin/HEAD)

@silabs-hfegran this fix also need implementing into the E40X using the same approach.


silabs-hfegran:
Thanks for this fix @eroom1966

silabs-hfegran (approved):

PR #2176, opened 2023-09-04
added asserts verifying m_none

silabs-mateilga:
These asserts verify that CV32E40S upholds M_NONE, MUL and DIV instructions should never retire without trapping.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_integration_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_integration_assert.sv
@@ -21,9 +21,15 @@
 
 module uvmt_cv32e40s_integration_assert
   import uvm_pkg::*;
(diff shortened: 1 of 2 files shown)


silabs-robin (approved):
Approving based on:
Reasons given in comments.
Looks like a good addition.
Both ci_check and formal pass, from what I can understand.

cv32e40s/tb/uvmt/uvmt_cv32e40s_integration_assert.sv
silabs-robin, line 1: @silabs-mateilga Question: Does any of the tests in ci_check run with M_NONE?
How much work would it be to create a config with M_NONE and run at least one test with mul/div?

silabs-mateilga: AFAIK not, we should consider this.

PR #2177, opened 2023-09-04
Obi req fifo

silabs-krdosvik:
Fixed bugs in support logic fifo.
Added fifo for obi reqs (both data and instr).

Formal compiles
ci check passes


cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
...
+  FIFO_TYPE [2:0] fifo;
   logic [1:0] pointer;
+  FIFO_TYPE zero = '0;
 
   assign is_req_attribute_in_response_o = rvalid ? fifo[2] : '0;
 
-  always @(posedge clk_i, negedge rst_ni) begin
+  always_ff @(posedge clk_i, negedge rst_ni) begin
     if(!rst_ni) begin
       fifo <= 3'b000;
```
silabs-robin, line 84: Does this give warnings in formal when the fifo type is greater than 1-bit?

The entire fifo should then be >3 bit, but this reset logic only sets 3-bits explicitly.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
...
-        fifo <= {fifo[1:0], '0};
+        fifo <= {fifo[1], fifo[0], zero};
 
       //This logic is demonstrated in time t8 and t9 in the figure above (and uses t7 to generate a situation where this part of the logic can be used)
       end else if ((gnt && req) && rvalid) begin
-        fifo[pointer] = req_attribute_i;
-        fifo <= {fifo[1:0], '0};
-
+        fifo <= {fifo[1], fifo[0], zero};
+        fifo[pointer+1] <= req_attribute_i;
```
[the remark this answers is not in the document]
silabs-krdosvik: Good question. If obi work as it should, rvalid can not be set high unless a req & gnt is set high prior to it. So in this state the pointer will never be 2, and hence pointer+1 will not overflow. (I had to think about this case myself too).

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
...
+  FIFO_TYPE zero = '0;
 
   assign is_req_attribute_in_response_o = rvalid ? fifo[2] : '0;
 
-  always @(posedge clk_i, negedge rst_ni) begin
+  always_ff @(posedge clk_i, negedge rst_ni) begin
     if(!rst_ni) begin
       fifo <= 3'b000;
-      pointer = 2'd2;
+      pointer <= 2'd2;
```
silabs-robin, line 85: If this resets to 2, does that mean the illustrations above are mirrored backwards?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -1436,10 +1436,8 @@ module uvmt_cv32e40s_tb;
         .lrfodi_bus_req (core_i.load_store_unit_i.buffer_trans_valid),
         .lrfodi_bus_gnt (core_i.load_store_unit_i.buffer_trans_ready),
 
-        .req_is_store (core_i.m_c_obi_data_if.req_payload.we),
```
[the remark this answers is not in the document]
silabs-krdosvik: Good question. I am not sure if I can use it in this case though (if that was what you were thinking about). Since obi transaction does not directly translate to ex/wb-state due to traps and debug.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Good point! Not sure about the warning part, but this should definitely be fixed.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Yess, that is correct, and confusing. Will fix the illustration. Thanks.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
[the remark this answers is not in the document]
silabs-robin: I hadn't read enough to realize the fifo counted downwards.
Can it _under_flow then?

Several `req && gnt && !rvalid`.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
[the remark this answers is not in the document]
silabs-robin: I just saw the removed line and started to wonder how the new decoder relates to this https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv#L973 .

Marking this conversation as solved, because that code looks good enough already.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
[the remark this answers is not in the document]
silabs-krdosvik: Somehow I have the idea that there cannot be more than 2 outstanding reqs at the same time, and therefore the fifo has only 3 elements (2 active and one extra for the counter to not underflow). I will check if this is correct. Now that you mention it I should probably parameterize it in case we decide to allow more or less outstanding reqs.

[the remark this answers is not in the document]
silabs-robin: If yes, what happens when `pointer = 2'd 3`? And is there any compilation warning on this?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-robin: The alternative is to count upwards.
I'm not sure which is better.
*If* it is decided to increase the fifo size, then the reset value would have to be reset to the max fifo size - 1 or something like that.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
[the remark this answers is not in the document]
silabs-robin: Resolution: Add comment after `The FIFO can only hold 2 requests at once.`, like `The FIFO can only hold 2 requests at once. (Because the instr and data OBI interfaces of the core have maximum 2 outstanding transactions.)`.

(Will resolve convo when added.)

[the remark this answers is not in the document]
silabs-krdosvik: That wont happen because there can only be two outstanding reqs.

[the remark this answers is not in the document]
silabs-krdosvik: I think I will not use time to parameterize the fifo now as it is not strictly necessary, but I will write it on the list of things to do when time.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-robin: Wrt conversation about underflow https://github.com/openhwgroup/core-v-verif/pull/2177#discussion_r1315075736, leaving the code as is (counting downwards) looks like the best choice.

(Would still be nice if the illustrations were flipped.)

silabs-krdosvik: I will leave it and fix the illustration yess. But yeah, I guess it is just a matter of preference, and it could indeed be counting upwards instead. (But agreeing leaving it for now as it serves its purpose).

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
     input logic rvalid,
 
     //Attribute in the current request
-    input logic [XLEN-1:0] req_attribute_i,
+    input FIFO_TYPE req_attribute_i,
 
     //Indicates if the response's request contained the attribute or not
-    output logic [XLEN-1:0] is_req_attribute_in_response_o
+    output FIFO_TYPE is_req_attribute_in_response_o
```
silabs-hfegran, line 73: is the input and output here always of the same type? (There is only one FIFO_TYPE parameter)

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -53,8 +53,9 @@ _____________  |  _____________  |  _____________  |
 
 
 module uvmt_cv32e40s_sl_req_attribute_fifo
+  import cv32e40s_pkg::*;
   #(
-    parameter int XLEN = 1
+    parameter type FIFO_TYPE = obi_inst_req_t
```
silabs-hfegran, line 58: I _think_ that I would prefer FIFO_TYPE_T here, as it denotes a type not a variable.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
...
+    input FIFO_TYPE req_attribute_i,
 
     //Indicates if the response's request contained the attribute or not
-    output logic [XLEN-1:0] is_req_attribute_in_response_o
+    output FIFO_TYPE is_req_attribute_in_response_o
   );
 
-  logic [2:0][XLEN-1:0] fifo;
+  FIFO_TYPE [2:0] fifo;
   logic [1:0] pointer;
```
silabs-hfegran, line 77: 2:0 and 1:0 represents what? maybe use some constant to represent these magic numbers?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
```diff
@@ -532,12 +528,12 @@ interface uvmt_cv32e40s_support_logic_module_o_if_t;
          cnt_irq_ack,
          cnt_rvfi_irqs,
 
-         req_was_store,
+         data_obi_req_was,
+         instr_obi_req_was,
```
silabs-hfegran, line 532: _was? is this perhaps _previous or _past would be a clearer suffix here

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -53,8 +53,9 @@ _____________  |  _____________  |  _____________  |
```
silabs-hfegran, line 53: There is a lot of information on why this module is here, but not so much on how to use it - it would be good to start with a few sentences on how this module should be used.

and a minor typo above ( i cannot comment there as there are no changes ) : 
> ((gnt && reg)==1)

should be `req` not `reg`

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-hfegran: Reading the fine print I think this is correct - as an idea for the future, perhaps it would be good if we could just pass an entire request/response-struct as one item in the fifo output? Then we would not need a separate module for every attribute we are interested in? It would only be necessary to hook one module up to one interface, and that would keep track of all completed obi-transactions

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
```diff
@@ -66,39 +67,40 @@ module uvmt_cv32e40s_sl_req_attribute_fifo
     input logic rvalid,
 
     //Attribute in the current request
-    input logic [XLEN-1:0] req_attribute_i,
+    input FIFO_TYPE req_attribute_i,
 
     //Indicates if the response's request contained the attribute or not
-    output logic [XLEN-1:0] is_req_attribute_in_response_o
+    output FIFO_TYPE is_req_attribute_in_response_o
   );
```
silabs-hfegran, line 74: `is_req_attribute_in_response_o` The naming here seems to indicate that you only pass a boolean true/false to the presence of an attribute in the request phase. What about if you want to track address vs data-phase? would that work for this fifo? if so, the naming here does not make sense.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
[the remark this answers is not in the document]
silabs-robin: https://github.com/openhwgroup/core-v-verif/pull/2180

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Agreed. Will change.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Good Idea. I will let the FIFO only use one type as I believe that is more modular, but I will add a struct for a general OBI req-resp packet in the support logic.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Agree, will be changed

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Yes, have made it parameterized

silabs-krdosvik: And named the magic numbers

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-robin: > I will add a struct for a general OBI req-resp packet in the support logic

Do we already know where it shall be used?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: There should be no warnings in formal related to this added feature.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: The info is changed

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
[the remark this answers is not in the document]
silabs-krdosvik: I parameterized it for easier to use in the future.

silabs-krdosvik:
I have addressed all comments.
ci_check passes
Formal compiles
No warnings related to this PR in formal or simulation (as far as I can see).

silabs-robin (approved):
Approving based on:
All my comments and questions are resolved.

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_ifs.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

cv32e40s/tb/uvmt/support_logic/uvmt_cv32e40s_sl_req_attribute_fifo.sv
silabs-krdosvik: Is this resolved @silabs-hfegran ?

PR #2179, opened 2023-09-04
Updated 40x-dv hash to include rv32e assert file

silabs-mateilga:


bin/clonetb
silabs-robin, line 1: ci_check passes and formal still runs?

silabs-mateilga: It did when we merged the PR to 40x-dv, I can rerun the checks now, if necessary?

silabs-robin: Oh ok. If it was ran on cv32e40x/dev then its the same.

silabs-robin (approved):
Approving based on:
Nice to get the hash updated.
ci_check and formal was good on https://github.com/openhwgroup/cv32e40x-dv/pull/35 .

PR #2183, opened 2023-09-05
Add rvfi memory signals to rvfi interface

silabs-krdosvik:
Add rvfi memory signals to rvfi interface. The usermanual is not up to date, but the added rvfi memory signals are official signals in the same manner as rvfi_mem_prot. This PR is connected to other PRs that is not yet made, that is why it looks so "empty". It should not cause any problem to merge it though.

instr/data_prop is not as the other signals, that is because I am lazy and has not changed it. I could do it if that is preferred though.

### DIFF
--- a/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
+++ b/lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
@@ -74,7 +74,14 @@ interface uvma_rvfi_instr_if_t
     input logic [(NMEM*XLEN/8)-1:0]  rvfi_mem_wmask,
 
     input logic [2:0]                instr_prot,
-    input logic [NMEM*3-1:0]         mem_prot
+    input logic [1:0]                rvfi_instr_memtype,
+    input logic                      rvfi_instr_dbg,
+    input logic [ NMEM*3-1:0]        mem_prot,
+    input logic [ 1*NMEM-1:0]        rvfi_mem_exokay,
+    input logic [ 1*NMEM-1:0]        rvfi_mem_err,
+    input logic [ 6*NMEM-1:0]        rvfi_mem_atop,
+    input logic [ 2*NMEM-1:0]        rvfi_mem_memtype,
+    input logic [ NMEM-1  :0]        rvfi_mem_dbg
   );
 
   typedef logic[4*NMEM-1:0] mem_mask_t;


lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -74,7 +74,14 @@ interface uvma_rvfi_instr_if_t
...
     input logic [2:0]                instr_prot,
-    input logic [NMEM*3-1:0]         mem_prot
+    input logic [1:0]                rvfi_instr_memtype,
+    input logic                      rvfi_instr_dbg,
+    input logic [ NMEM*3-1:0]        mem_prot,
+    input logic [ 1*NMEM-1:0]        rvfi_mem_exokay,
+    input logic [ 1*NMEM-1:0]        rvfi_mem_err,
+    input logic [ 6*NMEM-1:0]        rvfi_mem_atop,
+    input logic [ 2*NMEM-1:0]        rvfi_mem_memtype,
+    input logic [ NMEM-1  :0]        rvfi_mem_dbg
```
silabs-robin, line 84: What is the reason for not doing the tb bind connections now?

https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv#L166

bilde

silabs-robin (commented):
> This PR is connected to other PRs that is not yet made, that is why it looks so "empty". 

Interesting.

I just have the one question below.

silabs-robin (approved):
Approving based on:
Since the new signals are *official* we want them in.
Changes seem safe (compilation wise).
I'm curious about the PR strategy.
Even if the signals are unconnected and if anyone should use them erronously, it will give us insight in safe assertion writing habits (and the signals will soon be connected in upcoming PRs anyway).

I will merge once the comment is resolved, unless anyone objects.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-robin, line 1: I just noticed _(since the merge button wasn't gated by open conversations)_ that this PR is to x/dev.
That should be fine. (Changes pertain to top-level files and all of that is perfectly merge'able across S <-> X.)

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
silabs-krdosvik: That is a good point. I think I thought this should wait for the signals to actually exist in the rvfi module (there will come a PR there too), but if it is no problems I should include it here so we dont need yet another PR.

silabs-robin: Resolving based on IRL discussion.

silabs-krdosvik: I have actually had problems with that before in simulation, so I think it must wait for now

PR #2184, opened 2023-09-06
Updated cv32e40x-dv hash

silabs-hfegran:


silabs-robin (approved):
Approving based on:
The hash update seems to only pull in https://github.com/openhwgroup/cv32e40x-dv/pull/36 , which only updates `debug_test_trigger.c`.
So I think it is safe to assume that ci_check has been maintained all the way through that point, that this change can't have a negative impact on ci_check, and that ci_check will be passing after this.

PR #2188, opened 2023-09-07
ISACOV : Fix rd_toggle for LBU & LHU instructions

AyoubJalali:
This PR, fix an issue related to rd_toggle for LBU & LHU instruction, because the destination register store only 1 byte (LBU), 2 bytes (LHU), and add to this the instructions are unsigned so we can't toggle all the raining bits in the register, so it's make more sense to toggle only 1 byte or 2 byte.


AyoubJalali:
@MikeOpenHWGroup I changed the target branch

MikeOpenHWGroup (approved):

PR #2193, opened 2023-09-11
JSON Vplans - Generate From CSV

silabs-robin:
This PR adds ".json" variants of all 40s vplans.

The conversion is not perfect*, but I think it will be useful for 1) evaluating vplan update PRs, and 2) exploring vplan format requirements.
_(*I.e. there can be empty fields with MISSING data, but there shouldn't be any WRONG data in any fields. Just ask me if anyone is curious about the details of that.)_


bin/csv2json
```diff
@@ -57,11 +57,16 @@ csv_reader   = csv.DictReader(csv_file)
 csv_rows         = []
 csv_row_previous = None
 
+def should_fill_cell(cell_value, previous_row, row_key):
+    cell_is_empty        = not cell_value
+    previous_row_exists  = previous_row
+    column_should_repeat = row_key in ['Feature', 'Verification Goal']
+    return  cell_is_empty  and  previous_row_exists  and  column_should_repeat
```
silabs-robin, line 64: This filtering doesn't handle every possible case, because the ".xlsx" files are haphazardly and inconsistently formatted.

(I have ideas for how to fix this in the future. But let's compromise for now since this is just a test run?)

MikeOpenHWGroup (approved):
As discussed, this looks "OK" to me.  I will approve and defer to @silabs-hfegran for the merge.

PR #2196, opened 2023-09-13
use correct csr privlevel path

silabs-krdosvik:
RTL has added an if statement that needs to be included in the path.

Formal compiles
Ci check is all red
But after this bug is fixed: https://github.com/openhwgroup/cv32e40s/issues/499
It is all green except clic config tests (clic and debug_test2) (which I think is a clic problem and not something caused by this change, as that would be weird)

### DIFF
--- a/cv32e40s/sim/ExternalRepos.mk
+++ b/cv32e40s/sim/ExternalRepos.mk
@@ -15,7 +15,7 @@ export SHELL = /bin/bash
 
 CV_CORE_REPO   ?= https://github.com/openhwgroup/cv32e40s
 CV_CORE_BRANCH ?= master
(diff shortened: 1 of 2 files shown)


silabs-mateilga:
Looks good, but we should probably delay merging until the mentioned bug is fixed.

silabs-krdosvik:
Agreed! Now it is fixed! So would be nice to merge this asap

silabs-hfegran:
This requires hash update of the S as well right? that's missing from this PR

silabs-hfegran (changes requested):
Which hash of the 40S broke this? and which hash  fixes it? core-v-verif needs to be updated to match the OK hash before this PR can be merged

silabs-hfegran (approved):

PR #2199, opened 2023-09-13
Merge S->X

silabs-robin:
TODO


bin/xlsx2csv
silabs-krdosvik, line 1: Why have you removed the .py? Personally I prefer the ".<file-type>", but I guess it doesn't really matter.

silabs-krdosvik (approved):
Looks good. Have only one question, but doesn't matter what the outcome is (and has probably been done in a previous PR), so approve

bin/xlsx2csv
silabs-krdosvik: Make sense!

PR #2200, opened 2023-09-13
Merge X->S

silabs-robin:
TODO


silabs-robin:
This diff is so small I don't think it is worth doing until later.

PR #2202, opened 2023-09-15
Updated cv32e40x-dv hash to 4ba93acf850

silabs-hfegran:


bin/clonetb
silabs-robin, line 1: ci_check? Formal?

silabs-hfegran: OK

silabs-robin (approved):
Approving based on:
ci_check passes.
(Assuming formal worked too, when the changes were PRed into x-dv.)
Updating the hash is always nice.

PR #2204, opened 2023-09-18
Updated hpmcounter tests to account for csr stall optimizations

silabs-hfegran:


silabs-hfegran:
only c-test changes, fixes tests that are broken with the current hash, passes with latest core hash

silabs-robin (approved):
Approving based on:
Offline discussion.
CSR optimalization caused changes in cycle counts.

PR #2207, opened 2023-09-19
Disabled buserr ifault assertion

silabs-hfegran:
Disabled assertion that frequently gives false positives in sim, needs a rewrite


silabs-robin (approved):
Approving based on:
This was indeed causing unnecessary problems for the regressions.
Assuming the TODO will be fixed in the not too far future.

PR #2209, opened 2023-09-22
Disabled buserr sb assertion due to false positives

silabs-hfegran:


silabs-hfegran:
diff easier to read if you exclude whitespace changes

silabs-hfegran:
ci_check and formal OK

silabs-robin (approved):
Approving based on:
The regression problem this fixes is known, and this removes the problem.
The commented-out code has a todo, so it cannot be completely forgotten.
The code in question is not mission critical.

PR #2210, opened 2023-09-22
more xsecure expression coverage waving

silabs-krdosvik:
Added more xsecure glitch waving, since we dont have xsecure simulation tests


silabs-robin (approved):
Approving based on:
We have indeed decided to do xsecure in formal, so the sim coverage is irrelevant and ok to .
Since the tool mangles the diff I will simply trust it.

PR #2211, opened 2023-09-25
Updated cv32e40x-dv hash to latest

silabs-hfegran:
ci_check, formal OK


silabs-robin (approved):
Approving based on:
Hash update, always wanted.
Test status, ok.

PR #2212, opened 2023-09-25
Explicitly clear minhv at the end of mret_after_minhv test

silabs-hfegran:
Only C-test change, passes without ISS (known ISS mismatch)


silabs-robin (approved):
Approving based on:
This fixes a timeout from an infinite loop.
Rest of ci_check and formal can't have been affected by 1 file changed.
I don't fully understand the CLIC implications nor the test flow, so I just trust it.

PR #2217, opened 2023-09-28
Toggle Coverage

silabs-robin:
This PR facilitates getting that top-level toggle coverage.

Test status:
* Hello world - Compiles and passes without new warnings.
* ci_check - Only fails on "clic" test.
* Formal - N/A.


mk/uvmt/xrun.mk
silabs-robin, line 1: ci_check was indeed not affected.

silabs-hfegran (approved):

PR #2219, opened 2023-09-28
Added cfg to disable data independent timing and pc hardening

silabs-hfegran:
* Added possibility and cfg to disable data independent timing and pc hardening features for random tests
* Added this cfg to rand instr test and rand arithmetic test to cover holes
* Bugfix for unused dummy and no hint cfg

No files affecting formal changed
Full regression pass


cv32e40s/env/corev-dv/cv32e40s_asm_program_gen.sv
```diff
@@ -43,8 +43,8 @@ class cv32e40s_asm_program_gen extends corev_asm_program_gen;
       end
       2'b01: begin
         instr = {
-          $sformatf("add x%0d, zero", cfg.gpr[0]),
-          $sformatf("addi x%0d, 0x4", cfg.gpr[0]),
+          $sformatf("add x%0d, x%0d, zero", cfg.gpr[0], cfg.gpr[0]),
+          $sformatf("addi x%0d, x%0d, 0x4", cfg.gpr[0], cfg.gpr[0]),
```
silabs-robin, line 47: Was 1 operand missing before?
How could it compile? It defaults to some short-hand pseudo instruction?

cv32e40s/env/corev-dv/cv32e40s_asm_program_gen.sv
```diff
@@ -67,6 +67,36 @@ class cv32e40s_asm_program_gen extends corev_asm_program_gen;
       end
     endcase
 
+    case ({corev_cfg.disable_pc_hardening, corev_cfg.disable_data_independent_timing})
```
silabs-robin, line 70: Since this is in `gen_program_header`, does it mean we will not toggle it randomly during runtime?
Do we not want that, or does it not make sense?

silabs-krdosvik: > Since this is in gen_program_header, does it mean we will not toggle it randomly during runtime?

It was so that random tests should be able to set the CSR by using random sequences, right? If so, at least coverage analyzise shows that it has not happened.

cv32e40s/env/corev-dv/cv32e40s_asm_program_gen.sv
silabs-hfegran: sva compiled, but not the resulting assembly - this combination has never been used so it was never caught until I copy-pasted this code to the newly added section and discovered that it would not compile.

cv32e40s/env/corev-dv/cv32e40s_asm_program_gen.sv
silabs-hfegran: Normally we do not expect these bits to be turned off, but at least this case covers the case where we actually turn them off and run without.

cv32e40s/env/corev-dv/cv32e40s_asm_program_gen.sv
```diff
@@ -67,6 +67,36 @@ class cv32e40s_asm_program_gen extends corev_asm_program_gen;
...
+        instr = {
+          $sformatf("add x%0d, x%0d, zero", cfg.gpr[0], cfg.gpr[0]),
+          $sformatf("addi x%0d, x%0d, 0x8", cfg.gpr[0], cfg.gpr[0]),
+          $sformatf("csrrc x0, 0xbf0, x%0d", cfg.gpr[0])
+        };
+        gen_section(get_label("disable_pc_hardening_data_ind_timing", hart), instr);
+      end
+      2'b11: begin
+        instr = {
+          $sformatf("add x%0d, x%0d, zero", cfg.gpr[0], cfg.gpr[0]),
```
silabs-hfegran, line 92: Adds are wrong here, need fix

silabs-robin (approved):
Approving based on:
Test status reported to be ok.

PR #2223, opened 2023-10-03
Enabled wfe_test and xsecure_* tests in full regression

silabs-hfegran:
Added missing hpmcounter-flag to ISS settings

ci_check ok,
full regression w/o ref. model shows no direct test breakage
fv no impact


silabs-robin (approved):
Approving based on:
Full regression now covers more features.
Test status reported to be ok.

PR #2228, opened 2023-10-03
TB & corev-dv stream generations fixes 

XavierAubert:
Main topics for this PR are fixes and updates to the stream generation made by @dd-baoshan & @dd-vaibhavjain, following our first regressions runs and they mainly concern E40P only. 

The only file non-specific to CV32E40P updated is the cv_regress script where I updated the use of skip_sim field temporarily to be able to prevent tests to be run under specific cfgs.


bin/cv_regress
```diff
@@ -173,6 +168,16 @@ def read_file(args, file):
         elif test.num != 1:
             test.num = int(args.num or test.num)
 
+        try:
```
MikeOpenHWGroup, line 171: I can see the utility of skipping a specific simulation for a given configuration, but it is not clear to me why we would skip a simulaton based on the simulator used.  Having said that, this `test.simulator` check has been in this script for a long time, so I will not gate approval of this PR on it.

Nevertheless, I am curious - @silabs-hfegran can you comment?

cv32e40p/env/corev-dv/custom/isa/cv32e40p_riscv_compressed_instr.sv
```diff
@@ -12,46 +15,33 @@
...
 
 class cv32e40p_riscv_compressed_instr extends riscv_compressed_instr;
 
-  rand riscv_fpr_t fs1;
   rand riscv_fpr_t fs2;
   rand riscv_fpr_t fd;
-  rand bit[4:0]  reg_idx_fs2; //FIXME : dd-vaibhavjain
-  rand bit[4:0]  reg_idx_fd; //FIXME : dd-vaibhavjain
 
-  //FIXME: dd-vaibhavjain. Remove after toolchain fix
```
MikeOpenHWGroup, line 31: Wow.  I did not imagine that we would ever be able to say the toolchain is fixed.  ;-)

cv32e40p/env/corev-dv/cv32e40p_rand_instr_stream.sv
```diff
@@ -109,7 +109,18 @@ class cv32e40p_rand_instr_stream extends riscv_rand_instr_stream;
 
         rand_cnt++;
         if (rand_cnt >= 200) begin
-          `uvm_fatal(`gfn, $sformatf("rand_cnt hit limit %0d and not able to find space for instr placement. Please revise the stream", rand_cnt))
+          int idx_placement[$];
+          `uvm_info(`gfn, $sformatf("placement limit %0d reached. Place the stream at begining of instr_list", rand_cnt), UVM_NONE)
```
MikeOpenHWGroup, line 113: Nice solution.

cv32e40p/regress/cv32e40pv2_benchmarks.yaml
```diff
@@ -16,21 +16,45 @@ builds:
...
 # List of tests
 tests:
+  dhrystone:
+    build: uvmt_cv32e40p
+    description: dhrystone
+    dir: cv32e40p/sim/uvmt
+    cmd: make test TEST=dhrystone CFG_PLUSARGS="+UVM_TIMEOUT=1000000"
+    num: 1
+
+  fibonacci:
```
MikeOpenHWGroup, line 31: Does this `fibonacci` test add to our total coverage?  It was "found" in a directory of old RIS5CY test-programs and I copied it over.  Maybe it is time to retire it.  (Of course, it doesn't hurt to keep it...)

Maybe if we increased the length of the calculated sequence and enabled random interrupts.

cv32e40p/tb/uvmt/uvmt_cv32e40p_imperas_dv_wrap.sv
```diff
@@ -455,8 +455,8 @@ module uvmt_cv32e40p_imperas_dv_wrap
 
     // Select processor name
     void'(rvviRefConfigSetString(IDV_CONFIG_MODEL_NAME, "CV32E40P"));
-    // Worst case propagation of events 4 retirements (actually 3 observed)
-    void'(rvviRefConfigSetInt(IDV_CONFIG_MAX_NET_LATENCY_RETIREMENTS, 4));
+    // Worst case propagation of events 14 retirements (14 due to long fpu multicycle instr)
```
MikeOpenHWGroup, line 458: It might be a good idea to get a review from @duncangraham-Imperas on these updates.  (I won't gate the approval on this, but it would be a good idea going forward.)

MikeOpenHWGroup (commented):
Looks good @XavierAubert.   I will run a local CI before approving.

MikeOpenHWGroup (approved):
Approving based on my previous review comments and passing local CI (both vsim and vcs).

cv32e40p/regress/cv32e40pv2_benchmarks.yaml
pascalgouedo: There are tests for verification and tests for benchmarking.
The second category is not there for any verification purpose.
But I agree I never used that fibonacci test in the past.
And we are going to use Embench additionally to Dhrystone and Coremark for performance benchmarking.

bin/cv_regress
silabs-hfegran: I was not aware of this code in the cv_regress script.

That said, the only thing that comes to mind is if a specific test triggers a known issue with a certain simulator - I guess this was more relevant back when we had multiple tools running the various testbenches to avoid having to do a lot of unrelated maintenance.

PR #2229, opened 2023-10-06
[ci] Add codebuild

zarubaf:


buildspec.yml
```diff
@@ -0,0 +1,8 @@
+version: 0.2
+
+phases:
+
+  build:
+    commands:
+      - chown -R florian:florian .
+      - su florian -c "source /synopsys/scripts/env.sh; echo $PATH"
```
MikeOpenHWGroup, line 8: Will this command require the runner to provide the password for user `florian`?

MikeOpenHWGroup, line 8: Looking at log AWS logs I see this:
```
[Container] 2023/10/10 20:38:27 Running command su florian -c "source /synopsys/scripts/env.sh; echo $PATH"

/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/codebuild/user/bin
```
It seems that "sourcing" the script does not propagate the env vars to the invoking shell.  How about this:
```
     commands:
      - chown -R florian:florian .
      - export VCS_HOME=/synopsys/vcs/S-2021.09-SP1
      - export PATH=$VCS_HOME/bin:$PATH
      - vcs -id
 ```

zarubaf: If it works, please change whatever works!

PR #2232, opened 2023-10-10
Add wrongly removed tdata3 coverage

silabs-krdosvik:
Formal compiles
hello-world passes
ci check is running

The coverage is removed in this PR: https://github.com/openhwgroup/core-v-verif/pull/2226
But the vplan ask to have these coverage points:
image
So therefore reinserting the coverage points.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
@@ -370,6 +370,32 @@ module uvmt_cv32e40s_triggers_assert_cov
     p_dt_tcsr_not_implemented(ADDR_TDATA3)
   ) else `uvm_error(info_tag, "Access to tdata3 does not cause an illegal exception (when no higher priority exception has occured)\n");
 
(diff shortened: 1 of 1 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -370,6 +370,32 @@ module uvmt_cv32e40s_triggers_assert_cov
     p_dt_tcsr_not_implemented(ADDR_TDATA3)
   ) else `uvm_error(info_tag, "Access to tdata3 does not cause an illegal exception (when no higher priority exception has occured)\n");
 
+  // Assertions and coverages for when there are debug triggers:
```
silabs-hfegran, line 373: I disagree with this reintroduction - This creates cover properties for accessing a CSR that does not exist, an assertion should suffice to formally very that this register can never actually be read/written (i.e. always illegal) and its associated cover/assertion coverage metric should be sufficient to say that this has been covered either by formal or by sim

silabs-hfegran (changes requested):
Suggest we drop this PR and rather change the vplan

PR #2233, opened 2023-10-10
ASSERT_OFF

silabs-robin:
This PR makes compilation of assertions optional.

For blazing fast re-compilation and next level productivity iteration times,  simply use `+define+COREV_ASSERT_OFF` _(and optionally comment away the ifndef for a specific assertion set you want to work on)_.

Compilation times (mm:ss):
* 00:20 - No assertion sets
* 00:23 - One assertion set  (umode_assert)
* 01:30 - All assertion sets
* 02:10 - Internal formal

Test results:
* `ci_check` - All pass (except "clic").
* Cvverif formal - Compiles and runs decently.
* Internal formal - No reds after a few minutes.

--------

I got tired of formal taking so long to recompile just for simple changes, so that is why I implemented this.

An additional bonus, is that the warnings log is much smaller, so you specifically see all warnings related to your module.

The reason for using `ifndef` is because 1) this is consistent with [the way it is already done in the RTL repo](https://github.com/openhwgroup/cv32e40s/blob/master/bhv/cv32e40s_wrapper.sv#L172), and 2) we don't want to change the hierarchy with several additional paths with generates.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
@@ -441,6 +441,7 @@ module uvmt_cv32e40s_tb;
                                                                      .rvfi_csr_wdata(rvfi_i.rvfi_csr_dscratch_wdata[0])
     );
 
+
   // dscratch1
   bind cv32e40s_wrapper
     uvma_rvfi_csr_if_t#(uvmt_cv32e40s_base_test_pkg::XLEN) rvfi_csr_dscratch1_if(.clk(clk_i),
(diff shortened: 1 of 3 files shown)


cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
silabs-robin, line 1: One moment... I realized 2 assertion sets need to be exemptet from this guard.

silabs-robin: Done.

silabs-hfegran:
> This is a very tricky diff to read, but I don't see any issues. I approve of the motivation for this change, so as long as nothing breaks this is a good change.
> 
> Approving

Much easier to read this diff with the "hide whitespace" option set.

silabs-hfegran:
Looks good to me, thanks @silabs-robin

silabs-hfegran (approved):

PR #2234, opened 2023-10-11
Fix comment, and xsecure prefix.

silabs-krdosvik:
Remove comment about already removed code (see https://github.com/openhwgroup/core-v-verif/pull/2232)
Fix xsecure prefix.

Formal compiles
Hello-world passes
Didnt see the need for a ci check.

### DIFF
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
@@ -359,11 +359,9 @@ module uvmt_cv32e40s_triggers_assert_cov
 
   //- Vplan:
   //Check that attempts to access "tdata3" raise an illegal instruction exception, always. (Unless overruled by a higher priority.)
-  //Verify that tdata3 is illegal for all tdata2 types.
 
   //- Assertion verification:
   //1) Check that attempts to access "tdata3" raise an illegal instruction exception, always. (Unless overruled by a higher priorit
-  //2) Verify that tdata3 is illegal for all tdata2 types.
 
   //1)
   a_dt_tdata3_not_implemented: assert property (
--- a/cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_interface_integrity_assert.sv
+++ b/cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_interface_integrity_assert.sv
@@ -706,7 +706,7 @@ module uvmt_cv32e40s_xsecure_interface_integrity_assert
   ) else `uvm_error(info_tag_glitch, "The NMI caused by an associated parity/checksum error does not have exception code 1027 or 1026.\n");
 
   //Load instructions
-  c_glitch_xsecure_security_parity_checksum_fault_NMI_load_instruction: cover property (
+  c_glitch_xsecure_integrity_parity_checksum_fault_NMI_load_instruction: cover property (
 
     obi_data_rvalid
     && data_integrity_err
@@ -715,7 +715,7 @@ module uvmt_cv32e40s_xsecure_interface_integrity_assert
   );
 
   //Store instructions
-  c_glitch_xsecure_security_parity_checksum_fault_NMI_store_instruction: cover property (
+  c_glitch_xsecure_integrity_parity_checksum_fault_NMI_store_instruction: cover property (
 
     obi_data_rvalid
     && data_integrity_err


cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -359,11 +359,9 @@ module uvmt_cv32e40s_triggers_assert_cov
 
   //- Vplan:
   //Check that attempts to access "tdata3" raise an illegal instruction exception, always. (Unless overruled by a higher priority.)
-  //Verify that tdata3 is illegal for all tdata2 types.
 
   //- Assertion verification:
   //1) Check that attempts to access "tdata3" raise an illegal instruction exception, always. (Unless overruled by a higher priorit
-  //2) Verify that tdata3 is illegal for all tdata2 types.
 
   //1)
```
silabs-robin, line 366: I can't see clearly what is happening.

Will `a_dt_tdata3_not_implemented` provide the same coverage as the covers in those two previous PRs?

And do we fully check and cover everything that the vplan item is asking for?
bilde

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -359,11 +359,9 @@ module uvmt_cv32e40s_triggers_assert_cov
 
   //- Vplan:
   //Check that attempts to access "tdata3" raise an illegal instruction exception, always. (Unless overruled by a higher priority.)
-  //Verify that tdata3 is illegal for all tdata2 types.
```
silabs-robin, line 362: Why is this comment removed? It seems like exactly the thing the vplan is asking for us to test?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_interface_integrity_assert.sv
```diff
@@ -715,7 +715,7 @@ module uvmt_cv32e40s_xsecure_interface_integrity_assert
   );
 
   //Store instructions
-  c_glitch_xsecure_security_parity_checksum_fault_NMI_store_instruction: cover property (
+  c_glitch_xsecure_integrity_parity_checksum_fault_NMI_store_instruction: cover property (
```
silabs-robin, line 718: Seems to now be consistent with `c_glitch_xsecure_integrity_update_register_parity_checksum_error`. Ok.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-krdosvik: This PR (https://github.com/openhwgroup/core-v-verif/pull/2226) removes the coverpoints, without removing the comments.
All the comments are copied from the vplan and specify what vplan point each assertion/coverpoint covers.

Since the vplan says: "Verify that tdata3 is illegal for all tdata2 types", I made a PR to reinsert the coveragepoints, as I thought they should be there (https://github.com/openhwgroup/core-v-verif/pull/2232)

However, Henrik argues that they are not needed after all and that they could be removed, and we should therefore also change the vplan (see comments here: https://github.com/openhwgroup/core-v-verif/pull/2232), and therefore I will also remove the comments to keep it updated with a upcomming vplan change.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-hfegran: Comment is fine as long as there is an assertion in place that simply checks that tdata3 accesses can never succeed - then _all_ tdata2 types will be covered

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-robin: > are they really needed if we have formal proof that tdata3 accesses can never succeed

The vplan reasonably asks to check that the failure happens in various scenarios.

Ever since coverage found that over-constraint bug where half the instruction space was impossible to execute in formal I don't trust "I think this is probably covered by that assert".

silabs-robin: To make it clear:

--------

Summary of my understanding of this case:

The blue part of the vplan screenshot below _can_ be removed from the vplan (or we can also save time by not doing it).
That vplan item is not very important so it is not worth more time.

The red part of the vplan screenshot below should be removed from the vplan, because the covers do not exist anymore.

bilde

--------

Therefore, the answers to my questions are:

> Will a_dt_tdata3_not_implemented provide the same coverage as the covers in those two previous PRs?

No, it doesn't give the level of assurance that the vplan asks for.

> And do we fully check and cover everything that the vplan item is asking for?

No, but it doesn't matter, because we now deem that part of the vplan to not be worth any more effort.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-robin: For completeness:

> Why is this comment removed? It seems like exactly the thing the vplan is asking for us to test?

Because the corresponding vplan item is unimportant and we don't want to explicitly check it.

silabs-robin (approved):
Approving based on:
Changes matches what we have discussed regarding following the vplan.

PR #2236, opened 2023-10-11
Prototype VCS CI running on AWS

MikeOpenHWGroup:


.github/workflows/aws.yml
```diff
@@ -0,0 +1,28 @@
+name: aws
+on: push
+
+# on:
+#   push:
+#     branches:
+#       - 'dev'
```
zarubaf, line 7: You might want to change that, depending on when to trigger the job. Most l likely on PRs to `dev` I guess?

buildspec.yml
```diff
@@ -0,0 +1,37 @@
...
+      - export PATH=$VCS_HOME/bin:$PATH
+      - vcs -id
+#      - export LM_LICENSE_FILE=29000@ip-172-31-46-244
+      - export SNPSLMD_LICENSE_FILE=27020@ip-172-31-46-244.eu-west-1.compute.internal
+# Questasim (vsim) environment
+#      - export QUESTAROOT=/onespin/questasim
+#      - export LM_LICENSE_FILE=29000@ip-172-31-46-244
+#      - export PATH=${QUESTAROOT}/bin:${PATH}
+#      - export MTI_VCO_MODE=64
+#      - wget https://buildbot.embecosm.com/job/corev-gcc-centos7/19/artifact/corev-openhw-gcc-centos7-20230905.tar.gz
```
zarubaf, line 26: Maybe remove the commented lines?

buildspec.yml
```diff
@@ -0,0 +1,37 @@
...
+#
+# Build-specification for the aws.yml workflow.
+
+version: 0.2
+
+phases:
+
+  build:
+    commands:
+      - chown -R florian:florian .
```
silabs-hfegran, line 12: I do not fully comprehend everything here, but are these scripts tied to a specific users account? should/could this be generalized?

silabs-hfegran (commented):
ci_check script changes looks good to me - no immediate concerns there; as for the AWS related files I think that is best left up to others to review.

.github/workflows/aws.yml
MikeOpenHWGroup: Yes, you are correct.  I will be discussing the details of this with @rickoco (to understand our legal constraints) and the Members (to understand their expectations) and will update the trigger accordingly.

buildspec.yml
MikeOpenHWGroup: Yes, the commented lines will be removed before the Draft status of this PR is dropped.  These were left in place to show you what I did to get Questasim (vsim) to run.  It did not work.  Perhaps the value of `LM_LICENSE_FILE` needs to be set to the same value as we are using for `SNPSLMD_LICENSE_FILE`?  Any thoughts on this?

buildspec.yml
MikeOpenHWGroup: This is a good question.  The CI runs on a AWS VM instance that only OpenHW staff have access to.  It _could_ be any of us, but it cannot be "just anyone".  It probably makes sense to create a generic "openhw-ci-runner" account that has its access privileges tuned to the CI requirements (and no more).

.github/workflows/aws.yml
suppamax: I think it relates to the process which had been setup for cv32e40p: https://github.com/openhwgroup/cv32e40p#contributing

suppamax (approved):

PR #2240, opened 2023-10-13
fixing Interrupt wiring, imperas wiring, and updating to hash 9e0615b

LeeHoff:
fixing MIE wiring and updating to hash 9e0615b to get newest interrupt config.

### DIFF
--- a/cv32e20/bsp/handlers.S
+++ b/cv32e20/bsp/handlers.S
@@ -49,6 +49,7 @@
 .global m_fast15_irq_handler
 .global m_nmi_irq_handler
 
+.weak __no_irq_handler
 .weak u_sw_irq_handler
 .weak m_software_irq_handler
 .weak m_timer_irq_handler
--- a/cv32e20/bsp/handlers.S
+++ b/cv32e20/bsp/handlers.S
@@ -47,6 +47,7 @@
 .global m_fast13_irq_handler
 .global m_fast14_irq_handler
 .global m_fast15_irq_handler
+.global m_nmi_irq_handler
 
 .weak u_sw_irq_handler
 .weak m_software_irq_handler
@@ -68,7 +69,7 @@
 .weak m_fast13_irq_handler
 .weak m_fast14_irq_handler
 .weak m_fast15_irq_handler
-
+.weak m_nmi_irq_handler
 
 /* exception handling */
 __no_irq_handler:
@@ -131,6 +132,9 @@ m_fast14_irq_handler:
 	j __no_irq_handler
 
 m_fast15_irq_handler:
+	j __no_irq_handler
+
+m_nmi_irq_handler:
 	j __no_irq_handler
 
 u_sw_irq_handler:
--- a/cv32e20/sim/ExternalRepos.mk
+++ b/cv32e20/sim/ExternalRepos.mk
@@ -12,7 +12,7 @@ export SHELL = /bin/bash
 
 CV_CORE_REPO   ?= https://github.com/openhwgroup/cve2
 CV_CORE_BRANCH ?= main
-CV_CORE_HASH   ?= 00f7eb1
+CV_CORE_HASH   ?= 9e0615b
 CV_CORE_TAG    ?= none
 
 RISCVDV_REPO    ?= https://github.com/google/riscv-dv
--- a/cv32e20/tests/programs/custom/hello-world/hello-world.c
+++ b/cv32e20/tests/programs/custom/hello-world/hello-world.c
@@ -28,14 +28,14 @@
 #include <stdlib.h>
 
 // MVENDORID CSR: 0x602 is the value assigned by JEDEC to the OpenHW Group
-#define EXP_MVENDORID 0x00000000
-//#define EXP_MVENDORID 0x00000602
+//#define EXP_MVENDORID 0x00000000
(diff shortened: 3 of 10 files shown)


cv32e20/tests/programs/custom/hello-world/hello-world.c
```diff
@@ -28,14 +28,14 @@
 #include <stdlib.h>
 
 // MVENDORID CSR: 0x602 is the value assigned by JEDEC to the OpenHW Group
-#define EXP_MVENDORID 0x00000000
-//#define EXP_MVENDORID 0x00000602
+//#define EXP_MVENDORID 0x00000000
```
MikeOpenHWGroup, line 31: In a future PR it would be good to remove these commented-out EXP_ values as it might create confusion.  It is not worth gating this PR.

MikeOpenHWGroup (commented):
This all looks good to me @LeeHoff (with one minor comment on `debug_test.c` which should not gate a merge.

This change passes local CI under VCS, but not Questasim (vsim).  Investigating...

cv32e20/sim/ExternalRepos.mk
```diff
@@ -12,7 +12,7 @@ export SHELL = /bin/bash
 
 CV_CORE_REPO   ?= https://github.com/openhwgroup/cve2
 CV_CORE_BRANCH ?= main
-CV_CORE_HASH   ?= 00f7eb1
+CV_CORE_HASH   ?= 9e0615b
```
MikeOpenHWGroup, line 15: This version of the core has an additional interrupt input.  The testbench does not currently drive this.

MikeOpenHWGroup (changes requested):
I spoke too soon, this PR cannot be merged as-is.  An update to `cv32e20/tb/uvmt/uvmt_cv32e20_dut_wrap.sv` is required.  In that file we have:
```
  // Interrupt inputs
         .irq_software_i         ( irq_uvma[3]),
         .irq_timer_i            ( irq_uvma[7]),
         .irq_external_i         ( irq_uvma[11]),
         .irq_fast_i             ( irq_uvma[30:16]),
         .irq_nm_i               ( irq_uvma[31]),       // non-maskeable interrupt
```
As of commit `9e0615b` for the cv32e20 core, `irq_fast_i` is now 16-bits, not 15-bits.  The above code means that `irq_fast_i[15]` is left unconnected.  This causes Questa (vsim) simulations to fail due to an assertion check in the RTL `cve2_top.sv` ([link](https://github.com/openhwgroup/cve2/blob/9e0615bc3b9503152e1a223794eb996f5b469cea/rtl/cve2_top.sv#L270)).

I propose using `irq_uvma[0]` for the `irq_nm_i`, which would give us this:
```
  // Interrupt inputs
         .irq_software_i         ( irq_uvma[3]),
         .irq_timer_i            ( irq_uvma[7]),
         .irq_external_i         ( irq_uvma[11]),
         .irq_fast_i             ( irq_uvma[31:16]),
         .irq_nm_i               ( irq_uvma[0]),       // non-maskeable interrupt
```
Currently, the enivronment does not drive irq_uvma[0] (other than to set it to 0 at reset), so that capability will need to be added to the virtual peripheral.  For now, it allows all test-programs in the `ci_check` to pass cleanly under all simulators.

cv32e20/tests/programs/custom/hello-world/hello-world.c
LeeHoff: I will revise this PR and remove the commented out values.

cv32e20/sim/ExternalRepos.mk
LeeHoff: I will implement your suggested change to the interrupts to drive the extra interrupt input.

LeeHoff (commented):
Test out these changes against what you expected.

cv32e20/bsp/handlers.S
```diff
@@ -68,7 +69,7 @@
 .weak m_fast13_irq_handler
 .weak m_fast14_irq_handler
 .weak m_fast15_irq_handler
-
+.weak m_nmi_irq_handler
```
MikeOpenHWGroup, line 72: The `debug_test` wants to overload `__no_irq_handler`, so please add the following line below this one:
```
.weak __no_irq_handler
```

MikeOpenHWGroup (changes requested):
Ah, _much_ better @LeeHoff, thanks.  With these changes I have a local CI regression passing with both VCS and VSIM, and the interrupt_test runs and passes as well.

Unfortunately, the debug_test test-program cannot compile because it needs to overload `__no_irq_handler`.  See my comment on `cv32e20/bsp/handlers.S`.

cv32e20/bsp/handlers.S
LeeHoff: I put this line in at a different spot (line 52).  It works for me. I must not have simulated the debug_test prior to the last change.  It works for me now.

MikeOpenHWGroup (approved):
Nice!  With the latest update to `cv32e20/bsp/handler.S`, a local CI regression (sans ImperasDV) is now working with both VCS and VSIM.  Also, manually invoked `debug_test` and `interrupt_test` also passing with VCS and VSIM (again, sans ImperasDV).

Thanks for playing ball @LeeHoff!

PR #2252, opened 2023-10-18
Atomic sanity test

silabs-krdosvik:
1) Add exokay obi behavior
2) Out comment atomic uvm error

ci check: passes 
formal: compiles

This PR is connected to: https://github.com/openhwgroup/cv32e40x-dv/pull/49

Touching UVM code. I think what I have done is correct, but would appreciate if reviewers took a closer look.

### DIFF
--- a/lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
@@ -245,7 +245,17 @@ class uvma_obi_memory_cfg_c extends uvm_object;
    /**
     * Calculate a random atomic exokay response from random knobs
     */
-   extern function bit calc_random_exokay(bit[31:0] addr);
+   extern function bit calc_random_exokay(bit[31:0] addr, bit is_SCW);
+
+   /**
+    * Invalidate RL.W reservation set
+    */
+   extern function void invalidate_reservation();
+
+   /**
+    * Set RL.W reservation set
+    */
+   extern function void set_reservation(bit[31:0] addr, int nr_words_reserved);
 
    /**
     * Returns 1 if this OBI agent supports version 1.2 or higher
@@ -271,6 +281,8 @@ function uvma_obi_memory_cfg_c::new(string name="uvma_obi_memory_cfg");
    else if ($test$plusargs("rvalid_singles_stall")) begin
       rvalid_singles_stall = 1;
    end
+
+   directed_slv_exokay_valid = 0;
 
 endfunction : new
 
@@ -353,14 +365,15 @@ function bit uvma_obi_memory_cfg_c::calc_random_err(bit[31:0] addr);
 
 endfunction : calc_random_err
 
-function bit uvma_obi_memory_cfg_c::calc_random_exokay(bit[31:0] addr);
+function bit uvma_obi_memory_cfg_c::calc_random_exokay(bit[31:0] addr, bit is_SCW);
 
    bit exokay;
 
    // Check for a directed error reponse first
-   if (directed_slv_exokay_valid &&
-       (addr <= directed_slv_exokay_addr_min) &&
-       (addr <= directed_slv_exokay_addr_min)) begin
+   if (is_SCW &&
+      (!directed_slv_exokay_valid ||
+      ((addr < directed_slv_exokay_addr_min) ||
+      (addr > directed_slv_exokay_addr_max)))) begin
       return 0;
    end
 
@@ -377,6 +390,22 @@ function bit uvma_obi_memory_cfg_c::calc_random_exokay(bit[31:0] addr);
    return exokay;
 
 endfunction : calc_random_exokay
+
+function void uvma_obi_memory_cfg_c::invalidate_reservation();
+
+   directed_slv_exokay_valid = 0;
+
+endfunction : invalidate_reservation
+
+function void uvma_obi_memory_cfg_c::set_reservation(bit[31:0] addr, int nr_words_reserved);
+
+   if (addr[1:0] == 2'b0) begin
+      directed_slv_exokay_valid = 1;
+      directed_slv_exokay_addr_min = addr;
+      directed_slv_exokay_addr_max = addr + 4*(nr_words_reserved-1);
+   end
+
+endfunction : set_reservation
 
 function bit uvma_obi_memory_cfg_c::is_1p2_or_higher();
 
--- a/lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
@@ -93,6 +93,11 @@ class uvma_obi_memory_cfg_c extends uvm_object;
    bit [31:0]                                    directed_slv_exokay_addr_max;
    bit                                           directed_slv_exokay_valid;
 
+   bit [31:0]                                    directed_slv_reservation_addr_min;
+   bit [31:0]                                    directed_slv_reservation_addr_max;
+   bit                                           directed_slv_reservation_valid;
+   bit [31:0]                                    directed_slv_nr_reserved_words = 32'b1;
+
 
    `uvm_object_utils_begin(uvma_obi_memory_cfg_c)
       `uvm_field_int (                         enabled          , UVM_DEFAULT)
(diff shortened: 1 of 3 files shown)


lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-robin, line 1: > ci check: passes except pma_2

Does regression history say if it used to pass before?
(I cannot remember this failing before.)
(Also, "pma_2" is the config name and not the test name.)

What is failing?
ISS? Assert? Self-check?

silabs-robin, line 1: Why is the PR a draft?

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
```diff
@@ -245,7 +245,17 @@ class uvma_obi_memory_cfg_c extends uvm_object;
    /**
     * Calculate a random atomic exokay response from random knobs
     */
-   extern function bit calc_random_exokay(bit[31:0] addr);
```
silabs-robin, line 248: I see this was here before, but I wonder why utility functions (or driver functions?) are part of the configuration class.
Config, to me, means "this is enabled, that is disabled, some var is tuned to this value, etc".
But I guess it is convenient because the config object is often available.

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
```diff
@@ -353,14 +365,15 @@ function bit uvma_obi_memory_cfg_c::calc_random_err(bit[31:0] addr);
...
 
    // Check for a directed error reponse first
-   if (directed_slv_exokay_valid &&
-       (addr <= directed_slv_exokay_addr_min) &&
-       (addr <= directed_slv_exokay_addr_min)) begin
+   if (is_SCW &&
+      (!directed_slv_exokay_valid ||
+      ((addr < directed_slv_exokay_addr_min) ||
+      (addr > directed_slv_exokay_addr_max)))) begin
       return 0;
```
silabs-robin, line 377: This looks strange to me (tho I'm not familiar with the existing code).

The old code had a condition for aborting the function.
That condition is now narrower.
So several of the cases that were previously reasons to abort, will no longer abort the function.

Is that how it should be?

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
```diff
@@ -493,11 +493,11 @@ function void uvma_pma_sb_c::check_obi_d_mapped_region(uvma_obi_memory_mon_trn_c
...
-      if (obi.atop) begin
-         `uvm_error("PMAOBID", $sformatf("OBI D %s address: 0x%08x, region: %0d atop is not zero, OBI: 0x%0x",
-                                          obi.access_type.name(), obi.address, index,
-                                          obi.atop));
-      end
+      //if (obi.atop) begin
+      //   `uvm_error("PMAOBID", $sformatf("OBI D %s address: 0x%08x, region: %0d atop is not zero, OBI: 0x%0x",
+      //                                    obi.access_type.name(), obi.address, index,
+      //                                    obi.atop));
+      //end
```
silabs-robin, line 500: If this is commented away, it should maybe either be deleted or have a TODO comment?

lib/uvm_agents/uvma_obi_memory/src/seq/uvma_obi_memory_slv_base_seq.sv
```diff
@@ -165,13 +165,21 @@ endfunction : add_err
 
 function void uvma_obi_memory_slv_base_seq_c::add_exokay(uvma_obi_memory_mon_trn_c mon_req, uvma_obi_memory_slv_seq_item_c slv_rsp);
 
-   // Only respond exokay == 1 to SC or LR as signaled by atop
+   int nr_words_reserved = 1;
```
silabs-robin, line 168: When is `nr_words_reserved` different from `1`?

silabs-krdosvik: Good question. Atomic spec state that the reservation set can any size, so the size I chose was 1. But is should be able to change, which makes me think that this configuration should probably be somewhere else. (But where?? Hem.. I must think about that, unless someone already knows).

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-krdosvik: I see, hehe, thanks.
I have not seen it failing either.
The problem seems to have been some system problem, and now after running it again it passes.

silabs-krdosvik: Because ci check was not fully green without me knowing the problem, bit now that it is solved I will open it again.

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-robin: My bad. I didn't notice the details of how the old code was modified.

I still don't understand it, but since the new sanity check has self-checking that gives sufficient confidence IMO.

silabs-krdosvik: Good point. I might have misunderstood too. For me it would make sense if exokay is always cleared if the SCW dont write to reservation region. I have done some changes to fulfill that interpretation, but I seems this code is based on some other interpretation, I am not 100% sure.

lib/uvm_agents/uvma_obi_memory/src/seq/uvma_obi_memory_slv_base_seq.sv
silabs-robin: I just caught my eye, since we are adding a new feature but not really having a use for it.

But I'll just resolve this comment because the code seems to be working and now we have the feature if needed later.

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
```diff
@@ -353,14 +365,15 @@ function bit uvma_obi_memory_cfg_c::calc_random_err(bit[31:0] addr);
...
    bit exokay;
 
    // Check for a directed error reponse first
-   if (directed_slv_exokay_valid &&
-       (addr <= directed_slv_exokay_addr_min) &&
-       (addr <= directed_slv_exokay_addr_min)) begin
+   if (is_SCW &&
+      (!directed_slv_exokay_valid ||
+      ((addr < directed_slv_exokay_addr_min) ||
+      (addr > directed_slv_exokay_addr_max)))) begin
```
silabs-hfegran, line 376: is this address range supposed to be non-inclusive?

silabs-hfegran:
I think this looks good, though I do not fully understand all the code

silabs-krdosvik:
> I think this looks good, though I do not fully understand all the code

Let me know if there is something you want me to explain :-)

lib/uvm_agents/uvma_obi_memory/src/seq/uvma_obi_memory_slv_base_seq.sv
silabs-krdosvik: Yeah. Btw. the feature is needed in this PR: https://github.com/openhwgroup/cv32e40x-dv/pull/49

silabs-robin: My feedback is resolved.

silabs-krdosvik: Fixed in new commit

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
silabs-krdosvik: Fixed in new commit

silabs-robin (approved):
Approving based on:
Looks like a good addition.
Test status good.

PR #2253, opened 2023-10-18
Directed Test - Pushpop Debug Triggers

silabs-robin:
This PR adds a directed sanity test for Zc push/pop triggering debug entry.

Test status:
* The new test - Passes.
* ci_check - N/A.
* "full" - Launches and passes the new test.

### DIFF
--- a/cv32e40s/bsp/bsp.h
+++ b/cv32e40s/bsp/bsp.h
@@ -1,0 +1,44 @@
+// Copyright 2022 Silicon Laboratories Inc.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+//
+// Licensed under the Solderpad Hardware License v 2.1 (the "License"); you
+// may not use this file except in compliance with the License, or, at your
+// option, the Apache License version 2.0.
+//
+// You may obtain a copy of the License at
+// https://solderpad.org/licenses/SHL-2.1/
+//
+// Unless required by applicable law or agreed to in writing, any work
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+typedef union {
+  struct {
+    volatile uint32_t  load     : 1;
+    volatile uint32_t  store    : 1;
+    volatile uint32_t  execute  : 1;
+    volatile uint32_t  u        : 1;
+    volatile uint32_t  s        : 1;
+    volatile uint32_t  res_5_5  : 1;
+    volatile uint32_t  m        : 1;
+    volatile uint32_t  match    : 4;
+    volatile uint32_t  chain    : 1;
+    volatile uint32_t  action   : 4;
+    volatile uint32_t  size     : 4;
+    volatile uint32_t  timing   : 1;
+    volatile uint32_t  select   : 1;
+    volatile uint32_t  hit      : 1;
+    volatile uint32_t  vu       : 1;
+    volatile uint32_t  vs       : 1;
+    volatile uint32_t  res_26_25: 2;
+    volatile uint32_t  dmode    : 1;
+    volatile uint32_t  type     : 4;
+  } __attribute__((packed)) volatile fields;
+  volatile uint32_t raw;
+} __attribute__((packed)) mcontrol6_t;
+
--- a/cv32e40s/regress/cv32e40s_full.yaml
+++ b/cv32e40s/regress/cv32e40s_full.yaml
@@ -386,12 +386,6 @@ tests:
     dir: cv32e40s/sim/uvmt
     cmd: make test TEST=debug_test_reset
 
-  debug_test_trigger:
-    description: Debug trigger test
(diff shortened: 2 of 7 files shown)


cv32e40s/regress/cv32e40s_full.yaml
```diff
@@ -386,12 +386,6 @@ tests:
     dir: cv32e40s/sim/uvmt
     cmd: make test TEST=debug_test_reset
 
-  debug_test_trigger:
-    description: Debug trigger test
-    builds: [ uvmt_cv32e40s, uvmt_cv32e40s_clic]
-    dir: cv32e40s/sim/uvmt
-    cmd: make test TEST=debug_test_trigger
```
silabs-robin, line 393: Note: This was overwritten further down.

cv32e40s/tests/programs/custom/debug_test_trigger/test.yaml
```diff
@@ -1,6 +1,3 @@
-# Test definition YAML for test
-
-# Debug trigger directed test
```
silabs-robin, line 3: Note: These lines were just useless.

cv32e40s/regress/cv32e40s_full.yaml
```diff
@@ -526,6 +520,12 @@ tests:
     cmd: make test TEST=zc_test
     num: 5
 
+  pushpop_debug_triggers:
+    description: Zc push/pop vs debug triggers
+    builds: [ uvmt_cv32e40s_clic ]
+    dir: cv32e40s/sim/uvmt
+    cmd: make test TEST=pushpop_debug_triggers
```
silabs-robin, line 527: Note: Adding this new test to the "full" regression.

cv32e40s/regress/cv32e40s_full.yaml
silabs-krdosvik: What do you mean overwritten further down? Does it mean that this test was not run? And that it only run with the uvmt_cv32e40s_debug_trigger_cfgs, and not uvmt_cv32e40s_clic? I am not really understanding it...

cv32e40s/bsp/bsp.h
```diff
@@ -0,0 +1,44 @@
...
+    volatile uint32_t  select   : 1;
+    volatile uint32_t  hit      : 1;
+    volatile uint32_t  vu       : 1;
+    volatile uint32_t  vs       : 1;
+    volatile uint32_t  res_26_25: 2;
+    volatile uint32_t  dmode    : 1;
+    volatile uint32_t  type     : 4;
+  } __attribute__((packed)) volatile fields;
+  volatile uint32_t raw;
+} __attribute__((packed)) mcontrol6_t;
```
silabs-hfegran, line 43: should we perhaps put this in some other file in the bsp folder rather than including it directly in bsp.h; otherwise this will get very cluttered very fast (previously these headers have generally been added to the tests directly)

silabs-hfegran: nevermind the first comment - I realized this is a new file

silabs-hfegran (approved):
Looks good to me

cv32e40s/tests/programs/custom/pushpop_debug_triggers/pushpop_debug_triggers.c
```diff
@@ -0,0 +1,287 @@
+// Copyright 2023 Silicon Labs, Inc.
```
silabs-krdosvik, line 1: Looks good. At first sight it seems complexed, but when looking deeper into it, it seems like it has the just the complexity it needs (and that it is not that complexed) :-) .

silabs-krdosvik (approved):
Looks good. At first sight it seems complexed, but when looking deeper into it, it seems like it has the just the complexity it needs (and that it is not that complexed) :-) .

cv32e40s/regress/cv32e40s_full.yaml
silabs-robin: > What do you mean overwritten further down?

Further down, in the same file, that test was defined twice.

The second definition takes precedence, so the first one is superfluous (and also erronous).

bilde

silabs-robin: > And that it only run with the uvmt_cv32e40s_debug_trigger_cfgs, and not uvmt_cv32e40s_clic? I am not really understanding it...

I found the problem like this:
I wanted to run the test, so I confered the "full" yaml to see what config to run with.
Then it failed.
It can seemingly not run with those configs, it needs those configs who are in the second definition.

cv32e40s/tests/programs/custom/pushpop_debug_triggers/pushpop_debug_triggers.c
```diff
@@ -0,0 +1,287 @@
...
+// may not use this file except in compliance with the License, or, at your
+// option, the Apache License version 2.0.
+//
+// You may obtain a copy of the License at
+// https://solderpad.org/licenses/SHL-2.1/
+//
+// Unless required by applicable law or agreed to in writing, any work
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
```
silabs-robin, line 15: woops

PR #2254, opened 2023-10-23
Added flag to prevent xcelium from searching for cds.lib/hdl.var

silabs-hfegran:
… unless explicitly specified

- Prevents slow elaboration when library files are not available


silabs-robin (approved):
Approving based on:
I am familiar with the problem that this PR addresses. Nice to see it get fixed.

PR #2255, opened 2023-10-23
Updated wfe test to work correctly without user-mode/PMP, added flags…

silabs-hfegran:
… to fix ISS coverage issue

- Fixed: Wfe test did not support running without PMP/User mode
- ~Added: --excludem RISCV_CSR_UNIMP flag as per imperas' instructions to avoid errors when sampling coverage for illegal CSR instructions (w/ non-implemented CSRs)~


silabs-robin (approved):
Approving based on:
It looks like this will make the test passable on the X.

PR #2260, opened 2023-10-25
Added zcmt directed test (40S only), moved structs to bsp

silabs-hfegran:
* Initial commit to close some obvious mstateen0/umode/zcmt coverage holes (not relevant for 40X)
* Moved commonly used structs to bsp.h
* Cleanup in wfe and pmp_csr_access tests

### DIFF
--- a/cv32e40s/tests/programs/custom/zcmt_test/zcmt_test.c
+++ b/cv32e40s/tests/programs/custom/zcmt_test/zcmt_test.c
@@ -33,7 +33,7 @@
 // status, thus we are limited to 31 tests with this construct.
 #define NUM_TESTS 11
 // Set which test index to start testing at (for quickly running specific tests during development)
-#define START_TEST_IDX 5
+#define START_TEST_IDX 0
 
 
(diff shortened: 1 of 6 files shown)


silabs-hfegran:
Changed tests pass OK with configurations specified in cv32e40s_full.yaml regression list

cv32e40s/tests/programs/custom/zcmt_test/zcmt_test.c
silabs-robin, line 1: @silabs-mateilga I think you know Zc better than me. Could you have a look as well?

silabs-mateilga (approved):
The moves to bsp.h seems reasonable. The zcmt_test also looks good, I can't really see anything to comment on, but the high complexity could very well hide problems that are hard to see by review. However, the test seems to be checking things we need covered for zcmt, so assuming ci_check passes, I'm approving this PR.

silabs-robin (approved):
Approving based on:
Marton's approval.
Test results/status.

PR #2261, opened 2023-10-26
Enable toggle coverage only for ports

silabs-hfegran:


cv32e40s/sim/tools/xrun/covfile.tcl
silabs-robin, line 1: Does this get run when using COV=1?

silabs-hfegran: yes

silabs-robin (approved):
Approving based on:
Looks like a nice way to stop the simulator from collecting data we don't care about.

PR #2265, opened 2023-10-30
Fixed pmp-assertion

silabs-hfegran:


silabs-robin (approved):
Approving based on:
Offline discussion.

PR #2266, opened 2023-10-31
Updated pushpop_debug_triggers with sticky debug request

silabs-hfegran:
Fixed: Debug request was short pulse, could in some instances be missed. Changed to sticky debug req, where the request is disabled from debug handler


silabs-robin (approved):
Approving based on:
There was a bug (infinite loop on certain seeds).
These changes rectify that bug.

PR #2268, opened 2023-10-31
Updated 40s core hash

silabs-hfegran:
Update to latest hash

ci_check ok


silabs-robin (approved):
Approving based on:
Hash updates are usually good.
ci_check reported to be ok.

PR #2272, opened 2023-11-01
Updating interrupt test and Imperas integration.

LeeHoff:
Interim changes to interrupt_test to remove non-supported interrupt features.
Update to Imperas nmi connection. and attempt to configure default.yaml file correctly.

### DIFF
--- a/cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
+++ b/cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
@@ -272,7 +272,7 @@ int main(int argc, char *argv[]) {
     if (retval != EXIT_SUCCESS)
         return retval;
 
-// unused tests. to be removed at clean-up 
+// TODO unused tests. to be removed at clean-up 
 /*    // Repeat test1 (restore vector mode)
     retval = test7();
     if (retval != EXIT_SUCCESS)
@@ -619,9 +619,11 @@ int test6() {
 
 // Test 7 will repeat the basic interrupt test in test 1
 // But with a relocated vector table via mtvec CSR and DIRECT vector mode
-// CV32E20 does not support DIRECT mode.  Saved here until clean up
-int test7() {
+// CV32E20 does not support DIRECT mode.  
+// TODO Saved here until clean up
 /*
+int test7() {
+
     volatile uint32_t save_mtvec;
     int retval;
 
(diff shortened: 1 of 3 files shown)


cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
```diff
@@ -272,7 +272,8 @@ int main(int argc, char *argv[]) {
     if (retval != EXIT_SUCCESS)
         return retval;
 
-    // Repeat test1 (restore vector mode)
+// unused tests. to be removed at clean-up
```
MikeOpenHWGroup, line 275: Recommend adding the string "TODO" in this comment.  Most editors will automatically highlight it and it also makes it easier to grep code to find what needs to be cleaned-up.

cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
```diff
@@ -651,14 +617,56 @@ int test8() {
     return EXIT_SUCCESS;
 }
 
+// Test 7 will repeat the basic interrupt test in test 1
+// But with a relocated vector table via mtvec CSR and DIRECT vector mode
+// CV32E20 does not support DIRECT mode.  Saved here until clean up
+int test7() {
```
MikeOpenHWGroup, line 623: As far as I can tell, this function is never called.  Nevertheless, it is a integer function that does not return a value which can be a dangerous practise.  I recommend either commenting it out in it's entirety, or returning `EXIT_SUCCESS`.  In the future, if someone mistakenly calls this function, that will minimize the likelihood of unexpected behavior in the callee.

MikeOpenHWGroup (changes requested):
Hi @LeeHoff, this all looks good to me and passes a local CI regression with both VCS and Questa.  The updated interrupt_test (not yet part of the CI) also passes locally.  Check out my comments on `interrupt_test.c`, there are a couple of "maintainability" issues that I flagged.

cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
LeeHoff: I have added TODOs and moved the comment so it comments out the whole function.

cv32e20/tests/programs/custom/interrupt_test/interrupt_test.c
```diff
@@ -651,14 +617,56 @@ int test8() {
...
+    active_test = 8;
+
+    if (test1_impl(0) != EXIT_SUCCESS)
+        return ERR_CODE_TEST_8;
+
+    return EXIT_SUCCESS;
+ */
+ }
+
+// CV32E20 - this test appears to be a coverage filler test. Will wait until we see initial coverage numbers
```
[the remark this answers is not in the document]
LeeHoff: Done.

PR #2275, opened 2023-11-02
Fixed isa-decoder bitfield issues + made functions automatic

silabs-hfegran:
* Fixed: Two functions had wrong bit-widths specified
* Changed functions to automatic

Formal/Sim OK

### DIFF
--- a/lib/isa_decoder/isa_decoder.sv
+++ b/lib/isa_decoder/isa_decoder.sv
@@ -23,7 +23,7 @@
   // ---------------------------------------------------------------------------
   // Stack_adj for zcmp instructions
   // ---------------------------------------------------------------------------
-  function int get_stack_adj( rlist_t rlist, logic[5:4] spimm);
+  function automatic int get_stack_adj( rlist_t rlist, logic[5:4] spimm);
     int stack_adj_base;
     int stack_adj;
 
@@ -43,7 +43,7 @@
   // ---------------------------------------------------------------------------
   // Non-trivial immediate decoder
   // ---------------------------------------------------------------------------
-  function logic [20:1] get_sort_j_imm(instr_t instr);
+  function automatic logic [20:1] get_sort_j_imm(instr_t instr);
     get_sort_j_imm = {
       instr.uncompressed.format.j.imm[31],
       instr.uncompressed.format.j.imm[21:12],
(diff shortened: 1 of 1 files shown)


lib/isa_decoder/isa_decoder.sv
```diff
@@ -86,7 +86,7 @@
     };
   endfunction : get_sort_ci_imm_addi16sp
 
-  function logic [8:0] get_sort_cb_imm_not_sequential(instr_t instr);
+  function automatic logic [7:0] get_sort_cb_imm_not_sequential(instr_t instr);
```
silabs-robin, line 89: The CB format (as used by branch instructions) is actually `8:1`, but since 1) the CB format is "used" by other instructions that don't use it like 8:1, and 2) 7:0 should be(?) assign-compatible to 8:1 wherever it matters, I think this function signature might be alright.

bilde

lib/isa_decoder/isa_decoder.sv
```diff
@@ -23,7 +23,7 @@
   // ---------------------------------------------------------------------------
   // Stack_adj for zcmp instructions
   // ---------------------------------------------------------------------------
-  function int get_stack_adj( rlist_t rlist, logic[5:4] spimm);
+  function automatic int get_stack_adj( rlist_t rlist, logic[5:4] spimm);
```
silabs-robin, line 26: Automatic is good. Should maybe have been the default.

lib/isa_decoder/isa_decoder.sv
```diff
@@ -96,7 +96,7 @@
     };
   endfunction : get_sort_cb_imm_not_sequential
 
-  function logic [5:0] get_sort_cj_imm(instr_t instr);
+  function automatic logic [10:0] get_sort_cj_imm(instr_t instr);
```
silabs-robin, line 99: CJ is `11:1` in the spec. I'm not sure if there will be any problems with this, but it reminds me of the kinds of bugs we have seen in isacov.

bilde

silabs-hfegran: Yeah, this is the eternal discussion on how to handle these bits - regardless, the function needs to at least return the correct amount of bits

lib/isa_decoder/isa_decoder.sv
silabs-hfegran: Yeah, I would like to use the offset-values, but 8:1 is in non-strongly typed terms just a 8-bit vector 7:0, so either we need to add the low bits which would not be the raw value, or add it and already have an interpreted type.

silabs-robin: I think it would be better to write the same indices as the spec, so that everything is 1-1 and we don't have to maintain a separate system from the spec.
But since they are both 8-bit vectors I think the risk for bugs is low enough, and I don't know the full score of consequences, so I would be ok with either outcome.

silabs-robin: I guess... If isacov had a `8:1` signal, being assigned from this `7:0` signal like `a[8:1] = b[8:1]` there would be an error? So any potential risk that a conventions-mismatch introduces, will be automatically identified?

silabs-hfegran: I like the idea of this being consistent, which it can be for non-compressed. But take the ci-type for example 
Screenshot 2023-11-03 at 2 06 56 PM
how should these bits be represented?

silabs-robin: Darn, that complicates it. `[N:0]` seems ok.

silabs-robin (approved):
Approving based on:
(Reasoning given in in-line comments.)

PR #2278, opened 2023-11-03
Removed dret to umode clears mintthresh assertion (no longer req)

silabs-hfegran:
Spec no longer has this requirement. 
vplan update in later commit


silabs-mateilga (approved):
Looks good

silabs-robin (approved):
Approving based on:
Marton's approval.

PR #2279, opened 2023-11-06
Added more linked coverage to U-mode vplan

silabs-hfegran:


silabs-robin (approved):
Approving based on:
This fills in some gaps in umode link2cov.
The linked tests seems appropriate given the verification goals.

PR #2280, opened 2023-11-06
CV32E40P corev-dv debug gen fixes & HWLoop coverage

XavierAubert:
This PR brings up two modifications. 

First a contribution from @dd-vaibhavjain targetting corev-dv support for debug trigger and updates illegal instruction generation. More details can be found here XavierAubert#76

The second part contain @dd-baoshan contributions, first the addition of rvvi interface for coverage collection (see XavierAubert#79 for details) and following this, addition of hwloop coverage (refer to XavierAubert#81).


cv32e40p/env/corev-dv/cv32e40p_asm_program_gen.sv
```diff
@@ -550,6 +550,9 @@ class cv32e40p_asm_program_gen extends corev_asm_program_gen;
 
   // Override Illegal instruction handler - gen_illegal_instr_handler
   // Replace pop_gpr_from_kernel_stack with pop_regfile_from_kernel_stack
+  // With RV32X enabled, check for illegal instr on the last instr of hwloop
```
MikeOpenHWGroup, line 553: Is `RV32X` the proposed graphics ISA?  AFAIK, CV32E40P does not support this ISA.

cv32e40p/env/corev-dv/cv32e40p_asm_program_gen.sv
```diff
@@ -558,11 +561,38 @@ class cv32e40p_asm_program_gen extends corev_asm_program_gen;
 
     gen_signature_handshake(instr, CORE_STATUS, ILLEGAL_INSTR_EXCEPTION);
     gen_signature_handshake(.instr(instr), .signature_type(WRITE_CSR), .csr(MCAUSE));
-    instr = {instr,
-            $sformatf("csrr  x%0d, 0x%0x", cfg.gpr[0], MEPC),
-            $sformatf("addi  x%0d, x%0d, 4", cfg.gpr[0], cfg.gpr[0]),
-            $sformatf("csrw  0x%0x, x%0d", MEPC, cfg.gpr[0])
-    };
+    if (riscv_instr_pkg::RV32X inside {riscv_instr_pkg::supported_isa}) begin
+      instr = {instr,
```
MikeOpenHWGroup, line 565: A more detailed comment about what is happening here would be useful.

cv32e40p/env/corev-dv/cv32e40p_debug_rom_gen.sv
```diff
@@ -167,6 +178,169 @@ class cv32e40p_debug_rom_gen extends riscv_debug_rom_gen;
...
+    // iterations and overlaps its usage for single step routine as well
+    // So this input is to differentiate code generation for different
+    // scenrios with or without single step overlap in the tests
+    //
+    // Description: Setup trigger breakpoint setup with instruction address
+    // match. Uses random config item trigger_addr_offset to set trigger
+    // address adding this random offset to depc value
+    // Use config trigger_iterations with use_dscratch0=1 to setup trigger
+    // multiple times in a single test.
+    virtual function void gen_dbg_trigger_setup_with_addr_match(int use_dscratch0=0);
```
MikeOpenHWGroup, line 193: This is a lot of complexity to verify this feature.  This code-segment looks very much like "production code" for handling address match breakpoints.  Do we really need to do all of this?

cv32e40p/env/corev-dv/cv32e40p_instr_sequence.sv
```diff
@@ -275,11 +277,48 @@ class cv32e40p_instr_sequence extends riscv_instr_sequence;
...
     prefix = format_string($sformatf("%0d:", i), LABEL_STR_LEN);
     if(!is_main_program) begin
       generate_return_routine(prefix);
     end
   endfunction
 
+  // Similar to base class insert_illegal_hint_instr function
+  // Added cv32_illegal_instr.cv32e40p_init() to include cv32e40p
+  // specific legal opcodes
+  function void cv32e40p_insert_illegal_hint_instr();
```
MikeOpenHWGroup, line 290: I'll also need some help with this one.

cv32e40p/env/uvme/cov/uvme_rv32x_hwloop_covg.sv
```diff
@@ -0,0 +1,920 @@
+///////////////////////////////////////////////////////////////////////////////
```
MikeOpenHWGroup, line 1: This is great!  I wonder if we should consider moving it to ISACOV?  Do you have an opinion about that?

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
```diff
@@ -0,0 +1,624 @@
+// Copyright 2020 OpenHW Group
```
MikeOpenHWGroup, line 1: This looks like a copy-n-paste of [cv32e40p_tracer_pkg.sv](https://github.com/openhwgroup/cv32e40p/blob/e70418e0139ef4ab65a4b9dc691ac23f920fbe7d/bhv/include/cv32e40p_tracer_pkg.sv#L1), is that right?

MikeOpenHWGroup (commented):
This PR looks (mostly) good to me @XavierAubert and I see no reason _not_ to merge it.  It is passing a local CI regression under both VCS and Questasim.  However, before merging I have a few questions and it will be a lot easier to discuss them "in person" at our weekly meeting.  Please see my comments.

Also, this was done months ago, but I noticed that `cv32e40p/env/uvme/vseq/uvme_cv32e40p_random_debug_vseq.sv` has two classes in a single file.  Can we fix that in a subsequent PR?

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
dd-baoshan: Yes, there is also note inside this file as information on parameter references.

// note: 
// 1) below instructions parameters are mirror copy of cv32e40p_pkg.sv and cv32e40p_tracer_pkg.sv
// 2) cv32e40p_pkg.sv is available at /cv32e40p/rtl/include
// 3) cv32e40p_tracer_pkg.sv is available at /cv32e40p/bhv/include

cv32e40p/env/uvme/cov/uvme_rv32x_hwloop_covg.sv
dd-baoshan: I need to seek for your clarification related to ISACOV.  My understanding is current tb has uvme_rv32isa_covg.sv: which used to capture instructions related coverage. it uses uvmt_cv32e40p_isa_covg_if interface for coverage sampling.
May I know what is the meaning of "moving it to ISACOV" here?

cv32e40p/env/corev-dv/cv32e40p_asm_program_gen.sv
pascalgouedo: RV32X stands for PULP custom instructions.
CV32E40P is RV32IMC[F|Zfinx]X.

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
pascalgouedo: There should exist another way than a "mirror copy" which will immediately become out-of-date if one of the 2 packages is changed.
For sure this should not stay like that.

cv32e40p/env/corev-dv/cv32e40p_debug_rom_gen.sv
dd-vaibhavjain: Yes this is needed to be added in the corev-dv random test generator to support debug trigger verification.  

This is on same lines as how single step, and ebreak, which were already supported within this corev-dv. I have only extended this to also support trigger in same manner.  

Since corev-dv does not have toolchain injected code to manage all such configurations, the support needs to be added explicitly, like this logic in assembly instructions, to configure the relevant CSR.

This code is not enabled by default and needs an explicit config(plusargs) to be applied to enable this logic, so using or not is upto the test writers.
It does not alter/effect any previously existing test or any future test as long as the relevant configs are never applied.

cv32e40p/env/corev-dv/cv32e40p_instr_sequence.sv
dd-vaibhavjain: This function basically injects illegal instructions as part of the generated instruction stream to verify the behaviour of illegal instructions.

This is same as the one in base class, but I had to add another one explicitly here  to update the legal instruction space specific to CV32E40P v2 mainly for custom pulp Instructions. (And base class function is not declared virtual).

Previously without this function, the part of custom instruction opcode space (which is legal for CV32E40P due to custom PULP instructions) were injected as "illegal" instructions by the base class function, since the original function did not had any knowledge of legal custom instruction space specific to this core.

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
dd-vaibhavjain: Yes I agree with Pascal. This should not be a mirror copy of tracer_pkg.

I had requested for maintaining such defines for opcodes within core-v-verif repo's cv32e40p directory independently. (But not a plain copy)
I feel useasyto by copying it like this too. Though I support the idea of having such defines added explicitly in core-v-verif.

cv32e40p/env/uvme/cov/uvme_rv32x_hwloop_covg.sv
dd-vaibhavjain: Specifics of which file this code should exist in, I would leave for everyone to conclude. 
I just want to point out that this functional coverage is written on same lines and basis as that of ISACOV and relies entirely on RVFI, so it is written as if it was an extended part of ISACOV model but at the moment in a different class.

cv32e40p/env/corev-dv/cv32e40p_asm_program_gen.sv
dd-vaibhavjain: This part of logic is based on the comment above regarding HWLOOP with RV32X extension.

We encountered a problem in some random tests where the illegal instruction was placed at last instruction of a HWLOOP, and the instruction handler was not supporting such specific case for HWLOOPs, where if last instruction on hwloop was illegal,  the handler was just incrementing the MEPC value by 4, this caused hwloop to be exited (not-intended), rather than jump to 1st hwloop instruction next.

So this part of illegal handler code is added here to add a check for:
1. If the illegal exception happened within a hwloop.
2. If yes , check if the illegal was the last instruction of an hwloop with atleast 1 more iteration left.
3. If yes, then set MEPC to first instruction of hwloop instead of just incrementing by 4.
4. Else MEPC is just incremented by 4

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
MikeOpenHWGroup: I also agree with @pascalgouedo.  In the interest of time we can merge this in now and fix it in a future pull-request.

cv32e40p/env/uvme/cov/uvme_rv32x_hwloop_covg.sv
MikeOpenHWGroup: > May I know what is the meaning of "moving it to ISACOV" here?

The idea would be to move `uvme_rv32x_hwloop_covg.sv` to `lib/uvm_agents/uvma_isacov/cov`.  It could be available for any future cores that might want to implement the PULP ISA.

> this functional coverage is written on same lines and basis as that of ISACOV and relies entirely on RVFI

Excellent!  So it should be "easy" (haha!) to move the file to ISACOV.

We do not need to it now.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for answering all my question team.  Merging...

cv32e40p/env/uvme/uvme_cv32e40p_param_all_insn.sv
dd-baoshan: I can insert fixme note into this file for future fix. 

There are 2 options:

1)	Import tracer pkg and reuse all the defined insn parameters for coverage purpose. (This was my initial implementation)
a.	Pros: 
i.	one package to maintain for all instruction under supported standards.
ii.	this package is used in rvfi and tracer. It must be maintained from time to time.

2)	Define instruction parameters into a common package that used for core-v-verif. (Currently implementation based on feedback)

The reason I use the term “mirror copy” is because even though option (2) is chosen, all instruction parameters MUST have same assigned value due to specification. The only different is parameter naming.
The output would be:
a)	<DIFF params naming than tracer pkg> + <SAME assign value as tracer pkg>  in separate file
b)	<SAME params naming than tracer pkg> + <SAME assign value as tracer pkg> in separate file    this is mirror copy
As for me, To be easy cross checktracer pkg with meld or tkdiff , I choose option 2b.

@pascalgouedo , please advise which option (1 or 2) is your preference? if you have other options than 1 or 2, please also let me know.

pascalgouedo: Hi,
I would go to import packages from cv32e40P repo and add a separate core-v-verif package for any new core-v-verif specific definitions.

PR #2281, opened 2023-11-07
Coverage & decoder fixup

silabs-hfegran:
- Fixed rvfi.trap and rvfi.intr usage in exception covergroup
- Fixed outdated exception causes and support for clic in exception covergroup
- Fixed missing c.andi instruction field assignments in decoder
- Fixed rdrs1 assignment to check for valid register from decoder (fixes issue with c.jalr and c.jr in isacov)


cv32e40s/env/uvme/cov/uvme_exceptions_covg.sv
```diff
@@ -21,11 +21,11 @@ covergroup cg_exceptions
...
   `per_instance_fcov
 
-  cp_trap : coverpoint isacov.instr.rvfi.trap {
+  cp_trap : coverpoint isacov.instr.rvfi.trap.trap {
     bins trap = {1};
   }
-  cp_intr : coverpoint isacov.instr.rvfi.intr {
-    bins intr = {1};
+  cp_intr : coverpoint isacov.instr.rvfi.intr.interrupt {
+    bins interrupt = {1};
```
silabs-robin, line 28: Makes sense given the "new" rvfi_trap/_intr structs.

cv32e40s/env/uvme/cov/uvme_exceptions_covg.sv
```diff
@@ -41,15 +41,17 @@ covergroup cg_exceptions
   cp_no_ebreakm : coverpoint (isacov.instr.rvfi.csrs["dcsr"].get_csr_retirement_data()[15]) {
     bins no_ebreakm = {0};
   }
-  cp_mcause : coverpoint isacov.instr.rvfi.csrs["mcause"].get_csr_retirement_data() {
+  cp_mcause : coverpoint isacov.instr.rvfi.csrs["mcause"].get_csr_retirement_data()[10:0] {
```
silabs-robin, line 44: I think this will cause trouble with VCS, but since this file is not top-level it should not cause trouble for other teams.

lib/isa_decoder/isa_decoder.sv
```diff
@@ -635,6 +635,22 @@
           asm.rs1.valid           = 1;
           asm.rs1.valid_gpr_rvc   = 1;
           asm.imm.valid           = 1;
+        end else if (name inside { C_ANDI }) begin
```
silabs-robin, line 638: This was missing indeed.
Setting both `rd` and `rs1` might not be ideal, but it is the safest.

silabs-robin (approved):
Approving based on:
Comments given in-line.

(And... assuming sim/fv is still ok.)

PR #2282, opened 2023-11-07
Tweak Scripts -  `merge` and `clonetb`

silabs-robin:
This PR makes a few small tweaks to the `merge` and `clonetb` scripts.
(See in-line comments in the diff.)

### DIFF
--- a/bin/merge.sh
+++ b/bin/merge.sh
@@ -33,6 +33,7 @@ usage() {
   echo "--x-dv_into_s     Do a merge of cv32e40x-dv main into core-v-verif cv32e40s/dev cv32e40s (not yet developed)"
   echo "--sdev_into_xdev  Do a merge of core-v-verif cv32e40s/dev into core-v-verif cv32e40x/dev"
   echo "--xdev_into_sdev  Do a merge of core-v-verif cv32e40x/dev into core-v-verif cv32e40s/dev"
+  echo "--rejection-diff  Merge s/dev to x-dv, using 'theirs'"
 
   exit 1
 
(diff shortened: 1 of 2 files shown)


bin/merge.sh
```diff
@@ -124,38 +125,33 @@ merge_xdev_into_sdev () {
 }
 
 
-check_cv32e40x_repo() {
+clone_x_dv() {
```
silabs-robin, line 128: The S->X merge now starts with cloning x-dv.
It asks for y/n confirmation first.

I think this streamlines the most typical use case of this script.
Please let me know if I overlooked any reason this might be a bad idea.

bin/merge.sh
```diff
@@ -176,7 +172,7 @@ main() {
       usage
       ;;
   esac
+
```
silabs-robin, line 175: I also added newlines to match the dominating style convention in the rest of this file.

bin/merge.sh
```diff
@@ -124,38 +126,53 @@ merge_xdev_into_sdev () {
...
 
 
 check_merge_status() {
+
   git status
+
+}
+
+
+rejection_diff() {
```
silabs-robin, line 151: I realized I wanted to add this one as well, so I don't have to think as much the next time around.

bin/merge.sh
```diff
@@ -64,12 +66,12 @@ merge_cv32e40s_into_cv32e40x-dv () {
...
 
-  echo "=== Exchange 40x/X with 40s/S in file names ==="
+  echo "=== Replace 40s/S with 40x/S in file names ==="
 
   find . -type d | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | xargs -n1 dirname | awk '{gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 mkdir -p
-  find . -type d | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | awk '{printf $1; printf " "; gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 git mv
+  find . -type d | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | awk '{printf $1; printf " "; gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 git mv -f
   find . -type f | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | xargs -n1 dirname | awk '{gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 mkdir -p
-  find . -type f | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | awk '{printf $1; printf " "; gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 git mv
+  find . -type f | egrep -iv '\/\.|40sx|40xs' | grep -i 40s | awk '{printf $1; printf " "; gsub(/40s/, "40x"); gsub(/40S/, "40X"); print}' | xargs -n2 git mv -f
```
silabs-robin, line 74: `git mv` refused to copy certain files.
Added `-f` to force every move to take effect.

silabs-robin (commented):
Explaining the latest commit.

(I have no more commits planned.)

bin/merge.sh
```diff
@@ -64,12 +66,12 @@ merge_cv32e40s_into_cv32e40x-dv () {
 
 move_files_40s_into_40x () {
 
-  echo "=== Exchange 40x/X with 40s/S in file names ==="
+  echo "=== Replace 40s/S with 40x/S in file names ==="
```
silabs-krdosvik, line 69: Is it supposed to be Replace 40s/S with 40x/**X** ?

bin/merge.sh
silabs-krdosvik: Nice!

bin/merge.sh
silabs-krdosvik: That is very nice!

MikeOpenHWGroup (commented):
The `bin/merge.sh` shell script seems to be growing more than we might have expected.  Would it be easier to maintain as a python scipt?

bin/merge.sh
silabs-robin, line 1: > The bin/merge.sh shell script seems to be growing more than we might have expected. Would it be easier to maintain as a python scipt?

@MikeOpenHWGroup 
I think the complexity is still very low.
There are 2 control flow elements (2 case statements, 0 loops).
There are 2 inter dependencies (`main` calls other functions, `rejection_diff` calls other functions).
The functions mostly just call `git` in various ways, and `sed` etc.
I actually don't think there would be any benefit of going to Python, as there is no complexity it could remove and it would add overhead with subprocesses (which is approximately 100% free in shell).

bin/merge.sh
silabs-robin: Fixed.

silabs-krdosvik: RESOLVED

bin/merge.sh
MikeOpenHWGroup: Fair enough.  Let's keep an eye on those regexs.  :wink:

silabs-oysteink (approved):
Approving in Henrik's absence

PR #2283, opened 2023-11-07
Add possibility for external binds, comment out failing assertions

silabs-krdosvik:
- Add possibility for external binds
- Comment out failing assertions

removing assertions should not impact ci check, so not testes, but hello-world runs.
Formal compiles.


silabs-robin (approved):
Approving based on:
External binds are useful.
The guarding constants used seem appropriate.
Commented-away asserts have (named) todos.

PR #2287, opened 2023-11-09
Add paramset 1

silabs-krdosvik:
Add paramset 1.


silabs-robin (approved):
Approving based on:
Looks innocuous, and useful.

PR #2290, opened 2023-11-14
Organizing acknowledgements in the repo.

jquevremont:
Following a meeting with @davideschiavone (TRISTAN) and Flo, for a solution common to all projects (and alignment with TRISTAN project).


davideschiavone:
I would leave this to @MikeOpenHWGroup as mainteiner of the repo

jquevremont:
We should be all set now with the acknowledgements.

MikeOpenHWGroup (approved):

PR #2294, opened 2023-11-16
Cv32e20/dev

emgens:
Added User mode support to the CSR access test.


cv32e20/tests/programs/custom/riscv_csr/custom_handlers.S
```diff
@@ -0,0 +1,222 @@
+/*
+**
+** Copyright 2023 Caddence
```
MikeOpenHWGroup, line 3: Misspelled Cadence

cv32e20/tests/programs/custom/riscv_csr/riscv_csr.c
```diff
@@ -0,0 +1,69 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include "custom_handlers.h"
+#include "riscv_csr_test_0.h"
+
+volatile int glb_expect_illegal_insn    = 0;
+
+volatile int  glb_csr_address = 0;
+
+#define TEST_FAILED  *(volatile int*)(0x20000000) = 1
```
MikeOpenHWGroup, line 41: Redundant already defined in BSP

MikeOpenHWGroup (commented):
Thanks for this @emgens.  I pointed out a couple of typos.  Also, could you add a README.md to explain how an end-user will use these tests?

cv32e20/tests/programs/custom/riscv_csr/custom_handlers.S
emgens: I have pushed a fix for this.  Thanks!

cv32e20/tests/programs/custom/riscv_csr/riscv_csr.c
emgens: I could not find these in the BSP.  Neither could the compiler (it returned an undeclared symbol error when I removed these lines.) 

I agree that this functionality is a good candidate to move to the BSP, but unless there is something I'm missing, it's not there now.

MikeOpenHWGroup: Yipe!  It seems a lot of bit-rot has crept into our test-programs over time.  This rot hasn't broken anything, but that is just dumb luck.  Here is the situation as I currently understand it:
- CORE-V-VERIF's BSP is built on [newlib](https://sourceware.org/newlib/), a C library for embedded systems.  This allows us to write C programs with support for nice things such as **argc/argv**, **printf**, etc.
- In order for `newlib` to work on your system, we implemented some low-level functions such as `_putc()` and `_exit()`.  This is done in [bsp/syscalls.c](https://github.com/openhwgroup/core-v-verif/blob/69fca5157545245bba95b6bb49c3a51d1f606e58/cv32e20/bsp/syscalls.c#L1).
- With the above in place, returns from `main()` using either `EXIT_SUCCESS` or `EXIT_FAILURE` will eventually invoke CORE-V-VERIF's implementation of `_exit()`.  Check out [hello-world.c](https://github.com/openhwgroup/core-v-verif/blob/69fca5157545245bba95b6bb49c3a51d1f606e58/cv32e20/tests/programs/custom/hello-world/hello-world.c#L141) for examples of this.
- CORE-V-VERIF's implementation of `_exit()` writes either 0 or 1 to `EXIT_REG` which is defined as 0x20000004 (look familiar?).
- Not all CORE-V-VERIF C test-programs use these exit codes.  For example, [debug_test.c](https://github.com/openhwgroup/core-v-verif/blob/69fca5157545245bba95b6bb49c3a51d1f606e58/cv32e20/tests/programs/custom/debug_test/debug_test.c#L342) defines `TEST_FAILED` and `TEST_PASSED` macros and simply uses these.  The macros are in-line writes to 0x20000000:
```
#define TEST_FAILED  *(volatile int *)0x20000000 = 1
#define TEST_PASSED  *(volatile int *)0x20000000 = 123456789
```
Over time, several other test-programs have copied the practise of locally defining `TEST_FAILED` and `TEST_PASSED` macros.  Some test-programs hard-code the value rather than using macros.

I have create a [task](https://github.com/openhwgroup/cve2/issues/172) to get this cleaned up.  In the meantime, this pull-request is following a path used by others, so it is not adding too much chaos into the code-base so let's merge it in now and clean up later.

MikeOpenHWGroup (approved):

PR #2295, opened 2023-11-17
Debug vplan: zalrsc

silabs-krdosvik:
Removed zalrsc part of the TODO.


silabs-robin (approved):
Approving based on:
Updates to the vplan have value.
I trust the rationale of the link2cov.

PR #2296, opened 2023-11-20
Xdev2sdev merge 2023 11 20

silabs-robin:


bin/gen_csr_access_test.py
silabs-robin, line 1: This is a script update I did on X some time ago. Coming over to S now.

cv32e40s/tb/assertions/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -932,21 +932,21 @@ module  uvmt_cv32e40s_umode_assert
 
   // vplan:InstrProt
 
-  a_instr_prot: assert property (
+  a_rvfi_instr_prot: assert property (
     rvfi_valid
     |->
-    (rvfi_if.instr_prot[2:1] == rvfi_if.rvfi_mode)  ||
+    (rvfi_if.rvfi_instr_prot[2:1] == rvfi_if.rvfi_mode)  ||
```
silabs-robin, line 938: The interface was previously updated on the X, adding `rvfi_` prefix to data/instr prot. Had to update this file.

cv32e40s/tb/assertions/uvmt_cv32e40s_umode_assert.sv
```diff
@@ -932,21 +932,21 @@ module  uvmt_cv32e40s_umode_assert
 
   // vplan:InstrProt
 
-  a_instr_prot: assert property (
+  a_rvfi_instr_prot: assert property (
```
silabs-robin, line 935: Woops, search-and-replace got this name as an innocent bystander. Not a problem though.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -164,45 +164,55 @@ module uvmt_cv32e40s_tb;
...
+      .rvfi_mem_wmask (rvfi_i.rvfi_mem_wmask[uvma_rvfi_pkg::NMEM*uvmt_cv32e40s_base_test_pkg::XLEN/8*0  +:uvma_rvfi_pkg::NMEM*uvmt_cv32e40s_base_test_pkg::XLEN/8]),
+      .rvfi_instr_prot (rvfi_i.rvfi_instr_prot),
+      .rvfi_instr_memtype (rvfi_i.rvfi_instr_memtype),
+      .rvfi_instr_dbg (rvfi_i.rvfi_instr_dbg),
+      .rvfi_mem_prot (rvfi_i.rvfi_mem_prot),
+      .rvfi_mem_exokay (rvfi_i.rvfi_mem_exokay),
+      .rvfi_mem_err (rvfi_i.rvfi_mem_err),
+      .rvfi_mem_atop (rvfi_i.rvfi_mem_atop),
+      .rvfi_mem_memtype (rvfi_i.rvfi_mem_memtype),
+      .rvfi_mem_dbg (rvfi_i.rvfi_mem_dbg)
```
silabs-robin, line 214: Connecting data/instr prot, and some other new signals.

(Also reformatted, because I found it so difficult to read.)

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb_files.flist
```diff
@@ -18,39 +18,39 @@
 ${DV_UVMT_PATH}/uvmt_cv32e40s_dut_wrap.sv
 ${DV_UVMT_PATH}/uvmt_cv32e40s_tb.sv
 
+${DV_UVMT_PATH}/../assertions/uvmt_cv32e40s_fencei_assert.sv
+${DV_UVMT_PATH}/../assertions/uvmt_cv32e40s_pmp_assert.sv
+${DV_UVMT_PATH}/../assertions/uvmt_cv32e40s_pmprvfi_assert.sv
+${DV_UVMT_PATH}/../assertions/uvmt_cv32e40s_umode_assert.sv
```
silabs-robin, line 24: Moved umode and pmp asserts to the assertions dir.

(The other changes are just `sort` in alphabetical order.)

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-robin, line 1: Not sure what this is.
Seem atomics-related.
Was previously reviewed and approved on the X.

bin/clonetb
```diff
@@ -81,8 +81,7 @@ clone() {
 clone_cv32e40x() {
   CV_CORE=cv32e40x
   VERIF_ENV_REPO=https://github.com/openhwgroup/cv32e40x-dv.git
-  VERIF_ENV_REF=443f11c
-
+  VERIF_ENV_REF=be17b8902002f91803abde4bfb8caa91088575e1
```
silabs-krdosvik, line 84: Maybe this will be fixed in another commit, but the hash is not the updated one.

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-krdosvik: True! Yes, this is zalrsc related. The code makes it possible to generate correct exokay value on obi. This is X specific because no zalrsc på S, but since we aim for making the code bases similar I think it is okay to add this code.

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
```diff
@@ -290,14 +290,6 @@ function void uvma_pma_sb_c::check_obi_i_default_region(uvma_obi_memory_mon_trn_
                                          obi.access_type.name(), obi.address, override_region,
```
silabs-krdosvik, line 290: On X this file was modify to allow obi.atop signal to be set, which is allowed on X but not S. I dont think we should allow the same on S. Soo, simplest method, revert the file back to what it was? :-)

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-robin: > but since we aim for making the code bases similar I think it is okay to add this code

Also, this is top-level "lib/", which is common to all cores using core-v-verif.

bin/clonetb
silabs-robin: It was the latest when I made the PR.
And the newer commits are seemingly just vplan updates.

> Maybe this will be fixed in another PR

I'd prefer this one.

bilde

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
silabs-robin: This is top-level "lib/".

silabs-krdosvik: Okei. I dont think it is that big deal really, but I was thinking that the lib could be different due to the different branches cv32e40xs/dev. If we add it to S too, it will just be a little less security features around the atop feature.

lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
silabs-krdosvik: I see. RESOLVED.

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
silabs-krdosvik: I am fine with both ways here, so RESOLVED.

bin/clonetb
silabs-robin, line 1: > This branch has conflicts that must be resolved

I'll fix this too...

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
silabs-robin: > it will just be a little less security features around the atop feature

I don't know the implications of this. Are we sure it will be fine?

bin/clonetb
silabs-robin: Done.

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_sb.sv
silabs-krdosvik: If S somehow uses atop signal (which is not possible at the moment), we dont get an uvme error (as we remove the if(obi.atop) statement). I think it should be ok.

silabs-robin: What about other cores using core-v-verif?

silabs-krdosvik: Yea.. I guess if they depend on this check is is not good if we remove it.
But, if we must chose between keeping the if statement or remove it given that s and x should have the same code then we must remove it because it will be a stopper for x. But if we dont need to remove it, it is best not to.

PR #2297, opened 2023-11-21
Updated disassembler to handle sreg representation of compressed regs

silabs-hfegran:
- mva01s, mvsa01 missed sreg-representaitons of GPRs, fixed


silabs-hfegran:
Recommend review with hidden whitespace changes

silabs-robin (approved):
Approving based on:
Offline discussion explained how this was not handled before.
Presuming it compiles and runs without problem.

PR #2301, opened 2023-11-23
Updated clic test to cover mret+minvh+access fault trap

silabs-hfegran:
- Added directed test to cover the case where software sets minhv, executes mret and the table access is blocked
- Added directed test to cover the case where software sets minhv, executes mret and the table pointer derefence is blocked


silabs-hfegran:
Note: This PR needs a hash update, as it will mismatch ISS with earlier hashes

PR #2304, opened 2023-11-28
fix assert

silabs-krdosvik:
Fix assert that had failing coverage.
Passes in formal.
ci check:


cv32e40s/tb/assertions/uvmt_cv32e40s_pmprvfi_assert.sv
```diff
@@ -507,26 +509,86 @@ module uvmt_cv32e40s_pmprvfi_assert
...
+
+    support_if.asm_rvfi.csr.address == MSECCFG &&
+    support_if.asm_rvfi.csr.valid &&
+
+    (support_if.asm_rvfi.instr == CSRRW ||
+    support_if.asm_rvfi.instr == CSRRC ||
+    support_if.asm_rvfi.instr == CSRRWI ||
+    support_if.asm_rvfi.instr == CSRRCI) &&
+
+    !rvfi_trap)[->1]
```
silabs-mateilga, line 547: [->1] isn't necessary here, the entire antecedent is in one cycle

cv32e40s/tb/assertions/uvmt_cv32e40s_pmprvfi_assert.sv
```diff
@@ -507,26 +509,86 @@ module uvmt_cv32e40s_pmprvfi_assert
...
+    (support_if.asm_rvfi.instr == CSRRWI && support_if.asm_rvfi.imm.imm_value[2] == 1'b0) ||
+    (support_if.asm_rvfi.instr == CSRRCI && support_if.asm_rvfi.imm.imm_value[2] == 1'b1))
+
+    |->
+    pmp_csr_rvfi_wmask.mseccfg.rlb          &&  // Must attempt
+    (pmp_csr_rvfi_wdata.mseccfg.rlb == 1'b0)     // Must succeed
   ) else `uvm_error(info_tag, "RLB must be changeable after reset");
 
+
   a_until_reset_notbefore_1: assert property (
```
silabs-robin, line 563: Is this the same code as above, but duplicated instead of letting the property take an argument, in order to avoid the formal compilation warning about non-lrm semantics for cover properties?

silabs-krdosvik: The code is similar but not equal. I found it difficult to distinguish between the CSRRC and CSRRS logic in just one property, so I just splited them.

cv32e40s/tb/assertions/uvmt_cv32e40s_pmprvfi_assert.sv
silabs-krdosvik: True! FIXED

silabs-robin (approved):
Approving based on:
Test status ok.
Feedback resolved.

PR #2308, opened 2023-12-01
updated latest link2cov

silabs-mateilga:


silabs-robin (commented):
Approving based on:
Looks good and matches our latest fixes/discussions about clic/pma/zcmt tests.

PR #2310, opened 2023-12-01
update write to mhpmcounterh

silabs-krdosvik:


silabs-robin (approved):
Approving based on:
Seems like that change makes the test more meaningful.
Presuming the test passes.

PR #2316, opened 2023-12-10
Random-test Fixup - `obi_err_debug`

silabs-robin:
This merge fixes a problem with the `corev_rand_data_obi_err_debug` test where early bus errors would cause lockup.

Test status:
* `>100` of the 4 "obi_err" tests - Passed.
* "full" testlist - All pass (except timeouts, acceptable)

--------

Details:

`riscv-dv` uses two stack pointers ("kernel" and "user").
If bus errors appear early, those stack pointers have not been set up.
The bus errors' handler still use those stack pointers, so when their prologs try to backup the GPRs to stack, they just overwrite instructions on address 0x0 etc.
Eventually the flow enters an infinite trap-loop executing illegal instructions from address 0x0.

With these changes, the `riscv-dv` init will have completed the stack pointer setup before bus errors start going off.
A plusarg is added to let the bus error generation await a "go-ahead" flag, and a VP is used to let the assembly de-assert that flag.

bilde


lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv
```diff
@@ -191,6 +193,7 @@ class uvma_obi_memory_cfg_c extends uvm_object;
       soft drv_slv_err_mode               == UVMA_OBI_MEMORY_DRV_SLV_ERR_MODE_OK;
       soft drv_slv_err_one_shot_mode      == 0;
       soft drv_slv_exokay_mode            == UVMA_OBI_MEMORY_DRV_SLV_EXOKAY_MODE_SUCCESS;
+      soft random_err_await_goahead       == 0;
```
silabs-robin, line 196: Note: Default behavior is unchanged. New behavior only active upon explicit request.

silabs-robin (commented):
Adding some notes to aid review.

silabs-mateilga (approved):
Looks good, nice functionality to have, and seems to solve our immediate problem.

silabs-oysteink (approved):

PR #2320, opened 2023-12-11
Cv32e40s/vsim take2

MikeOpenHWGroup:
Hi @silabs-robin, as per your suggestion, here is a PR from my core-v-verif fork (branch cv32e40s/vsim-take2) to openhwgroup branch cv32e40s/release.


silabs-robin:
I'm on the case. Doing some preparations first.

backup
silabs-robin, line 1: Something is wrong with Github, but I found a possible fix that I will try https://stackoverflow.com/a/46782679

silabs-robin: That worked.

silabs-robin:
See https://github.com/openhwgroup/core-v-verif/pull/2314#issuecomment-1852138624

PR #2325, opened 2023-12-13
added necessary overrides for clic

silabs-mateilga:


cv32e40s/tests/cfg/pma_test_cfg_2.yaml
silabs-robin, line 1: Does ci_check pass?

silabs-mateilga: yes

silabs-robin (approved):
Approving based on:
Fixes known problem.
Also discussed offline.

Issue #2326, opened 2023-12-13
ISACOV : Pseudo-instructions


AyoubJalali:
Hello, 
@AnouarZajni create functional coverage for all supported csr in the embedded config using uvm_reg and using the decode instruction coming from ISACOV, what we observe that the come-out result wasn't that good for us, because we have a large tests for CSR.
the problem was that the ISACOV doesn't support pseudo-code, and almost all the csr tests was writing using CSR pseudo-code (CSRR, CSRW, CSRC, CSRS)
here is a screen-shot of ISACOV log : 
![image](https://github.com/openhwgroup/core-v-verif/assets/110232072/bb1f643d-5cbe-4704-abf1-634e59a31a2d)

So could confirm the bug, so I can create a PP to fix it.

silabs-robin:
> the problem was that the ISACOV doesn't support pseudo-code

I'm not convinced there is any issue with that. (You can ask Mike if you'd like more details on the rationale there.)

--------

> here is a screen-shot of ISACOV log :

"unknown"? "ILLEGAL"?

It seems ISACOV lacks support for whatever those instructions are.
Have you decoded them (or have the objdump) and know what these instructions are?

AyoubJalali:
> It seems ISACOV lacks support for whatever those instructions are.

these are csrr xx mcause (it's a csr instruction)

silabs-robin:
> > It seems ISACOV lacks support for whatever those instructions are.
> 
> these are csrr xx mcause (it's a csr instruction)

That csrr pseudo instr is ran as an actual csrrs.
![bilde](https://github.com/openhwgroup/core-v-verif/assets/75431602/e1f329ad-65ee-4a3c-b3f5-22f0754cc06f)

But what is the hex of it?
The isacov logger only prints the PC, but it would be convenient if it printed the hex.
![bilde](https://github.com/openhwgroup/core-v-verif/assets/75431602/2b12ed1d-accf-4fa8-8525-846481df5888)

But if its only csrrs I don't understand why isacov wouldn't accept it.
Maybe it is just the convert2string function that doesn't support it.
![bilde](https://github.com/openhwgroup/core-v-verif/assets/75431602/b502fdf5-b078-46ef-bdec-cbd6799d648a)

AyoubJalali:
> I'm not convinced there is any issue with that. (You can ask Mike if you'd like more details on the rationale there.)



[...]

PR #2327, opened 2023-12-13
Add CSR pseudo-instructions to ISACOV

AyoubJalali:
This PR fix [CSR Pseudo-instruction issue](https://github.com/openhwgroup/core-v-verif/issues/2326)

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -2709,12 +2709,13 @@ function void uvma_isacov_cov_model_c::sample (uvma_isacov_instr_c instr);
   if (!have_sampled && is_normal_instr && cfg.core_cfg.ext_zicsr_supported) begin
     have_sampled = 1;
     case (instr.name)
-      CSRRW:   rv32zicsr_csrrw_cg.sample(instr);
-      CSRRS:   rv32zicsr_csrrs_cg.sample(instr);
-      CSRRC:   rv32zicsr_csrrc_cg.sample(instr);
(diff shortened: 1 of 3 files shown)


MikeOpenHWGroup:
As per my comment in #2326, there is no need for functional coverage of pseudo-instructions.

AyoubJalali:
> As per my comment in #2326, there is no need for functional coverage of pseudo-instructions.

I'm not trying to add functional coverage for psudo-code, I'm trying to add the capability to ISACOV to know this CSR pesudo-code and not drive me an UNKNOWN instruction.

silabs-hfegran:
Is there a way to have spike output the proper instruction names instead of the pseudo names? I think that would be cleaner than adding support for pseudo-instructions to isacov.

AyoubJalali:
> Is there a way to have spike output the proper instruction names instead of the pseudo names? I think that would be cleaner than adding support for pseudo-instructions to isacov.

No idea to be honest, but I think after the spike tandem version, something changed in spike-dasm (I think we don't use spike-dasm in lib/ or something else)

silabs-hfegran:
Completely untested, but I suspect it could be as easy as removing these four lines of code from `disasm/disasm.cc` in spike
```  
  add_insn(new disasm_insn_t("csrr", match_csrrs, mask_csrrs | mask_rs1, {&xrd, &csr}));
  add_insn(new disasm_insn_t("csrw", match_csrrw, mask_csrrw | mask_rd, {&csr, &xrs1}));
  add_insn(new disasm_insn_t("csrs", match_csrrs, mask_csrrs | mask_rd, {&csr, &xrs1}));
  add_insn(new disasm_insn_t("csrc", match_csrrc, mask_csrrc | mask_rd, {&csr, &xrs1}));
  ```

AyoubJalali:
As waiting for the decoder, I'll close this PR

PR #2330, opened 2023-12-14
update hash

silabs-krdosvik:
Update hash


bin/clonetb
silabs-robin, line 1: Approving based on:
Presuming that ci_checked passed or will be promptly fixed.

silabs-robin (approved):
Approving based on:
Presuming that ci_checked passed or will be promptly fixed.

PR #2331, opened 2023-12-14
merge script update

silabs-krdosvik:
Add file substitution option

### DIFF
--- a/bin/merge.sh
+++ b/bin/merge.sh
@@ -90,6 +90,12 @@ substitute_file_content_40s_into_40x () {
 
   echo "=== Exchange 40x/X with 40s/S in file content ==="
 
+  read -p "Exchange 40x/X with 40s/S in file content? y/n " yn
+  case $yn in
+    [Yy]* ) ;;
+    * ) echo "Skip content substitution"; return;;
+  esac
+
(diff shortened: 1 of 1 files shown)


bin/merge.sh
```diff
@@ -90,6 +90,12 @@ substitute_file_content_40s_into_40x () {
 
   echo "=== Exchange 40x/X with 40s/S in file content ==="
 
+  read -p "Exchange 40x/X with 40s/S in file content? y/n " yn
+  case $yn in
+    [Yy]* ) ;;
+    * ) echo "Skip content substitution"; return;;
+  esac
```
silabs-robin, line 97: So this is "opt in"?
Defaults to _not_ substituting?

Maybe the y/n message should say something about the default, hinting at the default being a safe recommendation?
`(default: n)` or something short and user-friendly, so people don't have to stop and think.

silabs-robin (approved):
Pre-approving.
Whether my comment results in a change or not are both ok outcomes.
I'm ok with this PR being merged after the comment has been resolved.

bin/merge.sh
silabs-krdosvik: Good point. I will fix it in another PR.

PR #2335, opened 2023-12-15
[SPIKE] Add ZCB isa to RVFI utils and add priv parameter to Spike

MarioOpenHWGroup:


lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
```diff
@@ -46,13 +46,14 @@ localparam config_pkg::cva6_cfg_t CVA6Cfg = cva6_config_pkg::cva6_cfg;
...
     function automatic void rvfi_initialize_spike(bit testharness);
-       string binary, rtl_isa;
+       string binary, rtl_isa, rtl_priv;
 
         if ($value$plusargs("elf_file=%s", binary))
             `uvm_info("spike_tandem", $sformatf("Setting up Spike with binary %s...", binary), UVM_LOW);
 
-       rtl_isa = $sformatf("RV%-2dIM",
-                            riscv::XLEN);
+        rtl_priv = "M";
```
MikeOpenHWGroup, line 54: Is the strategy here to set the priv level to M by default because every core must support M-mode and then add modes as required by specific cores?   Does that imply that debug-mode is considered optional?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
```diff
@@ -65,17 +66,29 @@ localparam config_pkg::cva6_cfg_t CVA6Cfg = cva6_config_pkg::cva6_cfg;
         if (CVA6Cfg.RVC) begin
             rtl_isa = {rtl_isa, "C"};
         end
-       if (cva6_config_pkg::CVA6ConfigBExtEn) begin
-           rtl_isa = $sformatf("%s_zba_zbb_zbc_zbs", rtl_isa);
-       end
+        if (cva6_config_pkg::CVA6ConfigBExtEn) begin
```
MikeOpenHWGroup, line 69: I just noticed this now.  Once we added this line:
```
```
we made the cva6-package a required package for the RVFI.  The RVFI agent needs to be useable by any environment, not just environments that support the CVA6.

MikeOpenHWGroup (commented):
Hi @MarioOpenHWGroup.  I cannot comment on the extensions/configurations to the Spike model, other than to say it looks good.  I will ask @zchamski to comment.

I do have a comment/question about the RVFI agent.  Please have a look at that.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
MarioOpenHWGroup: I totally agree, that was what I was thinking, when I return I will modify this to use a generic config struct which will be independent from the specific core, meanwhile I think we can merge it as it is as it is already cva6 dependant

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
MarioOpenHWGroup: We need to discuss this for sure, right now I just wanted to parameterize the priv mode, but for sure the presence of  debug mode should be also a param.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
MikeOpenHWGroup: Good.  I have crrated #2336 is track this.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
MikeOpenHWGroup: Agreed.  For the CV32E60MX (the "embedded application" of the CVA6), debug mode is not supported, so we do need to treat debug mode as a param.  However, I believe that machine mode _must_ be supported by all cores, so your strategy seems OK to me.

@silabs-hfegran and @YoanFournier , what are your thoughts on this?

zchamski: The Privileged Spec v20211203, Section 1.2, page 3 states explicitly that

> _All hardware implementations must provide M-mode, as this is the only mode that has unfettered
access to the whole machine._

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
zchamski, line 1: LGTM

vendor/riscv/riscv-isa-sim/riscv/Simulation.cc
zchamski, line 1: LGTM.  The cleanup is a Good Thing.

vendor/riscv/riscv-isa-sim/spike_main/spike.cc
zchamski, line 1: LGTM

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
zchamski, line 1: Given the explicit statement about mandatory M-mode in the Privileged Spec document, this LGTM.

zchamski (approved):
From my side the PR looks all good.

PR #2337, opened 2023-12-17
Small fixes

silabs-krdosvik:
make pma 2 clic config
clock sequences for questa
remove todos, 
remove old comment
add comment to make assertions more understandable.

Addresses some of the comments in this PR https://github.com/openhwgroup/core-v-verif/pull/2314

Formal compiles
Ci check: pass


silabs-hfegran (approved):
Looks good to me. 

One comment about the `pma_test_cfg_<x>` configs, these test configs are not intended to be used with directed tests (at least not without a matching linker script, but in this case the configuration should be ok to use and should work fine for this specific test).

PR #2343, opened 2023-12-21
New AXI agent

AEzzejjari:
The agent supports slave mode only, and can be used as active or passive.
The new AXI5 agent is based on AXI4 plus atomic transactions from AXI5.
The agent support out of order and outstanding transaction, Channel delay, Multiple region signaling, Access permissions.


JeanRochCoulon:
@MikeOpenHWGroup and @MarioOpenHWGroup  The Today verification meeting was great. This is the related AXI5 PR to be merged in core-v-verif. Giving us your feedback would be great.

lib/uvm_agents/uvma_axi5/src/uvma_axi_pkg.flist
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Alae Eddine EZ ZEJJARI (alae-eddine.ez-zejjari@external.thalesgroup.com) – sub-contractor MU-Electronics for Thales group
+// Co-Author: Abdelaali Khardazi
+
+// Directories
++incdir+${DV_UVMA_AXI_PATH}
```
MikeOpenHWGroup, line 12: These variables conflict with variables of the same name in `lib/uvm_agents/uvma_axi`.    We either need to rename these variables (e.g. `DV_UVMA_AXI5`) or delete the AXI Agent in `lib/uvm_agents/uvma_axi`.

lib/uvm_agents/uvma_axi5/src/uvma_axi_pkg.sv
```diff
@@ -0,0 +1,77 @@
...
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Alae Eddine EZ ZEJJARI (alae-eddine.ez-zejjari@external.thalesgroup.com) – sub-contractor MU-Electronics for Thales group
+// Co-Author: Abdelaali Khardazi
+
+/***** AXI4 slave agent package *****/
+
+`ifndef __UVMA_AXI_PKG_SV__
```
MikeOpenHWGroup, line 13: These macros names are the same as is used for `lib/uvm_agents/uvma_axi/src/uvma_axi_pkg.sv`.  This means that these two Agents cannot be used in the same UVM environment.  If the `lib/uvm_agents/uvma_axi` is to be deleted that may be OK, but I would recommend uniquifying these macros to some like `__UVMA_AXI4_PKG_SV__`.

This problem exists for all SV files in this PR.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_fw_preload_seq.sv
```diff
@@ -0,0 +1,91 @@
...
+//
+// Original Author: Alae Eddine EZ ZEJJARI (alae-eddine.ez-zejjari@external.thalesgroup.com) – sub-contractor MU-Electronics for Thales group
+
+
+`ifndef __UVMA_AXI_FW_PRELOAD_SEQ_SV__
+`define __UVMA_AXI_FW_PRELOAD_SEQ_SV__
+
+
+/**
+ * Virtual sequence implementing the cv32e40x virtual peripherals.
```
MikeOpenHWGroup, line 16: Please update this comment.  This virtual sequence preloads a memory.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_fw_preload_seq.sv
```diff
@@ -0,0 +1,91 @@
...
+ * TODO Move most of the functionality to a cv32e env base class.
+ */
+class uvma_axi_fw_preload_seq_c extends uvm_sequence;
+
+   static uvm_cmdline_processor uvcl = uvm_cmdline_processor::get_inst();
+            // Agent handles
+   uvma_axi_cfg_c    cfg;
+   uvma_axi_cntxt_c  cntxt;
+
+   uvml_mem_c mem;
```
MikeOpenHWGroup, line 26: Why is the memory a member of a sequence?  The memory should be a member of the Agent's context (cntxt).

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
```diff
@@ -0,0 +1,430 @@
+#AXI Agent documentation
+
+## Contents
```
MikeOpenHWGroup, line 3: This is **_outstanding_** @AEzzejjari.  You have set the standard for UVM component documentation in CORE-V-VERIF.  Good work (and thanks).

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
```diff
@@ -0,0 +1,430 @@
+#AXI Agent documentation
```
MikeOpenHWGroup, line 1: Please add a license header.  For Markdown files it can look something like:
```
<!--
Copyright (c) Thales 2024
SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
-->
```

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Directory Structure
+
+-   docs: Documents describing the AXI.
+
+-   examples: Samples for users
+
+-   src: Source code for this package
+
+## Packages dependencies
```
MikeOpenHWGroup, line 420: There is also a dependency on `read_elf()`, `get_section()` and `read_section_sv()` (although I am not sure we should keep these).

lib/uvm_agents/uvma_axi5/src/comps/uvma_axi_agent.sv
```diff
@@ -0,0 +1,144 @@
...
+   `uvm_component_utils(uvma_axi_agent_c)
+
+   uvma_axi_mon_c             monitor;
+   uvma_axi_drv_c             driver;
+   uvma_axi_vsqr_c            vsequencer;
+   uvma_axi_seq_item_logger_c seq_item_logger;
+
+   // Objects handles
+   uvma_axi_cfg_c      cfg;
+   uvma_axi_cntxt_c    cntxt;
```
MikeOpenHWGroup, line 28: The memory that is currently a member of `uvma_obi_memory_fw_preload_seq_c` should be a member of this `cntxt` member.

lib/uvm_agents/uvma_axi5/src/comps/uvma_axi_drv.sv
```diff
@@ -0,0 +1,311 @@
...
+class uvma_axi_drv_c extends uvm_driver #(uvma_axi_slv_seq_item_c);
+
+   `uvm_component_utils(uvma_axi_drv_c)
+
+   // Objects
+   uvma_axi_cfg_c                      cfg;
+   uvma_axi_cntxt_c                    cntxt;
+
+   //Handle to the items to be drived
+   uvma_axi_slv_seq_item_c             item;
```
MikeOpenHWGroup, line 28: I do not think you need this.  The `uvm_driver` base class already has a member called `req` which you have (properly) parameterized to be of type `uvma_axi_slv_seq_item_c`.   So everywhere that this driver uses `item`, you should be using `req`.

Is there a reason to declare `item`?

lib/uvm_agents/uvma_axi5/src/obj/uvma_axi_cfg.sv
```diff
@@ -0,0 +1,131 @@
...
+      soft is_active                   == UVM_ACTIVE;
+      soft trn_log_enabled             == 1;
+      soft axi_lock_enabled            == 1;
+      soft randomization_enabled       == 0;
+      soft cov_model_enabled           == 1;
+      soft ordering_read_mode          == UVMA_AXI_ORDERING_MODE_RANDOM;
+      soft ordering_write_mode         == UVMA_AXI_OUTSTANDING_MODE;
+      soft max_outstanding_write_trs   == 2;
+      soft max_outstanding_read_trs    == 2;
+      soft write_response_latency      == 20;
```
MikeOpenHWGroup, line 78: I recommend that the latency values be ranges, rather than fixed values.

lib/uvm_agents/uvma_axi5/src/obj/uvma_axi_cntxt.sv
```diff
@@ -0,0 +1,45 @@
...
+ * Object encapsulating all state variables for all AXI agent
+ * (uvma_axi_agent_c) components.
+ */
+class uvma_axi_cntxt_c extends uvm_object;
+
+   // Handle to agent interface
+   virtual uvma_axi_intf  axi_vi;
+
+   // Handle to memory storage for active slaves
+   uvml_mem_c mem;
```
MikeOpenHWGroup, line 23: This is the memory the test-program machine code can be loaded into.  It should be sufficent to use `$readmemh()` for this purpose.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_base_seq_item.sv
```diff
@@ -0,0 +1,140 @@
...
+
+
+`ifndef __UVMA_AXI_BASE_SEQ_ITEM_SV__
+`define __UVMA_AXI_BASE_SEQ_ITEM_SV__
+
+
+/**
+ * Object created by AXI agent monitor extending uvma_axi_seq_base_c.
+ */
+class uvma_axi_base_seq_item_c extends uvml_trn_seq_item_c;
```
MikeOpenHWGroup, line 18: Most of the members of this sequence item are actually ports on the AXI physical interface.  That is to be expected, but there should also be meta-members that can be used to define latency, control error generation, etc.

Having these meta-members available in the sequence_item makes things like functional coverage, scoreboarding, debugging and logging a lot easier.

MikeOpenHWGroup (commented):
This is great work @AEzzejjari!  Of course, I have lots of comments and questions:
- Thanks for doing the additional work to implement the seq_item_logger.  The team will thank you for it!
- Why is this Agent called `axi5` when it is mostly `axi4` with just one axi5 feature?
- Have you considered creating a self-testing environment?  The idea would be to create a simple environment class that connects two of these UVM Agents together. I expect you will find it very useful.

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
```diff
@@ -0,0 +1,430 @@
...
+
+In the sv testbench instantiate the VIF and connect it to the DUT master
+interface:
+
+   `uvma_axi_intf                axi_if(
+                                         .clk(clknrst_if.clk),
+                                         .rst_n(clknrst_if.reset_n)
+                                      );`
+
+//DUT connection Active mode
```
MikeOpenHWGroup, line 174: GitHub's formatting for this is a little messed up.

lib/uvm_agents/uvma_axi5/src/uvma_axi_assert.sv
```diff
@@ -0,0 +1,344 @@
...
+         if(axi_assert.psv_axi_cb.w_valid && axi_assert.psv_axi_cb.w_ready) begin
+            data_count[0]++;
+            if(burst_len[0] == (data_count[0]-1)) begin
+               w_data_num : assert(axi_assert.psv_axi_cb.w_last) begin          // Check if The Master assert the WLAST signal when it is driving the final write transfer in the burst
+                  cov_w_data_num++;
+               end else begin
+                  `uvm_error("AXI4 protocol checks assertion", "Violation of w_data_num");
+                  cov_w_data_num++;
+               end
+               data_count.pop_front();
```
MikeOpenHWGroup, line 76: The `pop_front()` function built into SystemVerilog queues returns the first element from the queue and your code does not "absorb" this return value.  For some simulators, this is a compile-time error.  (I believe VCS generates a warning.) There are two possible ways you can resolve this:
1. Declare an `int` to absorb the return value:
```
int data_count[$];
int data_not_used;
...
data_not_used = data_count.pop_front(); // remove, and discard first element of data_count[$]
```
2. Cast the function to void:
```
void'( data_count.pop_front() );
```
I prefer option 2 since some lint tools will not like option 1 (unused variable), and it is obvious that the return value is not used.

**Note**: there are 8 instances of this in this file.

JeanRochCoulon:
Thanks @MikeOpenHWGroup for this appreciated review. 

Maybe calling it axi4plus UVM agent could be better ?

@AEzzejjari can I ask you to not use the "git commit -force" to fix the PR. It is better to follow the modifications.

MikeOpenHWGroup:
> Maybe calling it axi4plus UVM agent could be better ?

It would be better to simply call it an AXI Agent and make the AXI feature support controllable via the configuration object.  This is "the UVM way" and you can see an example of how this is done by looking at the [OBI Agent](https://github.com/openhwgroup/core-v-verif/tree/master/lib/uvm_agents/uvma_obi_memory).  That Agent's configuration object has a random enumerated member called [version](https://github.com/openhwgroup/core-v-verif/blob/4da8b11327634a2c22de9fd6b9923f28bdc19109/lib/uvm_agents/uvma_obi_memory/src/obj/uvma_obi_memory_cfg.sv#L43).  The environment constrains this to fit the needs of the DUT.

We can do something similar for this AXI Agent.  Given that this Agent only supports AXI4, the version can be hard-coded for now.  We can add a member to the config object that enables select AXI5 features, such as the atomic transactions.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_fw_preload_seq.sv
AEzzejjari: Yes, the memory must be a member of cntxt and that's what I've done. This mem here declared to initialize the cntxt memory at the end of the preload sequence.
Capture1

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
AEzzejjari: Thank you @MikeOpenHWGroup

lib/uvm_agents/uvma_axi5/src/obj/uvma_axi_cfg.sv
AEzzejjari: Yes, I agree and that's what I did for the ready-valid latency, but this latency is the maximum cycle we can have between two outstanding transactions, but we can't control the exact latency because it depends on the CPU. It might be better to change the name to max_write_response_latency.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_base_seq_item.sv
AEzzejjari: This sequence item contains all the signals from the AXI master, as well as some variables used to identify the state of the transaction. This sequence is therefore only necessary to capture the request made by the CPU and give the synchronizer enough information to generate the right response. 
But if you see the uvma_axi_slv_seq_item_c here, I have the meta-members like delay and response error signal and the signals we need for coverage and scoreboard.

lib/uvm_agents/uvma_axi5/docs/axi5_agent_documentation.md
AEzzejjari: I don't think this is necessary because these functions are related to CVA6.

lib/uvm_agents/uvma_axi5/src/seq/uvma_axi_fw_preload_seq.sv
```diff
@@ -13,8 +13,7 @@
 
 
 /**
- * Virtual sequence implementing the cv32e40x virtual peripherals.
- * TODO Move most of the functionality to a cv32e env base class.
+ * Virtual sequence preload the CVA6 memory.
```
MikeOpenHWGroup, line 16: If this sequence is _only_ for the CVA6, then it must not be in `lib`.  Rather, it should be in a CVA6-specific virtual sequence.  A good example is in the CORE-V-VERIF repo at `core-v-verif/cv32e40p/env/uvme/vseq`, which is a set of CV32E40P-specific virtual sequences.

If this sequence is needed for CVA6, and it is CVA6-specific, it should be placed in in the CVA6 repo at `cva6verif/env/uvme/vseq`.

Having said that, I still think we should not use this method to populate the test-bench memory.

AEzzejjari: After reading your comment, I agree that we shouldn't put something that depends on CVA6 in the agent.
To do this, I change the preload sequence to something generic like in obi agent and add a field in the configuration class to control whether we want to use the agent's preloading sequence or not.
If we don't create the hex file, we can override the agent seq with one created in the environment, and in this new sequence we'll use the CVA6-dependent methods.

lib/uvm_agents/uvma_axi5/src/obj/uvma_axi_cfg.sv
AEzzejjari: After yesterday's discussion, I tried adding response latency to the sequence element, but found that it's very difficult, if not impossible, to delay the response if we have latency in the sequence.
In the future, we'll modify the agent to be transaction-based rather than cycle-based as it is today. It will then be easy to add this latency to the sequence element and make it random for each transaction.

MikeOpenHWGroup: OK.  Let's discuss at the CVA6 Workshop in a couple of weeks.

AEzzejjari:
Thanks Mike, for reviewing my PR and for your comment. I think all the point are solved, it’s remains only the response delay randomization, that will be possible to integrate it in the next step with the transaction based agent.
For that I like to know if we can merge the agent since all the issues have been solved.

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
@AEzzejjari, yes I have approved this PR and will leave it to either @JeanRochCoulon or @ASintzoff to do the merge.

There are still a few outstanding items, but we can update these after the merge.

Issue #1084, opened 2021-12-24
Duplicate code in uvmt_cv32e40p_debug_assert.sv


MikeOpenHWGroup:
Hi @silabs-oysteink, this is a somewhat unfair question about code you contributed back in September of 2020.

### Type
* Duplicated code

### Issue Description
The file in question is [uvmt_cv32e40p_debug_assert.sv](https://github.com/openhwgroup/core-v-verif/blob/65624082195461d1d631645c53673fef377acb2e/cv32e40p/tb/uvmt/uvmt_cv32e40p_debug_assert.sv).  The duplicated code is shared by property `p_cebreak_exception`, starting on [line 121](https://github.com/openhwgroup/core-v-verif/blob/65624082195461d1d631645c53673fef377acb2e/cv32e40p/tb/uvmt/uvmt_cv32e40p_debug_assert.sv#L121) and property `p_ebreak_exception`, starting on [line 136](https://github.com/openhwgroup/core-v-verif/blob/65624082195461d1d631645c53673fef377acb2e/cv32e40p/tb/uvmt/uvmt_cv32e40p_debug_assert.sv#L136).

These are complex properties with only a single clause difference between them (and that difference is only one character, p_**c**ebreak_exception vs. p_ebreak_exception.  Given that, future maintenance of these properties could be problematic.

Perhaps these properties could be combined into a single property:
```
    // ebreak or c.ebreak without dcsr.ebreakm results in exception at mtvec
    // Exclude single stepping as the sequence gets very complicated
    property p_cebreak_exception;
        disable iff(cov_assert_if.debug_req_i | !cov_assert_if.rst_ni)
        $rose(cov_assert_if.is_ebreak||cov_assert_if.is_cebreak) && cov_assert_if.dcsr_q[15] == 1'b0 && !cov_assert_if.debug_mode_q  && cov_assert_if.is_decoding && cov_assert_if.id_valid &&
        !cov_assert_if.debug_req_i && !cov_assert_if.dcsr_q[2]
        |-> (decode_valid) [->1:2] ##0  !cov_assert_if.debug_mode_q && (cov_assert_if.mcause_q[5:0] === cv32e40p_pkg::EXC_CAUSE_BREAKPOINT) 
                                                                && (cov_assert_if.mepc_q == pc_at_ebreak) &&
                                                                   (cov_assert_if.id_stage_pc == cov_assert_if.mtvec);
    endproperty
```

### Additional context
[...]

PR #2346, opened 2024-01-03
Flag duplicate properties for future FIXME

MikeOpenHWGroup:
Temporary resolution of #1084.


YoanFournier (approved):
LGTM

pascalgouedo:
Those 2 assertions are not duplicate as they are using 2 different instructions decoding signals.
One is using i_ebreak and the other one i_cebreak which differs depending on cov_assert_if.id_stage_is_compressed value.

MikeOpenHWGroup:
Good eye @pascalgouedo!  This duplicate code" was flagged by a Linter and I missed the  `i_ebreak` vs. `i_cebreak` difference.  I will create a new PR to remove the "FIXME" and update the comment to point out the difference.

PR #2347, opened 2024-01-04
properties p_ebreak_exception and p_cebreak_exception are not duplicates

MikeOpenHWGroup:
Updated the "FIXME" comment header that incorrectly flagged `p_ebreak_exception` and `p_*c*ebreak_exception` as duplicates.

### DIFF
--- a/cv32e40p/tb/uvmt/uvmt_cv32e40p_debug_assert.sv
+++ b/cv32e40p/tb/uvmt/uvmt_cv32e40p_debug_assert.sv
@@ -117,15 +117,14 @@ module uvmt_cv32e40p_debug_assert
             `uvm_error(info_tag,$sformatf("Debug mode with wrong cause after ebreak, case = %d",cov_assert_if.dcsr_q[8:6]));
 
     ////////////////////////////////////////////////////////////////////////////
-    // FIXME: start
-    //        It appears that the properties "p_cebreak_exception" and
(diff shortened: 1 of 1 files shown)


pascalgouedo (approved):

YoanFournier:
In issue https://github.com/openhwgroup/core-v-verif/issues/1084 the combined property takes into account the different decoding signals by using $rose(cov_assert_if.is_ebreak||cov_assert_if.is_cebreak). Since both properties have the same effect, I guess it should be equivalent in theory.

MikeOpenHWGroup:
> In issue #1084 the combined property takes into account the different decoding signals by using $rose(cov_assert_if.is_ebreak||cov_assert_if.is_cebreak). Since both properties have the same effect, I guess it should be equivalent in theory.

Agree.  If you would like to make that change that would be good.  However, I worry about me making a change to assertions while you are trying to close coverage.  It sounds like a good way to break something.

pascalgouedo:
Yes today in CV32E40P they are not combined.
Request from #1084 end of 2021 was to combine them and now get the point of #2346.
But as stated by @MikeOpenHWGroup we already started to collect coverage and changing that now is too risky.
Mike you can restore PR 2346 but with a more detailed text explaining the combination.
It was not the case and lead to misunderstanding from my side.

Issue #2134, opened 2023-08-18
Merge Script Improvements


silabs-robin:
### Task Outcome
Merge script gets additional features and tweaks (as per list below).

### Background information
We have started using the new script for doing S->X merges, and ideas for improvements are appearing.

### Location Information
https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/bin/merge.sh

### Completion Criteria
All feature suggestions are either implemented or rejected.

### Additional context

New suggestions:
* (All done.)

Implemented suggestions:
* Don't require `clonetb` to be run manually beforehand.
    * Rationale: It makes it less easy to use, it has caused confusion, and it might be unnecessary.
    * Comment: Could do `git init` instead?
    * https://github.com/openhwgroup/core-v-verif/pull/2282
* Option to generate "rejection diff".
    * Rationale: x-dv PR reviews are much easier to do with a rejection diff.
    * Comment: Example https://github.com/openhwgroup/cv32e40x-dv/pull/20
    * https://github.com/openhwgroup/core-v-verif/pull/2282
* Abort if merge is `Already up to date`.
    * Rationale: Otherwise the script will look like it did a merge, and confusion ensues.
* Generated branch names should be PR-title'able.
    * Rationale: Github defaults the PR title to the merge commit description, no need to manually come up with a title.
* "Press any key to continue" before major operations.
    * Rationale: Sometimes one would like more control and monitoring of what the script is about to do.
    * Comment: Could be an option to enable it (default is off).

Rejected suggestions:
* Echo what command is run ("set -x").
    * Rationale: It is not easy to understand what the script does when it runs.
    * (It prints to much, substitutions and everything.)
* Checkout cv32e40x-dv to a folder not called cv32e40x.
    * Rationale: Preserve work-in-progress that may be present in the cv32e40x folder.
    * Comment: Relates do the "don't require clonetb" suggestion.
    * (There is no reason to multitask like that. Do a fresh clone of cvverif, run the script, be done.)
* `git branch --unset-upstream` after cloning/branching operations.
    * Rationale: The default tracking is the upstream repo, and it is slightly impractical.
    * (One can easily get this idea again when the situation reoccurs, and it is easy to implement.)

kristinedosvik:
Agree with all the suggustions!

silabs-krdosvik:
[...]

PR #2349, opened 2024-01-15
address issue 2134

silabs-krdosvik:
Address issue https://github.com/openhwgroup/core-v-verif/issues/2134

Add check if merge is up to date
Better branch names, 
Add checks befor/after operations. 
Remove unused option

### DIFF
--- a/bin/merge.sh
+++ b/bin/merge.sh
@@ -21,6 +21,7 @@
 
 #Variables:
 date_time=$(date +%Y.%m.%d-%H.%M)
+your_cvv_branch=$(git rev-parse --abbrev-ref HEAD)
 
 
 usage() {
@@ -165,7 +166,7 @@ merge_xdev_into_sdev () {
 
 clone_x_dv() {
 
-  echo "=== Cloning x-dv ==="
+  echo "======= Cloning x-dv ======="
 
   read -p "This overwrites 'cv32e40x/'. Continue? (default: y, n) " yn
   continue_check $yn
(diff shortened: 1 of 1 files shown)


bin/merge.sh
```diff
@@ -183,20 +202,95 @@ rejection_diff() {
 
 }
 
+need_40s_40x-dv_merge(){
```
silabs-robin, line 206: I'm wondering if this needs to be so complicated.

Would something like this be equivalent?
```
theirs_head_commit=$(git rev-parse theirs_branch_name)
theirs_is_parent=$(git branch --contains $theirs_head_commit)
if [ $theirs_is_parent ]; then
    echo "No need to merge"
    exit 0
fi
```

bin/merge.sh
```diff
@@ -183,19 +203,112 @@ rejection_diff() {
...
+
+}
+
+main() {
   case $1 in
     "--s_into_x-dv")
       clone_x_dv
+      checkout_update_cv32e40s_dev_branch
+      need_40s_40x-dv_merge
+      checkout_your_ccv_branch
```
silabs-robin, line 306: core-c-verif

bin/merge.sh
```diff
@@ -183,19 +203,112 @@ rejection_diff() {
...
+  echo "   Compare the 40s and 40x version of the same file and know there are new changes in the 40s version"
+  echo "   that must manualy be merged to the 40x version. Delete the 40s version of the file after the manual merge is finished."
+
+}
+
+main() {
   case $1 in
     "--s_into_x-dv")
       clone_x_dv
+      checkout_update_cv32e40s_dev_branch
```
silabs-robin, line 304: Why is this needed?

bin/merge.sh
```diff
@@ -19,18 +19,15 @@
 # limitations under the License.
```
silabs-robin, line 19: Before merge, update https://github.com/openhwgroup/core-v-verif/issues/2134

(Let's leave this comment up until right before merge.)

PR #2352, opened 2024-01-22
Add ZCB extension to ISACOV

AyoubJalali:
This MR add to ISACOV the ZCB extension

### DIFF
--- a/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
+++ b/lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
@@ -1510,6 +1510,270 @@ covergroup cg_cj(
 
 endgroup : cg_cj
 
+covergroup cg_zcb_lbu(
+    string name,
+    bit reg_crosses_enabled,
+    bit reg_hazards_enabled,
+    bit rs1_is_signed,
+    bit rd_is_signed,
+    bit imm_is_signed
+) with function sample (
+    uvma_isacov_instr_c instr
+);
+  option.per_instance = 1;
+  option.name = name;
+
+  cp_rs1_value: coverpoint instr.rs1_value_type {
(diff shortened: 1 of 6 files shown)


MikeOpenHWGroup (commented):
LGTM @AyoubJalali, thanks for this.

I am not in a position to try to compile and run this, so I will defer merging to @ASintzoff and/or @JeanRochCoulon

JeanRochCoulon (approved):

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
```diff
@@ -1510,6 +1510,270 @@ covergroup cg_cj(
...
+    bins RD[] = {[0:7]} iff (instr.c_rd == instr.c_rs1);
+  }
+
+  `ISACOV_CP_BITWISE(cp_rs1_toggle, instr.rs1_value, 1)
+  `ISACOV_CP_BITWISE(cp_rd_toggle, instr.rd_value, 1)
+  `ISACOV_CP_BITWISE_0_0(cp_imm_toggle, instr.get_field_imm(), 1)
+
+endgroup : cg_zcb_lh
+
+covergroup cg_zcb_sh(
```
silabs-robin, line 1681: I wonder if we some time in the future could have less covergroups, if we find a smart way to structure them.

silabs-robin (commented):
This new code seems to follow the existing pattern well.
I have not fine-read but overall it looks good.

lib/uvm_agents/uvma_isacov/cov/uvma_isacov_cov_model.sv
AyoubJalali: I think the only way to do that, is to think of a way to remove the **ISACOV_CP_BITWISE** macros

PR #2353, opened 2024-01-24
Update cv32e40p/dev OBI Fcov to match master

MikeOpenHWGroup:
As per a request from BeeNee.Lim@dolphin.fr


YoannPruvost (approved):
All good on my side. Merging

PR #2354, opened 2024-01-24
ISACOV : Coverage Update and tuning

AyoubJalali:
As in the CVA6 we don't have this trap signal (with 12 bit) to identify ecall_ebreak exception code, to work-arroud this I add cause flied and assign it ti rvfi_cause to terminate witch exception code, now we can cover ECALL, EBREAK, C_EBREAK instructions.
The second change is about the DRET instruction, so I conditional it to be define in the coverage if only we're in debug mode, if not is an illegal instruction (and no need to cover it).
The last change related to ZCB instruction, I fix C.MUL instruction covergroup & remove ZCB_GROUP (isn't used)


MikeOpenHWGroup (approved):
LGTM.  I will defer to @ASintzoff for the final merge.

PR #2360, opened 2024-02-02
Illegal instruction generator for CV32E40P V1

GTumbush:
Illegal instruction generator for CV32E40P V1


MikeOpenHWGroup (approved):
Thanks @GTumbush, this is terrific!

PR #2363, opened 2024-02-06
Fix vsim compilation

MikeOpenHWGroup:
It seems that the Questasim (vsim) scriptware has fallen behind and no longer properly compiles and runs.  This PR attempts to restore that.


cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -458,11 +458,15 @@ module uvmt_cv32e40s_triggers_assert_cov
     //3) see a_dt_instr_trigger_hit_*, a_dt_load_trigger_hit_*, a_dt_store_trigger_hit_*, a_dt_exception_trigger_hit_*, a_dt_enter_dbg_reason
 
     //4)
+    // MIKET: this assertion fires, and I think it should (so maybe that's an error?).
     a_dt_tselect_higher_than_dbg_num_triggers: assert property(
       rvfi_if.rvfi_valid
       |->
-      tselect_post_state < CORE_PARAM_DBG_NUM_TRIGGERS
-    ) else `uvm_error(info_tag, "The CSR tselect is set to equal or higher than the number of trigger.\n");
+      rvfi_if.rvfi_rd1_wdata < CORE_PARAM_DBG_NUM_TRIGGERS
```
silabs-hfegran, line 465: @silabs-krdosvik Do you agree with this change?

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_data_independent_timing_assert.sv
```diff
@@ -75,6 +75,8 @@ module uvmt_cv32e40s_xsecure_data_independent_timing_assert
     //second branch instruction is retired.
     //There is therefor only 1 empty cycle after a branch instruction.
 
+`ifndef QUESTA_VSIM
+  // Questasim says: "Use of a method on an unclocked sequence is illegal"
```
silabs-hfegran, line 79: unclocked? does `@(posedge clk_i)` not imply clocked?

lib/isa_decoder/isa_decoder.sv
```diff
@@ -1534,9 +1534,12 @@
   // ---------------------------------------------------------------------------
   // Identify if a given instruction matches an expected instruction name
   // ---------------------------------------------------------------------------
+  // FIXME: is this a redundant declaration of match_instr?
```
silabs-hfegran, line 1537: Not redundant, should (eventually) replace the other function as this is much easier to use.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -414,7 +414,12 @@ interface uvma_rvfi_instr_if_t
     is_pma_instr_fault               = is_pma_instr_fault_f();
   end
 
+  // TODO: always_comb does not work here with VSIM
+`ifndef QUESTA_VSIM
```
silabs-hfegran, line 418: Does this not imply functionally different behavior? If I am not mistaken, always@ here is only updated on signal change, while always_comb is also evaluated at time 0.

cv32e40s/tests/uvmt/base-tests/uvmt_cv32e40s_base_test_elaboration_workarounds.sv
```diff
@@ -0,0 +1,33 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// Depreciate select errors to warnings.
+//
+// function void set_report_severity_id_override(uvm_severity cur_severity, string id, uvm_severity new_severity);
+// Note that "id" is the first argument in `uvm_error(), `uvm_warning(), etc.
+
+// A bunch of assertions in cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
+// are firing with Questasim (do they fire with other simulators?).
+`ifdef QUESTA_VSIM
```
silabs-hfegran, line 28: Do these asserts fire at time 0? If so then I suspect my alway_comb-comment to be related

lib/uvm_libs/uvml_sb/uvml_sb_cntxt.sv
```diff
@@ -51,12 +51,18 @@ class uvml_sb_cntxt_c#(
...
    realtime          total_latency          = 0;
    realtime          avg_latency            = 0;
-   
+
    // Events
-   uvm_event#(T_TRN)  exp_observed_e;
-   uvm_event#(T_TRN)  act_observed_e;
-   
-   
+   // FIXME: why doesn't Questa like parameterized uvm_events?
```
silabs-hfegran, line 56: This is UVM 1.2 - are you compiling with UVM 1.1?

mk/uvmt/vsim.mk
```diff
@@ -94,7 +108,7 @@ VLOG_LDGEN_FLAGS ?= \
 VOPT_LDGEN_FLAGS ?= \
                     -debugdb \
                     -fsmdebug \
-                    -suppress 7034 \
+					$(VSIM_SUPPRESS) \
```
silabs-hfegran, line 111: whitespace issues,  line 101 as well

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -633,14 +648,14 @@ interface uvma_rvfi_instr_if_t
   // Check if instruction is of a certain type
   // Usage: instr_mask sets the parts of the instruction you want to compare,
   //        instr_ref is the reference to match
-  function automatic logic match_instr(
+  function automatic logic rvfi_match_instr(
```
silabs-hfegran, line 651: If there is a naming conflict that questa cannot resolve here? From what I can see the 23.8.1 section of the LRM should cause this to be resolved correctly.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -947,6 +951,7 @@ module uvmt_cv32e40s_triggers_assert_cov
 
 
     //1)
+    // MIKET: how does this property work?
```
silabs-krdosvik, line 954: Oh.. I added a comment about it in the consequent, but it is still not clear?

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
```diff
@@ -956,6 +961,7 @@ module uvmt_cv32e40s_triggers_assert_cov
       rvfi_if.is_csr_write(ADDR_DCSR)
     ) else `uvm_error(info_tag, "Action is taken when there is a trigger match while in debug mode (dcsr is changed even though we dont do a dcsr write operation).\n");
 
+    // MIKET: how does this property work?
```
silabs-krdosvik, line 964: Same ^

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_data_independent_timing_assert.sv
MikeOpenHWGroup: Agreed.  The default clocking should be sufficient.  As per @silabs-hfegran suggestion I will install the latest version of Questa and see what we get.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
MikeOpenHWGroup: You are right @silabs-hfegran, the functional difference between `always_comb` and `always @(complete, sensitivity, list)` is that the former is evaluated at t=0 and the later is not.  In this case I submit we are safe because the conditions we are looking for will not be happening at t=0 (by definition all instructions are retired after the first instruction is fetched).

cv32e40s/tests/uvmt/base-tests/uvmt_cv32e40s_base_test_elaboration_workarounds.sv
MikeOpenHWGroup: Could be.  In any case, the need for this workaround goes away using Questa 2023.3_2.

lib/uvm_libs/uvml_sb/uvml_sb_cntxt.sv
MikeOpenHWGroup: This error when away with Questa 2023.3_2.  I did not alter the version of UVM and am not sure if the new version of Questa points to the new version of UVM or not.

mk/uvmt/vsim.mk
MikeOpenHWGroup: Fixed.  Thanks.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
MikeOpenHWGroup: Yes, you did! That comment is an older hash of `cv32e40s/dev`, I'll need to merge it in.  Thanks.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
MikeOpenHWGroup: Same

cv32e40s/tb/uvmt/uvmt_cv32e40s_xsecure_assert/uvmt_cv32e40s_xsecure_data_independent_timing_assert.sv
silabs-krdosvik: It is okey for me at least. But isnt this sequence clocked? or have I misunderstood something? 
image

lib/uvm_libs/uvml_sb/uvml_sb_cntxt.sv
silabs-hfegran: Sounds like it uses 1.2 (or IEEE std version) by default then, otherwise this should fail to compile.

cv32e40s/tb/uvmt/uvmt_cv32e40s_triggers_assert_cov.sv
silabs-robin: Not really relevant anymore, after https://github.com/openhwgroup/core-v-verif/pull/2367

But here's the answer:

`tselect_post_state` is the newest code (which this PR tries to remove).
`rvfi_if.rvfi_rd1_wdata` is the oldest code (which this PR wants in).

bilde

MikeOpenHWGroup:
@silabs-robin said:
>  Not really relevant anymore, after https://github.com/openhwgroup/core-v-verif/pull/2367

Agree.  I will close this PR once #2367 is merged in.

PR #2365, opened 2024-02-13
[Xcelium support] Bugfix uvm sysverilog compliant

CoralieAllioux:
Those minor fixes are suggested in the context of supporting Cadence Xcelium simulator for cva6 verification.
Those modifications allows Xcelium to compile and does not impact VCS usage.

This has been check with VCS 2021.09.

### DIFF
--- a/lib/uvm_agents/uvma_rvfi/uvma_rvfi_agent.sv
+++ b/lib/uvm_agents/uvma_rvfi/uvma_rvfi_agent.sv
@@ -95,10 +95,10 @@ class uvma_rvfi_agent_c#(int ILEN=DEFAULT_ILEN,
     */
    extern function void connect_trn_loggers();
 
-//    /**
-//     * Connects reference model port to monitor analysis ports.
-//     */
-//    extern function void connect_reference_model();
+   /**
+    * Connects reference model port to monitor analysis ports.
+    */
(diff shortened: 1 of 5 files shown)


CoralieAllioux:
Hi @MikeOpenHWGroup , the ECA for Isabelle is now solved for this PR

lib/uvm_agents/uvma_rvfi/uvma_rvfi_agent.sv
```diff
@@ -98,7 +98,7 @@ class uvma_rvfi_agent_c#(int ILEN=DEFAULT_ILEN,
    /**
     * Connects reference model port to monitor analysis ports.
     */
-   extern function void connect_reference_model();
+   //extern function void connect_reference_model();
```
MarioOpenHWGroup, line 101: Instead of commenting it, you can remove it as it is not implemented nor used anywhere

MarioOpenHWGroup (changes requested):
Besides the commented function, LGTM!

lib/uvm_agents/uvma_rvfi/uvma_rvfi_agent.sv
CoralieAllioux: This is done, thank you for the comment

PR #2364, opened 2024-02-13
Bugfix/unsupported with

CoralieAllioux:
This fix is suggested in the context of supporting Cadence Xcelium simulator for cva6 verification.

There is a VCS bug which does not support correctly the difference between with and iff keywords. A missing define needs to be added in order to allow xcelium to compile.

This has been check with VCS 2021.09.

PR #2366, opened 2024-02-13
[Xcelium support] Remove void from DPI definition

CoralieAllioux:
Those minor fixes are suggested in the context of supporting Cadence Xcelium simulator for cva6 verification.
Xcelium does not allow the DPI function to be to return a void value.

This means in CVA6 repository, the reference to this function must be updated! If a positive feedback is given, I should created a PR should on cva6 repository and update the core-v-verif hash submodule.

This has been check with VCS 2021.09: the behavior is the same than previous on our side.

### DIFF
--- a/lib/uvm_agents/uvma_axi/src/uvma_axi_pkg.sv
+++ b/lib/uvm_agents/uvma_axi/src/uvma_axi_pkg.sv
@@ -35,7 +35,7 @@ package uvma_axi_pkg;
 
    import "DPI-C" function read_elf(input string filename);
    import "DPI-C" function byte get_section(output longint address, output longint len);
-   import "DPI-C" context function void read_section_sv(input longint address, inout byte buffer[]);
+   import "DPI-C" context function read_section_sv(input longint address, inout byte buffer[]);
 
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Hi @CoralieAllioux.  This PR looks good, but as with #2364 your colleague Isabelle Schmid is the author of the Commit and must be covered by the ECA.
image

CoralieAllioux:
Hi @MikeOpenHWGroup, the ECA for Isabelle as the PR author is now solved.

MikeOpenHWGroup (commented):
@MarioOpenHWGroup can you comment?  The [call to read_section_sv](https://github.com/openhwgroup/core-v-verif/blob/c7d2077e8b4a1493aaecb89534562893064d5e0c/lib/uvm_agents/uvma_axi/src/seq/uvma_axi_fw_preload_seq.sv#L68) expects it to be void - will this cause problems?

MarioOpenHWGroup (approved):

MarioOpenHWGroup:
LGTM :smile:

PR #2368, opened 2024-02-16
Adding make targets needed to run riscof arch-tests for cv32e40x

silabs-kjetil:
Porting over the RISC-V architecture tests from E40P over to the cv32e40x. This makes it possible to run the full riscof complience test suite for the cv32e40x.

### DIFF
--- a/bin/clonetb
+++ b/bin/clonetb
@@ -81,7 +81,7 @@ clone() {
 clone_cv32e40x() {
   CV_CORE=cv32e40x
   VERIF_ENV_REPO=https://github.com/openhwgroup/cv32e40x-dv.git
-  VERIF_ENV_REF=a78635844f929fe7a852e3ddcd15dd809c377850
+  VERIF_ENV_REF=d2ef40f9338a05af4ed9551a1c75d53370744e64
   clone
 
   ignore_cloned_directory
--- a/mk/Common.mk
+++ b/mk/Common.mk
@@ -232,7 +232,7 @@ endif
 ###############################################################################
 # cfg
 CFGYAML2MAKE = $(CORE_V_VERIF)/bin/cfgyaml2make
-CFG_YAML_PARSE_TARGETS=comp ldgen comp_corev-dv gen_corev-dv test hex clean_hex corev-dv sanity-veri-run bsp compliance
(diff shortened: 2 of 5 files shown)


mk/riscof.mk
```diff
@@ -0,0 +1,95 @@
+###############################################################################
+#
+# Copyright 2023 OpenHW Group
+# Copyright 2023 Dolphin Design
```
MikeOpenHWGroup, line 4: This is a separate file.  If you are modifying the file substantially, please update the Copyright by adding "Copyright 2024 Silicon Labs".

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Approved.  I will defer to @silabs-robin or @silabs-hfegran  to execute the merge.

mk/riscof.mk
silabs-kjetil: Added copyright to the riscof.mk file, this is based on the e40p work and modified slightly.

mk/Common.mk
```diff
@@ -232,7 +232,7 @@ endif
 ###############################################################################
```
silabs-robin, line 232: We could also update the default hash for x-dv, if we want.

Though, we don't _need_ to do that for every x-dv update, so it's fine to leave it for now.

bilde

silabs-kjetil: How do we sequence this correctly? We first have to merge the cv32e40x-dv repo and then update this core-v-verif repo or is there another way.

silabs-robin: Alternative 1:
* Merge in arbitrary order.
* We can simply use `clonetb --x-main` to get the latest x-dv.

Alternative 2:
1. Merge the `x-dv` PR.
2. Update `VERIF_ENV_REF` in this `x/dev` PR.
    * (Note: Step 1 and 2 can be done in reverse order, since the currently latest commit on the x-dv PR will enter x-dv anyway. But that's an aside.)
3. Merge this PR in `x/dev`.

silabs-kjetil: I updated the VERIF_ENV_REF in this PR

bin/clonetb
```diff
@@ -81,7 +81,7 @@ clone() {
 clone_cv32e40x() {
```
silabs-robin, line 81: @silabs-kjetil I actually forgot to ask one thing, `ci_check` passes, right?

PR #2370, opened 2024-02-19
Cv32e40p/bsm update tb files ww07

dd-baoshan:
Add functional coverage exclusion files - assert, directive and interrupt


dd-baoshan:
clsoed due to PR wrong target

PR #2372, opened 2024-02-20
Some updates for easier Embench runs (+ some coremark/matmul_32b_float improvements)

pascalgouedo:


MikeOpenHWGroup (commented):
LGTM.  Only two files are non-embench-test specific (`mk/Common.mk` and `mk/uvmt/uvmt.mk`) and these changes look good.

I will defer to others who are more familiar with Embench to give the final approval and merge.  Unfortunately, that is probably **Marton Teilgard** who is no longer at SiLabs.  @silabs-hfegran, can you have a look?

cv32e40p/tests/embench/pylib/run_corev32.py
```diff
@@ -100,16 +100,11 @@ def build_benchmark_cmd(bench, args):
...
     cpu_per = float(1/(args.cpu_mhz*1_000_000))
     
-    if args.cfg == 'default':
-      core_config = f'CFG=default'
-    else:
-      core_config = f'CFG=pulp'
-
     #Utilize "make test" environment in core-v-verif
     if args.sim_parallel == 'YES':
-      return ['make', '-C', args.make_path, 'test', f"TEST=emb_{bench}", f"SIMULATOR={args.simulator}", 'USE_ISS=NO', core_config, 'COMP=NO', 'USER_RUN_FLAGS=+rand_stall_obi_disable']
```
silabs-hfegran, line 110: Why was this flag removed? I do not think the flag should be set in the python script directly but rather in the cfg yaml - but now it is in neither place, would this not cause the introduction of random obi bus stalls?

silabs-hfegran (commented):
Looks good to me, but I have one question wrt. random bus stalls that I would like answered before approving this PR

cv32e40p/tests/embench/pylib/run_corev32.py
silabs-hfegran: Thank you for the explanation - that should resolve my only concern for this PR

silabs-hfegran (approved):

PR #2374, opened 2024-02-22
Generating a .map file when building an elf executable

silabs-kjetil:
The map file makes it easier to see how we layout sections in memory and what symbols are included and dropped.

### DIFF
--- a/mk/Common.mk
+++ b/mk/Common.mk
@@ -505,6 +505,7 @@ else
 		-nostartfiles \
 		$(TEST_FILES) \
 		-T $(LD_FILE) \
+		-Xlinker -Map=$*.map \
 		$(LD_LIBRARY) \
 		-lcv-verif
 endif


mk/Common.mk
```diff
@@ -505,6 +505,7 @@ else
 		-nostartfiles \
 		$(TEST_FILES) \
 		-T $(LD_FILE) \
+		-Xlinker -Map=$*.map \
```
silabs-robin, line 508: `-Xlinker` passes the rest of this line to `ld`?

`-Map=$*.map` causes a new `mytest.elf.map` file to be written next to the elf?

silabs-kjetil: This gcc option passes the next option/token to the linker. This is used to pass options down to the linker that gcc does not understand. https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-Xlinker

The `$*` automatic variable will match the stem of the target which is the full path before `.elf` in our case. If for instance the test output/target is called `mytest.elf` then the map file will be called `mytest.map`. This also means that the map file will be placed next to the elf file like the other metadata files we generate. https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html

silabs-robin: Perfect

silabs-robin (approved):
Approving based on:
Seems useful.
Matches the typical way similar things are done.

PR #2375, opened 2024-02-22
Adding zalrsc support to uvma core config

silabs-kjetil:
Adding zalrsc config to uvma_core_cntrl_cfg in order to make it possible to differentiate between a core configuration where A extension is not enabled, but ZALRSC is enabled. This is needed in order to get the ISS misa configuration to behave correctly.

### DIFF
--- a/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
+++ b/lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
@@ -57,6 +57,7 @@
    rand bit                      ext_v_supported;
    rand bit                      ext_f_supported;
    rand bit                      ext_d_supported;
+   rand bit                      ext_zalrsc_supported;
    rand bit                      ext_zba_supported;
    rand bit                      ext_zbb_supported;
    rand bit                      ext_zbc_supported;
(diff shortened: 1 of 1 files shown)


lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
```diff
@@ -153,6 +154,7 @@
       `uvm_field_int(                          ext_f_supported                , UVM_DEFAULT          )
       `uvm_field_int(                          ext_d_supported                , UVM_DEFAULT          )
       `uvm_field_int(                          ext_v_supported                , UVM_DEFAULT          )
+      `uvm_field_int(                          ext_zalrsc_supported           , UVM_DEFAULT          )
```
silabs-hfegran, line 157: Can you add a default disabled soft contraint to the `constraint riscv_cons_soft { ... } ` as well? (you can override this soft constraint on the core-specific level if it needs to have another value)

silabs-hfegran (changes requested):

lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_cfg.sv
silabs-hfegran: you might want to check if the `get_misa` function on the bottom of this file is correct here as well

silabs-kjetil: I added a default disabled soft constraint to the value.

The `get_misa` function is correct. It should only report A extension if the full A extension is enabled, when only ZALRSC is enabled bit 0 should be zero.

Issue #2345, opened 2024-01-03
Use of two-state variables in RVFI Agent


MikeOpenHWGroup:
Hi @MarioOpenHWGroup, you are getting this issue because git says you are the author of this code.

In `uvma_rvfi_tdefs.sv` there is a packed struct called [st_rvfi](https://github.com/openhwgroup/core-v-verif/blob/4da8b11327634a2c22de9fd6b9923f28bdc19109/lib/uvm_agents/uvma_rvfi/uvma_rvfi_tdefs.sv#L29) that is used to create arguments to a couple of DPI functions to implement Spike Tandem operation.

All of the members of this struct are `longint unsigned` which are two-state variables.  In general, two-state variables are to be avoided as uninitialized data will be mapped to zero which may mask problems.

Can these be implemented as `logic[63:0]`?

MarioOpenHWGroup:
I think so, it should be a problem at all :+1:

PR #2379, opened 2024-02-26
[RVFI] CSRs comparison

MarioOpenHWGroup:
Modifications:
* `rvfi_agent` CSRs comparison.
* `riscv-isa-sim` CSRs modifications retrieve.

Note: Currently in Spike only 10 CSRs modifications are retrieved to avoid bloat in DPI.


lib/uvm_agents/uvma_core_cntrl/uvma_core_cntrl_tdefs.sv
```diff
@@ -729,17 +731,22 @@ typedef struct packed {
    // Common parameters
    int unsigned                 num_mhpmcounters;
    //uvma_core_cntrl_pma_region_c  pma_regions[];
+    //
+   bit  unsigned                 dram_valid;
```
MikeOpenHWGroup, line 735: These appear to be properties of the Spike ISS, not the core.  As far as I can tell, the `dram_base` and `dram_size` are not used to control or configure the RTL.  Should these members be in another configuration class elsewhere in the environment?

lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
```diff
@@ -29,7 +29,6 @@ class uvma_rvfi_instr_seq_item_c#(int ILEN=DEFAULT_ILEN,
    rand bit [ORDER_WL-1:0]       order;
    rand bit [ILEN-1:0]           insn;
    rand bit [TRAP_WL-1:0]        trap;
-   rand bit [XLEN-1:0]           cause;
```
MikeOpenHWGroup, line 32: Why delete `cause`?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_cfg.sv
```diff
@@ -33,12 +33,15 @@ class uvma_rvfi_cfg_c#(int ILEN=DEFAULT_ILEN,
 
    // Common options
    rand bit                      enabled;
-   rand bit                      csr_enabled;
+   rand bit                      intr_enabled;
```
MikeOpenHWGroup, line 36: It looks like `intr_enabled` replaces `csr_enabled`.  Don't we need both?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
```diff
@@ -35,7 +35,6 @@ interface uvma_rvfi_instr_if
     input [ORDER_WL-1:0]       rvfi_order,
     input [ILEN-1:0]           rvfi_insn,
     input [TRAP_WL-1:0]        rvfi_trap,
-    input [XLEN-1:0]           rvfi_cause,
```
MikeOpenHWGroup, line 38: Same question as before (why remove "cause"?).

lib/uvm_agents/uvma_rvfi/uvma_rvfi_tdefs.sv
```diff
@@ -25,51 +25,56 @@ typedef enum bit[MODE_WL-1:0] {
...
-
-   longint unsigned         rd2_addr;
-   longint unsigned         rd2_wdata;
-
-   longint unsigned         mem_addr;
-   longint unsigned         mem_rdata;
-   longint unsigned         mem_rmask;
-   longint unsigned         mem_wdata;
-   longint unsigned         mem_wmask;
+   bit [MAX_XLEN-1:0]         nret_id;
```
MikeOpenHWGroup, line 29: I think these members must have been defined as `longint unsigned` for a reason.  Maybe they do not work well with ImperasDV otherwise.  Do we need to change their type for Spike?

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
```diff
@@ -48,15 +64,21 @@ import "DPI-C" function void spike_step_struct(inout st_rvfi core, inout st_rvfi
 
     endfunction : rvfi_spike_step
 
+    st_core_cntrl_cfg m_core_cfg;
+
     function automatic void rvfi_initialize_spike(string core_name, st_core_cntrl_cfg core_cfg);
```
MikeOpenHWGroup, line 69: I know you have not added this in this PR, but it seems we have a dependency between RVFI and Spike.  This does not sound like a good idea.

lib/uvm_libs/uvml_sb/uvml_sb_cntxt.sv
```diff
@@ -51,12 +51,12 @@ class uvml_sb_cntxt_c#(
...
    realtime          total_latency          = 0;
    realtime          avg_latency            = 0;
-   
+
    // Events
-   uvm_event#(T_TRN)  exp_observed_e;
-   uvm_event#(T_TRN)  act_observed_e;
-   
-   
+   uvm_event exp_observed_e;
```
MikeOpenHWGroup, line 56: Why do we want to get ride of the parameterization?

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -18,6 +18,7 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
   memset(&rvfi, 0, sizeof(st_rvfi));
 
   this->taken_trap = false;
+  this->which_trap = 0;
```
MikeOpenHWGroup, line 21: This is a modification of Spike, is it not?  I thought we were _extending_ Spike, not modifying it.

MikeOpenHWGroup (commented):
Good stuff @MarioOpenHWGroup!  I have lots of questions, mostly about making sure we have a proper seperation of concerns.  Also, I cannot comment on the changes to the Spike model itself.  Perhaps @szbieg can comment.

lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
MarioOpenHWGroup: `cause` is not a field of the specification. It was made up by me so the time has come to delete it :smile:

lib/uvm_agents/uvma_rvfi/uvma_rvfi_cfg.sv
MarioOpenHWGroup: The `csr_enabled` was a field that I introduced some weeks ago but I realized that we have another field that is used for the same

lib/uvm_agents/uvma_rvfi/uvma_rvfi_instr_if.sv
MarioOpenHWGroup: Not part of the original spec

MikeOpenHWGroup: But it is used/needed by the CV32E40S version (and for good reason).

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
MarioOpenHWGroup: This is not a base spike file, so not a modification itself

lib/uvm_agents/uvma_rvfi/uvma_rvfi_utils.sv
MarioOpenHWGroup: Currently I think we do not have any dependency between rvfi and spike, this will only be executed if the uvmc_rvfi_reference_model calls it.

lib/uvm_agents/uvma_rvfi/uvma_rvfi_tdefs.sv
MarioOpenHWGroup: This was my code previously and I changed it to address the following issue https://github.com/openhwgroup/core-v-verif/issues/2345

MarioOpenHWGroup:
TODO once it is ready add `riscv-isa-sim` patches

lib/uvm_agents/uvma_rvfi/uvma_rvfi_cfg.sv
MarioOpenHWGroup: TODO double check the use of `intr_enabled`

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
MarioOpenHWGroup: TODO delete `which_trap`

lib/uvm_agents/uvma_obi_memory/src/uvma_obi_memory_assert.sv
```diff
@@ -218,8 +218,18 @@ module uvma_obi_memory_assert
     endcase
   endfunction : get_addr_lsb
 
+  // R-8 Data address LSBs must be consistent with byte enables on writes
```
MarioOpenHWGroup, line 221: @MikeOpenHWGroup double check if you can :smile:

lib/uvm_agents/uvma_rvfi/uvma_rvfi_constants.sv
```diff
@@ -42,11 +42,14 @@ localparam TRAP_DBG_CAUSE_WL     = 3;
...
 localparam DEFAULT_XLEN     = 32;
+localparam MAX_XLEN         = 64;
 localparam DEFAULT_NRET     = 1;
 
+localparam CSR_QUEUE_SIZE = 10;
+
 const string format_header_str = "%8s | RVFI | %8s | %6s | %8s | %8s | %s | %03s | %08s | %03s | %08s | %03s | %08s | %03s | %08s | %08s | %s";
-const string format_instr_str  = "%8s | RVFI | %8d | %6d | %8x | %8s | %s | x%-2d | %08x | x%-2d | %08x | x%-2d | %08x";
-`define FORMAT_INSTR_STR_MACRO "%8s | RVFI | %8d | %6d | %8x | %8s | %s | x%-2d | %08x | x%-2d | %08x | x%-2d | %08x"
+const string format_instr_str  = "%8s | RVFI | %8d | %6d | %8x | %8s | %s | x%-2d | %8x | x%-2d | %8x | x%-2d | %10x";
```
MarioOpenHWGroup, line 51: Align the alignments `%10x`

lib/uvm_agents/uvma_obi_memory/src/uvma_obi_memory_assert.sv
MikeOpenHWGroup: This is more tricky than it would seem!   A few comments:
1. `get_min_be()` is declared as `bit[1:0]`, but the `return` statements attempt to return a 4-bit value.   Probably a typo.
2. `get_min_be()` only works when DATA_WIDTH is 32.  I doubt this is worth fixing, but maybe you should call a `uvm_error()` if `get_min_be()` is called when DATA_WIDTH != 32.
3. OBI Requirement `R-8` was changed in v1.5.0.  The only version check in this module is for v1.2.0 (`IS_1P2`). The short term fix here is to add a `IS_1P5_OR_GREATER` instantiation parameter.  A much better fix is to refactor the code to use an enumerated type that declares the OBI version. I have created #2383 to discuss.

I think we should not wait to fix item 3 before merging this, but we do need to fix items 1 and 2.

MikeOpenHWGroup (approved):
Updates look good.  Approving.  @mario feel free to merge when ready.

MarioOpenHWGroup:
I need to add more changes to allow not to build spike

MikeOpenHWGroup:
Thanks for the updates @MarioOpenHWGroup.  I have approved this PR.  However, I think the final merge decision should be made by either @ASintzoff or @JeanRochCoulon.

PR #2384, opened 2024-03-07
Cleaning up the riscof makefile and fixing cwd issue

silabs-kjetil:
The riscof tool requires you to run the riscof tool from a certain path as the config file paths are relative to the location where you execute riscof from. So in this change we are more explicit on setting the current working directory before running any of the riscof commands.

Also removing much of the headers/banner from the makefile to make it more readable and compact.

### DIFF
--- a/mk/riscof.mk
+++ b/mk/riscof.mk
@@ -26,64 +26,16 @@
 
 RISCOF_ARCH_TEST_SUITE_PKG   := $(CORE_V_VERIF)/$(CV_CORE_LC)/vendor_lib/riscof/riscof_arch_test_suite
 RISCOF_SIM ?= NO
-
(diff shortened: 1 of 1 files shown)


silabs-kjetil:
@silabs-hfegran, @silabs-robin and @MikeOpenHWGroup  can you have a look at this PR?

silabs-robin (approved):
Approving based on:
`cwd` issue seems like a real problem, deserving of a fix.
The cleanup seems to make it easier to quickly understand what is where in the code.

However, I will not merge until @MikeOpenHWGroup has had a look.
Maybe the other users of riscof want to take a peek?

silabs-kjetil:
@silabs-hfegran and @silabs-robin this PR seems to be approved, is there anything missing before this PR can be merged?

PR #2387, opened 2024-03-18
Control Spike parameters from cmdline.

zchamski:
* vendor/riscv/riscv-isa-sim/riscv/Params.h (Param::setFromCmdline): New.
* vendor/riscv/riscv-isa-sim/spike_main/spike.cc (main): Declare 'params'<br>
  variable ahead of cmdline options.  Add cmdline option '--param PATH=VALUE'.<br>
  Add corresponding help string.

### DIFF
--- a/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
+++ b/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
@@ -240,7 +240,7 @@ void Processor::default_params(string base, openhw::Params &params) {
   params.set_bool(base, "misaligned", false, "false",
              "Support for misaligned memory operations");
 
-  params.set(base, "csr_counters_injection", std::any(false), "bool", "false",
+  params.set_bool(base, "csr_counters_injection", false, "false",
              "Allow to set CSRs getting values from a DPI");
 }
 
--- a/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
+++ b/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
@@ -106,8 +106,8 @@ Processor::Processor(
   Processor::default_params(base, this->params);
   Params::parse_params(base, this->params, params);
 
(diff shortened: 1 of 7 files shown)


MikeOpenHWGroup:
Hi @zchamski, who would be the best person to review this?  (I am unqualified!)

zchamski:
Hi @MikeOpenHWGroup , I think @MarioOpenHWGroup is the expert here 😃

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -207,42 +203,41 @@ void Processor::take_trap(trap_t &t, reg_t epc) {
...
              "MISA CSR Write Enable param enable");
-  params.set(base, "misa_we", any(false), "bool", "false",
+  params.set_bool(base, "misa_we", false, "false",
              "MISA CSR Write Enable value");
 
-  params.set(base, "misaligned", std::any(false), "bool", "false",
+  params.set_bool(base, "misaligned", false, "false",
              "Support for misaligned memory operations");
 
   params.set(base, "csr_counters_injection", std::any(false), "bool", "false",
```
valentinThomazic, line 243: This should probably be : 
`params.set_bool(base, "csr_counters_injection", false, "false",
             "Allow to set CSRs getting values from a DPI");`
(Will not build otherwise)

valentinThomazic (changes requested):

MarioOpenHWGroup:
Did you manage to reproduce the bug with the `any` field ?

zchamski:
> Did you manage to reproduce the bug with the `any` field ?

Hi @MarioOpenHWGroup, using `std::any` repeatedly prevented changing previously set param values, at least with the GCC from RHEL8.  Switching to explicitly typed setters (and separate per-type fields in `Param` struct) introduced in commit b14c6f3 solves the problem.

JeanRochCoulon (approved):

PR #2392, opened 2024-03-20
CV32E40Pv2 Verification update w12

XavierAubert:
This PR contains updates done by @YoannPruvost @pascalgouedo @dd-baoshan @dd-BeeNee and I during the last week. 

Since the last update, we have licences for the reference model again, so we were able to focus on fixes on pending topics. 

- @dd-BeeNee continued to work on waivers (PR XavierAubert#192)

- @dd-baoshan fixed issues in corner cases scenarios (4e538f800ff23d6a6143645447494c6ff6084866), improved functional coverage for fp instruction (1f9df153b0f5fc781d1d544d97d597502e13d636), and in general some fiddling with regression issues. 

- @pascalgouedo fixed an issue with riscvISACOV location and GCC toolchain  (commits ca2aa2a6b401cf540726254656d7611277ca2a5c and c51d8b50618026c14c723ff31969ed2fda7f82e5)

- @YoannPruvost fixed an issue with CSRs counters RVFI-side and added a new test for jalr (see PR XavierAubert#188 and XavierAubert#189)

- I worked on getting v1 tests back in our regression and added a new regress file for this purpose (see af010c0b735b3602037ba05e9483571176d692fc and XavierAubert#195)

PR #2393, opened 2024-03-21
Incorporate UNSUPPORTED_WITH kludge for VCS

MikeOpenHWGroup:
This pull-request resolves an issue with #2392 that impacts VCS.   It incorporates the so-called `UNSUPPORTED_WITH` work-around that is in use by multiple CORE-V-VERIF projects.  It is ugly, but works and needs to be in place until at least the next release of VCS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ VRMDATA
 *whetstone*
 *.do
 *eembc*
+*.code-workspace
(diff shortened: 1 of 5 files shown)


XavierAubert (approved):
LGTM, thank you for your help on this Mike

dd-baoshan:
For #2392 , I did not use this workaround. Instead, I changed with statement clause in the way that accepted in vcs.

dd-baoshan:
> > One question - I noticed you applied the workaround but I tested on my side without this workaround and I did not get any compilation/simulation error. Just curious, is that my fix is not fully work on your side?
> 
> Interesting. What version of VCS are you using? We have `VCS S-2021.09-SP1`.

We here are using vcs_mx "U-2023.03-SP1".

MikeOpenHWGroup:
> We here are using vcs_mx "U-2023.03-SP1".

Thanks @dd-baoshan.  I will work to obtain this version of VCS.  This will take some time, so would you be willing to accept this PR in its current state?  If so , let's merge it in and I will assign and Issue to myself to get the latest version of VCS.

dd-baoshan (approved):

PR #2396, opened 2024-03-25
Adding support for providing tests that are placed outside of tree

silabs-kjetil:
This change will make it possible to build and run simulation on testfolders that are placed in other directories than the hardcoded test subfolders.


silabs-kjetil:
@silabs-hfegran and @silabs-robin can you have a look at this PR?

bin/yaml2make
```diff
@@ -70,6 +70,11 @@ def read_file(test, type, run_index):
     matches = [os.path.join(TOPDIR, p, test, type) for p in TEST_PATHS 
                 if os.path.exists(os.path.join(TOPDIR, p, test, type))]
 
+    # Allow the user to specify full path to test folder
```
MikeOpenHWGroup, line 73: Where would these test folders exist?

mk/uvmt/uvmt.mk
```diff
@@ -86,7 +86,7 @@ SIM_RESULTS             ?= $(if $(CV_RESULTS),$(abspath $(CV_RESULTS))/$(SIMULAT
 SIM_CFG_RESULTS          = $(SIM_RESULTS)/$(CFG)
 SIM_COREVDV_RESULTS      = $(SIM_CFG_RESULTS)/corev-dv
 SIM_LDGEN_RESULTS        = $(if $(LDGEN),$(SIM_CFG_RESULTS)/$(LDGEN),$(SIM_CFG_RESULTS))
-SIM_TEST_RESULTS         = $(SIM_CFG_RESULTS)/$(TEST)
+SIM_TEST_RESULTS         = $(SIM_CFG_RESULTS)/$(TEST_NAME)
```
MikeOpenHWGroup, line 89: AFAIKT, this is the only use of the variable `$TEST_NAME`.   How does it work?

MikeOpenHWGroup (changes requested):
Left a couple of comments.

mk/uvmt/uvmt.mk
silabs-kjetil: The `$(TEST_NAME)` variables is already in used multiple places, even inside the general test target: https://github.com/openhwgroup/core-v-verif/blob/cv32e40x/dev/mk/uvmt/xrun.mk#L436

The TEST_NAME value is the simplified name of the test which is derived from the required `name` member inside of the test yaml file.

bin/yaml2make
silabs-robin, line 1: Do normal tests still work too?

(They probably do, but I cannot see that the PR says so explicitly.)

bin/yaml2make
MikeOpenHWGroup: Does this imply there are test-programs being used on CV32E40X that are **_not_** in this repo?

silabs-kjetil: The tests for CV32E40X are placed in this or the cv32e40x-dv repo, this change will make it possible to  for instance build temporary test programs for debugging purposes without having to place them into a hardcoded path.

bin/yaml2make
silabs-kjetil: ci_check is passing

silabs-robin (approved):
Approving based on:
I don't know what the `matches` variable signifies, or where `TEST_NAME` is defined, but I trust that this code has been tested and demonstrated to work.
And since `ci_check` is passing we know that it the change is non-breaking.

PR #2397, opened 2024-03-26
Restore DSim

MikeOpenHWGroup:
This PR re-establishes [Metrics DSim](https://www.metrics.ca/) as one of the SV simulators supported by CORE-V-VERIF.  Changes fall into one of the following categories:
- Fixes to Makefiles and associated scripts.
- Workarounds for specific DSim parser issues.  These are clearly identified in comments (look for "Metrics DSim") and do not alter the function or the code.
- Increased the verbosity level for printing the testcase and environment configurations (from UVM_NONE to UVM_DEBUG).

Tested these changes with VCS, Questasim and DSim.

### DIFF
--- a/cv32e40p/env/corev-dv/instr_lib/cv32e40p_float_instr_lib.sv
+++ b/cv32e40p/env/corev-dv/instr_lib/cv32e40p_float_instr_lib.sv
@@ -419,7 +419,13 @@ class cv32e40p_float_zfinx_base_instr_stream extends cv32e40p_base_instr_stream;
     if (more_weight_for_fdiv_fsqrt_gen || use_only_for_fdiv_fsqrt_gen) begin
       if (select_fp_instr) // is fp
         if ($urandom_range(1) || use_only_for_fdiv_fsqrt_gen)
-          include_instr = new[1] ($urandom_range(1) ? {FDIV_S} : {FSQRT_S});
+          // Metrics DSim throws an "InvalidTypeAssign" error given the conditional
+          // assigment below, so it has been expanded into two assignments.
+          //     include_instr = new[1] ($urandom_range(1) ? {FDIV_S} : {FSQRT_S});
+          if ($urandom_range(1))
+            include_instr = new[1] ({FDIV_S});
+          else
+            include_instr = new[1] ({FSQRT_S});
     end
 
     if (!include_load_store_base_sp) begin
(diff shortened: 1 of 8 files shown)


MikeOpenHWGroup:
Hi @XavierAubert, @YoannPruvost, @dd-baoshan and @dd-BeeNee, can one or more of you have a look at this?  The DSim-specific scriptware will not affect you, but changes to `cv32e40p/env/corev-dv/instr_lib/cv32e40p_float_instr_lib.sv`, `cv32e40p/env/uvme/cov/uvme_cv32e40p_cov_model.sv`, `cv32e40p/env/uvme/vseq/uvme_cv32e40p_random_debug_vseq.sv` and `cv32e40p/tb/uvmt/uvmt_cv32e40p_macros.sv` _might_ (I do not think so, but we should check).

I would also like your opinion on the change to verbosity for printing the testcase and environment configurations.

dd-baoshan (changes requested):
All files changes are fine for me except uvmt_cv32e40p_base_test.sv.
I think it is fine to always print one time for test and env cfg during simulation. The performance impact is not huge and by it gives users a glace on the configurations whether for test/tb debugging or developing purpose. 
Is there any concern that driven the changes of verbosity to UVM_DEBUG in base test?

MikeOpenHWGroup:
> All files changes are fine for me except uvmt_cv32e40p_base_test.sv. ... Is there any concern that driven the changes of verbosity to UVM_DEBUG in base test?

You are right @dd-baoshan, that there is no performance impact of printing the test and testbench configurations to stdout, and it is a useful debug aid.  Now that the environment is stable, this information is rarely used, so I increased the verbosity level for simply to reduce the amount of text printed to the screen.  If you would prefer to keep it, that is fine, I will restore the printing of these configs.

dd-baoshan:
> > All files changes are fine for me except uvmt_cv32e40p_base_test.sv. ... Is there any concern that driven the changes of verbosity to UVM_DEBUG in base test?
> 
> You are right @dd-baoshan, that there is no performance impact of printing the test and testbench configurations to stdout, and it is a useful debug aid. Now that the environment is stable, this information is rarely used, so I increased the verbosity level for simply to reduce the amount of text printed to the screen. If you would prefer to keep it, that is fine, I will restore the printing of these configs.

It would be great to retain this printing because users can view the configurations details in case for development or debugging new tests that extend this base test. I agreed that some configuration details are not really useful but we are using the sprint function to ease the print out of all properties that pass into objects util macro.

MikeOpenHWGroup:
> It would be great to retain this print out

Done.  I have pushed in an update to this PR that restores the original verbosity of the cfg dumps.  Please have a look.

dd-baoshan (approved):

PR #2413, opened 2024-04-18
Add riscv-isa-sim libraries for VCS

MarioOpenHWGroup:
* Add riscv-isa-sim libraries (riscv,fesvr,disasm)


MikeOpenHWGroup (approved):
LGTM, thanks @MarioOpenHWGroup

PR #2419, opened 2024-05-02
Revert PR #2418

JeanRochCoulon:
This commit creates a new dependency with a new library. As GitHub Actions and Thales CI do not support it, the CI fails. The new tool needs to be installed to use the master core-v-verif.


ASintzoff (approved):
LGTM

PR #2427, opened 2024-05-10
Merge master to srelease

silabs-robin:
This PR merges `master` into `cv32e40s/release`.

`ci_check` passes.

Conflict resolution diff: https://github.com/silabs-robin/core-v-verif/compare/merge_master_to_srelease__conflicts..silabs-robin:core-v-verif:merge_master_to_srelease

PR #2432, opened 2024-05-14
[WIP] Reference model

torjeikenes:
This draft PR brings in some of my work implementing a reference model with the goal of verifying asynchronous events. The RM contains a core-specific pipeline shell, built around a core-independent ISS. The ISS is currently Spike from vendor/riscv/, with modifications done on another branch from master. This PR is therefore not currently functional and depends on  #2427 before both branches can be merged. As discussed with @silabs-robin, I am submitting this PS as a draft in order to get feedback and track what changes need to be made.


mk/uvmt/vsim.mk
```diff
@@ -177,8 +176,25 @@ VSIM_UVM_ARGS      = +incdir+$(UVM_HOME)/src $(UVM_HOME)/src/uvm_pkg.sv
...
+GCC_PATH			= $(HOME)/opt/gcc-11.4.0/bin/gcc
+
+VSIM_FLAGS        	+= -dpicpppath $(GCC_PATH)
+VSIM_FLAGS 			+= -noautoldlibpath
+
+# This must currently always be exported, since the reference_model_wrap is used in uvmt_cv32e40s_tb.sv
+# TODO: Make a "dummy" reference model, that is loaded when it is disabled, like ImperasDV. 
+# Or find another solution to handle swapping different ISSs
+export FILE_LIST_RM      ?= -f $(DV_UVMT_PATH)/reference_model.flist
+
```
torjeikenes, line 188: Since ImperasDV support is fairly hard-coded, "disabling" IDV and enabling the RM this way is not optimal. Suggestions on how this can be done is very welcome

cv32e40s/tests/programs/custom/custom_interrupt_test/custom_interrupt_test.c
```diff
@@ -0,0 +1,882 @@
...
+
+/*
+
+    retval = memTest();
+    if (retval != EXIT_SUCCESS) {
+      return retval;
+    }
+
+  // Test 1
+  retval = test1();
```
silabs-robin, line 265: All the tests are commented out?

cv32e40s/tests/programs/custom/custom_interrupt_test/README.md
```diff
@@ -0,0 +1,3 @@
+CV32E40S Core Interrupt Testcases
+=============================
+This directory contains manually written directed testcases for interrupt testing of CV32E40S.
```
silabs-robin, line 3: Could you add some more info in the readme, about why this test exists and approximately what it does?
(I know that is more documentation than some of the other tests have, but I think it would be useful for a future person that stumbles upon this code.)

mk/uvmt/vsim.mk
```diff
@@ -177,8 +176,25 @@ VSIM_UVM_ARGS      = +incdir+$(UVM_HOME)/src $(UVM_HOME)/src/uvm_pkg.sv
...
+ifeq ($(call IS_YES,$(USE_RM)),YES)
```
silabs-robin, line 189: I'm struggling to see what is completely new vs what is copied from "master" and will disappear from this diff after the other PR is merged.

lib/reference_model/pipeline_shell.sv
silabs-robin, line 1: The directory of the reference model could maybe use a readme as well.

silabs-robin (commented):
Hi. Thanks for the PR.
I have added comments for some of my immediate thoughts.

cv32e40s/tb/uvmt/reference_model.flist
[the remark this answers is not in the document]
torjeikenes: I took inspiration from similar ImperasDV files for these types of files, which is why it is placed here. But I agree that it could make sense to have it in lib/

lib/uvm_agents/uvma_rvfi/uvma_rvfi_tdefs.sv
```diff
@@ -18,6 +18,71 @@
...
+   bit [MAX_XLEN-1:0]         mem_wmask;
+
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_valid;
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_addr;
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_rdata;
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_rmask;
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_wdata;
+   bit [CSR_QUEUE_SIZE-1:0] [MAX_XLEN-1:0] csr_wmask;
+
+} st_rvfi;
```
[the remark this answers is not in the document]
torjeikenes: Yes, that is correct.

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
```diff
@@ -0,0 +1,293 @@
...
+);
+
+`define ASSERT_EQUAL(NAME, NAME_STR) \
+    NAME``_a: assert property( @(posedge rvfi_core.clk) \
+      rvfi_rm.valid |-> (rvfi_rm.NAME == rvfi_core.NAME)) \
+      else `uvm_error(NAME_STR, $sformatf("PC: %8h RM=%8h CORE=%8h", rvfi_rm.pc_rdata, $sampled(rvfi_rm.NAME),  $sampled(rvfi_core.NAME)))
+
+`define ASSERT_EQUAL_RANGE(NAME, NAME_STR, FROM, TO) \
+    NAME``_a: assert property( @(posedge rvfi_core.clk) \
+      rvfi_rm.valid |-> (rvfi_rm.NAME``[FROM:TO] == rvfi_core.NAME``[FROM:TO])) \
```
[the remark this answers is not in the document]
torjeikenes: It is used because the current spike RVFI output does not support multiple memory operations in a single instruction, so spike only outputs on the last 4 bits of mem_xmask.

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
```diff
@@ -0,0 +1,293 @@
...
+ * Compare the RVFI interface of the core and the reference model.
+ * Use assertions to support potentially support formal verification
+ */
+module rvfi_compare(
+  rvfi_if_t rvfi_core,
+  rvfi_if_t rvfi_rm
+);
+
+`define ASSERT_EQUAL(NAME, NAME_STR) \
+    NAME``_a: assert property( @(posedge rvfi_core.clk) \
```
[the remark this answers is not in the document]
torjeikenes: `` is a delimiter to separate between the argument and the rest of the string

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
```diff
@@ -0,0 +1,293 @@
...
+  `ASSERT_EQUAL(trap, "RVFI_TRAP")
+  `ASSERT_EQUAL(halt, "RVFI_HALT")
+  `ASSERT_EQUAL(dbg, "RVFI_DBG")
+  `ASSERT_EQUAL(dbg_mode, "RVFI_DBG_MODE")
+  `ASSERT_EQUAL(nmip, "RVFI_NMIP")
+  `ASSERT_EQUAL(intr, "RVFI_INTR")
+  `ASSERT_EQUAL(mode, "RVFI_MODE")
+  `ASSERT_EQUAL(rd1_addr, "RVFI_RD1_ADDR")
+
+  // TODO: Spike outputs wrong values for rs1_addr and rs2_addr
```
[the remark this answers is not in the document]
torjeikenes: It is a problem with the current spike RVFI implementation where it  reads rs1_addr and rs2_addr directly from the the insn bits, regardless of instruction type. For example reading rs2 for an I-type instruction without the rs2 field, it instead reads the same bits from the IMM field.

mk/uvmt/vsim.mk
torjeikenes: All of these changes are new, and not a part of master

cv32e40s/tests/programs/custom/custom_interrupt_test/README.md
torjeikenes: This directed test was only used while debugging to test for some specific cases. I am not sure if it really is necessary, so I am considering removing it

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-robin, line 1: bilde

The Eclipse "eca" check is failing, and I suspect it is because the commits are done from different emails.
One way to fix it is using interactive rebase to update the author information.

torjeikenes: I think that fixed it

lib/reference_model/reference_model.flist
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+
+//+incdir+${RM_HOME}
+${RM_HOME}/iss_wrap_pkg.sv
+${RM_HOME}/pipeline_shell_stage.sv
+${RM_HOME}/pipeline_shell_controller_cv32e40s.sv
```
silabs-robin, line 21: With this, the common "reference_model" component has a hard dependency on a 40s-specific file?

I think that as many files as possible should be common, but the common code should not have a hard dependency.
Whether or not the core-specific file could live in a sort of "add-on"-like directory in the common dir or whether it should be in a core-specific dir (e.g. "tb/"), I am not sure.

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-robin: bilde

Awesome :)

lib/reference_model/reference_model.flist
torjeikenes: The idea is to encapsulate all the core-specific implementation details into the controller module, so only this would have to be modified for different cores. I am also unsure where it makes the most sense to place this file and how to choose between different cores. Since the controller is a module inside pipeline_shell, which is inside reference_model, it might be "ugly" to place this in "tb/"?

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
```diff
@@ -1764,6 +1764,10 @@ module uvmt_cv32e40s_tb;
       uvmt_cv32e40s_imperas_dv_wrap imperas_dv (rvvi_if);
     `endif
 
+    `ifdef USE_RM
```
torjeikenes, line 1767: @silabs-robin After some testing, it looks like this does not work properly and does not compile reference_model, even though USE_RM is set. Do you know what could be wrong or a better way to enable or disable the reference model?

mk/uvmt/vsim.mk
silabs-robin: > enabling the RM this way is not optimal

Could you point me specifically to what way you mean?

Do you mean setting args on the cli to say no to one and yes to the other? Or are you referring to the defines used in the systemverilog code?

cv32e40s/tests/programs/custom/custom_interrupt_test/README.md
silabs-robin: I lean towards adding the new test, because the cost of adding it is relatively low and it provides additional test stimuli.

lib/reference_model/iss_wrap_pkg.sv
```diff
@@ -0,0 +1,98 @@
...
+
+import "DPI-C" function bit spike_set_debug(bit debug_req, int unsigned revert_steps, bit debug_allowed);
+
+
+    function automatic void iss_init(string binary);
+
+        string rtl_isa;
+        string rtl_priv;
+
+        rtl_isa = "rv32imc_zicsr_zifencei_zca_zcb_zcmp_zcmt_zba_zbb_zbc_zbs";
```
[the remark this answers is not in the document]
torjeikenes: That was an overlook on my part, and "c" should not be set. The ISA string should probably also be configurable depending on the core configuration, instead of static like it is now.
But to answer the question, it looks like spike only uses zca internally, but enables zca if c is set while parsing the string.

mk/uvmt/vsim.mk
torjeikenes: Now, USE_ISS is synonymous with ImperasDV, which is a bit unspecific. This could instead be USE_IDV, or different ISSs could be chosen with ISS=IMPERASDV or ISS=RM, etc. Additionally, the name "ISS" is not that accurate either since both of these are more than an ISS. However, is this probably not within the scope of this PR?

silabs-robin: I like the idea of keeping `USE_ISS`, and to let some additional variable default to `IDV` but be overridable to e.g. `RM`.
Because then the old behavior will be preserved (no breakage), and the new behavior will be possible to enable.s

lib/reference_model/iss_wrap_pkg.sv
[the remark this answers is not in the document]
MarioOpenHWGroup: Probably you would like to take a look to the reference model code that is already in master to avoid duplication https://github.com/openhwgroup/core-v-verif/blob/master/lib/uvm_components/uvmc_rvfi_reference_model/rvfi_spike.sv

lib/reference_model/pipeline_shell.sv
silabs-robin: Which, preferably, should contain usage instructions.

For instance:
```
For running with cv32e40s, run `make` as normal, but include the following.
Set <name of required env var>.
[...]
Call make as normal.
```

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
silabs-robin, line 1: > depends on https://github.com/openhwgroup/core-v-verif/pull/2427

That one got merged just now.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
silabs-robin: It is not too easy to get an overview of how everything connects right now.
If I could try it out hands-on that would probably help. (ref https://github.com/openhwgroup/core-v-verif/pull/2432#discussion_r1607880003)

cv32e40s/tests/programs/custom/custom_interrupt_test/custom_interrupt_test.c
```diff
@@ -0,0 +1,882 @@
...
+            // Select a wakeup interrupt and enable only it
+            mie_enable(irq);
+
+            // Prep the IRQ ID Q to be empty, we need to detect if any interrupts (or none) taken
+            irq_id_q[0] = -1;
+            irq_id_q_ptr = 0;
+
+            // Set the global MSTATUS.MIE
+            // Note that WFI should ignore this (but subsequent ISR will not be taken if MSTATUS.MIE == 0)
+            if (gmie)
```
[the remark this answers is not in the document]
torjeikenes: I removed these tests now. Since they were direct copies of `interrupt_test`, the same problem is probably still in `interrupt_test`?

lib/reference_model/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+To run the reference model alongside cv32e40s, run `make` as normal, but include the following:
+
+- set `USE_RM=YES` to enable the reference model.
+
+- set `SPIKE_HOME=<path to riscv/riscv-isa-sim>` to use another spike version.
+If not set, spike from `vendor/riscv/riscv-isa-sim` will be used.
+
+For example:
+`make test TEST=hello-world USE_RM=YES SPIKE_HOME=~/core-v-verif/vendor/riscv/riscv-isa-sim`
```
silabs-robin, line 19: When running `make comp USE_RM=1` I get
`uvmt_cv32e40s_tb.sv,1771|53): instance 'uvmt_cv32e40s_tb.reference_model' of design unit 'uvmt_cv32e40s_reference_model_wrap' is unresolved in 'worklib.uvmt_cv32e40s_tb:sv'.`

Is this a known problem with a known fix?

silabs-robin: I think I know the problem... One moment, I shall look for a fix.

silabs-robin: I will not fix it now. But for the future:
This should be ported to Xcelium, or rather could live in `mk/uvmt/uvmt.mk` instead of `mk/uvmt/vsim.mk` as much as possible.
But, I don't know about the `-dpicpppath` vsim option and how to port this etc, so I conclude that now is not the time to fix that.
(I will use vsim while testing.)

silabs-robin, line 19: When running `make test TEST=hello-world USE_RM=1 SIMULATOR=vsim` I get
`# ** Error: (vsim-3763) SystemVerilog DPI cannot access file '/work/ropeders/src/core-v-verif-tne/tools/spike/lib/libriscv.so'`.

Spike was compiling for quite a time, but it seems it didn't succeed.
(If it fails at compiling, then I don't know why that isn't flagged to make, but that might be a problem with the "vendorization" and not with this PR.)
Here are a few messages that seemed relevant and might shine light on the situation:
```
make[1]: Leaving directory `/work/ropeders/src/core-v-verif-tne/vendor/riscv/riscv-isa-sim/build'
Makefile:349: warning: overriding recipe for target `disasm.o'
Makefile:349: warning: ignoring old recipe for target `disasm.o'
make[1]: Entering directory `/work/ropeders/src/core-v-verif-tne/vendor/riscv/riscv-isa-sim/build'
[...]
make[1]: Circular libfesvr.so <- libfesvr.so dependency dropped.
make[1]: Circular libcustomext.so <- libcustomext.so dependency dropped.
make[1]: Circular libsoftfloat.so <- libsoftfloat.so dependency dropped.
../scripts/mk-install-dirs.sh /work/ropeders/src/core-v-verif-tne/tools/spike/lib
```

@torjeikenes is this expected to work on the branch as it is now, or does it require a merge from `cv32e40s/release` first?

torjeikenes: Just merged in the new changes from srelease, which changes spike compilation slightly. Is this still a problem?

mk/uvmt/vsim.mk
torjeikenes: This functionality was added with the merge from master

lib/reference_model/README.md
silabs-robin: Thanks. I am having other technical difficulties on my end, so I will try it out again once that is resolved.

silabs-robin: I'm trying `make test TEST=hello-world SIMULATOR=vsim USE_ISS=YES ISS=RM`,
after having done `git clean -xdff .` at the top directory,
then I get to:
```
../fesvr/fesvr_dpi.cc:22:10: fatal error: svdpi.h: No such file or directory
   22 | #include <svdpi.h>
```

torjeikenes: Right. You have to do `export QUESTASIM_HOME=<path to questasim>`. I think svdpi.h is included into spike [here](https://github.com/openhwgroup/core-v-verif/blob/3206c283fc50d32c4691df50eb9554e2584e2c03/vendor/riscv/riscv-isa-sim/Makefile.in#L72)

torjeikenes: Hmm, this does not happen to me... Could it be something [here](https://github.com/openhwgroup/core-v-verif/blob/3206c283fc50d32c4691df50eb9554e2584e2c03/mk/uvmt/vsim.mk#L200) since I don't have access to IMPERAS_HOME, but you do?

silabs-robin: > Right. You have to do `export QUESTASIM_HOME=<path to questasim>`. I think svdpi.h is included into spike [here](https://github.com/openhwgroup/core-v-verif/blob/3206c283fc50d32c4691df50eb9554e2584e2c03/vendor/riscv/riscv-isa-sim/Makefile.in#L72)

This is something from upstream "master" in core-v-verif, right?

Because spike started compiling fine before, and I did not get that error.
Plus, `lib/dpi_dasm` uses that same `svdpi.h` file and `make dpi_dasm SIMULATOR=vsim` works fine.
It just seems like it shouldn't be a technical necessity.

--------

> IMPERAS_HOME

I unsat everything IMPERAS and it got me even further.
`version 'GLIBC_2.28' not found`

This might be from the Red Hat version I am using.
If one specifies `ISS=RM` I don't think this system should be sensitive to whether imperas variables are set in the environment.

silabs-robin: @torjeikenes do you know how I could test just the vanilla spike "vendorization" without the refmodel? E.g. some other project like cva6 or e20. Or, if you know some other way, I would like to run something with just the vendorizationa and not the refmodel to see if the reason I can't run this stems from the vendorization itself or from something particular to the refmodel.

silabs-robin: Making sure I'm using the latest questasim and gcc (2024.1_1, 11.4.0).
I'm getting all kinds of different error messages no matter what I try.

```
make test TEST=hello-world SIMULATOR=vsim USE_ISS=YES ISS=RM
[...]
make[1]: Entering directory '/work/ropeders/src/core-v-verif-tne/vendor/riscv/riscv-isa-sim/build'
../scripts/mk-install-dirs.sh /work/ropeders/src/core-v-verif-tne/tools/spike/include
make[1]: *** No rule to make target 'config.h', needed by 'install-config-hdrs'.  Stop.
```

torjeikenes: > do you know how I could test just the vanilla spike "vendorization" without the remodel? 
You could attempt to build spike using the normal [build steps](https://github.com/riscv-software-src/riscv-isa-sim?tab=readme-ov-file#build-steps) from /vendor/riscv/riscv-isa-sim/ in either master or s/release

These steps work for me in s/release:

`cd core-v-verif/vendor/riscv/riscv-isa-sim`
`mkdir build`
`cd build`
`../configure --prefix=<core-v-verif path>/tools/spike`
`make`
`make install`

silabs-robin: > These steps work for me in s/release

Do they work after a fresh clone or `git reset . ; git checkout . ; git clean -xdff .`?

I get:
```
../fesvr/elfloader.cc:3:10: fatal error: config.h: No such file or directory
 #include "config.h"
          ^~~~~~~~~~
compilation terminated.
```
But when I grep for `config.h` there is no readme mentioning that I need to create it myself or anything like that.

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
silabs-robin: I cannot run the system itself, and the fact that the testbench atm unconditionally compiles the reference_model blocks normal xcelium runs, so we need to move the reference_model compilation back into the conditional and figure out why the define is not working, and I cannot do it as I can't run the system.

> does not compile reference_model, even though USE_RM is set

How do you know it is set?
If you use the conditional in the tb and add some lines of illegal SystemVerilog it should reveal whether it is set or not.
If not set, then let's examine the makefile to figure out what is missing.
If set, then the tb at least attempts to instantiate the rm, and some other files might not be compiled.

lib/reference_model/README.md
torjeikenes: It seems like `config.h` is generated when running `../configure`. Do you have a `config.h` file in ..../riscv-isa-sim/build/ after running configure?

silabs-robin: No.

`configure` says:
```
config.status: creating config.h
config.status: config.h is unchanged
```

But there is no such file:
```
$ pwd
/work/ropeders/src/core-v-verif-tne/vendor/riscv/riscv-isa-sim
$ find | grep config.h
./config.h.in
```

torjeikenes: Hmm, that is weird. Are you able to clone and build vanilla spike https://github.com/riscv-software-src/riscv-isa-sim?

silabs-robin: No.
Maybe someone (Mike or Mario maybe) who are already able to compile that could sanity check this PR branch afterwards.
In the meantime, I think it is best to fix everything else, and then come back to this in the end.
I need to at least be able to run hello world and pass `ci_check` with Xcelium without changing anything in my standard setup, and all other review conversations should be resolved.

silabs-robin: For the record: I created this ticket https://github.com/riscv-software-src/riscv-isa-sim/issues/1683

cv32e40s/tb/uvmt/uvmt_cv32e40s_tb.sv
torjeikenes: This should be solved now

silabs-robin: Was the problem the type of whitespace?

torjeikenes: I also added the USE_RM define to VLOG_FLAGS

lib/reference_model/README.md
silabs-robin: Well, I am back to errors like `failed: /lib64/libstdc++.so.6: version 'GLIBCXX_3.4.20' not found`.
I was close to getting it to work yesterday, but I don't know which planets need to be in alignment for this to work, so I will focus on testing Xcelium with Imperas for now.

silabs-robin: `ci_check` passes for me.
So when all feedback conversations are resolved I think we will be good.

cv32e40s/tests/programs/custom/custom_interrupt_test/custom_interrupt_test.c
torjeikenes: Done

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
```diff
@@ -0,0 +1,293 @@
...
+ */
+module rvfi_compare(
+  rvfi_if_t rvfi_core,
+  rvfi_if_t rvfi_rm
+);
+
+`define ASSERT_EQUAL(NAME, NAME_STR) \
+    NAME``_a: assert property( @(posedge rvfi_core.clk) \
+      rvfi_rm.valid |-> (rvfi_rm.NAME == rvfi_core.NAME)) \
+      else `uvm_error(NAME_STR, $sformatf("PC: %8h RM=%8h CORE=%8h", rvfi_rm.pc_rdata, $sampled(rvfi_rm.NAME),  $sampled(rvfi_core.NAME)))
```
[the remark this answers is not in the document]
silabs-robin: There are some cases that are not updated to that format. E.g. this:
https://github.com/openhwgroup/core-v-verif/pull/2432/files#diff-1b3730da7a533ab04f3049cdf44cc04ea4ec2b6259c9489259c28079d12d957eR81

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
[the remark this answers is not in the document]
silabs-robin: I searched for `property` and found one more case:
https://github.com/openhwgroup/core-v-verif/pull/2432/files?show-viewed-files=true&file-filters%5B%5D=#diff-1b3730da7a533ab04f3049cdf44cc04ea4ec2b6259c9489259c28079d12d957eR48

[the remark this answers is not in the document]
torjeikenes: Fixed now

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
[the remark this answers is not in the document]
torjeikenes: Fixed now

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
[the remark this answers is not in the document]
silabs-robin: Thanks.

lib/reference_model/reference_model.flist
silabs-robin: Then there is just this and the other thread remaining.

> it might be "ugly" to place this in "tb/"?

If it logically can be made to fit there, that would be better than for other cores to be forced to use something 40s-specific.

torjeikenes: Moved the controller to the core-specific files now

cv32e40s/tb/uvmt/uvmt_cv32e40s.flist
```diff
@@ -65,5 +65,10 @@ ${DV_UVMT_PATH}/uvmt_cv32e40s_pkg.sv
 // Depends on env
 ${FILE_LIST_IDV}
 
+// Reference model wrapper files
+${FILE_LIST_RM}
+${TBSRC_HOME}/uvmt/uvmt_cv32e40s_reference_model_controller.sv
+//${TBSRC_HOME}/uvmt/uvmt_cv32e40s_reference_model_wrap.sv
```
silabs-robin, line 71: This line is commented out. Can it be deleted?

lib/reference_model/reference_model.flist
silabs-robin: Looks good. Thanks.
(I will try it out with Xcelium and our typical flow. Will start new thread if anything goes wrong.)

lib/reference_model/iss_wrap_pkg.sv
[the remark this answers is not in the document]
silabs-robin: I think this PR is near ready to be merged.
What is missing to resolve this thread?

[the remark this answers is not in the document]
silabs-robin: > What is missing to resolve this thread?

If there are more commits incoming, I will wait.
Otherwise, I will run `ci_check` and prepare to merge.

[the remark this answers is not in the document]
silabs-robin: This change is deceptively simple and has far-reaching implications that are more effort atm that it is worth. So at this point, it will be considered a nice-to-have. The functionality of the content added in the PR is the main priority.
Resolving the thread.

cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_controller.sv
silabs-robin, line 1: @torjeikenes 

```
$ git status
On branch referenceModel
Your branch is up to date with 'tne/referenceModel'.

nothing to commit, working tree clean


$ git log -1
commit 692db72488497c27b45db149c54bad6edc46a3f3 (HEAD -> referenceModel, tne/referenceModel, cv32e40s/release__merge_test_ci_check__tne/referenceModel)
Author: torjeikenes <torje.n.eikenes@gmail.com>
Date:   Tue Jun 11 14:11:42 2024 +0200

    move core-specific controller module to cv32e40s specific files


$ cd cv32e40s/sim/uvmt/


$ make test TEST=hello-world

[...]

xmvlog: *E,NOPBIND (/work/ropeders/src/core-v-verif-tne/cv32e40s/tb/uvmt/uvmt_cv32e40s_reference_model_controller.sv,31|22): Package iss_wrap_pkg could not be bound.
        input pipe_stage_t if_id_pipe_i,
```

torjeikenes: Hmm. These commands work on my setup running vsim, but I don't have access to Xcelium. Can you see if you can find the cause since I am not able to reproduce the error?

silabs-robin: It works on the next to latest commit.

silabs-robin: But I can indeed run it with vsim...

silabs-robin: @torjeikenes Do you recognize from the error message what could be wrong?
Do you know which files I can expect to find the related files?
And do you know the desired order they should be compiled in?

torjeikenes: It might be because I only modified vsim.mk but not xrun.mk. [Here](https://github.com/torjeikenes/core-v-verif/blob/692db72488497c27b45db149c54bad6edc46a3f3/mk/uvmt/vsim.mk#L219) in vsim.mk I export `FILE_LIST_RM` which is used in `uvmt_cv32e40s.flist`. This must probably also be exported for xrun

silabs-robin: Awesome! Thanks. Running `ci_check` now.

silabs-robin (approved):
Approving based on:
* All feedback resolved.
* ci_check passes (how we typically run it with xrun).

PR #2437, opened 2024-05-16
Fix tb issues

dd-baoshan:
1) Update bhavioral logic on collecting clk_cycle_window for b2b addr phase scnearios
2) Fix incorrect constraint from parent after factory overidden by child class


MikeOpenHWGroup (approved):
LGTM @dd-baoshan.  Passes local CI on Questasim (vsim).

PR #2438, opened 2024-05-16
New define to enable ImperasDV model CV32E40P patch.

pascalgouedo:


MikeOpenHWGroup (approved):
LGTM, thanks @pascalgouedo.  I am not able to test the changes to `xsim.mk`, but it looks fine and the `vsim.mk` changes work.

PR #2439, opened 2024-05-16
CV32E40P fix for mem stress scenario with compressed instructions

XavierAubert:
The following code generation command would fail (see attached file for error message);

```
make gen_corev-dv TEST=corev_rand_illegal_instr_test CV_CORE=cv32e40p CFG=pulp_fpu_zfinx_1cyclat TEST_CFG_FILE=floating_pt_zfinx_instr_en,disable_all_trn_logs SIMULATOR=vsim SEED=1248233755
```

This was due to memory stress streams randomly reserving all allowed registers for compressed instructions (S0 to A5) 

The fix consists in checking the number of S0:A5 registers that have been already reserved before randomization, and making sure that there are less registers taken in that range.  

[mem_stress_with_compress_failure.log](https://github.com/openhwgroup/core-v-verif/files/15335288/mem_stress_with_compress_failure.log)

### DIFF
--- a/cv32e40p/env/corev-dv/cv32e40p_instr_base_test.sv
+++ b/cv32e40p/env/corev-dv/cv32e40p_instr_base_test.sv
@@ -45,6 +45,7 @@ class cv32e40p_instr_base_test extends corev_instr_base_test;
     override_privil_reg();
     override_debug_rom_gen();
     override_instr_stream();
+    override_load_store_lib();
     super.build_phase(phase);
   endfunction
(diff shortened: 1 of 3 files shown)


MikeOpenHWGroup (approved):
LGTM @XavierAubert.  Passes a local regression and the test invocation you provided (`make gen_corev-dv TEST=corev_rand_illegal_instr_test ...`) also compiles and runs under Questasim.

PR #2443, opened 2024-05-17
(Redo) Add store restore SP for fpu streams when load store with sp is enabled

dd-baoshan:
…s enabled


MikeOpenHWGroup (approved):
LGTM and passes local CI regression.

PR #2446, opened 2024-05-23
s/release Vsim Fixes

silabs-robin:
This PR fixes Questasim compilation after the previous PR.

After https://github.com/openhwgroup/core-v-verif/pull/2427, compilation did not work for me using vsim because of this need for an explicit cast.

Test status:
* xrun ci_check - pass
* vsim hello world - pass

### DIFF
--- a/lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
+++ b/lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
@@ -469,7 +469,7 @@ function void uvma_rvfi_instr_seq_item_c::rvfi2seq(st_rvfi rvfi);
     trap = rvfi.trap;
     halt = rvfi.halt;
     intr = rvfi.intr;
-    mode = rvfi.mode;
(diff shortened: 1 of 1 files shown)


lib/uvm_agents/uvma_rvfi/seq/uvma_rvfi_instr_seq_item.sv
```diff
@@ -469,7 +469,7 @@ function void uvma_rvfi_instr_seq_item_c::rvfi2seq(st_rvfi rvfi);
     trap = rvfi.trap;
     halt = rvfi.halt;
     intr = rvfi.intr;
-    mode = rvfi.mode;
+    mode = uvma_rvfi_mode'(rvfi.mode);
```
silabs-robin, line 472: @torjeikenes I thought the problem was bigger, because I was getting errors from `cv32e40s_rvfi_sim_trace.sv` as well, but that turned out to be the vsim version I was using defaulted to an older simulator version.

This one line seems to be the entire issue.

silabs-hfegran (approved):
Looks good to me, thanks for the fix!

PR #2448, opened 2024-05-27
Fix incorrect stack flow during nested irq

dd-baoshan:


MikeOpenHWGroup (approved):
LGTM @dd-baoshan and passes a local CI.  Mergining....

Issue #2450, opened 2024-05-29
QUESTASIM_HOME for "spike_lib" Neither Documented nor Checked


silabs-robin:
### Type
* Compile error (hopefully nobody has committed anything that doesn't compile!)
* Confusing or extraneous status or error messages
* Other.

### Steps to Reproduce
Try to build `spike_lib`.
E.g. approximately `cd cv32e40s/sim/uvmt && make spike_lib`.

### Additional context
Here is the error message you get:
`../fesvr/fesvr_dpi.cc:22:10: fatal error: svdpi.h: No such file or directory`
There is no clue that QUESTASIM_HOME is missing.
`grep`'ing from top-level reveals no documentation.
What should happen is that whenever you run a target that requires this env var to be set, then it should automatically be checked and then printed an error if it is not set ([example](https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/dev/bin/clonetb#L51), in shell but you can do the same in make or any other language).

MarioOpenHWGroup:
Hi Robin, makes sense I will open a PR with the check :+1:

MarioOpenHWGroup:
Probably the best solution is to not let the user compile spike_lib if the SIMULATOR var is not set as it requires any of the three big ones

silabs-robin:
I had SIMULATOR set

MarioOpenHWGroup:
Okeey I will submit a fix :+1:

MikeOpenHWGroup:
A little bit of digging produced this:
```
ifeq ($(ENVXXX),)
   $(error ENVXXX is undefined)
endif
```
Adding this to the top of `vsim.mk` will ensure that the check runs for every invocation of make, regardless of the target. **_Important_**: the `ifeq` and `endif` must start in column 0 and the indentation of `$error` must be with spaces, not tabs.

PR #2452, opened 2024-05-29
Fix riscv-isa-sim compilation in Makefiles

MarioOpenHWGroup:
This PR fixes https://github.com/openhwgroup/core-v-verif/issues/2450

### DIFF
--- a/mk/uvmt/vcs.mk
+++ b/mk/uvmt/vcs.mk
@@ -52,6 +52,7 @@ VCS_UVM_VERBOSITY ?= UVM_MEDIUM
 #VCS_UVMHOME_ARG ?= /opt/uvm/1800.2-2017-0.9/
 #VCS_UVMHOME_ARG ?= /opt/synopsys/vcs-mx/O-2018.09-SP1-1/etc/uvm
 VCS_UVMHOME_ARG  ?= /synopsys/vcs/S-2021.09-SP1/etc/uvm
+export VCS_HOME  ?= $(abspath $(shell which $(VCS))/../../)
 VCS_UVM_ARGS     ?= +define+UVM +incdir+$(VCS_UVMHOME_ARG)/src $(VCS_UVMHOME_ARG)/src/uvm_pkg.sv +UVM_VERBOSITY=$(VCS_UVM_VERBOSITY) -ntb_opts uvm-1.2
 
 VCS_COMP_FLAGS  ?= -lca -sverilog \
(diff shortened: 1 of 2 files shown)


MarioOpenHWGroup:
Also adding $(LIBS) variables in `gen_corev-dv`

mk/uvmt/vcs.mk
```diff
@@ -52,6 +52,7 @@ VCS_UVM_VERBOSITY ?= UVM_MEDIUM
 #VCS_UVMHOME_ARG ?= /opt/uvm/1800.2-2017-0.9/
 #VCS_UVMHOME_ARG ?= /opt/synopsys/vcs-mx/O-2018.09-SP1-1/etc/uvm
 VCS_UVMHOME_ARG  ?= /synopsys/vcs/S-2021.09-SP1/etc/uvm
+export VCS_HOME  ?= $(abspath $(shell which $(VCS))/../../)
```
MikeOpenHWGroup, line 55: Will this work reliably for every flavour of Linux?

mk/uvmt/vsim.mk
```diff
@@ -34,6 +34,7 @@ VCOVER                  = vcover
 VWORK     				= work
 VSIM_COV_MERGE_DIR      = $(SIM_CFG_RESULTS)/$(CFG)/merged
 UVM_HOME               ?= $(abspath $(shell which $(VLIB))/../../verilog_src/uvm-1.2/src)
+export QUESTASIM_HOME  ?= $(abspath $(shell which $(VLIB))/../../)
```
MikeOpenHWGroup, line 37: Same question.

MikeOpenHWGroup (commented):
Hi @MarioOpenHWGroup.  Do we know that the export of VCS_HOME will work in all cases?  Perhaps we should simply check that VCS_HOME is set, rather than attempt to set it ourselves.

mk/uvmt/vcs.mk
silabs-robin: Note that this is how it has been done for `UVM_HOME` for 4 years:
https://github.com/openhwgroup/core-v-verif/blame/0036e69b4d37388c86b3138ca2e31d0e889e11f7/mk/uvmt/vsim.mk#L36
`UVM_HOME               ?= $(abspath $(shell which $(VLIB))/../../verilog_src/uvm-1.2/src)`

silabs-robin: With that in mind, I think we could trust this way of doing it.
It is anyway better to have this than today's current failure that simply appears completely unrelated.
I would vote for Mario's fix, and rather change it to the "user must specify this env var" _if_ someone comes along and reports problems.
4 years of this way of doing it suggests that it is quite robust.

MarioOpenHWGroup: Yes, this way has been the way of the Makefiles all this time

MikeOpenHWGroup: Alrighty then.   Let's merge it as-is.

MikeOpenHWGroup (approved):

PR #2454, opened 2024-05-29
Questasim - Fix ADV_DEBUG

silabs-robin:
This PR fixes `ADV_DEBUG=1` to work for vsim.

--------

Without `ADV_DEBUG=1` I get an ancient GUI (where I can't even find the "start" button).

Without this fix, I run
`make test TEST=hello-world SIMULATOR=vsim USE_ISS=0  GUI=1 ADV_DEBUG=1 WAVES=1`
and get:
```
Fatal Error :
Design File ../design.bin not found ...
```
Because the `../design.bin` path is slightly off.

--------

Test status:
* Hello world with GUI (ADV_DEBUG) - Pass
* `ci_check` - Pass, except...
    * ... 4 runs say failed even though they print pass in the log, because Questa prints a message about not liking seeds with negative values.
    * I think it is extremely unlikely that this PR has introduced this, as it touches something unrelated, so since all logs print "SIMULATION PASSED" I deem it as passed.

### DIFF
--- a/mk/uvmt/vsim.mk
+++ b/mk/uvmt/vsim.mk
@@ -42,9 +42,9 @@ DPI_INCLUDE            ?= $(abspath $(shell which $(VLIB))/../../include)
 
 # Default flags
 VSIM_USER_FLAGS         ?=
-VOPT_COV  				?= +cover=setf+$(RTLSRC_VLOG_TB_TOP).
-VSIM_COV 				?= -coverage
-VOPT_WAVES_ADV_DEBUG    ?= -designfile design.bin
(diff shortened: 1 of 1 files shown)


mk/uvmt/vsim.mk
silabs-robin, line 1: Hi @MikeOpenHWGroup. I figured out the issue with `vsim` and `ADV_DEBUG=1`.

However, I put the PR in "Draft" mode because I have not run `ci_check`.

silabs-hfegran (approved):
Looks good to me

PR #2457, opened 2024-05-31
Add RTL code coverage waiver for some instances in FPU

dd-BeeNee:
Just a brief explanation for why this update: the RTL code coverage for these design modules are already fully covered in earlier reports (if we check file by file per 'Design Unit'). However on close look, some instances are not individually fully covering for each use. It is also due to how the information is displayed by the tools misleaded us when we scrub the coverage analysis by going thru 'Design Unit'.


pascalgouedo:
Yes

Issue #1933, opened 2023-05-25
Mention of UVM run_phase in uvm_tests.rst is incorrect


MikeOpenHWGroup:
Email from dave.rich@siemens.com:
```
I was reading the text on this page: https://github.com/openhwgroup/core-v-verif/blob/master/docs/VerifStrat/source/uvm_tests.rst

In the run phase the base test will assert the fetch_en input to the core which signals it to start running. The timing of this is randomized but keep in mind that it will always happen after reset is de-asserted (because resets are done in the reset phase, which always executes before the run phase).

This is not correct. The UVM’s run_phase starts at time 0 in parallel with the reset_phase. You use run_phase with components that do not need to be aware of other phases, or if your sequences are set up to be unaware of phases. (i.e. you would start a reset_sequence in the run_phase, followed by other sequences. Usually monitors are not aware of other phases.

But if you do choose to use the reset_phase, then you would use the main_phase instead of the run_phase.

See https://www.chipverify.com/uvm/uvm-phases

-Dave
```

subbu009:
As Dave said, in UVM the `run_phase `starts in parallel with other run time phases like `reset_phase `and `configure_phase`. One of the base testcase [here ](https://github.com/openhwgroup/core-v-verif/blob/master/cv32e40p/tests/uvmt/base-tests/uvmt_cv32e40p_base_test.sv) uses `run_phase `which only contains `watchdog_timer()`, which is essentially used to stop simulation if timeout is reached. So, according to this testcase it works fine even the other run time phases are used along with `run_phase`. But if some functional sequence needs to be started in `run_phase`, then as Dave said `main_phase `must be used to run after `configure_phase`. Correct me if I am wrong.

subbu009:
Yes, I am interested. Help me get started

subbu009:
Hi @MikeOpenHWGroup , 
If we replace the `run_phase `with `main_phase`, then all the three phases `reset_phase`, `configure_phase `and `main_phase `will run one after the other as expected. I guess this will fix the issue explained by Dave.

About the "on-the-fly" reset, could you please elaborate more on the task as what is expected?

MikeOpenHWGroup:
> If we replace the run_phase with main_phase, then all the three phases reset_phase, configure_phase and main_phase will run one after the other as expected.

Yes, that is my understanding as well.

> About the "on-the-fly" reset, could you please elaborate more on the task as what is expected?

[...]

PR #2458, opened 2024-05-31
modified run_phase -> main_phase

subbu009:
As per issue #1933 , I have modified `run_phase` to `main_phase` to be in consistent with the use of other run time phases (`reset_phase`, `confgure_phase`) in the base test, firmware test and firmware riscof test. 
Now the problem with line 154 in [uvmt_cv32e40p_firmware_test.sv](https://github.com/openhwgroup/core-v-verif/blob/cv32e40p/dev/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv) is statements below are waiting on posedge of resetn. If we use `run_phase` it works fine because `run_phase` starts at time 0 in parallel with `reset_phase` and senses the posedge of resetn whenever it happens. But when I use `main_phase` instead, I observed the line 154 waits on for resetn posedge and it never happens because the `reset_phase` has already completed. Due to this the test hangs and never finishes. So, to run ci_check, I have commented the line 154, and checks passed as expected. The question : is it okay if we do not wait for posedge of resetn? if we were to wait on reset, then it is better to fork off the current reset, configure and run phase tasks in only one run_phase so the reset posedge in one thread can be detected correctly by other forked thread.

On the other hand, if we would like to add _on-the-fly reset_ feature as @MikeOpenHWGroup explained, then it is better to use reset, configure and main phases instead of `run_phase`.

### DIFF
--- a/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
+++ b/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
@@ -76,7 +76,7 @@ class uvmt_cv32e40p_firmware_test_c extends uvmt_cv32e40p_base_test_c;
    /**
     *  Enable program execution, wait for completion.
     */
-   extern virtual task run_phase(uvm_phase phase);
+   extern virtual task main_phase(uvm_phase phase);
 
    /**
    * Start random debug sequencer
@@ -122,10 +122,10 @@ task uvmt_cv32e40p_firmware_test_c::configure_phase(uvm_phase phase);
 endtask : configure_phase
 
 
-task uvmt_cv32e40p_firmware_test_c::run_phase(uvm_phase phase);
+task uvmt_cv32e40p_firmware_test_c::main_phase(uvm_phase phase);
 
    // start_clk() and watchdog_timer() are called in the base_test
-   super.run_phase(phase);
+   super.main_phase(phase);
 
    if ($test$plusargs("gen_random_debug") || $test$plusargs("gen_reduced_rand_dbg_req")) begin
     fork
@@ -151,7 +151,7 @@ task uvmt_cv32e40p_firmware_test_c::run_phase(uvm_phase phase);
    end
 
    phase.raise_objection(this);
-   @(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
+   //@(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
    repeat (33) @(posedge env_cntxt.clknrst_cntxt.vif.clk);
    `uvm_info("TEST", "Started RUN", UVM_NONE)
    // The firmware is expected to write exit status and pass/fail indication to the Virtual Peripheral
@@ -165,7 +165,7 @@ task uvmt_cv32e40p_firmware_test_c::run_phase(uvm_phase phase);
    `uvm_info("TEST", $sformatf("Finished RUN: exit status is %0h", vp_status_vif.exit_value), UVM_NONE)
    phase.drop_objection(this);
 
-endtask : run_phase
+endtask : main_phase
 
 task uvmt_cv32e40p_firmware_test_c::reset_debug();
     uvme_cv32e40p_random_debug_reset_c debug_vseq;
--- a/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_riscof_firmware_test.sv
+++ b/cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_riscof_firmware_test.sv
@@ -84,7 +84,7 @@ class uvmt_cv32e40p_riscof_firmware_test_c extends uvmt_cv32e40p_base_test_c;
     /**
      *  Enable program execution, wait for completion.
      */
-    extern virtual task run_phase(uvm_phase phase);
+    extern virtual task main_phase(uvm_phase phase);
 
     extern function void write_riscof_signature();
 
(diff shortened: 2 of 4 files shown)


cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
```diff
@@ -151,7 +151,7 @@ task uvmt_cv32e40p_firmware_test_c::run_phase(uvm_phase phase);
    end
 
    phase.raise_objection(this);
-   @(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
+   //@(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
```
MikeOpenHWGroup, line 154: I am not sure this is safe.  Typically, the environment will de-assert reset very early in the simulation, but this may not be the general case.  Here is a simplified version of the code-snipit:
```
phase.raise_objection(this);
@(posedge reset_n);
repeat (33) @(posedge clk);
wait (exit_valid)
repeat (100) @(posedge clk);
phase.drop_objection(this);
```
If we remove `@(posedge reset_n);` **and** reset is not de-asserted in the first 33 clock cycles, it is possible that `exit_valid` will be 'bX.  I _think_ we need something like this:
```
phase.raise_objection(this);
if (first_reset) @(posedge reset_n);
repeat (33) @(posedge clk);
wait (exit_valid)
repeat (100) @(posedge clk);
phase.drop_objection(this);
```
Of course, the above assumes that we add a new member, `first_reset`, that is 'b1 from t=0 until the first posedge of reset_n.

cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_riscof_firmware_test.sv
```diff
@@ -139,12 +139,12 @@ function void uvmt_cv32e40p_riscof_firmware_test_c::post_randomize();
...
 
-task uvmt_cv32e40p_riscof_firmware_test_c::run_phase(uvm_phase phase);
+task uvmt_cv32e40p_riscof_firmware_test_c::main_phase(uvm_phase phase);
 
-    super.run_phase(phase);
+    super.main_phase(phase);
 
     phase.raise_objection(this);
-    @(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
+    //@(posedge env_cntxt.clknrst_cntxt.vif.reset_n);
```
MikeOpenHWGroup, line 147: Same comment.

MikeOpenHWGroup (commented):
Hi @subbu009, this looks _mostly_ good.  I have a question about how the "first reset" is handled.

Hi @datum-dpoulin, can you comment?

cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_firmware_test.sv
subbu009: Yes. But my understanding is the main_phase is not detecting posedge on reset_n because it is not yet started when during reset_phase. This works fine if we use run_phase as it starts at time 0 along with reset_phase.
I would also like to know how to open the gui of Questa Sim for the simulation?

MikeOpenHWGroup: Right!  I have not looked at the UVM run-flow in some time and forgot that both `reset_phase` and `main_phase` are sub-phases of the `run_phase`, and that `reset_phase` is guaranteed to execute before `main_phase.

Given that, please delete the `@(posedge reset_n);` line.  Leaving it there commented-out is not necessary and will eventually confuse somebody (like me!) in the future.

As for Questasim GUI, you are on your own.  I rarely use simulator GUIs and always post-process waveform viewing.

subbu009: Removed the line please check.

I would like to know how to enable waveform files for QuestaSim in this environment. Can see waveforms later by viewing the .wlf file

MikeOpenHWGroup:
This all looks good to be @subbu009, and it passes a local CI regression.  This branch is in active use by a team at Dolphin who are trying to close off coverage, so I will want their approval before merging.

@XavierAubert, @YoannPruvost, @dd-BeeNee, can one or more of you have a look?

cv32e40p/tests/uvmt/compliance-tests/uvmt_cv32e40p_riscof_firmware_test.sv
dd-baoshan: to retain the same behavior of "@(posedge env_cntxt.clknrst_cntxt.vif.reset_n);" when changing run_phase to main_phase, 
we can change  this line to "wait(env_cntxt.clknrst_cntxt.vif.reset_n === 1'b1);". 
Although it seems reduncant but at least it will ensure the dut is out of reset prior main execution (similar to intention in run_phase)

subbu009: This makes sense @dd-baoshan. Added this line in latest [commit](https://github.com/openhwgroup/core-v-verif/pull/2458/commits/b3ed47838d4b972763fa4ed69f1b75d4fff449c8).

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Great suggestion @dd-baoshan, thanks a lot!

I have approved this pull-request, so it can be merged at any time.  However, I will defer to the Dolphin before doing so because I know the team is currently very focused on completing this project within the next couple of weeks.

PR #2459, opened 2024-05-31
 updated test list & added missing test cases in DV Plans

XavierAubert:
The test list file is now aligned with the v2 regress files content (i.e. I cross-checked that every test inside a v2 regress is recorded inside the test list file).

Meanwhile, I also added a part of missing links between test list items and their related DVPlans. Some of them are still missing due to them being v1 tests not necessarily linkable to any of v2 DVPlans, and also test items that have been put in place to improve code coverage only (e.g. illegal instructions) that are actually not inside any existing DV Plan...


MikeOpenHWGroup (approved):
Thanks for the update @XavierAubert!

PR #2462, opened 2024-06-05
Spike modifications for cv32e40s reference model

torjeikenes:
This PR adds the changes to Spike necessary to support the reference model in #2432.
I have attempted to split it into two main contributions:
- Added spike support for CV32E40S 
  - Configuration of memory regions
  - Modifications to CSRs
  - Modifications to RVFI
- Modified spike to work with the reference model
  - Injection of interrupts and debug requests
  - Modifications to interrupt handling in spike, adding an interrupt_allowed check and external interrupts
  - Store and revert to previous states

(@silabs-robin, I can't add reviewers)

### DIFF
--- a/vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
+++ b/vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
@@ -1,0 +1,415 @@
+diff --git a/vendor/riscv/riscv-isa-sim/riscv/Proc.cc b/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
+index c3d4149..ce0078c 100644
+--- a/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
++++ b/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
+@@ -20,15 +20,23 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
+   this->taken_trap = false;
+   this->which_trap = 0;
+ 
+-  rvfi.pc_rdata = this->get_state()->pc;
++  // Store the state before stepping
++  state_t prev_state = *this->get_state();
++
+   processor_t::step(n);
+ 
++  rvfi.pc_rdata = this->last_pc;
++
++  rvfi.pc_wdata = this->get_state()->pc; // Next predicted PC
++
+   rvfi.mode = this->get_state()->last_inst_priv;
+   rvfi.insn =
+       (uint32_t)(this->get_state()->last_inst_fetched.bits() & 0xffffffffULL);
+ 
+   // TODO FIXME Handle multiple/zero writes in a single insn.
+   auto &reg_commits = this->get_state()->log_reg_write;
++  auto &mem_write_commits = this->get_state()->log_mem_write;
++  auto &mem_read_commits = this->get_state()->log_mem_read;
+   int xlen = this->get_state()->last_inst_xlen;
+   int flen = this->get_state()->last_inst_flen;
+ 
+@@ -37,13 +45,69 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
+   rvfi.rs2_addr = this->get_state()->last_inst_fetched.rs2();
+   // TODO add rs2_value
+ 
+-  rvfi.trap = this->taken_trap;
+-  rvfi.trap |= (this->which_trap << 1);
+ 
++  if(this->next_rvfi_intr){
++    rvfi.intr = next_rvfi_intr;
++    this->next_rvfi_intr = 0;
++
++  }
++  
++  // Output dbg caused from EBREAK the previous instruction
++  if(this->next_debug) {
++    rvfi.dbg = this->next_debug;
++    this->next_debug = 0;
++  }
++
++  if(this->state.debug_mode  && (prev_state.debug_mode == 0)){
++    // New external debug request
++    if((this->halt_request != HR_NONE)){ 
++      rvfi.dbg = this->get_state()->dcsr->cause;
++      rvfi.dbg_mode = 1;
++
++    // EBREAK
++    } else if(this->get_state()->dcsr->cause == DCSR_CAUSE_SWBP) {
++      // EBREAK will set debug_mode to 1, but we should report this at the next instruction
++      rvfi.trap |= 1 << 0; //trap [0]
++      rvfi.trap |= 1 << 2; //debug [2]
++      rvfi.trap |= 0xE00 & ((DCSR_CAUSE_SWBP) << 9); //debug cause [11:9]
++      
++      this->next_debug = DCSR_CAUSE_SWBP;
++    }
++  }
++
++  // Set dbg_mode to 1 the first instruction in debug mode, but delay turning 
++  // off dbg_mode to the next instruction after turning off to keep dbg_mode on for dret
++  if( (this->halt_request != HR_NONE)  && (prev_state.debug_mode == 0)) {
++    rvfi.dbg_mode = 1;
++  } else {
++    rvfi.dbg_mode = prev_state.debug_mode;
++  }
++
++
++  if(this->taken_trap) {
++    //interrrupts are marked with the msb high in which_trap
++    if(this->which_trap & ((reg_t)1 << (isa->get_max_xlen() - 1))) { 
++      //Since spike steps two times to take an interrupt, we store the intr value to the next step to return with rvfi
++      this->next_rvfi_intr |= 1 << 0; //intr [0]
++      this->next_rvfi_intr |= 1 << 2; //interrupt [2]
++      this->next_rvfi_intr |= 0x3FF8 & ((this->which_trap & 0xFF) << 3); //cause[13:3]
++    } else{
++      rvfi.trap |= 1 << 0; //trap [0]
++      rvfi.trap |= 1 << 1; //exception [1]
++      rvfi.trap |= 0x1F8 & ((this->which_trap) << 3); //exception_cause [8:3]
++      //TODO:
++      //debug_cause     [11:9] debug cause
++      //cause_type      [13:12]
++      //clicptr         [14]  CLIC interrupt pending
++      this->next_rvfi_intr = rvfi.trap; //store value to return with rvfi.intr on the next step
++    }
++  }
++
++  uint64_t last_rd_addr = 0;
++  uint64_t last_rd_wdata = 0;
+   bool got_commit = false;
+   for (auto &reg : reg_commits) {
+-      if ((reg.first >> 4) > 32) {
+-          if ((reg.first >> 4) < 0xFFF) {
++    if((reg.first & 0xf) == 0x4) { //If CSR
+             for (size_t i = 0; i < CSR_SIZE; i++) {
+                 if (!rvfi.csr_valid[i]) {
+                     rvfi.csr_valid[i] = 1;
+@@ -54,14 +118,59 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
+                 }
+             }
+           }
++    else {
++      if (got_commit) {
++        last_rd_addr = reg.first >> 4;
++        last_rd_wdata = reg.second.v[0];
++        continue;
+       }
+-      else {
+         // TODO FIXME Take into account the XLEN/FLEN for int/FP values.
+         rvfi.rd1_addr = reg.first >> 4;
+         rvfi.rd1_wdata = reg.second.v[0];
+         // TODO FIXME Handle multiple register commits per cycle.
+         // TODO FIXME This must be handled on the RVFI side as well.
++      got_commit = true; // Only latch first commit
++    }
++  }
++  // popret(z) should return rd1_addr = 0 instead of the SP to match with the cv32e40s core
++  if (((this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRET) == MATCH_CM_POPRET) ||
++      ((this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRETZ) == MATCH_CM_POPRETZ)) {
++    rvfi.rd1_addr = 0;
++    rvfi.rd1_wdata = 0;
++  }
++
++  bool mem_access = false; // TODO: support multiple memory writes/reads
++  int read_len;
++  for (auto &mem : mem_read_commits) {
++    //mem format: (addr, 0, size) (value is not stored for reads, but should be the same as rd)
++    if(!mem_access) {
++      rvfi.mem_addr = std::get<0>(mem);
++      if ((this->get_state()->last_inst_fetched.bits() & MASK_CM_POP) == MATCH_CM_POP         ||
++          (this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRET) == MATCH_CM_POPRET   ||
++          (this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRETZ) == MATCH_CM_POPRETZ ){    
++        rvfi.mem_rdata = last_rd_wdata; // During pop, rd1 returns sp, so instead return value read from memory 
++      } else {
++        rvfi.mem_rdata = rvfi.rd1_wdata; 
+       }
++      //mem_rmask should hold a bitmask of which bytes in mem_rdata contain valid data
++      read_len = std::get<2>(mem);
++      rvfi.mem_rmask = (1 << read_len) - 1;
++      mem_access = true;
++    }
++  }
++
++  int write_len;
++  for (auto &mem : mem_write_commits) {
++    //mem format: (addr, value, size)
++    if(!mem_access) {
++      rvfi.mem_addr = std::get<0>(mem);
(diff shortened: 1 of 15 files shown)


vendor/riscv/riscv-isa-sim/riscv/riscv_dpi.cc
```diff
@@ -50,6 +50,40 @@ extern "C" void spike_set_default_params(const char *profile) {
 
     params.set_string("/top/core/0/", "name", std::string("cva6"));
     params.set_string("/top/core/0/", "isa", std::string("RV64GC"));
+  } 
+  else if (strcmp(profile, "cv32e40s") == 0)
```
silabs-robin, line 54: This is a very nice part. It is cleanly contained within a 40s-specific block.

vendor/riscv/riscv-isa-sim/riscv/processor.cc
```diff
@@ -643,7 +644,7 @@ void processor_t::set_mmu_capability(int cap)
 void processor_t::take_interrupt(reg_t pending_interrupts)
 {
   // Do nothing if no pending interrupts
-  if (!pending_interrupts) {
+  if (!pending_interrupts || !interrupt_allowed) {
```
silabs-robin, line 647: Is the default case for `interrupt_allowed` handled such that other users won't get surprised by this?

vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
silabs-robin, line 1: For [the other PR](https://github.com/openhwgroup/core-v-verif/pull/2432), even if I couldn't compile even vanilla Spike I am still about to run `ci_check` for 40s to test the PR branch.

But here, I don't know how I can try this out myself.

@MikeOpenHWGroup and @MarioOpenHWGroup, the least net use of energy is most likely if one of you could run some check that you are already used to running, which uses some of the files involved in this PR.

@torjeikenes Would one very minimal way to test this simply be to `mkdir build; cd build; ../configure; make` as in building spike within the correct directory in `vendor/`?

vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
```diff
@@ -0,0 +1,415 @@
+diff --git a/vendor/riscv/riscv-isa-sim/riscv/Proc.cc b/vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```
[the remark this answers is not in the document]
torjeikenes: I did not use the script. I just used git diff redirected to a .patch file and applied this.

vendor/riscv/riscv-isa-sim/riscv/Types.h
```diff
@@ -28,8 +28,7 @@ typedef struct {
    uint64_t                 insn_interrupt;
    uint64_t                 insn_interrupt_id;
    uint64_t                 insn_bus_fault;
-   uint64_t                 insn_nmi_store_fault;
-   uint64_t                 insn_nmi_load_fault;
+   uint64_t                 insn_nmi_cause;
```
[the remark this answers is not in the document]
torjeikenes: I changed this to match the same struct in: https://github.com/openhwgroup/core-v-verif/blob/cv32e40s/release/lib/uvm_agents/uvma_rvfi/uvma_rvfi_tdefs.sv

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
```diff
@@ -992,13 +993,15 @@ bool wide_counter_csr_t::unlogged_write(const reg_t val) noexcept {
   // takes precedence over the increment to instret.  However, Spike
   // unconditionally increments instret after executing an instruction.
   // Correct for this artifact by decrementing instret here.
-   openhw::reg::unlogged_write( openhw::reg::unlogged_read()-1);
+
+  // disable since we (at the moment) check mcounterinhibit
+  //openhw::reg::unlogged_write( openhw::reg::unlogged_read()-1);
```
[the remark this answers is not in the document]
torjeikenes: I modified this and the +1 below to add support for disabling `mcycle` counting with `mcountinhibit` to match cv32e40s. I have not tested this much with `mcountinhibit` enabled since `mcycle` would likely mismatch with the core anyway.

vendor/riscv/riscv-isa-sim/riscv/processor.cc
torjeikenes: Added a default value now

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -92,6 +217,144 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
...
+      //(old_en_irq == 0 ) && 
+      (new_en_irq != 0)) 
+  {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+bool Processor::interrupt(reg_t mip, reg_t mie, uint32_t revert_steps, bool interrupt_allowed) {
```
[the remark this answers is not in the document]
torjeikenes: Added some doxygen comments now

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -37,33 +59,136 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
   rvfi.rs2_addr = this->get_state()->last_inst_fetched.rs2();
   // TODO add rs2_value
 
-  rvfi.trap = this->taken_trap;
-  rvfi.trap |= (this->which_trap << 1);
 
+  if(this->next_rvfi_intr){
```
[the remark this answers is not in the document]
torjeikenes: Also curious about this script, and if other parts of the code modify cva6 execution

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
```diff
@@ -1239,9 +1242,10 @@ reg_t dcsr_csr_t::read() const noexcept {
   v = set_field(v, DCSR_EBREAKH, ebreakh);
   v = set_field(v, DCSR_EBREAKS, ebreaks);
   v = set_field(v, DCSR_EBREAKU, ebreaku);
-  v = set_field(v, DCSR_STOPCYCLE, 0);
+  v = set_field(v, DCSR_STOPCYCLE, 1); //TODO: Make configurable
```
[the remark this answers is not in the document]
torjeikenes: Not sure if these changes also affect cva6, or if they need to be configurable to each core

vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
torjeikenes: yes, `mkdir build; cd build; ../configure; make` in `vendor/riscv/riscv-isa-sim` should test that it builds ok.

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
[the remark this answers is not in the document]
silabs-robin: Awesome

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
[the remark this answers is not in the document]
silabs-robin: Since there has been more than a week since I notified that "we only have one week" and there has been 0 external involvement since then, I think we can justify assuming that nothing breaks, particularly since this is on a 40s-specific branch so the rest of the system is shielded for now.
Additional testing will need to be done for a future merge to master.

vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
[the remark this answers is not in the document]
silabs-robin: How do we know that everything new in this PR is compatible with the existing script-based flow?

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -37,33 +59,136 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
...
+    if(this->which_trap & ((reg_t)1 << (isa->get_max_xlen() - 1))) { 
+      //Since spike steps two times to take an interrupt, we store the intr value to the next step to return with rvfi
+      this->next_rvfi_intr |= 1 << 0; //intr [0]
+      this->next_rvfi_intr |= 1 << 2; //interrupt [2]
+      this->next_rvfi_intr |= 0x3FF8 & ((this->which_trap & 0xFF) << 3); //cause[13:3]
+    } else{
+      rvfi.trap |= 1 << 0; //trap [0]
+      rvfi.trap |= 1 << 1; //exception [1]
+      rvfi.trap |= 0x1F8 & ((this->which_trap) << 3); //exception_cause [8:3]
+      //TODO:
```
silabs-robin, line 114: Just flagging all new TODOs in review comments, for visibility.

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -37,33 +59,136 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
...
+    }
+  }
+  // popret(z) should return rd1_addr = 0 instead of the SP to match with the cv32e40s core
+  if (((this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRET) == MATCH_CM_POPRET) ||
+      ((this->get_state()->last_inst_fetched.bits() & MASK_CM_POPRETZ) == MATCH_CM_POPRETZ)) {
+    rvfi.rd1_addr = 0;
+    rvfi.rd1_wdata = 0;
+  }
+
+  bool mem_access = false; // TODO: support multiple memory writes/reads
```
silabs-robin, line 158: Just flagging all new TODOs in review comments, for visibility.

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
```diff
@@ -92,6 +217,144 @@ st_rvfi Processor::step(size_t n, st_rvfi reference) {
...
+    return true;
+  } else {
+    return false;
+  }
+}
+
+bool Processor::interrupt(reg_t mip, reg_t mie, uint32_t revert_steps, bool interrupt_allowed) {
+  state_t *state = this->get_state();
+
+  reg_t mask = 0xFFFF0888; // TODO: automatically generate this
```
silabs-robin, line 304: Just flagging all new TODOs in review comments, for visibility.

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
```diff
@@ -1239,9 +1242,10 @@ reg_t dcsr_csr_t::read() const noexcept {
...
   v = set_field(v, DCSR_STOPTIME, 0);
   v = set_field(v, DCSR_CAUSE, cause);
+  v = set_field(v, DCSR_MPRVEN, 1); // TODO: Make configurable
```
silabs-robin, line 1248: Just flagging all new TODOs in review comments, for visibility.

vendor/riscv/riscv-isa-sim/riscv/debug_rom_defines.h
```diff
@@ -17,7 +17,7 @@
 
 // These needs to match the link.ld         
 #define DEBUG_ROM_WHERETO 0x300
-#define DEBUG_ROM_ENTRY   0x800
+#define DEBUG_ROM_ENTRY   0x1A110800 //TODO: make configurable
```
silabs-robin, line 20: Just flagging all new TODOs in review comments, for visibility.

silabs-robin (commented):
Just flagging all new TODOs in review comments, for visibility.

vendor/riscv/riscv-isa-sim/riscv/processor.cc
silabs-robin: But is it possible for the surrounding logic to change this signal so that users of other processor cores get their Spike-based simulations tampered with unexpectedly?

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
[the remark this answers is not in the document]
silabs-robin: Is the old behavior preserved unless the new behavior is explicitly asked for?

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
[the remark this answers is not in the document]
silabs-robin: `DCSR_STOPCYCLE` is bit 10, AKA `dcsr.STOPCOUNT`, and it is set to `1` to reflect https://docs.openhwgroup.org/projects/cv32e40s-user-manual/en/latest/control_status_registers.html#debug-control-and-status-dcsr ?
Is that correct?

If the entire Spike source now gets hardcoded to 40s-specifics, then I agree with the comment that it should be made configurable.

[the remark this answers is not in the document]
torjeikenes: That is correct.
If these changes interfere with cva6, a parameter could be added to configure this. DCSR has changes both to `DCSR_STOPCYCLE` and `DCSR_MPRVEN`, so maybe one parameter could contain the full `DCSR`, or there could be multiple individual parameters to configure stopcycle and mprven .

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
```diff
@@ -992,13 +993,15 @@ bool wide_counter_csr_t::unlogged_write(const reg_t val) noexcept {
...
+
+  // disable since we (at the moment) check mcounterinhibit
+  //openhw::reg::unlogged_write( openhw::reg::unlogged_read()-1);
   return true;
 }
 
 reg_t wide_counter_csr_t::written_value() const noexcept {
   // Re-adjust for upcoming bump()
-  return  openhw::reg::unlogged_read() + 1;
+  return  openhw::reg::unlogged_read(); //+ 1; // Disable since we dont always bump
```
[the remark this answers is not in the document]
torjeikenes: The +1 caused errors in `cv32e40s_csr_access_test` which reads the `mcycle` csr. Spike did not support disabling and enabling counting of `mcycle` with `mcountinhibit`. In 40S counting is disabled by default, so this caused a mismatch when the `mcycle` csr was read. Here `mcycle` counted up in Spike but not in the core, since `mcountinhibit` was enabled (disabling counting).

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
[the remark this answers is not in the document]
torjeikenes: I think so, but I have not done much testing with `mcountinhibit` enabled.
Since Spike "sees" fewer cycles than the core, `mcycle` would likely have a mismatch to a core anyway unless larger modifications are made.

vendor/riscv/riscv-isa-sim/riscv/processor.cc
torjeikenes: Since `interrupt_allowed` is set to true by default and only modified by the added `interrupt()` method, this should not influence old behavior.

vendor/riscv/riscv-isa-sim/riscv/csrs.cc
torjeikenes: See the other comment about dcsr_stopcycle above

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
torjeikenes: At the moment, I only support reporting one memory read or write over RVFI per instruction. To account for this, I only compare the "lowest" bits in the comparison module. 40S can report multiple memory operations per instruction, so this should be added to fully match the core, but I did not prioritize implementing this since it is not essential.

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
torjeikenes: The `rvfi_trap` signal is not fully implemented for debug mode like [this](https://docs.openhwgroup.org/projects/cv32e40s-user-manual/en/latest/rvfi.html#rvfi-trap-type).

vendor/riscv/riscv-isa-sim/riscv/Proc.cc
torjeikenes: This might not be needed since a similar mask is generated when `mie` is reset

vendor/patches/riscv/riscv-isa-sim/0021-cv32e40s-support.patch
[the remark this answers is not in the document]
torjeikenes: With my current setup on a server with an older git version, I cannot run the script in this branch or on master. Can you check if it runs for you? I get a git error when running this `python3 util/vendor.py vendor/riscv_riscv-isa-sim.vendor.hjson`

PR #2469, opened 2024-06-12
update prev_is_trap to deassert when exit from exception, irq or debug

dd-baoshan:


dd-baoshan:
passed v1 and v2 Ci check in local run

PR #2472, opened 2024-06-19
Fix issue found in random test

dd-baoshan:


pascalgouedo:

PR #2478, opened 2024-06-21
Reduced interrupt delay (long scenario)

XavierAubert:
Reduced a bit interrupt delay scenario as some wfi stress tests result in the core just waiting 90% of the simtime (70ms simtime...)


pascalgouedo:

Issue #2484, opened 2024-06-27
[BUG] UVMA core cntrl agent is not aware of H (hypervisor) extension


zchamski:
**Functionally incorrect behavior:**  the core cntrl cfg agent does not mark CSRs `mtinst` and `mtval2` as unsupported in the absence of the `H` (hypervisor) extension.  As a result, Spike tandem simulations of CVA6 consider these registers as implemented in RTL (**incorrect** behavior) while Spike reports them as not found (correct behavior).

The effect can be seen in tandem scoreboard checks.

### Steps to Reproduce
Please provide:
1. URL to the branch that exhibits the issue: [cva6 master@33ab2efa83454e5b4f90c2156863b47adb0c56ba](https://gitlab.thales-invia.fr/riscv-ci/cva6/-/commit/33ab2efa83454e5b4f90c2156863b47adb0c56ba)
2. Command line: `export DV_SIMULATORS=vcs-testharness SPIKE_TANDEM=1; bash verif/regress/smoke-tests.sh`, target `cv32a65x`.
3. Logfile excerpt  from `verif/sim/out_2024-06-27/vcs-testharness_sim/rv32i-I-ADD-01.cv32a65x.log.iss`, CSRs 34a (`mtinst`) and 34b (`mtval2`):
```
UVM_INFO /gitlab-runner/runner_riscv-public/builds/yD5zmwgi3/1/riscv-ci/cva6/verif/core-v-verif/lib/uvm_components/uvmc_rvfi_scoreboard/uvmc_rvfi_scoreboard_utils.sv(198) @ 12210: reporter [spike_tandem] 
CSR 341  Mismatch   [REF]: 0xffffffff8000031a [CORE]: 0x8000031a        
CSR 34a  not found  [REF]: 0x0                [CORE]: 0x0               
CSR 34b  not found  [REF]: 0x0                [CORE]: 0x0               
UVM_INFO /gitlab-runner/runner_riscv-public/builds/yD5zmwgi3/1/riscv-ci/cva6/verif/core-v-verif/lib/uvm_components/uvmc_rvfi_scoreboard/uvmc_rvfi_scoreboard_utils.sv(199) @ 12210: reporter [spike_tandem]                12210 | RVFI |        3 |      0 | 80000004 | 34202f73 | M | x0  | 00000000 | x2  | 00000000 | x30 | 000000000000000b | csrrs   t5, mcause, zero
[...]

PR #2486, opened 2024-06-27
[RVFI scoreboard] Compare only XLEN least significant CSR bits from ref model.

zchamski:
Ensure that the comparison of refmodel and RTL CSR values on the scoreboard is performed on XLEN least-significant bits of the refmodel value.

This removes the false mismatches observed in tandem tests (see issue core-v-verif #2484.)


MarioOpenHWGroup:
LGTM

MarioOpenHWGroup (approved):

PR #2487, opened 2024-06-28
V2 regression scripts & other updates

XavierAubert:
In this PR, I added two files (`cv32e40p/docs/CV32E40Pv2_full_nr.sh` and `cv32e40p/docs/CV32E40Pv2_full_nr_questa_vrun.sh`) to execute the regression with the same parameters that has been used during the project, with shell and Questa VRUN. 

I also included an interactive script to help generate all regress files needed to run CV32E40Pv2 regression (shell or Questa  Vrun). See the file `bin/cv32e40pv2_nr_generator.sh`. 
When I wrote this script, I saw an issue with the cv_regress python script outputting info-level messages into stderr instead of stdout, so I fixed it. 

I also changed the status of Vplans capture to Complete has all sign-off reviews are now complete.


pascalgouedo:
Merged as needed by other files for correct linking.

PR #2504, opened 2024-07-09
Only cross with cp_group in csr_hazard coverpoint

AyoubJalali:
This MR is related to update csr_hazard cross in isacov coverage model, instead of cross csr_hazard with cp_instr (this give more then 20000 bin to cover), I replace it with cp_group this reduce a little bit the combination to be cover.

N.B : I don't konw if this cross is necessary for our verification goals


AyoubJalali:
@silabs-robin could you review this PR please !!

JeanRochCoulon (approved):

silabs-robin (commented):
> instead of cross csr_hazard with cp_instr (this give more then 20000 bin to cover), I replace it with cp_group this reduce a little bit the combination to be cover.

Makes sense to me.

(Sorry for late reply, I was on vacation.)

PR #2511, opened 2024-07-19
Added file list to be able to get rid of internal flow for compilatio…

ludovicpion:
…n, [mem_rsp_model]: fixes related to atomic transaction, Updated translator script

Added file list to be able to get rid of internal flow for compilation, 
[mem_rsp_model]: fixes related to atomic transaction, 
Updated translator script


lib/cv_dv_utils/uvm/Files.f
```diff
@@ -0,0 +1,30 @@
+${CV_DV_UTILS_DIR}/uvm/memory_partition/memory_partitions_pkg.sv
```
MikeOpenHWGroup, line 1: A couple of comments about this file:
1. It should have a license header (two-liner is sufficient).
2. It seems that the user us expected to define `CV_DV_UTILS_DIR` shell env variable.  Is that correct?  If so, that should be documented in a README.

MikeOpenHWGroup (changes requested):

lib/cv_dv_utils/uvm/Files.f
ludovicpion: Hi,

License has been added on uvm/Files.f

Regarding CV_DV_UTILS_DIR, you are rigth. Actually the README indicated that the user must define CV_DV_UTILS variable. I changed it to CV_DV_UTILS_DIR.

Regards,
Ludovic

ludovicpion: Hi @MikeOpenHWGroup ,

Is it ok now?

Thanks,
Ludovic

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
Thanks for the updates @ludovicpion. I have approved and will merge in this PR.

A couple of things:
- A two-line SPDX header is preferable to the complete Apache header:
```
// Copyright 2024 Commissariat a l'Energie Atomique et aux Energies Alternatives (CEA)
// SPDX-License-Identifier: Apache-2.0
```
- Although it is not worth gating the pull-request, I would rather not have a requirement for the user to define `CV_DV_UTILS_DIR` in their shell environment.  We strive to keep the number of shell variables to a minimum.  Our Makefiles typically determine the value of `CORE_V_VERIF` ([example](https://github.com/openhwgroup/core-v-verif/blob/8e25885aa9596fce0a47c53eb0965c981bf4047c/cv32e40p/sim/uvmt/Makefile#L48)) and then define tool or manifest dependent variables based on that ([example](https://github.com/openhwgroup/core-v-verif/blob/8e25885aa9596fce0a47c53eb0965c981bf4047c/mk/Common.mk#L339)).  I will create an issue for you to add that to the `tp_translator.`

PR #2514, opened 2024-07-25
Generic PMA agent for multiple memory interfaces (AXI / OBI)

abdelhak-chkirid:
Modified CV32E40 PMA agent to be generic for multiple memory interfaces such as AXI and OBI
  - add monitor for AXI transactions 
  - add monitor for OBI transactions
  - Update scoreboard checks to be based on PMA transaction

### DIFF
--- a/lib/uvm_agents/uvma_pma/src/comps/uvma_pma_region_cov_model.sv
+++ b/lib/uvm_agents/uvma_pma/src/comps/uvma_pma_region_cov_model.sv
@@ -153,7 +153,6 @@ task uvma_pma_region_cov_model_c::run_phase(uvm_phase phase);
    fork
       // Monitor transactions
       forever begin
-         `uvm_info("PMA REGIONS COV MODEL", $sformatf("waitong axi transaction"), UVM_LOW)
          mon_trn_fifo.get(mon_trn);
          if (cfg.enabled && cfg.cov_model_enabled) begin
             sample_mon_trn();
@@ -170,7 +169,6 @@ function void uvma_pma_region_cov_model_c::sample_mon_trn();
    if (!mon_trn.is_default && mon_trn.region_index == this.region_index) begin
       pma_access_covg.sample(mon_trn);
    end
-   `uvm_info("PMA REGIONS COV MODEL ", $sformatf("sample coverage model"), UVM_LOW)
 
 
 endfunction : sample_mon_trn
--- a/lib/uvm_agents/uvma_pma/src/comps/uvma_pma_region_cov_model.sv
+++ b/lib/uvm_agents/uvma_pma/src/comps/uvma_pma_region_cov_model.sv
@@ -13,9 +13,9 @@
 `define __UVMA_PMA_REGION_COV_MODEL_SV__
 
 covergroup cg_pma_access(
-   string name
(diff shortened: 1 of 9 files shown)


MikeOpenHWGroup:
Thanks for this PR @abdelhak-chkirid.

Hi @silabs-hfegran, I know you did not write this code, but AFAIK, SiLabs is the only user of the PMA Agent.  Can you (or your deligate) add a review to this PR? Thanks.

lib/uvm_agents/uvma_pma/src/obj/uvma_pma_cfg.sv
```diff
@@ -76,13 +87,34 @@ function int uvma_pma_cfg_c::get_pma_region_for_addr(bit[XLEN-1:0] addr);
...
+      end
+      else if (memory_intf == UVMA_PMA_OBI_INTF) begin
+         if (regions[i].is_addr_in_region(addr))
+            return i;
+      end
+     else begin
+     `uvm_fatal("CFG", "Configuration does not contain valid memory interface")
+     end
+   end
+   for (int i = 0; i < regions.size(); i++) begin
```
silabs-hfegran, line 102: This looks incorrect to me, why is this code duplicated? (the code inside ` if (memory_intf == UVMA_PMA_AXI_INTF) ` is repeated below)

silabs-hfegran (changes requested):
In general it looks good to me, though with one part that looks functionally incorrect (or redundant?), see my review comment.

One improvement I think that we could benefit from would be to introduce more type parameterization to avoid duplicating files and code for AXI/OBI, as most of the code is fairly similar. That would reduce the maintenance effort.

lib/uvm_agents/uvma_pma/src/obj/uvma_pma_cfg.sv
abdelhak-chkirid: thanks for the review , i think now it's ok

abdelhak-chkirid:
@silabs-hfegran  the duplicated part of code is deleted now , can check please .

abdelhak-chkirid:
@silabs-hfegran the duplicated part of code is deleted now , can you check please .

MikeOpenHWGroup:
Pinging @silabs-hfegran

lib/uvm_agents/uvma_pma/src/comps/uvma_pma_region_cov_model.sv
```diff
@@ -13,9 +13,9 @@
 `define __UVMA_PMA_REGION_COV_MODEL_SV__
 
 covergroup cg_pma_access(
-   string name
-) with function sample(uvma_pma_mon_trn_c           trn,
-                       uvma_core_cntrl_pma_region_c pma_region);
+   string name , bit is_main ,  bit support_atomic ,  bit support_cacheable ,  bit support_buff )
```
silabs-hfegran, line 16: I would prefer this to remain a type that is overridable in the core-specific code. 
E.g. we have the cv32e40s which overrides the atomic bit to be an integrity bit instead, and thus explicitly naming it here makes the code less flexible and readable for other than this specific case.

This applies to other code in these files as well, not only on this line.

silabs-hfegran (changes requested):

PR #2529, opened 2024-09-03
[SPIKE] Add sign extension to csr injection

MarioOpenHWGroup:


zchamski (commented):
Please add a quick description to the commit explaining why signed reg (sreg_t) is used instead of unsigned reg (reg_t).  This is not obvious in the light of RISC-V specs stating that e.g. counter CSR values are treated as unsigned (see Priv. ISA 20240411, Section 3.2.1 Machine Time Registers).

MarioOpenHWGroup:
> Please add a quick description to the commit explaining why signed reg (sreg_t) is used instead of unsigned reg (reg_t). This is not obvious in the light of RISC-V specs stating that e.g. counter CSR values are treated as unsigned (see Priv. ISA 20240411, Section 3.2.1 Machine Time Registers).

Sorry for not putting it, when using 32 bits the registers are also 64 bits so the values have to be normalized to 64 when injected. It has more to do with the internal representation of Spike than to the spec.

ASintzoff (approved):

PR #2533, opened 2024-09-18
AXI assertions: Comment the assertions that check the outstanding transactions...

AEzzejjari:
Comment the assertions that check the outstanding transactions at the end of the simulation.

### DIFF
--- a/lib/uvm_agents/uvma_axi5/src/uvma_axi_assert.sv
+++ b/lib/uvm_agents/uvma_axi5/src/uvma_axi_assert.sv
@@ -48,10 +48,6 @@ module  uvma_axi_assert (uvma_axi_intf axi_assert);
    int upper_wrap_boundary;   // The highest address within a wrapping burst
    int bus_width;   // The number of byte lanes in the data bus
    int valid_resp[];  //Store  the number of Wlast for each outstanding transaction ID
-   //Variable for checking assertion coverage
-   int cov_w_data_num = 0;
-   int cov_w_data_num_not = 0;
-   int cov_errm_wstrb = 0;
    int status = 0;
    int aw_enable = 0;
    int w_enable = 0;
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (changes requested):
In a UVM environment there should be one-and-only-one `final` block.  The reason for this is that when `final` blocks are invoked, the UVM run-flow has already been shutdown.  Thus, the call to `uvm_error` is not caught by the reporter in the UVM messaging service (because that object no longer exists), so there is a risk that any calls of `uvm_error()` in the final block will not cause the simulation to report a failure.

The "UVM way" to handle these end-of-simulation checks is to implement the `check_phase()` method in the Agent and add the check for the number of outstanding burst responses there.

Also, I do not think we need an `assert` here, a simple `if...else` will do:
```
function void check_phase( uvm_phase phase );
  foreach(burst_resp[i]) begin
         // Check if All write transaction addresses are matched with a corresponding buffered response (Section A3.4.5)
         if (burst_resp[i] !== 0) begin
            `uvm_error("AXI4 protocol checks assertion", "Violation of bresp_all_done");
         end
  end
endfunction
```

AEzzejjari:
The assertions here are not implemented in a UVM class, they're just a SystemVerilog module that is bound to the interface in the testbench module, so we can't implement a check_phase task here.

The assert statement is only used for coverage.

MikeOpenHWGroup:
> The assertions here are not implemented in a UVM class, they're just a SystemVerilog module that is bound to the interface in the testbench module, so we can't implement a check_phase task here.

Are you saying that `lib/uvm_agents/uvma_axi5/src/uvma_axi_assert.sv` is a SystemVerilog module and that the user of this AXI IP is expected to know that they must bind this to the interface in their testbench module?  This is not a good design.   Constructing a UVM Agent should bring in everything that is required.  Please add these assertions to the monitor class of the agent.

> The assert statement is only used for coverage.

No, you have an "else... uvm_error", so it is also used for error detection.   The best way to implement this is with an explicit `cover property` and `assert property`, to make you intention clear.

AEzzejjari:
These assertions are related to the AXI protocol. The idea behind creating an assertion module is to use it in any testbench, even without an UVM agent.

MikeOpenHWGroup:
> These assertions are related to the AXI protocol. The idea behind creating an assertion module is to use it in any testbench, even without an UVM agent.

That is not obvious to anyone who might use these assertions.  In any case, the usage for UVM environments that use the Agent is broken and I will not approve this PR unless and until the phasing issue is resolved.

JeanRochCoulon (approved):

MikeOpenHWGroup (approved):

MikeOpenHWGroup:
You are right @JeanRochCoulon, this is better, so I have approved.

However, we should carefully consider using immediate asserions (`assert()`) for coverage.  It is not obvious to the user that we need to have coverage for this.  If we want functional coverage, we should be explicit about it by create a `cover property`.

PR #2541, opened 2024-10-15
Clone 40s-dv  (Delete 40s directory)

silabs-robin:
This PR _deletes_ `cv32e40s/*`.

--------

Overview:

After https://github.com/openhwgroup/cv32e40s-dv/pull/3, the external "dv" repo can replace the directory `cv32e40s/`.
_(Just like it was done [for the x](https://github.com/openhwgroup/core-v-verif/tree/cv32e40s/dev/cv32e40x).)_

The core-specific directories in core-v-verif (i.e. `cv32e40s/` and `cv32e40x/`) will be empty, except for a README with instructions for getting the actual content.

--------

Changes:
* `bin/clonetb` gets new flags, to help clone cv32e40s-dv.
* `cv32e40s/README.md` just explains how to use bin/clonetb.
* `cv32e40s/*` is otherwise _purged_.

--------

Rationale:

This was already done for the X, because 1) it significantly eases merging S->X, and 2) it reduces the size of core-v-verif itself (so it just containts common code, libraries, agents, etc).

The reason for doing it on the S as well is because 1) it makes the system symmetrical (_both_ cores have a *-dv repo), and 2) merging will be even easier, 3) dealing with smaller repos are easier.

We already have the `cv32e40s-dv` repo established, and half-way up to date, and [soon it is full-way up to date](https://github.com/openhwgroup/cv32e40s-dv/pull/3).

--------

Reviewing:

Evaluate the "Changes:" above.

If you want to run hello world:
Clone this PR's branch.
See the instructions in `cv32e40s/README.md` but use `clonetb` like it was done in https://github.com/openhwgroup/cv32e40s-dv/pull/3.
Then run hello world as normal.


bin/clonetb
```diff
@@ -41,6 +41,7 @@ usage() {
   echo "usage: $0 [-x] [--clone] [--ignore] [--unignore]"
   echo "  -x          clone cv32e40x subtree, known stable hash"
   echo "  --x-main    clone cv32e40x subtree, latest main branch"
+  echo "  --s-main    clone cv32e40s subtree, latest main branch"
```
silabs-robin, line 44: There should be an `-s` flag too!

But I want it to point to the merge-commit from https://github.com/openhwgroup/cv32e40s-dv/pull/3 after that PR is merged.

silabs-robin:
Hi, @MikeOpenHWGroup, @silabs-hfegran, @silabs-krdosvik.
This PR is in draft, but only because it is pending on a different PR to be merged.
Hence, I am requesting an early review, to inspect the principal changes of _this_ PR.

bin/clonetb
silabs-robin, line 1: > Do you plan to change the merge.sh script: https://github.com/openhwgroup/core-v-verif/blob/cv32e40x/dev/bin/merge.sh, at least I think it would need a change, or be deleted to avoid confusion.

_(Moving this to a thread, to keep related comments bundled together, and to make it resolveable. @silabs-krdosvik )_

Yes, that should be changed as well. Good point.

It _should_ be easy?
Where `git subtree` is used, we can simply add a remote to the `cv32e40s-dv` repo instead.
Then everything else should work as before.

Do you think it should be done as part of this PR, or could it wait until later?
I also plan to merge `cv32e40s-dv` -> `cv32e40x-dv`, so at that time the script would be useful again and prompt a change.

silabs-robin:
The related changes in https://github.com/openhwgroup/cv32e40s-dv/pull/3 has now been merged, so this PR is ready to be reviewed/merged.

PR #2552, opened 2024-11-14
Updated cv dv utils for VCS

khandelwaltanuj:
Hello,

I made changes in some of the packages and filelist of cv_dv_utils, so that it can compile with VCS. 

Thanks and Regards
Tanuj Khandelwal


ludovicpion (approved):

MikeOpenHWGroup:
Thanks for the review @ludovicpion.  I will merge this in now.

AEzzejjari:
The modification on the AXI agent caused problems in the CVA6 pipeline, we need to fix it before doing the merge.

PR #2560, opened 2024-11-22
Remove non-utf-8 characters

MikeOpenHWGroup:
The LICENSE.md file in the OBI directory has some non-UTF-8 characters in it.  This doesn't cause problems for GitHub, but it does for some python scripts.  This PR simply removes these characters.


MikeOpenHWGroup:
Opps, a few additional files got into the PR.  I will close this and open a new one.

PR #2581, opened 2025-01-29
CVXIF UVM agent : remove wrong assertion

AyoubJalali:
This is some changes on cvxif uvm agent :
remove wrong assertion
add support to the agent to see that the cpu is retracting transaction


MikeOpenHWGroup (approved):
LGTM @AyoubJalali.  I will defer to @ASintzoff for final review and merge.

PR #2583, opened 2025-02-03
[Spike] Introduce support of all instructions in v1.0.0 CV-X-IF insn spec.

zchamski:
Add support of solo and tandem Spike configurations using an explicit 'Xcvxif' ISA option rather than just a Spike functional extension.
Presence of the CV-X-IF support is signaled by adding the '_xcvxif' member to the ISA string:

* for solo mode: either on the command line or through the ISA string param in the Yaml file;
* for tandem mode: in the ISA string param constructed by Core Cntrl agent (UVM setups) or in the Yaml file (all other setups).


JeanRochCoulon (approved):
LGTM

PR #2585, opened 2025-02-10
Restore Verilator compile/simulation

### LINKED ISSUE #2584: Followed the quick start for cv32e40p but got hit with 7 warnings
*https://docs.openhwgroup.org/projects/core-v-verif/en/latest/quick_start.html

I just ran the quick-start and was it with the following error:

%Warning-COMBDLY: /home/vboxuser/Documents/core-v-verif/core-v-cores/cv32e40p/rtl/../bhv/cv32e40p_sim_clock_gate.sv:25:31: Non-blocking assignment '<=' in combinational logic process
                                                                                                                         : ... This will be executed as a blocking assignment '='!
   25 |     if (clk_i == 1'b0) clk_en <= en_i | scan_cg_en_i;
      |                               ^~
                  *** See https://verilator.org/warn/COMBDLY before disabling this,
                  else you may end up with different sim results.
%Error: Exiting due to 7 warning(s)
make[1]: *** [Makefile:491: testbench_verilator] Error 1
make[1]: Leaving directory '/home/vboxuser/Documents/core-v-verif/cv32e40p/sim/core'
make: *** [Makefile:485: sanity-veri-run] Error 2

MikeOpenHWGroup:
Resolves #2584, and cleans up DSim compiles as well.

### DIFF
--- a/cv32e40p/sim/core/Makefile
+++ b/cv32e40p/sim/core/Makefile
@@ -134,7 +134,8 @@ SIMV                    = ./simv
 # dsim is the Metrics Technologies SystemVerilog simulator  (https://metrics.ca/)
 DSIM                    = dsim
 DSIM_HOME               = /tools/Metrics/dsim
-DSIM_CMP_FLAGS          = +define+CORE_TB -timescale 1ns/1ps $(SV_CMP_FLAGS) -suppress MultiBlockWrite +define+CV32E40P_APU_TRACE
+DSIM_CMP_FLAGS          = +define+CORE_TB +define+CV32E40P_APU_TRACE +acc+b -timescale 1ns/1ps $(SV_CMP_FLAGS)
+DSIM_SUPPRESS           = -suppress MultiBlockWrite -suppress IneffectiveDynamicCast
 DSIM_RUN_FLAGS          =
 DSIM_UVM_ARGS           = +incdir+$(UVM_HOME)/src $(UVM_HOME)/src/uvm_pkg.sv
 DSIM_RESULTS           ?= $(PWD)/dsim_results
@@ -142,7 +143,6 @@ DSIM_WORK              ?= $(DSIM_RESULTS)/dsim_work
 DSIM_IMAGE              = dsim.out
 
 ifneq (${WAVES}, 0)
-  DSIM_CMP_FLAGS += +acc
   DSIM_DMP_FILE  ?= dsim.fst
   DSIM_RUN_FLAGS += -waves $(DSIM_DMP_FILE)
 endif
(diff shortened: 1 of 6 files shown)


cv32e40p/sim/core/Makefile
```diff
@@ -276,9 +277,13 @@ mk_results:
 
 # Metrics dsim compile targets
 dsim-comp: mk_results CV_CORE_pkg tbsrc_pkg tbsrc
+	@echo "$(BANNER)"
+	@echo "* Compiling with Metrics DSIM"
```
cairo-caplan, line 281: It is nice that you have added a printing message to the target! So the user can have a stronger feedback that make is executing the right target/options

cv32e40p/sim/core/Makefile
```diff
@@ -490,13 +499,13 @@ testbench_verilator: CV_CORE_pkg $(TBSRC_VERI) $(TBSRC_PKG)
...
 		$(VERI_TRACE) \
-		--Wno-lint --Wno-UNOPTFLAT \
-		--Wno-MODDUP --top-module \
+		$(VERI_WARN_WAIVERS) \
+		--top-module \
 		tb_top_verilator $(TBSRC_VERI) \
 		-f $(CV_CORE_MANIFEST) \
 		$(CV_CORE_PKG)/bhv/$(CV_CORE_LC)_core_log.sv \
 		$(TBSRC_CORE)/tb_top_verilator.cpp --Mdir $(VERI_OBJ_DIR) \
-		-CFLAGS "-std=gnu++11 $(VERI_CFLAGS)" \
```
cairo-caplan, line 499: I suppose here "-std=gnu++11" was removed for it is no necessary on Verilator supported (or recent) versions.

I imagine this modification should also be done to the same targets for core-v-verif/cv32e40s/sim/core/Makefile:468 and  core-v-verif/cv32e40x/sim/core/Makefile

cv32e40p/tb/core/cv32e40p_tb_wrapper.sv
```diff
@@ -15,7 +15,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-0.51
 
-module cv32e40p_tb_wrapper
+module cv32e40p_core_tb_wrapper
```
cairo-caplan, line 18: Was there a particular reason for the renaming? I agree that adding _core looks better, but now the name of the file should also follow it :)

cairo-caplan (approved):
The modifications do allow Verilator to work again, solving https://github.com/openhwgroup/core-v-verif/pull/2585 - I tested it locally on Verilator 5.033.

I advise to change the filename of the cv32e40p/tb/core/cv32e40p_tb_wrapper.sv to cv32e40p/tb/core/cv32e40p_core_tb_wrapper.sv, and to check if the -std=gnu++11 CFLAG can be removed on Verilator targers for other cores too

cv32e40p/sim/core/Makefile
MikeOpenHWGroup: This was done as per the discussion of [Verilator Issue #4898](https://github.com/verilator/verilator/issues/4898#issuecomment-1986862011) which recommends removing `-std=c++11`.

You are right that we should do the same for cv32e40s and cv32e40x - those should be done on their respective `cv32e40[s|x]/dev` branches.

PR #2587, opened 2025-02-20
[Core cntrl agent] Do not pass 'xcvxif' to Spike if CV-X-IF is present.

zchamski:
Remove the systematic activation of `Xcvxif` ISA extension when the RTL instantiates CV-X-IF. The activation of the extension shall now be done **exclusively** through the Spike ISA string.


ASintzoff:
LGTM

ASintzoff (approved):

PR #2590, opened 2025-03-10
[Spike] Reset non-ISA extensions when registering them (as per comment!)

zchamski:
Explicitly reset non-ISA Spike extensions such as `cv32a60x` in the `Proc` constructor.

Such extensions contain no instruction definitions, only CSR definitions and hence, were not reset correctly. As a result, the CSRs provided by the `cv32a60x`extension were instantiated.


ASintzoff:
LGTM

ASintzoff (approved):
LGTM

PR #2600, opened 2025-03-24
ISACOV : ignore fencei group when it's disabled

AyoubJalali:
Disable fencei group from isacov cross_group when it's disabled


MikeOpenHWGroup (approved):
LGTM @AyoubJalali.  As always, I will defer to @ASintzoff for the final review and merge.

PR #2603, opened 2025-03-28
OBI Verif : add coverage for all properties

AyoubJalali:
This is a modification in obi assertion module, to add cover property for all properties.
This help use to be sure that we really passed by the scenario in the property.


MikeOpenHWGroup (approved):
LGTM, thanks @AyoubJalali.  Over to @ASintzoff for merging...

PR #2609, opened 2025-04-02
OBI assertion : be signal shouldn't be zero during address phase

AyoubJalali:
Byte enable should be check during address phase, not only for write transaction


MikeOpenHWGroup (approved):
LGTM, thanks @AyoubJalali.  Over to @ASintzoff for the merge.

PR #2612, opened 2025-04-07
ISACOV : no rd toggle for bext & bexti

AyoubJalali:
bext & bexti instructions extract only 1 bit from an index (rs2 or imm) and store it in RD, so we can't toggle all rd bits (with 1), so I don't think it's useful to keep this cover-point because doesn't make sense.


MikeOpenHWGroup (approved):
Good catch @AyoubJalali! I will defer to @ASintzoff for the final merge.

PR #2613, opened 2025-04-07
SPIKE DISAM: do not use pseudo instructions

AyoubJalali:
Fix in spike disam, to avoid pseudo code generation and can cover instructions


ASintzoff (approved):
LGTM

PR #2614, opened 2025-04-08
Spike disasm: don't optimize jump instructions

AyoubJalali:
This fix disam to not generate pseudo code for jal & jalr


ASintzoff (approved):
LGTM

PR #2618, opened 2025-04-14
ISACOV : no need to cover RO csr for csr_hazard

AyoubJalali:
NO need to cover RO csr for hazard csr as it raise an exception on a write on CSR RO


MikeOpenHWGroup (approved):
LGTM @AyoubJalali.  Overt to @ASintzoff for merging.

PR #2623, opened 2025-06-02
removed use_response_handler: not possible in a generic agent

khandelwaltanuj:
Hello,

In a generic agent, protocol is not known before hand. So it is not possible to correctly define the functionning of response handler. 

Hence I have removed this. 

Regards
Tanuj Khandelwal

### DIFF
--- a/lib/cv_dv_utils/python/sim_cmd/get_cmd.py
+++ b/lib/cv_dv_utils/python/sim_cmd/get_cmd.py
@@ -63,6 +63,11 @@ def cmp_cmd(yaml_file, outdir, opt, vopt_option, work):
       if 'svlog_source' in comp:
         src_list  = comp["svlog_source"]
         srcs      = src_list.split()
+        cmd       = "vlog -sv"
+      elif 'vlog_source' in comp:
+        cmd       = "vlog"
+        src_list  = comp["vlog_source"]
+        srcs      = src_list.split()
       else: 
         src_list  = ""
         srcs      = ""
--- a/lib/cv_dv_utils/python/sim_cmd/get_cmd.py
+++ b/lib/cv_dv_utils/python/sim_cmd/get_cmd.py
@@ -156,7 +156,7 @@ def get_cmd_opt(yaml_file):
     sim_cmd_opt = "{} {}".format(cmd, sim_opt)
     return sim_cmd_opt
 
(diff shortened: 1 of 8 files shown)


MikeOpenHWGroup (changes requested):
I agree that in a generic agent, protocol is not known before hand, so the generic agent should not define the functioning of response handler.  I'm not sure this is the way to due it.  Rather than comment out the call to the handlers, wouldn't it be better to define them as pure virtual functions?

khandelwaltanuj:
Hi Mike,

I have just removed the commented code. And I believe that user can define its own method in the derived class. 

Regards
Tanuj

khandelwaltanuj:
Hello @MikeOpenHWGroup 

I have been thinking how to implement this but I am not sure yet how to do it cleanly. Adding a warning may cause a problem in the case where response_handler is not used. It will give a warning for nothing. 
One option is I just update the READ saying response handler is disabled by default and not do anything else.

MikeOpenHWGroup:
> One option is I just update the READ saying response handler is disabled by default and not do anything else.

I like this option.  I assume enabling the response handler would be controlled via a member in the cfg object which would optionally enable it during the configure phase of the run phase.

khandelwaltanuj:
Hi @MikeOpenHWGroup 

I have made the necessary changes. 

Regards
Tanuj Khandelwal

khandelwaltanuj:
Hi @MikeOpenHWGroup 

Can we merge this PR please ? 

Regards
Tanuj

MikeOpenHWGroup:
There is still one out-standing question about `use_response_handler` in the README.

khandelwaltanuj:
Normally I have added following lines in the README

## Response Handler 

By default response handler is deactivated. Following setup needs to be done to enable the response handle

 ```
 uvm_config_db #( bit )::set(uvm_root::get(), "*", "generic_sequencer" , 1);

 over write the following task of generic_driver, otherwise it gives an uvm_warning 
 virtual task spy_and_drive_rsp(); 

 over write following function if generic_sequences
 virtual function void response_handler(uvm_sequence_item response);
 
 ```

MikeOpenHWGroup (approved):
Thanks for pointing out the update to the README @khandelwaltanuj.  All good.

Issue #2630, opened 2025-08-04
Suppress Verilator Warnings Causing Build Failure in Simulation


Uzair-90:
When running simulations with Verilator, the following warnings cause errors and halt simulation:

`BLKANDNBLK:` Detected mixed blocking/non-blocking assignments.

`MULTIDRIVEN:` Signal is driven from multiple sources (common in testbenches).

`COMBDLY:` Combinatorial delay path warnings.

These are benign in the context of testbench and simulation flow, but Verilator treats them as fatal unless explicitly suppressed.

Fix:

To resolve this, modify the Makefile to include the following flags:
`VERI_COMPILE_FLAGS += -Wno-BLKANDNBLK -Wno-MULTIDRIVEN -Wno-COMBDLY $(SV_CMP_FLAGS)
`
This ensures Verilator runs without stopping on expected testbench behavior.

Rationale:

Without these suppressions, simulation fails even when the RTL is correct and testbenches are safe. This is blocking users from verifying custom changes (e.g., CVA6 custom instructions, accelerator hooks, etc.).

Please let me know if you would like me to submit a patch or pull request to integrate this fix into the repository. I'm happy to contribute a permanent solution.

Uzair-90:
Hi @MikeOpenHWGroup please review the PR #2631 I have provided a fix thank you.

MikeOpenHWGroup:
Hi @Uzair-90, #2631 is now merged.  If that resolves this Issue, please close it.

PR #2631, opened 2025-08-07
FIX: Add Verilator warning suppression flags in cv32e40p/sim/core Mak…

Uzair-90:
Fix Verilator warnings that halt simulation in `cv32e40p/sim/core/Makefile`
**Problem**:
When running Verilator simulations, certain benign warnings are treated as fatal errors:
`BLKANDNBLK`: Mixed blocking/non-blocking assignments
`MULTIDRIVEN`: Multiple drivers (common in testbenches)
`COMBDLY`: Combinatorial delay paths
These warnings are expected in the testbench context but halt simulation unless explicitly suppressed.
**Fix**
Added the following flags to `VERI_COMPILE_FLAGS` in the Makefile:
`-Wno-BLKANDNBLK -Wno-MULTIDRIVEN -Wno-COMBDLY`
**Rationale**
These suppressions ensure simulation proceeds normally for valid RTL and testbenches, enabling users to verify custom changes (e.g., CVA6 extensions, accelerators) without Verilator halting on non-critical warnings.

PR #2632, opened 2025-08-07
Fix bitrot issues due to updated Verilator and Python versions

MikeOpenHWGroup:
Hi @Uzair-90, while reviewing your PR (#2631), I noticed a few other (non-fatal) issues.   Please have a look at these and let me know what you think.

### DIFF
--- a/mk/TOOLCHAIN.md
+++ b/mk/TOOLCHAIN.md
@@ -10,7 +10,7 @@ The recommended toolchain for all CORE-V cores is available from Embecosm
 It is recommended that you install this at `/opt/corev`.
 As detailed in [README](./README.md#required-corev-environment-variables), you will need to define a shell variable to point to it:
 ```
-$ export CV_SW_TOOLCHAIN="/opt/riscv"
+$ export CV_SW_TOOLCHAIN="/opt/corev"
 ```
 
 ### RISC-V Toolchain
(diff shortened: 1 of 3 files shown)


cv32e40p/sim/core/Makefile
Uzair-90, line 1: This is a solid cleanup, the build system is now easier to maintain and more flexible. Using `?=` lets people override settings without editing the Makefile, quoting in find makes it safer, and splitting the flags makes things clearer. The only thing to check is whether any CI scripts still use the old `VERI_FLAGS` name. But I think if all the tests are passed we are good.

bin/objdump2itb
Uzair-90, line 1: It very clean makes regex intent clear and avoids issues with backslash escaping.

Uzair-90 (commented):
Switching the regex patterns to raw strings makes them clearer and avoids any issues with escape sequences. The Makefile updates are a solid improvement variables can now be overridden more easily, the compile and run flags are separated for clarity, and quoting in the find commands makes things safer.
The only thing worth double-checking is whether any scripts or CI jobs still reference the old `VERI_FLAGS` name, since that’s now `VERI_RUN_FLAGS`. Other than that, this is a clean, safe, and maintainable refactor. But I think if all tests are passed we are good for now.

Uzair-90:
I have closed issue #2630 as my PR was merged for that. Please go on with the changes and you can assign me the review.

MikeOpenHWGroup:
OK @Uzair-90, I have resolved the merge conflict.  Please have a look (and thanks!).

PR #2633, opened 2025-08-11
[Spike] Do not re-configure PMPs based on device tree information.

zchamski:
This modification prevents last-instance reconfiguration of the PMPs during the elaboration phase of the simulation.

In the original implementation, PMP granularity and PMP region counts of individual cores were set based on the content of the built-in device tree _after_ applying all per-core parameters supplied in Yaml config file and/or on the Spike command line.  The PMP granularity value encoded in the default device tree is 4 and it was overriding the expected granularity of 8 bytes supplied via configuration parameter `/top/core/<id>/pmp_granularity`.


ASintzoff:
LGTM

ASintzoff (approved):

PR #2677, opened 2026-02-03
cv32e40p: add test for mscratch CSR register

sanjay-embedded:
add test to verify mscratch CSR 32 bit read/write access


sanjay-embedded:
@MikeOpenHWGroup , requesting your valuable input on the pull request this is specific to the LFX RISCV-ceritifaction-program.

PR #2679, opened 2026-02-03
cv32e40p: add mscratch CSR firmware test

ankit-cybertron:
The test writes different 32-bit values to mscratch and reads them back to verify correct operation.
It checks boundary values, single-bit patterns, repeating bit patterns, and random values.


MikeOpenHWGroup:
Hi @ankit-cybertron, thanks for your interest in OpenHW! As you know, the goal of this pull-request was to be part of a "coding challenge" for the RVI Mentorship program. The function of the `mscratch` CSR is tested elsewhere, so we don't need to merge this in.

PR #2695, opened 2026-02-08
Add mscratch CSR read/write access test with corresponding test.yaml

EmanNasar001:
The test is running fine


EmanNasar001:
> can you please help me . i am running hello world test but it is not running.

Hi Hassan! I myself had many issues specifically due to wrong toolchain ,most of the issues will be because of wrong toolchain and variable setting ,I will recommend to read the QUICK GUIDE thoroughly and there is link provided for toolchains and how to set the variables ,Still if you are facing a diff problem can you please specify i would love to answer and help if i could

PR #2700, opened 2026-02-08
cv32e40p: LFX mscratch_test coding assignment

juna-yednap:
This PR contains the coding assignment of LFX assignment of core-v-verif
This runs all the write, set and clear commands without using any pseudo commands being used.

The flow is as follows

1. some random patterns are checked
2. bits are set one by one using csrrw
3. bits are set cumulatively using csrrs
4. bits are cleared cumulatively using csrrc

The test can be run using make veri-test TEST=mscratch_test


juna-yednap:
Update!!
So, eclipse mailed back and said that their server was down yesterday and to perform the procedure today so i have completed the verification now.

PR #2719, opened 2026-02-28
docs: fix run_phase description and objection usage in uvm_tests.rst Fixes #1933

### LINKED ISSUE #1933: Mention of UVM run_phase in uvm_tests.rst is incorrect

Tharunbilluri:


MikeOpenHWGroup (approved):
Nice.  Thanks @Tharunbilluri.

PR #2732, opened 2026-04-03
[cv_dv_utils verif] added api set_is_passive in generic agent and correct an issue with axi2mem

khandelwaltanuj:
Hi 

This PR contains couple of fixes. 

1. Added function set_is_passive for generic uvm agent. 
2. fixed an issue with AXI2MEM. It was causing issues in the case where response starts arriving before the req LAST is inserted.


MikeOpenHWGroup (changes requested):

MikeOpenHWGroup (approved):
LGTM, thanks @khandelwaltanuj.  Approving and merging.

Issue #2735, opened 2026-05-08
uvma_obi_memory_mon: R-channel can mis-associate rvalid after reset with pre-reset outstanding transaction


adrianf0:
**Description:**
In `uvma_obi_memory_mon_c`, the R-channel monitor thread can incorrectly associate an `rvalid` response observed after a reset with an outstanding transaction sampled before reset. This can result in `uvma_obi_memory_mon_trn_c` objects containing mismatched address/data (or wrong access ordering).

**Expected behavior:**
When `reset_n` is asserted:
- all outstanding pre-reset transactions should be flushed/aborted, and
- the monitor should not match a post-reset `rvalid` to a pre-reset request.

**Actual behavior:**
`mon_chan_r_post_reset()` pops an outstanding transaction from `cntxt.mon_outstanding_reads_q` and then waits for `rvalid` in `mon_chan_r_trn()`. If reset is asserted while `mon_chan_r_trn()` is waiting, the task does not abort. After reset deassertion, the next `rvalid` (from a new post-reset request) can be incorrectly matched to the pre-reset transaction.

**How to reproduce:**
1. Configure OBI agent in SLV mode with non-zero randomized `rvalid_latency`.
2. Issue one request so it is pushed into `mon_outstanding_reads_q`.
3. Assert `reset_n==0` after the A-phase is observed but before the corresponding `rvalid`.
4. After reset is released, issue a new transaction.
5. Observe that the monitor can report a response for the old transaction using the `rvalid` of the new one (mismatched ordering/data).

**Impact:**
Scoreboards/checkers relying on `uvma_obi_memory_mon_trn_c` may report false data mismatches or “phantom” transactions around reset boundaries.

MikeOpenHWGroup:
Hi @adrianf0, thanks for generating this Issue (and also #2734).  We have known since the beginning that the OBI Agent lacks proper support for on-the-fly resets and you've motivated me to take another look at it.  Due to other activities it is unlikely I will have time for this for several weeks.  If you'd like to suggest a fix, we'd be happy to review your PR.

Be sure to review [CONTRIBUTING](https://github.com/openhwgroup/core-v-verif/blob/master/CONTRIBUTING.md) first.

PR #2741, opened 2026-06-18
Refactor: Adopt AXI5 reset pattern for OBI memory driver

### LINKED ISSUE #2734: uvma_obi_memory_drv (SLV mode) may still drive rvalid after reset asserted (cannot abort pending response)
**Description:**
When using `uvma_obi_memory_agent_c` in `UVMA_OBI_MEMORY_MODE_SLV` with randomized `drv_slv_rvalid_random_latency_*`, asserting `reset_n` on the `uvma_obi_memory_if` can still result in the driver asserting `rvalid` for a previously accepted slave response seq_item. The driver does not provide a way to cancel/flush an in-progress response once `try_next_item()` has returned an item and `drv_slv_req()` has started.

**Expected behavior:**
When `reset_n==0` (or when the agent enters `UVMA_OBI_MEMORY_RESET_STATE_IN_RESET`), the driver should immediately drive an idle response state (`rvalid==0`, etc.) and should **not** assert `rvalid` for any pending response seq_item. Pending items should be dropped/aborted cleanly (and not deadlock the sequencer).

**Actual behavior:**
If a slave response item has already been accepted and the driver is waiting inside:

- `repeat(req.rvalid_latency) @(slv_mp.drv_slv_cb);`

then a reset assertion does not abort the task. The driver can still reach:

- `slv_mp.drv_slv_cb.rvalid <= 1'b1;`

and produce `rvalid` during or after reset.

**How to reproduce:**
1. Configure agent in SLV mode:
   - `cfg.drv_mode = UVMA_OBI_MEMORY_MODE_SLV;`
   - `cfg.drv_slv_rvalid_mode = UVMA_OBI_MEMORY_DRV_SLV_RVALID_MODE_RANDOM_LATENCY;`
   - set `drv_slv_rvalid_random_latency_max > 0`.
2. Start any slave sequence (e.g. `uvma_obi_memory_slv_seq_c`).
3. Issue at least one request from the DUT.
4. Assert `obi.reset_n` low while the driver is in the latency delay (before it drives `rvalid`).
5. Observe that `obi.rvalid` can still pulse.

**Impact:**
This makes it difficult to verify “reset aborts in-flight transactions” behavior, and can cause protocol assertions or DUT-level assertions to fail when reset is asserted mid-transaction.

TheMonkeyG:
Resolves [#2734](https://github.com/openhwgroup/core-v-verif/issues/2734)

The OBI memory driver previously dispatched on `cntxt.reset_state`
inside per-task loops (`drv_pre_reset`/`drv_in_reset`/`drv_post_reset`).
Because `drv_post_reset()` called multi-cycle tasks (`drv_mstr_req`,
`drv_slv_read_req`, `drv_slv_write_req`), `reset_state` was only
re-checked *between* transactions, not during one — a reset asserted
mid-transaction would not abort it. That's the root cause of #2734.

This PR moves reset handling entirely into `run_phase`, using the
same `disable fork` pattern already used in `uvma_axi5`. Rationale
and an alternative approach that was considered (per-task
fork/join_any escapes) are written up in this comment:
[#2734 (comment)](https://github.com/openhwgroup/core-v-verif/issues/2734#issuecomment-4651504491)

## Reset architecture changes

- `run_phase` now spawns the gnt loop (`drv_slv_gnt`, `chan_a`) and
  the transaction loop (`drv_mstr_loop`/`drv_slv_loop`, `chan_r`)
  via `fork...join_none`, racing against a `wait(reset_n !== 1)` in
  the same scope. When reset asserts, `join_any` returns and
  `disable fork` kills both loops immediately, wherever they were
  in execution — including mid multi-cycle wait inside a
  transaction task.
- `drv_mstr_loop`/`drv_slv_loop` (formerly the MSTR/SLV branches of
  `drv_post_reset`) are now self-contained `forever` loops with no
  reset awareness of their own; they rely entirely on the outer
  `disable fork` to stop them.
- If the loop is killed between `get_next_item`/`try_next_item` and
  `item_done()`, the sequencer would otherwise deadlock. `run_phase`
  now checks `req != null` after `disable fork` and calls
  `item_done()` itself to release the sequencer.
- Added `force_idle_nets`/`release_nets` on `uvma_obi_memory_if`,
  operating on the bare signal names (force must target the
  physical net, not a clocking-block alias). This gives a hard
  guarantee the bus is idle while in reset, on top of the existing
  soft `<=` idle drive in `drv_idle()`.
- `drv_slv_gnt()` no longer dispatches on `cntxt.reset_state` — it's
  only ever invoked from inside the active-window fork now.

## Bundled protocol fixes 

Signed-off-by: Daniel <monkeyg400@gmail.com>

### DIFF
--- a/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
@@ -214,7 +214,7 @@ task uvma_obi_memory_drv_c::run_phase(uvm_phase phase);
 
          // Cleanup in one place
          if (req != null) begin
-            seq_item_port.item_done();
+            // seq_item_port.item_done();
             req = null;
          end
          mon_trn_fifo.flush();
@@ -265,7 +265,9 @@ task uvma_obi_memory_drv_c::drv_mstr_loop();
 
       // 3. Send out to TLM and tell sequencer we're ready for the next sequence item
       mstr_ap.write(mstr_req);
-      seq_item_port.item_done();
+      if (cntxt.reset_deasserted_ev.is_on()) begin
+         seq_item_port.item_done();
+      end
       req = null;
    end
 
@@ -289,7 +291,9 @@ task uvma_obi_memory_drv_c::drv_slv_loop();
 
          // 2. Send out to TLM and tell sequencer we're ready for the next sequence item
          slv_ap.write(slv_req);
-         seq_item_port.item_done();
+         if (cntxt.reset_deasserted_ev.is_on()) begin
+            seq_item_port.item_done();
+         end
          req = null;
       end else begin
          drv_slv_idle();
--- a/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
+++ b/lib/uvm_agents/uvma_obi_memory/src/comps/uvma_obi_memory_drv.sv
@@ -61,24 +61,24 @@ class uvma_obi_memory_drv_c extends uvm_driver#(
    extern virtual function void build_phase(uvm_phase phase);
 
    /**
-    * Oversees driving, depending on the reset state, by calling drv_<pre|in|post>_reset() tasks.
+    * Oversees driving, depending on the reset state.
     */
    extern virtual task run_phase(uvm_phase phase);
 
    /**
-    * Called by run_phase() while agent is in pre-reset state.
+    * Master mode: drive transactions in a loop
     */
-   extern task drv_pre_reset();
+   extern task drv_mstr_loop();
 
    /**
-    * Called by run_phase() while agent is in reset state.
+    * Slave mode: drive transactions in a loop
     */
-   extern task drv_in_reset();
+   extern task drv_slv_loop();
 
    /**
-    * Called by run_phase() while agent is in post-reset state.
+    * Drive idle state (called after reset)
     */
-   extern task drv_post_reset();
+   extern task drv_idle();
 
    /**
     * Drives the 'gnt' signal in response to 'req' being asserted.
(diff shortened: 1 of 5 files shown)


datum-dpoulin:
I would recommend keeping reset monitoring logic in the monitor and using a uvm_event in the context to share state with the driver.

TheMonkeyG:
Hi datum-dpoulin,
  Thanks for the feedback! I've updated the driver to use the  uvm_events   just like you suggested.
  Since I was already digging into mon, I went ahead and applied a similar fix to the monitor to address #2735 as well. 
  Let me know if this looks good to you!

MikeOpenHWGroup:
Hi @TheMonkeyG, thanks for the update.  This is a significant change that will impact multiple projects, so I am being very careful with it.  The OBI Agent is used by two active projects (CVE2 and CVA6) am actively working to test these changes in the CVE2 environment.  I'll let you know if I encounter any issues.

TheMonkeyG:
Hi @MikeOpenHWGroup,

I reviewed the reset handling and realized that using a fake `item_done()` to clean up aborted transactions in the driver was not the best idea :)

I've just pushed an update to fix this: I removed the manual `item_done()` call from the driver's reset cleanup. Instead, I added reset monitoring directly to the sequencer using `stop_sequences()`.

To check this, I wrote a quick smoke test that fires reset at different points during idle, right before item_done(), and mid-transaction. The approach mostly held up under reset, but the smoke test also found one narrow case: if item_done() happens to fire in the exact same delta cycle as reset asserting, it can slip through before the sequencer reacts — so the sequence sees a false "completed". Not sure yet if real OBI timing can actually land in that window, or if it's just an artifact of how the mock drives reset. Still looking into it.

Please let me know how it behaves in your CVE2 testing !

PR #2744, opened 2026-07-15
Add interrupt and debug_request throttle for CVE2

MikeOpenHWGroup:
The CVE2 lacks two features supported by the CVE4 cores:
1. Nested interrupts.
2. Handshake signals for interrupts and debug_requests (see CV32E40P example below)

```systemverilog
     // CV32E40P Interrupt interface
    .irq_i                    (),
    .irq_ack_o                (),
    .irq_id_o                 (),

    // CV32E40P Debug interface
    .debug_req_i              (),
    .debug_havereset_o        (),
    .debug_running_o          (),
    .debug_halted_o           ()
```
These handshake signals can be used to throttle interrupts.  As this feature is lacking in CVE2, this PR adds the ability to probe the RTL of the CVE2 to know when it is safe to re-assert and interrupt or debug_request.

### DIFF
--- a/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
+++ b/lib/uvm_agents/uvma_debug/uvma_debug_drv.sv
@@ -115,7 +115,12 @@ endtask : run_phase
 
 task uvma_debug_drv_c::drv_req(uvma_debug_seq_item_c req);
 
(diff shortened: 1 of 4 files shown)


cairo-caplan (approved):
Hi @MikeOpenHWGroup ,

Thanks for implementing hadshaking for interrupts and debug_requests for the CVE2 verification.
IT looks good to me.

PR #2746, opened 2026-09-02
Spike updates for CV32E20

MikeOpenHWGroup:
This PR involves one significant change and one minor change:
- The significant change updates `vendor/riscv/riscv-isa-sim/riscv/Proc.cc` and `vendor/riscv/riscv-isa-sim/riscv/Proc.h` to extend Spike to match the specified behavior of the CV32E20.
- The minor update modifies a generation constraint to ensure CVE2 can catch the `debug_request`.

### DIFF
--- a/lib/uvm_agents/uvma_debug/seq/uvma_debug_seq_item.sv
+++ b/lib/uvm_agents/uvma_debug/seq/uvma_debug_seq_item.sv
@@ -33,12 +33,10 @@ class uvma_debug_seq_item_c extends uvml_trn_seq_item_c;
    `uvm_object_utils_end
    
    
-   // TODO Add uvma_debug_seq_item_c constraints
-   //      Ex: constraint default_cons {
-   //             abc inside {0,2,4,8,16,32};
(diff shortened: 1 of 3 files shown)


lib/uvm_agents/uvma_debug/seq/uvma_debug_seq_item.sv
```diff
@@ -33,12 +33,10 @@ class uvma_debug_seq_item_c extends uvml_trn_seq_item_c;
    `uvm_object_utils_end
    
    
-   // TODO Add uvma_debug_seq_item_c constraints
-   //      Ex: constraint default_cons {
-   //             abc inside {0,2,4,8,16,32};
-   //          }
+   // Lower bound of 4 (rather than 1): a debug_req_i pulse that ends too soon after being
```
AlexChenIC: Could you clarify how the four-cycle minimum was determined? Is it derived from CVE2's WFI debug-request timing, a documented requirement, or regression observations?

PR #2748, opened 2026-09-12
cv_dv_utils: Fix out-of-order write response reconstruction in axi2mem

Jackie425:
Hi @khandelwaltanuj,

I found another issue in `axi2mem` when enabling out-of-order responses in `memory_rsp_model`. Memory-side write responses from different IDs may return out of order, but `axi2mem` currently uses a single shared burst-tracking state, which can cause responses to be matched with the wrong write request. This PR changes the tracking state to be maintained per AXI ID.

After this change, I no longer observe mismatches in my HPDcache verification environment.

Thanks and regards,
Jackie

### DIFF
--- a/lib/cv_dv_utils/uvm/memory_rsp_model/axi2mem/axi2mem.svh
+++ b/lib/cv_dv_utils/uvm/memory_rsp_model/axi2mem/axi2mem.svh
@@ -515,63 +515,53 @@ class axi2mem#(int unsigned w_addr = 0,  int unsigned w_data = 0, int unsigned w
     virtual task write_b_chan_fifo( );
        // Drive axi iterface
        b_chan_t        rsp;
-       int             cnt;
-       int             err_cnt;
-       int             ex_fail_cnt;
-       aw_ar_chan_t    req;
-       cnt     = 0; 
-       err_cnt = 0;
-       ex_fail_cnt = 0;
+       aw_ar_chan_t    req[integer];
+       int             cnt[integer];
+       int             err_cnt[integer];
+       int             ex_fail_cnt[integer];
+       int             id;
+
        forever begin
          @ (posedge mem_wr_vif.clk);
          if(mem_wr_vif.wr_res_valid) begin
+           id = int'(mem_wr_vif.wr_res_id);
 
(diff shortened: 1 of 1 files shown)


khandelwaltanuj:
Hi jackie,

I am a little confused here. 
AXI2MEM as such does not have a notion of out of order responses. It transfers whatever it gets from the memory. 
If let say there are 3 seccessives writes with ids 1, 2 and 3. And memory reponds with ID, 2, 3 and 1. AXI2MEM just transfer these responses as such ? This is not the case ?

Regards
Tanuj

Jackie425:
Hi Tanuj,

Yes, you are right.
For three single-beat writes with IDs 1, 2 and 3, responses returning as 2, 3 and 1 are handled correctly.

While the issue is with **multi-beat bursts** situation. `memory_rsp_model` may interleave responses from different IDs, while `axi2mem` currently shares `cnt` and `req` across all IDs.

For example, if ID 1 is a 2-beat burst and the responses from mem rsp model arrive as:
ID 1 beat 1 -> ID 2 -> ID 1 beat 2
The ID 2 response can modify the shared burst-tracking state for ID 1 and cause an earlier and incorrect B response reconstruction.

This change tries to track the burst count per ID to fix this.

Thanks and regards,
Jackie

khandelwaltanuj:
Hi Jackie,

I understand, that sounds good. I ignored this issue, at the time of coding. 

Regards
Tanuj

khandelwaltanuj (approved):

khandelwaltanuj:
by the way: AXI4 specs spicify very clearly that flit interleaving is not supported. Si it may not work in the case of HPDcache, if you are trying to do that. 

AXI4 does not support flit interleaving because the Advanced eXtensible Interface (AXI) is a packet-based/burst-based protocol, not a flit-based NoC (Network-on-Chip) protocol.However, AXI does support data interleaving (specifically write data interleaving in AXI3, which was deprecated in AXI4), and AXI transactions are often chopped into flits when converted by a Network-on-Chip (NoC) transport layer.

Jackie425:
Sure, I got it. Thanks for clarifying

